PR #110, opened 2017-03-23
modules: remove unneeded resources

crawford:
Setting "enable" to false in an Ignition config has no effect since
"disable" is a seperate (and not yet implemented) option.


crawford:
They were never enabled and even if they were, this config would have no effect. The services would need to be disabled (not supported yet) or masked.

alexsomesan:
sound good. thanks for the clarification.

Quentin-M:
Sooo.. it broke master.

PR #111, opened 2017-03-23
openstack: use coreos-metadata to fetch addresses

crawford:
coreos-metadata is a bit more robust toward network failures and will
fetch public addresses as well (making it easier to customize). This
probably shouldn't be used until coreos-metadata v0.7.0 (the first
version to support OpenStack) is more widely deployed.

### DIFF
--- a/modules/openstack/etcd/ignition.tf
+++ b/modules/openstack/etcd/ignition.tf
@@ -15,17 +15,21 @@ resource "ignition_systemd_unit" "etcd_member" {
     name = "40-etcd-cluster.conf"
 
     content = <<EOF
+[Unit]
+Requires=coreos-metadata.service
+After=coreos-metadata.service
+
 [Service]
+EnvironmentFile=/run/metadata/coreos
 Environment="ETCD_IMAGE_TAG=v3.1.0"
-ExecStartPre=/usr/bin/sh -c '/usr/bin/systemctl set-environment COREOS_PRIVATE_IPV4=$$(curl -s http://169.254.169.254/latest/meta-data/local-ipv4)'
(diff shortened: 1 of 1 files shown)


s-urbaniak:
Tentatively setting the "do not merge" label as long the metadata service is more widely deployed as mentioned above.

crawford:
Container Linux 1353.6.0 (Stable) now has coreos-metadata v0.7.0.

sym3tri:
@s-urbaniak @crawford what's up with this one? It's pretty stale.

s-urbaniak:
The decision to make here is whether we start using the new coreos-metadata v0.7.0. The impact is that Tectonic would not support Container Linux <1353.6.0.

If this is fine from the product side (/cc @robszumski) I gladly take this over.

crawford:
@s-urbaniak You'll need one more change to this PR. When `coreos-metadata` is invoked, you will need to override the provider (e.g. https://github.com/coreos/container-linux-config-transpiler/blob/913c48d7bd00dab8def3cf9acd7398fb9afc7da4/config/types/common.go#L39).

crawford:
This change isn't urgent. It helps improve stability (coreos-metadata is smart enough to retry) and aligns on the usage of coreos-metadata.

robszumski:
Sounds like we are ok pushing this out, removing from upcoming sprint

sym3tri:
Closing b/c this is 5 months old. Please resubmit if you'd like to get this in.

crawford:
I have no dog in the race. It just seemed like a good cleanup since the curl invocation has zero error checking or retry mechanism.

Issue #59, opened 2017-03-16
Use new master label


Quentin-M:
Now that we are finally in control of all the manifests!

sym3tri:
@Quentin-M @s-urbaniak 
Was this fully addressed by #195?

s-urbaniak:
@sym3tri indeed, this is fixed by #195, hence closing.

PR #112, opened 2017-03-23
Make node role labels consistent with other probjects

aaronlevy:
To be consistent with other tools in community:

https://github.com/kubernetes/kubeadm/issues/150#issuecomment-280986966
https://github.com/kubernetes/kubernetes/pull/41835

Original discussion: https://github.com/kubernetes/kubernetes/pull/39112

Originally a value of `true` was proposed, but looks like it might be extraneous and other tools (kubeadm, for example) are not populating the value.

/cc @philips

### DIFF
--- a/Documentation/generic-platform.md
+++ b/Documentation/generic-platform.md
@@ -33,8 +33,8 @@ Master nodes run most, if not all, control plane components including the API se
   - MUST have any necessary API access for k8s cloud plugin functionality (i.e. AWS node IAM Role) 
 
 - **Cloud-init/Ignition:**
-  - MUST run the kubelet with label node-role.kubernetes.io/master=true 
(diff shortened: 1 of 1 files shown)


Quentin-M:
Fixes #59

Quentin-M:
@aaronlevy Would you be willing to fix the labels in manifests and kubelets as well?

s-urbaniak:
@aaronlevy ping regarding the labels fix, otherwise we can take over.

aaronlevy:
Sorry - I wasn't planning to pick up any additional changes. Just wanted to make sure we're all on same page for which label to use.

Quentin-M:
Adding this to the milestone. This is an unique opportunity to make the whole project use these labels. Next opportunity after that would only be 1.7. We can take over the branch.

s-urbaniak:
closing in favor of #195

PR #121, opened 2017-03-24
Fixes for BYO VPC

alexsomesan:
BYO VPC slipped through the recent refactoring and wasn't properly tested.  It was not working before these changes.

With these changes clusters CAN now be built into existing VPC and will work as expected.

[The conditional operator always evaluates both branches](https://github.com/hashicorp/hil/issues/50). On one of the branches, the VPC resource may not be present when using an external VPC, because it's count is set to zero. Thus we have to wrap the potentially missing resource in a fake `join()`. Because the expression is now treated like an list of VPC resources, having an empty list is a valid state and the build works as expected.

Hashicorp is aware of the [issue](https://github.com/hashicorp/terraform/issues/11566) and intends to fix it.

This issue also affected @Quentin-M 's work on BYO external CA certs.

~~We run into a limitation of Terraform that forces us to take an unwanted side-effect.
Because [the conditional operator always evaluates both branches](https://github.com/hashicorp/hil/issues/50), we cannot skip creating a VPC resource as it would lead to one of the condition branches being invalid.
The resulting extra VPC will be empty and not actually used further on by the cluster, which will correctly build into the externally provided VPC. It doesn't incur any costs while it exists, but it will show up when listing VPCs. It also doesn't affect other resources in any way.~~

~~For the sake of providing the BYO VPC feature to users, I feel that it's worth living with this side-effect until the root issue gets addressed.~~

I would like to hear other opinions on this.

@philips @Quentin-M @s-urbaniak @bison

### DIFF
--- a/modules/aws/vpc/outputs.tf
+++ b/modules/aws/vpc/outputs.tf
@@ -1,5 +1,5 @@
 output "vpc_id" {
-  value = "${length(var.external_vpc_id) > 0 ? var.external_vpc_id : aws_vpc.new_vpc.id}"
+  value = "${length(var.external_vpc_id) > 0 ? var.external_vpc_id : join(" ", aws_vpc.new_vpc.*.id)}"
 }
 
 output "cluster_default_sg" {
--- a/modules/aws/vpc/vpc.tf
+++ b/modules/aws/vpc/vpc.tf
@@ -1,13 +1,8 @@
 data "aws_availability_zones" "azs" {}
 
 resource "aws_vpc" "new_vpc" {
-  # count                = "${length(var.external_vpc_id) > 0 ? 0 : 1}"
-  #
-  # We can't yet use the count gate here because of terraform issues:
-  # https://github.com/hashicorp/hil/issues/50
-  # https://github.com/hashicorp/terraform/issues/11566
-  # This should be re-enabled when above issues are fixed.
-  #
+  count = "${var.external_vpc_id == "" ? 0 : 1}"
+
   cidr_block = "${var.cidr_block}"
 
   enable_dns_hostnames = true
@@ -20,5 +15,5 @@ resource "aws_vpc" "new_vpc" {
 }
 
 data "aws_vpc" "cluster_vpc" {
-  id = "${var.external_vpc_id == "" ? aws_vpc.new_vpc.id : var.external_vpc_id }"
+  id = "${var.external_vpc_id == "" ? join(" ", aws_vpc.new_vpc.*.id) : var.external_vpc_id }"
 }
--- a/modules/aws/vpc/vpc.tf
+++ b/modules/aws/vpc/vpc.tf
@@ -1,14 +1,21 @@
 data "aws_availability_zones" "azs" {}
 
 resource "aws_vpc" "new_vpc" {
-  count                = "${length(var.external_vpc_id) > 0 ? 0 : 1}"
-  cidr_block           = "${var.cidr_block}"
+  # count                = "${length(var.external_vpc_id) > 0 ? 0 : 1}"
+  #
+  # We can't yet use the count gate here because of terraform issues:
+  # https://github.com/hashicorp/hil/issues/50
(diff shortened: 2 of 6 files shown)


s-urbaniak:
I am a bit nervous about the amount of logic we are putting in the templates, as HCL is not a proper templating language.

We have been brainstorming with alex today whether we could have an even more granular module structure and removing most of the ternary operators. In the case here we could have two separate vpc modules (`vpc`, `vpc-byo`) and potentially two separate etcd modules (`etcd`, `etcd-byo`) which would be composed in a given platform flavor.

I think we can live for the side-effect for now, so for me this LGTM for the time being.

sym3tri:
FYI this will be an issue for customer as many do not have 'VPC create' access.

alexsomesan:
@sym3tri I have a proper fix :)
I'm updating the PR in a few minutes.

alexsomesan:
Updated with a fix that works correctly.
Albeit the solution is bordering diabolical.

modules/aws/vpc/outputs.tf
```diff
@@ -1,5 +1,5 @@
 output "vpc_id" {
-  value = "${length(var.external_vpc_id) > 0 ? var.external_vpc_id : aws_vpc.new_vpc.id}"
+  value = "${length(var.external_vpc_id) > 0 ? var.external_vpc_id : join(" ", aws_vpc.new_vpc.*.id)}"
```
Quentin-M: 💯  💯  💯  💯  💯  💯  💯

modules/aws/vpc/vpc.tf
```diff
@@ -1,17 +1,19 @@
...
+  cidr_block = "${var.cidr_block}"
+
   enable_dns_hostnames = true
   enable_dns_support   = true
 
   tags {
-    Name              = "${var.cluster_name}"
-    KubernetesCluster = "${var.cluster_name}"
+    Name              = "${var.external_vpc_id == "" ? var.cluster_name : "${var.cluster_name}-side-effect"}"
+    KubernetesCluster = "${var.external_vpc_id == "" ? var.cluster_name : "${var.cluster_name}-side-effect"}"
```
Quentin-M: Do you still need this with the new trick?

platforms/aws/variables.tf
```diff
@@ -24,4 +24,14 @@ variable "tectonic_aws_az_count" {
 
 variable "tectonic_aws_external_vpc_id" {
   type = "string"
-}
\ No newline at end of file
+}
+
+variable "tectonic_aws_external_master_subnets" {
```
Quentin-M: These are new top-level variables for AWS? I'll add them to the upstream installer's var gen. Great catch.

Quentin-M: Can you add them to the Documentation.md too?

Quentin-M: Also, `terraform.tfvars.example`.

alexsomesan: Yeah, those are needed to import existing subnets. I noticed the installer is asking for them if you choose to install in an existing VPC.

I'll ad them to the docs, too.

modules/aws/vpc/vpc.tf
alexsomesan: Nope, not needed. I have to revert those lines.

Quentin-M:
LGTM after comments above :)

Quentin-M:
See https://github.com/coreos/tectonic-installer/pull/128 where the trick is now used too.

platforms/aws/variables.tf
Quentin-M: It seems that we expect to have subnet IDs here, right? Can you either add `_id` or document it where we can at least? This will become harder and harder overtime if we don't clearly specify the expected format (just like for etcd endpoints). Thank you!

alexsomesan: You're right. Variable names should contain some kind of hint (such as suffixes) as to what "type" of data is expected.
I like the idea. Let's stick to it for the future and also document it somewhere.

PR #137, opened 2017-03-27
Restructure files for release

robszumski:
Moving the install docs into `install` will cleanly merge with the other Tectonic installation docs that are also located within a `install` directory.

The line containing `git checkout tags/v0.0.1` will need to be updated per release. What are folks thoughts on accomplishing this?


robszumski:
Updated version numbers in https://github.com/coreos/tectonic-installer/pull/137/commits/a20caa63bdac6d564679c6e74a707c53c388206f

platforms/aws/terraform.tfvars.example
```diff
@@ -10,12 +10,15 @@ tectonic_etcd_count = 3
...
 tectonic_aws_ssh_key = "foobar"
 tectonic_aws_master_ec2_type = "t2.medium"
 tectonic_aws_worker_ec2_type = "t2.medium"
 tectonic_aws_etcd_ec2_type = "t2.medium"
-tectonic_aws_external_vpc_id = ""
+
+tectonic_aws_az_count = 3
+tectonic_aws_external_vpc_id = "vpc-123456"
+tectonic_aws_external_master_subnet_ids = ["subnet-111111", "subnet-222222", "subnet-333333"] # must match AZ count 
+tectonic_aws_external_worker_subnet_ids = ["subnet-111111", "subnet-222222", "subnet-333333"] # must match AZ count
```
robszumski: @s-urbaniak can you take a look at this? this worked in my testing, but what do you think about including this by default, instead of commented out?

Documentation/install/aws/aws-terraform.md
```diff
@@ -12,11 +12,19 @@ Generally, the AWS platform templates adhere to the standards defined by the pro
...
 ## Getting Started
 
-First, download Terraform with via `make`. This will download the pinned Terraform binary and modules:
+First, clone the Tectonic Installer repository in a convenient location:
 
 ```
-$ cd tectonic-installer
-$ make terraform-download
+$ git clone https://github.com/coreos/tectonic-installer.git
+$ git checkout tags/tectonic-1.5.5-tectonic.3
```
sym3tri: Why pin to a specific tag? I thought the point of this is so that people can use the latest version as it will be evolving quickly. Also supported versions will skew by platform.

Documentation/install/aws/aws-terraform.md
```diff
@@ -55,43 +63,51 @@ Now we're ready to specify our cluster configuration.
...
+$ cp platforms/aws/terraform.tfvars.example build/${CLUSTER}/terraform.tfvars
 ```
 
+A few fields require special consideration:
+
+ - **tectonic_base_domain** - domain name that is set in Route53, as described in the prerequisites.
+ - **tectonic_pull_secret_path** - path on disk to your downloaded pull secret. You can find this on your [Account dashboard][account].
+ - **tectonic_license_path** - path on disk to your downloaded Tectonic license. You can find this on your [Account dashboard][account].
+ - **tectonic_admin_password_hash** - generate a hash with the [bcrypt-hash tool][bcrypt] that will be used for your admin user.
+ - **tectonic_aws_external_vpc_id** - specify an existing VPC to use, or leave blank to create a new one
```
sym3tri: Why not put these descriptions inline in the example vars file too?

Documentation/install/aws/aws-terraform.md
robszumski: I was hesitant to do that for this initial sync until we get our testing process down more, but I guess that is the entire point..

Documentation/install/aws/aws-terraform.md
robszumski: We could. I also don't know what the future of the other variable docs that are already out of date. seems like we should remove them.

Documentation/install/aws/aws-terraform.md
robszumski: How do we sync that up with the customized binary?

robszumski:
Tested AWS. Microsoft has tested the Azure stuff. Ready to merge once we figure out the versioning issue above.

PR #152, opened 2017-03-30
modules/tectonic: stop templating console

squat:
This commit removes templated variables from the console deployment. This
allows us to manage the console deployment manifest with operators.


Quentin-M:
Nice! I did not know we could do that.

PR #156, opened 2017-03-30
Don't create new_vpc if using external_vpc_id

rji:
A new VPC should not be created if `external_vpc_id` was provided.


rji:
Compare with the desired output at https://github.com/coreos/tectonic-installer/blob/73d433414a0ead4bd55cd3786a7aca0d4ea35e9b/modules/aws/vpc/outputs.tf#L1-L3

rji:
Closing in favor of #151.

PR #159, opened 2017-03-31
Use same conditional format as in rest of module

alexsomesan:
`var.external_endpoints` is a list, so calling `length()` on it will return the number of elements.
No need to `join()` convert it to a string to check emptiness.

@Quentin-M @s-urbaniak


modules/aws/etcd/outputs.tf
```diff
@@ -1,4 +1,4 @@
 # We have to do this join() & split() 'trick' because the ternary operator can't output lists.
 output "endpoints" {
-  value = ["${split(",", join(",", var.external_endpoints) == "" ? join(",", aws_route53_record.etc_a_nodes.*.fqdn) :  join(",", var.external_endpoints))}"]
-}
\ No newline at end of file
+  value = ["${split(",", length(var.external_endpoints) == 0 ? join(",", aws_route53_record.etc_a_nodes.*.fqdn) :  join(",", var.external_endpoints))}"]
```
Quentin-M: Double space then LGTM.

PR #160, opened 2017-03-31
Add syntax formatting make targets

alexsomesan:
Terraform comes with a convenient syntax checking and formatting tool which can be used to ensure templates follow canonical syntax rules.

This change introduces two new make targets:
* `structure-check` will be used by the upcoming PR checker to validate canonical syntax
* `canonical-syntax` can be used at any time to align the complete repository to canonical syntax.


philips:
LGTM! Thanks!

PR #163, opened 2017-03-31
platforms/azure: add cloudapp dns name

sozercan:
Adds Azure <resourceGroupName>.cloudapp.azure.com FQDN if `tectonic_create_dns_zone = "false"` in tfvars which makes Azure DNS Zone creation optional. It is disabled by default.


modules/azure/master/output.tf
```diff
@@ -5,3 +5,19 @@ output "ip_address" {
 output "console_ip_address" {
   value = "${azurerm_public_ip.tectonic_console_ip.ip_address}"
 }
+
+output "ingress_external_fqdn" {
+  value = ["${azurerm_public_ip.tectonic_console_ip.*.fqdn}"]
```
Quentin-M: Discussed offline, looks like `.*` is not required. The resource will exist anyways. The FQDN might be an empty string but that's fine.

s-urbaniak:
@sozercan this seems to have a conflict with the current master changes. Do you mind to rebase? Thanks!

sozercan:
@s-urbaniak should be good now, thanks!

config.tf
```diff
@@ -163,6 +163,11 @@ variable "tectonic_ca_key_alg" {
   default     = "RSA"
 }
 
+variable "tectonic_create_dns_zone" {
+  description = "If set to true, create an Azure DNS zone"
```
Quentin-M: This is an Azure-specific variable. It should not be in that configuration file but in `platforms/azure/variables.tf`. Can we fix that please?

PR #171, opened 2017-04-03
modules/bootkube: disable alpha API groups

squat:
This commit changes the `--runtime-config` flags on the apiserver so
that only stable API groups and the RBAC APIs are enabled. in 1.6, RBAC
graduates to beta and we can remove that line as well.


s-urbaniak:
To be have a bit more clarity from my side, do we need to wait for 1.6 to be able to get this is or does this already apply to 1.5?

squat:
@s-urbaniak this applies to 1.5 as well! Hoping to get this in ASAP. The only alpha API group we use is rbac, and that has graduated in 1.6

PR #180, opened 2017-04-04
*: do not use paths for license/pull secret

Quentin-M:
This actually makes using the the SDK more complicated as multiple files need to be copied over and paths (relative/absolute) must be managed.

Tested on AWS.


alexsomesan:
I'm not quite sure what kind of complications do you mean to avoid with this change.
I was assuming copying files is a solved problem in this day and age.

**TL;DR;** No ops-minded engineer would agree that credentials backed in configuration are a good idea. Those ops-minded engineers are our users.

There are more implications here than just simplifying away a file copy.
The Tectonic license blob and Quay pull blob are essentially secrets.
Real-world users in production environments most often have strong processes around managing secrets, which are more strict and cumbersome then dealing with code or configuration.

With this change, we would be pushing these users to handle the entire cluster configuration as a secret, with all the extra effort around that. I don't expect this to be well received by most ops-minded users.

I'd like us to reconsider options here and fall-back on the side of ensuring security for our users (we are, after all, out here to "Secure the internet"). We should invest in the opposite direction, of linting the rest of the in-lined credentials form cluster configuration and the state files.

Quentin-M:
Sounds good~

ggreer:
`ca_key` is already in tfvars, so that ship has sailed. I'd very much like for this PR to be resurrected, as it would greatly simplify the code that converts UI state to TF vars.

philips:
What is the risk with using a relative path?

PR #186, opened 2017-04-05
Fixes ingress.yaml fqdn

coresolve:
- Changes the var in use so that fqdn is populated without the scheme.


s-urbaniak:
I think we fixed this already in master, hence we can close this, right @alexsomesan?

alexsomesan:
Yes, we fixed this yesterday because it was also blocking some work from the monitoring team.

PR #193, opened 2017-04-06
modules/{bootkube,tectonic}: use the tectonic provider

Quentin-M:
This commit is intended to be merged at the same time as the next Tectonic release. It will enable users to use the upstream TerraForm binaries, alongside with a plugin called `tectonic`, that is the Tectonic binary. That, until the plugins are upstreamed.

It'll require a documentation update in order to explain how the `.terraformrc` file has to be modified for this to work. At the same time, we will also be able to remove:
- The Makefile targets for TerraForm,
- The TerraForm fork.


Quentin-M:
Absolutely. Thanks!

PR #196, opened 2017-04-07
Jenkins pipeline first step: run Terraform syntax checker

alexsomesan:
This change introduces a simple Jenkins Pipeline DSL file that right now just runs one test: the Terraform syntax check.
Although not much of a test in itself, this serves a good base for tweaking Jenkins to run PR checks and other tests safely.
Once this is in place, we I will start adding Tectonic tests.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,7 +1,7 @@
 pipeline {
   agent {
     docker {
-      image 'quay.io/coreos/tectonic-terraform'
+      image 'quay.io/coreos/tectonic-terraform:v0.0.2'
       label 'worker'
     }
   }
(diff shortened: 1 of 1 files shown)


alexsomesan:
cc: @mxinden @sym3tri @Quentin-M @bison

Jenkinsfile
```diff
@@ -0,0 +1,21 @@
+pipeline {
+  agent {
+    docker {
+      image 'quay.io/coreos/tectonic-terraform'
```
Quentin-M: Tag?

Quentin-M:
It looks good as it seems very straight forward to integrate with the upcoming https://github.com/coreos-inc/tectonic/blob/master/Jenkinsfile (sometimes next week).

Jenkinsfile
alexsomesan: Done

Issue #61, opened 2017-03-16
Setup automated builds + smoke tests 


Quentin-M:
As soon as possible, we should come up with a plan and execute on running:

- [ ] Smoke tests for every PR
  - [x] `terraform fmt`
  - [ ] `terraform validate`
  - [ ] orphan variables / parameters check 
  - [x] `terraform apply` on AWS with basic parameters
- [ ] Nightly builds
- [ ] Daily scenarized tests (new VPC, existing VPC, external etcd, self-hosted etcd, ...)
- [ ] E2E tests

For the scenarized / E2E tests, we should sync with @pbx0, @ethernetdan and @bison to see how we can integrate with the existing tooling.

I think it could be interesting to have something like:
```
tests/scenarios/
test/scenerarios/aws-existing-vpc.tfvars <--- Input variables used to run tectonic-terraform-sdk
test/scenarios/aws-existing-vpc.tf      <--- Extra infrastructure to deploy before running 
tectonic-terraform-sdk (e.g. existing vpcs), output is used to substitute variables in associated .tfvars
test/scenarios/aws-self-hosted-etcd.tfvars
[...]
```

We have set up a public Jenkins to run all these:
- https://jenkins-tectonic-installer.prod.coreos.systems/
- https://jenkins-tectonic-installer-public.prod.coreos.systems/

pbx0:
I would be interested in making this a supported test platform in pluton. Its an open-question of how we would implement this and if pluton could be made flexible enough to deal with these platform specific scenarios. Some thoughts if anyone else is interested in this route.

In pluton, tests are defined as a go function with a signature `func(c *Cluster)` This `Cluster` is a platform-independant abstraction for managing a k8s cluster that also has the equivalent of a golang  `testing.T` embedded for logging output, skipping, having subtests, ect. 

Because test functions test against this abstraction, we need a declarative way to describe those testing scenarios if we want this integration to work.

Currently a test is declared:
```
pluton.Test{
        Name: "test.name",
        Run:  f,
        Options: pluton.Options{
                SelfHostEtcd:   false,
                InitialMasters: 1,
                InitialWorkers: 1,
        },
}
```
[...]

PR #207, opened 2017-04-10
Jenkins AWS smoke test

ethernetdan:
Adds AWS smoke testing to be eventually run on PRs

Creates a new folder for test configuration but we may want to get it working with the examples folder.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -35,7 +35,9 @@ pipeline {
               # Set required configuration
               export PLATFORM=aws
               export CLUSTER="tf-${PLATFORM}-${BRANCH_NAME}-${BUILD_ID}"
-              export TF_VAR_tectonic_cluster_name=${CLUSTER}
+
+              # s3 buckets require lowercase names
+              export TF_VAR_tectonic_cluster_name=$(echo ${CLUSTER} | awk '{print tolower($0)}')
 
               # AWS specific configuration
               export AWS_REGION="us-west-2"
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -23,12 +23,22 @@ pipeline {
         parallel (
           "AWS": {
               withCredentials([file(credentialsId: 'tectonic-license', variable: 'TF_VAR_tectonic_pull_secret_path'),
-                               file(credentialsId: 'tectonic-pull', variable: 'TF_VAR_tectonic_license_path')]) {
+                               file(credentialsId: 'tectonic-pull', variable: 'TF_VAR_tectonic_license_path'),
+                               [
+                                 $class: 'UsernamePasswordMultiBinding',
+                                 credentialsId: 'tectonic-aws-creds',
+                                 usernameVariable: 'AWS_ACCESS_KEY_ID',
+                                 passwordVariable: 'AWS_SECRET_ACCESS_KEY'
+                               ]
+                               ]) {
               sh '''
               # Set required configuration
               export PLATFORM=aws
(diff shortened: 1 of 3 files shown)


Quentin-M:
Have you seen https://github.com/coreos/tectonic-installer/issues/61?

Jenkinsfile
```diff
@@ -17,5 +17,54 @@ pipeline {
         sh 'make structure-check'
       }
     }
+
+    stage('Smoke Tests') {
+      steps {
+        parallel (
+          "AWS": {
+              withCredentials([file(credentialsId: 'tectonic-license', variable: 'TF_VAR_tectonic_pull_secret_path'),
```
dghubble: @ethernetdan If you check the Tectonic repo's Jenkinsfile, there is a tip about avoiding writing logic like this in the Jenkinsfile directly. Can you place this in a script which gets invoked please?

Jenkinsfile
```diff
@@ -17,5 +17,54 @@ pipeline {
...
+
+              # Create local config
+              make localconfig
+
+              # Use smoke test configuration for deployment
+              ln -sf ${WORKSPACE}/test/aws.tfvars ${WORKSPACE}/build/${CLUSTER}/terraform.tfvars
+
+              make plan
+
+              # always cleanup cluster
```
dghubble: You can use the Jenkins declarative pipeline `post` feature as well, to ensure a certain cleanup script or task is always run at the Jenkins level.

Makefile
```diff
@@ -26,7 +26,7 @@ apply: $(BUILD_DIR)/.terraform
 	cd $(BUILD_DIR) && terraform apply $(TOP_DIR)/platforms/$(PLATFORM)
 
 destroy:
-	cd $(BUILD_DIR) && terraform destroy $(TOP_DIR)/platforms/$(PLATFORM)
+	cd $(BUILD_DIR) && terraform destroy -force $(TOP_DIR)/platforms/$(PLATFORM)
```
mxinden: @ethernetdan thanks! Will need this for the prometheus operator e2e tests. :)

Jenkinsfile
mxinden: Having this in two scripts (e.g. 'bring-up-cluster', 'tear-down-cluster') would help us a lot with testing the prometheus operator.

mxinden:
@ethernetdan Thanks for this! We are about to use the tectonic-installer to bring up a cluster in the prometheus-operator e2e-tests. It works so far ([tectonic-installer-branch](https://github.com/mxinden/prometheus-operator/tree/tectonic-installer) & [jenkins-job](https://jenkins-monitoring.prod.coreos.systems/job/prometheus-operator-e2e-tests-tectonic-installer/lastFailedBuild/console)), but  is still  hacky. I will adapt most of this as it is a lot cleaner.

alexsomesan:
@dghubble I remember your suggestion from earlier on about following the example of the current tectonic Jenkinsfile to use K8S pods as workers. That's actually the path I had set out on, but have since heard from @chancez and @bison that we can't actually use that approach for open-source projects, because of security concerns around PR form outside CoreOS.
Given those restrictions, I think we need to keep the current setup until we can maybe configure an isolated cluster for these workers.

Jenkinsfile
alexsomesan: That's a good point! We should make use of the post step for cleanup.

dghubble:
@alexsomesan external contributors should not be running un-approved code on that CoreOS VM via docker or in pods on CoreOS cluster infrastructure. Neither is safe.

PR #224, opened 2017-04-12
docs: add missing modules note

robszumski:


robszumski:
Will make https://github.com/coreos/tectonic-forum/issues/96 easier to debug if someone google's the error

PR #225, opened 2017-04-12
Idempotent plan / apply for same cluster

alexsomesan:
The cluster_id is being newly generated on each terraform run. This lead terraform to believe that there are always changes to the assets parameters and prevents idempotent operations.

This change makes the cluster_id constant for the same cluster. 

@Quentin-M @s-urbaniak


s-urbaniak:
Looks fine from my side, but awaiting response from @squat due to the format change.

modules/tectonic/assets.tf
```diff
@@ -58,7 +58,7 @@ resource "template_folder" "tectonic" {
     kube_apiserver_url = "${var.kube_apiserver_url}"
     oidc_issuer_url    = "https://${var.base_address}/identity"
 
-    cluster_id            = "${uuid()}"
+    cluster_id            = "${sha256("${var.kube_apiserver_url}-${var.platform}")}"
```
squat: This will work fine and does. It violate any assumptions. We generate random UUIDs to ensure no PII was leaked; for all other reasons the cluster_id could just as well be `aleks_cluster`.

yifan-gu:
Hmm, changing the format is breaking channel operator, as it's expecting it to be uuid when connecting to omaha server.

Quentin-M:
```uuid() - Returns a UUID string in RFC 4122 v4 format. This string will change with every invocation of the function, so in order to prevent diffs on every plan & apply, it must be used with the ignore_changes lifecycle attribute.```

Quentin-M:
@yifan-gu fixing

PR #226, opened 2017-04-12
console: Adds flags for token revocation in console and update dex

amrutac:


modules/tectonic/resources/manifests/console/deployment.yaml
```diff
@@ -85,6 +85,15 @@ spec:
...
           value: /etc/tectonic/licenses/license
+        - name: BRIDGE_DEX_CLIENT_CERT_FILE
+          value: /etc/tectonic-identity-grpc-client-secret/tls-cert
+        - name: BRIDGE_DEX_CLIENT_KEY_FILE
+          value: /etc/tectonic-identity-grpc-client-secret/tls-key
+        - name: BRIDGE_DEX_API_HOST
+          valueFrom:
+            configMapKeyRef:
+              name: tectonic-config
+              key: dexAPIHost
```
Quentin-M: I don't see `dexAPIHost ` in https://github.com/coreos/tectonic-installer/blob/660bf8b/modules/tectonic/resources/manifests/config.yaml?

rithujohn191: yes you are right. we need an entry in this file similar to what we have in the tectonic repo:
data["dexAPIHost"] = fmt.Sprintf("%s:5557", tectonicIdentityAPIService)

rithujohn191: the tectonicIdentityAPIService variable will have to be defined in https://github.com/coreos/tectonic-installer/blob/master/modules/tectonic/assets.tf#L7 as well

Quentin-M: Variables to that file come from https://github.com/coreos/tectonic-installer/blob/660bf8b/modules/tectonic/assets.tf#L7

amrutac:
Closing this since it needs other modules to be updated. @rithujohn191 will create that PR.

modules/tectonic/resources/manifests/console/deployment.yaml
Quentin-M: As discussed online, I'd rename the `kube_apiserver_url` variable to `kube_apiserver_endpoint`, remove the scheme/port in the tectonic module's callers, and add it back in the templating - so you can also create the dex URL with that variable.

PR #231, opened 2017-04-12
console: Add flags for token revocation in console and update dex.

rithujohn191:
This change adds the required flags to the console deployment to enable Dex API calls. Also updates dex to v2.3.0


s-urbaniak:
I believe this was merged a bit too early, as `identity_api_service` in `modules/tectonic/assets.tf`, I will file a follow-up PR for that.

Issue #228, opened 2017-04-12
Subdomains  != Cluster Name


kans:
Clusters should be able to be located under a subdomain that differs from the cluster name.

PR #237, opened 2017-04-13
AWS: implement custom DNS prefixes

alexsomesan:
This fixes https://github.com/coreos/tectonic-installer/issues/228

@Quentin-M @s-urbaniak

### DIFF
--- a/modules/aws/master-asg/elb.tf
+++ b/modules/aws/master-asg/elb.tf
@@ -34,7 +34,7 @@ resource "aws_elb" "api-internal" {
 
 resource "aws_route53_record" "api-internal" {
   zone_id = "${var.internal_zone_id}"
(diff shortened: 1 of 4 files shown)


Quentin-M:
I am confused. How much do we need that? This was present due to the hard requirement by a customer that his CloudFormation name was specifically something (policies). We don't have CloudFormation anymore.

alexsomesan:
I'm not very familiar with the history of this feature, but we had an issue attached to the milestone to fix it.
To me it made sense since the UI exposes it now so we have to support it to claim feature parity.

@sym3try Thoughts?

alexsomesan:
@sym3tri ^

sym3tri:
Having this helps achieve feature parity WRT available customizations. 

IIRC customers have specific required formats they need to use for both CF and for their DNS.

Issue #221, opened 2017-04-11
Enable customization of Docker log rotation


robszumski:
Right now we're relying on each customer to configure Docker log rotation. Example:

```
DOCKER_OPTS=--log-opt max-size=50m --log-opt max-file=3
```

We should set a sane default that doesn't grow forever, and then offer a way to easily customize. Not sure if it makes more sense as a top level variable or an easy to understand section within an Ignition config.

colhom:
@robszumski I worry that allowing `max-file > 1` will not play well with the way many on-cluster log aggregators (ie: fluentd) currently work. Many use a regex to find all container log files in `/var/log/containers/*.log` and keep track of what logs have been spooled in based on filename. I feel that if we're doing rotation in this container log directory, aggregators with loose regexs may pick up the old log files and potentially re-spool those logs into the aggregator. I wouldn't be surprised if there are currently production deployments will assume all files in the set `/var/log/containers/*` are unique log files.

robszumski:
I just grabbed an example, I have no attachment to 3 :)

chancez:
I don't think that will be an issue. Upstream is doing the same in GCI. Additionally, since we use ignition, only new nodes in an existing cluster would have the config change applied, so we don't have to worry about an existing node's log files.

https://github.com/kubernetes/kubernetes/pull/40634/

s-urbaniak:
log rotation landed in #244, hence I am closing this issue for now. Feel free to reopen if we missed anything.

PR #244, opened 2017-04-14
modules/aws: limit docker log file sizes

sym3tri:
Untested. Do NOT merge yet.

First step towards fixing #221

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,7 +1,7 @@
 pipeline {
   agent {
     docker {
-      image 'quay.io/coreos/tectonic-terraform:v0.0.2'
+      image 'quay.io/coreos/tectonic-terraform:v0.8.8-coreos.1'
       label 'worker'
     }
   }
--- a/modules/aws/ignition/ignition.tf
+++ b/modules/aws/ignition/ignition.tf
@@ -19,6 +19,13 @@ resource "ignition_config" "main" {
 resource "ignition_systemd_unit" "docker" {
   name   = "docker.service"
   enable = true
+
+  dropin = [
+    {
+      name    = "10-dockeropts.conf"
+      content = "[Service]\nEnvironment=DOCKER_OPTS=--log-opt max-size=50m --log-opt max-file=3\n"
+    },
+  ]
 }
 
 resource "ignition_systemd_unit" "locksmithd" {
--- a/modules/aws/ignition/ignition.tf
+++ b/modules/aws/ignition/ignition.tf
@@ -23,7 +23,7 @@ resource "ignition_systemd_unit" "docker" {
   dropin = [
     {
       name    = "10-dockeropts.conf"
-      content = "[Service]\nEnvironment=DOCKER_OPTS=--log-opt max-size=50m --log-opt max-file=3\n"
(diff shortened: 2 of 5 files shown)


sym3tri:
Handing this over to @squat

chancez:
So the current commit isn't quite right, it's missing quotes around the value.

I used this and my nodes came up successfully:
```diff
+
+  dropin = [
+    {
+      name    = "10-dockeropts.conf"
+      content = "[Service]\nEnvironment=\"DOCKER_OPTS=--log-opt max-size=50m --log-opt max-file=3\"\n"
+    },
+  ]
```

I'll respond shortly once I've determined it actually limits the log sizes. It might be a little before I can personally say it works with fluentd, but I'm actually not all that worried about that.

chancez:
It might be a good idea to parameterize these, and set some reasonable defaults. I do expect some customers will need to tweak these settings since it has an impact on if logs get deleted before they're collected, depending on log throughput and collection throughput.

sym3tri:
@chancez yeah I agree. But first let's just get this in to avoid the urgent disk issues we've been seeing. We can parameterize later in a separate PR

s-urbaniak:
@sym3tri set the `do-not-merge` label for now as per initial comment. Please ping once we can get this in. Thanks!

sym3tri:
@squat thanks for fixing & testing. Is this ready to remove the "WIP" and "do-not-merge"?

squat:
yes, just finished testing and it is all looking good!

squat:
rebased and ready to merge @sym3tri

sym3tri:
@squat great. Can you summarize how you tested/validated the changes?

squat:
I primarily tested that the PR produces an ignition file that results in a valid configuration for Docker. To test this, I verified that the  `--log-opt max-size` parameter actually limits the logs size.

To reproduce:

* modify log options to be `--log-opt max-size=1k`
* terraform apply to create cluster
* ssh onto node
* test:
    ```shell
    # run test command to generate lots of logs:
    docker run -d --name=test -t busybox /bin/sh -c 'n=0; while true; do n=$((n+1));echo $n; done'
    sleep 2s && docker kill test
    # show all the logs. verify that this only every shows a few (~10) lines; most of the time this shows nothing as the logs are constantly being rotated:
    docker logs test --tail=all
    docker rm test
    # change log opts back to `50m`
    sudo vim /etc/systemd/system/docker.service.d/10-dockeropts.conf
    # restart docker:
    sudo systemctl daemon-reload
    sudo systemctl restart docker
    # run test command again:
    docker run -d --name=test -t busybox /bin/sh -c 'n=0; while true; do n=$((n+1));echo $n; done'
    sleep 2s && docker kill test
    # show all the logs. verify that all the data is shown:
    docker logs test --tail=all
    docker rm test
    ```

squat:
alternatively:

* modify log options to be `--log-opt max-size=1k`
* terraform apply to create cluster
* ssh onto node
* verify all logs are smaller than 1k:
    ```sh
    ls -hal /var/lib/docker/containers/*/*.log | awk '{print $5}'
    ```

chancez:
I've also tested my own patch of this which is identical. Fluentd/kubelet also handle the log rotation just fine.

sym3tri:
LGTM

PR #251, opened 2017-04-17
modules/aws: Add kubelet-env.service to get the kubelet image version.

yifan-gu:
Also make kubelet.service to use '/etc/kubernetes/kubelet.env' for
the image URL and VERSION. The '/etc/kubernetes/kubelet.env' will only
be set for the master node.

### DIFF
--- a/modules/aws/ignition/resources/services/kubelet-env.service
+++ b/modules/aws/ignition/resources/services/kubelet-env.service
@@ -1,0 +1,15 @@
+[Unit]
+Description=Determine the Kubelet Image Version
+ConditionPathExists=!/etc/kubernetes/kubelet.env
+
+[Service]
+
+ExecStartPre=/usr/bin/mkdir -p /etc/kubernetes
+ExecStartPre=/usr/bin/bash -c "/opt/s3-puller.sh ${kubeconfig_s3_location} /etc/kubernetes/kubeconfig"
+ExecStartPre=/usr/bin/bash -c "docker run --rm -v /etc/kubernetes/kubeconfig:/kubeconfig ${kube_version_image_url}:${kube_version_image_tag} --kubeconfig=/kubeconfig > /etc/kubernetes/kube.version"
+ExecStart=/usr/bin/bash -c "echo KUBELET_IMAGE_URL=${kubelet_image_url} > /etc/kubernetes/kubelet.env; echo KUBELET_IMAGE_TAG=$(tr '+' '_' < /etc/kubernetes/kube.version) >> /etc/kubernetes/kubelet.env; rm /etc/kubernetes/kube.version"
+Restart=on-failure
+RestartSec=10
+
+[Install]
+WantedBy=multi-user.target
--- a/modules/aws/ignition/resources/services/kubelet.service
+++ b/modules/aws/ignition/resources/services/kubelet.service
@@ -2,8 +2,7 @@
 Description=Kubelet via Hyperkube ACI
 
 [Service]
-Environment=KUBELET_IMAGE_URL=${aci}
-Environment=KUBELET_IMAGE_TAG=${version}
+EnvironmentFile=/etc/kubernetes/kubelet.env
 Environment="RKT_RUN_ARGS=--uuid-file-save=/var/run/kubelet-pod.uuid \
   --volume=resolv,kind=host,source=/etc/resolv.conf \
   --mount volume=resolv,target=/etc/resolv.conf \
--- a/config.tf
+++ b/config.tf
@@ -42,6 +42,7 @@ variable "tectonic_container_images" {
     etcd_operator                = "quay.io/coreos/etcd-operator:v0.2.5"
     kenc                         = "quay.io/coreos/kenc:48b6feceeee56c657ea9263f47b6ea091e8d3035"
     awscli                       = "quay.io/coreos/awscli:025a357f05242fdad6a81e8a6b520098aa65a600"
+    kube_version                 = "quay.io/coreos/kube-version:0.1.0"
   }
 }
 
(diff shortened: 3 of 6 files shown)


yifan-gu:
cc @Quentin-M @aaronlevy @philips 

@Quentin-M To finish the loop, we should modify the installer to only populate the bootstrap node's `/etc/kubernetes/kubelet.env` as we discussed last week.

Quentin-M:
For AWS, it's in there: https://github.com/coreos/tectonic-installer/blob/master/modules/aws/ignition/ignition.tf

yifan-gu:
@Quentin-M But how do I create the kubelet.env **ONLY** in the bootstrap nodes, not in the other nodes created by autoscale groups?

yifan-gu:
OK, modified the PR to only work for aws for now.
Testing it.

Quentin-M:
(Discussed in detail offline for the question above)

s-urbaniak:
Tentatively setting the do-not-merge label. @yifan-gu do you mind to ping here once the tests are finished so we can do the merge? Thanks!

yifan-gu:
@Quentin-M @s-urbaniak @aaronlevy I think this is working now for the aws installer:
```
 kubectl get nodes
NAME                                       STATUS    AGE       VERSION
ip-10-0-37-25.us-west-2.compute.internal   Ready     8m        v1.6.1+coreos.0
ip-10-0-79-74.us-west-2.compute.internal   Ready     7m        v1.6.1+coreos.0
```

The master node has the `kubelet.env` populated by the `init-assets.sh`, while the worker node uses `kubelet-env.service` to get the version.

yifan-gu:
But seems I also need to make other nodes in master group to use this, not only the worker group.

yifan-gu:
OK, I think this is working now, I created a cluster with 1 master 1 worker.
Then scale it to 2 masters, 2 workers.
Only the 1st master gets the kube-version from the seeded value, all others get the kube-version from the API server:
```
kubectl get nodes
NAME                                        STATUS    AGE       VERSION
ip-10-0-18-162.us-west-2.compute.internal   Ready     9m        v1.6.1+coreos.0
ip-10-0-33-13.us-west-2.compute.internal    Ready     1m        v1.6.1+coreos.0
ip-10-0-75-234.us-west-2.compute.internal   Ready     2m        v1.6.1+coreos.0
ip-10-0-94-170.us-west-2.compute.internal   Ready     8m        v1.6.1+coreos.0
```

dghubble:
We'll need to make the same changes to systemd units in the bare-metal Container Linux Configs, which do not use the terraform representation of Ignition.

yifan-gu:
@dghubble Sure, I will need help from you on testing a baremetal cluster :)

modules/aws/ignition/resources/services/kubelet.service
```diff
@@ -2,8 +2,7 @@
 Description=Kubelet via Hyperkube ACI
 
 [Service]
-Environment=KUBELET_IMAGE_URL=${aci}
-Environment=KUBELET_IMAGE_TAG=${version}
+EnvironmentFile=/etc/kubernetes/kubelet.env
```
Quentin-M: It'll fail if the file does not exist right? Should we add a dependency to avoid that?

Quentin-M:
@alexsomesan @s-urbaniak We need that for the release. Any concerns in this PR?

modules/aws/ignition/resources/services/kubelet.service
dghubble: I thought the plan was to move away from using the kubelet.env file? cc @yifan-gu

s-urbaniak: The kubelet will crash-loop until init-asset is done, and a general dependency on it will hurt on reboot, because init-asset is started during installation only. There could be a more elegant general solution, but until then I am fine with this dependency on the envelope file.

s-urbaniak:
From my side I have no general concerns, except the discussion mentioned above.

modules/aws/ignition/resources/services/kubelet-env.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=Determine the Kubelet Image Version
+ConditionPathExists=!/etc/kubernetes/kubelet.env
+
+[Service]
+
+ExecStartPre=/usr/bin/mkdir -p /etc/kubernetes
+ExecStartPre=/usr/bin/bash -c "/opt/s3-puller.sh ${kubeconfig_s3_location} /etc/kubernetes/kubeconfig"
+ExecStartPre=/usr/bin/bash -c "docker run --rm -v /etc/kubernetes/kubeconfig:/kubeconfig ${kube_version_image_url}:${kube_version_image_tag} --kubeconfig=/kubeconfig > /etc/kubernetes/kube.version"
+ExecStart=/usr/bin/bash -c "echo KUBELET_IMAGE_URL=${kubelet_image_url} > /etc/kubernetes/kubelet.env; echo KUBELET_IMAGE_TAG=$(tr '+' '_' < /etc/kubernetes/kube.version) >> /etc/kubernetes/kubelet.env"
```
aaronlevy: A nit might be to clean up `/etc/kubernetes/kube.version` file if it is never expected to be used again

aaronlevy:
On thing that we'll all need to discuss at some point is the plan for how we actually update the user-data of a cluster. At some point something about `kubelet.service` or `kube-env.service` will need to change -- and we don't have a good story around this.

Doesn't need to be addressed here -- but something we should start thinking about

modules/aws/ignition/resources/services/kubelet.service
yifan-gu: `kubelet.env` will be prepared either by `init-assets.sh` (on the leader master node), or `kubelet-env.service` (all other nodes, including non-leading master nodes).

`kubelet.service` will crashloop until the `kubelet.env` is prepared. A dependency will make `kubelet.service` fail permanently if the service that `kubelet.service` depends on fails.

yifan-gu:
test upgrade from master

yifan-gu:
test upgrade from master

yifan-gu:
Please ignore the `jenkins/upgrade test` failure, that's because I am testing my jenkins job.

Quentin-M:
How is it going? 👍

yifan-gu:
Should be good to go for aws, let me rebase.

PR #261, opened 2017-04-18
Update default IOPS for AWS gp2 volumes to 100

chancez:
The minimum value for gp2 IOPS is 100, this updates the default to
reflect this.

Without this, if a user runs terraform plan or apply, it would inform
the user that the volume needs to be recreated due to IOPs changing from
"100" => "0" after the refresh gathers AWS' default value of "100" from
the API.

### DIFF
--- a/platforms/aws/variables.tf
+++ b/platforms/aws/variables.tf
@@ -88,7 +88,7 @@ variable "tectonic_aws_etcd_root_volume_size" {
 
 variable "tectonic_aws_etcd_root_volume_iops" {
   type        = "string"
-  default     = "0"
+  default     = "100"
   description = "The amount of provisioned IOPS for the root block device of etcd nodes."
 }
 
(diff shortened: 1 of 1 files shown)


chancez:
cc @s-urbaniak for a review.

sym3tri:
nice catch

s-urbaniak:
Thanks, indeed a good catch. I did check the final IOPs being set to 100 on the AWS console but didn't see the messages you mentioned during my tests. Without deeper investigation the only difference I see that I tested this in eu-west-1, maybe the behavior is a little different here?!

Nevertheless merging this, since this is a sane default. Thanks!

chancez:
@s-urbaniak to see the effect, you would have had to run apply, then run plan/apply again afterwards. After the plan/apply refreshes the data, it would detect the current value is 100, but the desired is 0.

Easiest way to remember is check your work by running a `terraform plan` after `terraform apply` to make sure stuff is idempotent and not changing.

PR #265, opened 2017-04-19
modules/tectonic: add ingress role

robszumski:
Docs for using this are available in https://github.com/coreos-inc/tectonic/pull/1727


robszumski:
Depends on content that needs to land in https://github.com/coreos-inc/tectonic/pull/1723

robszumski:
Add this as a release blocker since the docs referencing this role have been merged...

Issue #214, opened 2017-04-10
Independent Worker/Controller AZ count


kans:

alexsomesan:
This is already possible.

Can you please document the specific issue a bit more? It's hard to understand from the current description :)

alexsomesan:
resolved by https://github.com/coreos/tectonic-installer/pull/267

PR #267, opened 2017-04-19
AWS: VPC subnets with custom CIDRs and AZs per workers / masters

### LINKED ISSUE #215: Support Custom Worker/Controller CIDRs
... for a given AZ when creating the VPC.

alexsomesan:
This change allows the VPC module to create subnets in arbitrary subsets of availability zones and using per-subnet custom CIDR ranges.

Resolves #215 and #214 

cc: @s-urbaniak @Quentin-M @sym3tri

### DIFF
--- a/modules/aws/vpc/vpc-private.tf
+++ b/modules/aws/vpc/vpc-private.tf
@@ -1,5 +1,5 @@
 resource "aws_route_table" "private_routes" {
-  count  = "${var.external_vpc_id == "" ? var.az_count : 0}"
+  count  = "${var.external_vpc_id == "" ? var.worker_az_count : 0}"
   vpc_id = "${data.aws_vpc.cluster_vpc.id}"
 
   tags = "${merge(map(
@@ -9,7 +9,7 @@ resource "aws_route_table" "private_routes" {
 }
 
 resource "aws_route" "to_nat_gw" {
-  count                  = "${var.external_vpc_id == "" ? var.az_count : 0}"
+  count                  = "${var.external_vpc_id == "" ? var.worker_az_count : 0}"
   route_table_id         = "${aws_route_table.private_routes.*.id[count.index]}"
   destination_cidr_block = "0.0.0.0/0"
   nat_gateway_id         = "${aws_nat_gateway.nat_gw.*.id[count.index]}"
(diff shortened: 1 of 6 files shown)


modules/aws/vpc/vpc-private.tf
```diff
@@ -9,28 +9,39 @@ resource "aws_route_table" "private_routes" {
...
-  cidr_block        = "${cidrsubnet(data.aws_vpc.cluster_vpc.cidr_block, 4, count.index + var.az_count)}"
-  vpc_id            = "${data.aws_vpc.cluster_vpc.id}"
-  availability_zone = "${data.aws_availability_zones.azs.names[count.index]}"
+  count = "${var.external_vpc_id == "" ? var.worker_az_count : 0}"
+
+  vpc_id = "${data.aws_vpc.cluster_vpc.id}"
+
+  cidr_block = "${lookup(var.worker_subnets,
+    "${element(concat(keys(var.worker_subnets), list("padding")), count.index)}",
+    "${cidrsubnet(data.aws_vpc.cluster_vpc.cidr_block, 4, count.index + var.worker_az_count)}")
```
s-urbaniak: The above interpolation (and the following ones) is pretty complex. I suggest we add a comment explaining what this does, especially the `list("padding")` workaround which makes the `element` function happy in case `worker_subnets` is empty and the `cidrsubnet` logic.

modules/aws/vpc/vpc-private.tf
```diff
@@ -9,28 +9,39 @@ resource "aws_route_table" "private_routes" {
...
+  }"
+
+  availability_zone = "${ "${length(keys(var.worker_subnets))}" > 0 ? 
+    "${element(concat(keys(var.worker_subnets), list("padding")), count.index)}" : 
+    "${data.aws_availability_zones.azs.names[count.index]}"
```
s-urbaniak: The same applies here, I suggest we add a doc comment above.

modules/aws/vpc/vpc-private.tf
alexsomesan: My plan is to move these expressions on the module inputs.
It would spread out the verbosity and make the resources read more cleanly.

I will then add documentation on the module inputs, which would all be in one place.

alexsomesan:
I renamed the top-level subnet input variables to explicitly mention "custom".

PR #271, opened 2017-04-20
frontend: Send more tfvars to backend.

ggreer:
Including:
- Storage info (iops, type, size)
- Public/private VPC
- Separate counts for master/worker AZs.
- Add TODOs for the few vars we're not yet passing to the backend.


squat:
It doesn't look like this undoes the decoupling of the worker/controller AZs.

ggreer:
It doesn't. The tfvars changed though, so the frontend has to send the new vars.

PR #273, opened 2017-04-20
Fix .gitignore and add missing vendors

Quentin-M:
Just built on a fresh GOPATH. Worked fine! Sorry about that.


.gitignore
```diff
@@ -6,5 +6,9 @@
 **/ssh-key.pem
 build/**
 config.tfvars
+terraform.tfvars
+pull-secret.json
+tectonic-license.txt
 generated/
 bin/
+
```
squat: extra

.gitignore
```diff
@@ -6,5 +6,9 @@
 **/ssh-key.pem
 build/**
 config.tfvars
+terraform.tfvars
+pull-secret.json
+tectonic-license.txt
 generated/
 bin/
```
ggreer: Won't lines 9 & 10 still match all kinds of stuff that we probably don't want to match?

derekparker:
Just ran into this!

.gitignore
Quentin-M: I think these names are pretty specific?

Quentin-M:
Thanks guys

.gitignore
ggreer: Any directory named "generated" or any directory named "bin"? I could see those existing in vendor.

squat: this takes care of line 3 in installer/.gitignore

installer/.gitignore
```diff
@@ -1,9 +1,7 @@
...
-tools/
-matchbox
-tests/scripts/aws/output/
-scripts/cluster/output/
+/assets/frontend/scripts/
+/bin/
+/binassets/assets.go
+/frontend/node_modules/
+/matchbox
+/tests/scripts/aws/output/
```
squat: I don't think any of these besides `/matchbox` should have the `/` prefix.

PR #269, opened 2017-04-20
Fix syntax error in Jenkinsfile

ethernetdan:
There missing a closing }

PR #276, opened 2017-04-20
Fix jenkinsfile

ethernetdan:
New from #269

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -141,8 +141,9 @@ pipeline {
 
             make apply
             '''
+            }
           }
-        }
+        )
       }
     }
   }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -85,7 +85,8 @@ pipeline {
             make plan
 
             # always cleanup cluster
-            function shutdown() {
+            shutdown()
+            {
               make destroy
             }
             trap shutdown EXIT
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -24,83 +24,38 @@ pipeline {
     }
 
     stage('Installer: Build & Test') {
-      agent none
+      agent {
+        docker {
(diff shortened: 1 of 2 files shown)


Quentin-M:
Answered in DM. This is due to the Kubernetes cluster being named differently in that Jenkins environment.

ethernetdan:
@Quentin-M what does it need to be named? or should I just not set it?

ethernetdan:
FYI, currently set [here](https://github.com/coreos/tectonic-installer/blob/fe15418/Jenkinsfile#L120)

Quentin-M:
It's not that. Jenkins is trying to find the configured cloud provider to launch Kubernetes pod to build the Tectonic Installer. In this Jenkins env, there is no `kubernetes` provider, but there is another one, named differently (by the actual cluster name). Either rename that provider to `kubernetes`, or either configure the Jenkinsfile to use the right name. It is not related to TF or to the variable you pointed out.

Jenkinsfile
```diff
@@ -19,88 +19,34 @@ pipeline {
...
-              containerTemplate(
-                name: 'webapp-agent',
-                image: 'quay.io/coreos/tectonic-builder:v1.4',
-                ttyEnabled: true,
-                command: 'cat',
-              )
-            ],
-            volumes: []
-          ) {
-            node('webapp-pod') {
```
dghubble: Is it intentional you've stopped running these containers on Kubernetes?

ethernetdan: For now until we can get the executors running correctly

ethernetdan:
Looks like this works!

@sym3tri @Quentin-M @dghubble good to merge?

ethernetdan:
Merging to get working validation running, will revisit with @dghubble comments from #207

PR #279, opened 2017-04-20
Default to AWS Terraform in GUI installer.

ggreer:
- Don't require dev mode for terraform in GUI.
- Move AWS Terraform to the top of the dropdown.
- Rename old AWS mode to AWS CloudFormation.

### DIFF
--- a/installer/frontend/platforms.js
+++ b/installer/frontend/platforms.js
@@ -4,26 +4,25 @@ export const AZURE = 'azure';
 export const BARE_METAL = 'bare-metal';
 export const OPENSTACK = 'openstack';
 
-export const isSupported = platform => [AWS, BARE_METAL].includes(platform) ||
-  (platform === AWS_TF && window.config.devMode);
+export const isSupported = platform => [AWS, AWS_TF, BARE_METAL].includes(platform);
 
 export const PLATFORM_NAMES = {
-  [AWS]: 'Amazon Web Services',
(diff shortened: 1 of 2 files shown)


installer/frontend/platforms.js
```diff
@@ -4,26 +4,25 @@ export const AZURE = 'azure';
...
 export const OPENSTACK = 'openstack';
 
-export const isSupported = platform => [AWS, BARE_METAL].includes(platform) ||
-  (platform === AWS_TF && window.config.devMode);
+export const isSupported = platform => [AWS, AWS_TF, BARE_METAL].includes(platform);
 
 export const PLATFORM_NAMES = {
-  [AWS]: 'Amazon Web Services',
-  [BARE_METAL]: 'Bare Metal',
+  [AWS]: 'Amazon Web Services (CloudFormation)',
```
sym3tri: Probably label this as "1.5.6-tectonic.1" too

ggreer: This has to fit in a dropdown, so it can't be much longer than it currently is. I made it 'Amazon Web Services (CloudFormation, legacy v1.5)'

Any better ideas to express to the user "this will install old k8s/terraform"?

sym3tri: SGTM. Abbreviating as `AWS` is also fine if you need more space.

PR #281, opened 2017-04-20
Migrate more manifests to k8s 1,6

squat:
This finishes migrating several tolerations and affinity definitions from annotations to properties in the pod spec.

cc @Quentin-M @aaronlevy


Quentin-M:
Thanks. Should we do that for some services like identity as it was discussed in the past?

squat:
aah, you mean give them master toleration?

PR #285, opened 2017-04-20
frontend: Hook up experimental operators in Terraform GUI.

ggreer:
Also, fix the app id for all platforms.

### DIFF
--- a/installer/frontend/cluster-config.js
+++ b/installer/frontend/cluster-config.js
@@ -175,7 +175,7 @@ export const DEFAULT_CLUSTER_CONFIG = {
   [UPDATER]: {
     server: 'https://tectonic.update.core-os.net',
     channel: 'tectonic-1.5',
-    appID: '',
+    appID: '6bc7b986-4654-4a0f-94b3-84ce6feb1db4',
   },
   [UPDATER_ENABLED]: false,
   [POD_CIDR]: "10.2.0.0/16",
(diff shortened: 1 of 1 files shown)


installer/frontend/cluster-config.js
```diff
@@ -175,7 +175,7 @@ export const DEFAULT_CLUSTER_CONFIG = {
   [UPDATER]: {
     server: 'https://tectonic.update.core-os.net',
     channel: 'tectonic-1.5',
-    appID: '',
+    appID: '6bc7b986-4654-4a0f-94b3-84ce6feb1db4',
```
amrutac: Where does this come from?

installer/frontend/cluster-config.js
```diff
@@ -386,6 +386,7 @@ export const toAWS_TF = (cc, FORMS) => {
       tectonic_worker_count: workers[NUMBER_OF_INSTANCES],
       // TODO: shouldn't hostedZoneID be specified somewhere?
       tectonic_dns_name: cc[CLUSTER_SUBDOMAIN],
+      tectonic_experimental: cc[UPDATER_ENABLED],
```
amrutac: Should there be a separate check if the customer is using the old installer?

installer/frontend/cluster-config.js
yifan-gu: @amrutac This is an hardcoded value in the coreupdate server.

installer/frontend/cluster-config.js
ggreer: Hmm... I'm not sure what you mean. This function (`toAWS_TF()`) is only called if using the terraform installer. The old cloudformation behavior remains the same.

amrutac: Oh OK, just wanted to make sure this doesn't apply for the old cloudformation clusters.

installer/frontend/cluster-config.js
amrutac: OK.

PR #287, opened 2017-04-20
frontend: Hook up custom subnets for AWS Terraform installer.

ggreer:
I'm pretty sure the structure is right. Terraform apply runs successfully and boots a cluster.

### DIFF
--- a/installer/frontend/cluster-config.js
+++ b/installer/frontend/cluster-config.js
@@ -321,6 +321,12 @@ export const toAWS = (cc, FORMS) => {
   return ret;
 };
 
+const toSubnetObj = (subnets, key) => _(subnets)
+  .keyBy(o => o.availabilityZone)
+  .mapValues(v => v[key])
+  .value();
+
+
 export const toAWS_TF = (cc, FORMS) => {
   const controllers = FORMS[AWS_CONTROLLERS].getData(cc);
   const etcds = FORMS[AWS_ETCDS].getData(cc);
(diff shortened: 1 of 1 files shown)


squat:
> Terraform apply runs successfully and boots a cluster.

You should be our CI tool :stuck_out_tongue:

installer/frontend/cluster-config.js
```diff
@@ -331,11 +337,23 @@ export const toAWS_TF = (cc, FORMS) => {
   let workerSubnets;
 
   if (cc[AWS_CREATE_VPC] === 'VPC_CREATE') {
-    controllerSubnets = toVPCSubnet(region, cc[AWS_CONTROLLER_SUBNETS], cc[DESELECTED_FIELDS][AWS_SUBNETS]);
-    workerSubnets = toVPCSubnet(region, cc[AWS_WORKER_SUBNETS], cc[DESELECTED_FIELDS][AWS_SUBNETS]);
+    controllerSubnets = toSubnetObj(
```
squat: we need to start migrating terminology s/controller/master/

installer/frontend/cluster-config.js
```diff
@@ -321,6 +321,12 @@ export const toAWS = (cc, FORMS) => {
   return ret;
 };
 
+const toSubnetObj = (subnets, key) => _(subnets)
+  .keyBy(o => o.availabilityZone)
+  .mapValues(v => v[key])
+  .value();
```
squat: so this creates a map that looks like:
```json
{
    "az1": "172.16.0.0/27",
    "az2": "172.16.0.32/27",
    ...
}
```

ggreer: Correct. That's the format that terraform wants.

installer/frontend/cluster-config.js
```diff
@@ -321,6 +321,12 @@ export const toAWS = (cc, FORMS) => {
   return ret;
 };
 
+const toSubnetObj = (subnets, key) => _(subnets)
```
squat: I almost think we should hardcode `instanceCIDR` rather than pass it as an argument to `toSubnetObj`. How abstract is this function really?

ggreer: Sometimes 'instanceCIDR' is passed in, sometimes 'id' is. This code will become cleaner once we remove cloudformation stuff (which needs the different keys).

PR #291, opened 2017-04-21
frontend: Add reset button on bootup page. 

ggreer:
Only show it if terraform isn't running.

This prevents users from getting stuck on the bootup page if terraform apply fails.


Quentin-M:
omg thanks

PR #292, opened 2017-04-21
frontend: Updates console to v1.4.0

amrutac:
https://github.com/coreos-inc/bridge/releases/tag/v1.4.0


squat:
Awesome! Can you also bump in the go installer? https://github.com/coreos/tectonic-installer/blob/master/installer/assets/console-deployment.yaml#L91

sym3tri:
@squat we are no longer maintaining that codepath and it will be removed soon

Quentin-M:
Thanks for the release, and for the quick tag before! 💯

PR #298, opened 2017-04-21
Makefile configures and uses installer plugins

alexsomesan:
This integrates the new pluginin-vending installer binary into the make based workflow.

After this change, `make [apply | destroy | plan]` will ensure that terraform is run with the correct plugin configuration.

/cc: @Quentin-M


alexsomesan:
That was fast! Thanks ;)

PR #304, opened 2017-04-22
Ensure bin directory exist before building tools

ethernetdan:
Fixes bug from #276

@Quentin-M @dghubble @sym3tri @squat  PTAL

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -37,7 +37,7 @@ pipeline {
         go version
         cd $GO_PROJECT/installer
 
-        echo "Sanity testing temporarily disabled" && touch bin/sanity
+        echo "Sanity testing temporarily disabled" && mkdir bin && touch bin/sanity
 
         make tools
         make build


ethernetdan:
Looks like this is working the failure is based on a quota issue:
```
aws_iam_instance_profile.worker_profile: Error creating IAM instance profile tf-aws-pr-304-2-worker-profile: LimitExceeded: Cannot exceed quota for InstanceProfilesPerAccount: 100
```

I'll switch to another AWS account which should fix the issue.

ethernetdan:
Fresh checks look good.

Jenkinsfile
```diff
@@ -37,7 +37,7 @@ pipeline {
         go version
         cd $GO_PROJECT/installer
 
-        echo "Sanity testing temporarily disabled" && touch bin/sanity
+        echo "Sanity testing temporarily disabled" && mkdir bin && touch bin/sanity
```
sym3tri: maybe add a `-p` just to be safe?

squat: -p should only be needed is workspace is not clean and bin exists. In that case, if the dir already exists then medir exits non zero and cause the whole script (with set -e) to fail. also, touch will fail if bin/sanity already exists

PR #305, opened 2017-04-23
installer/server/terraform/executor: add Zip() method w/ symlink support

### LINKED ISSUE #297: Installer UI: assets.zip cannot be extracted
I went through the UI wizard to test the installer. At the end, I wanted to download the assets.

The zip file that I received will not extract cleanly.

Log attached: [assets-unzip.txt](https://github.com/coreos/tectonic-installer/files/947658/assets-unzip.txt)

/cc: @sym3tri @Quentin-M

Quentin-M:
installer/server/terraform/executor: add Zip() method w/ symlink support  …
The library we were using does not support symlinks at all: creating
symlink headers with actual file contents, thus breaking unarchivers.

This new zipper preserve symlinks as long as their target will be part
of the generated archive. Otherwise, they are dropped and warnings are
issued.

Fixes #297


Quentin-M:
Failed to mention that the archiver we were using was unsafe to use against folders that may contain socket, named pipes, etc.

PR #314, opened 2017-04-24
Configure audit logging for kube-apiserver

chancez:
Adds audit related flags to the API server manifests for logging to a file in a host volume. I did not update the bootstrap API manifest because it seems unnecessary.

It is expected you would configure fluentd to tail this log file. 

An alternative approach is to make the API server container use `/dev/stdout` as the path to the logfile. This makes it harder and less efficient to configure capturing logs, since the streams will become interleaved, and the API server logs and the audit logs have different formats. This is the simplest approach currently.

If we wish to not use a host mount and not use `/dev/stdout`, then we could avoid it by running a sidecar pod which tails and prints the log file from an empty volume, but then pod restarts will cause some logs to get lost, and isn't really much better than a host mount. This is my least favorite of the options.

### DIFF
--- a/modules/bootkube/resources/manifests/kube-apiserver.yaml
+++ b/modules/bootkube/resources/manifests/kube-apiserver.yaml
@@ -54,6 +54,10 @@ spec:
         - --oidc-groups-claim=${oidc_groups_claim}
         - --oidc-ca-file=/etc/kubernetes/secrets/ca.crt
         - --cloud-provider=${cloud_provider}
+        - --audit-log-path=/var/log/kube-apiserver-audit.log
+        - --audit-log-maxage=30
+        - --audit-log-maxbackup=3
+        - --audit-log-maxsize=100
         volumeMounts:
         - mountPath: /etc/ssl/certs
           name: ssl-certs-host
@@ -63,6 +67,9 @@ spec:
           readOnly: true
         - mountPath: /var/lock
           name: var-lock
+          readOnly: false
+        - mountPath: /var/log
+          name: var-log
           readOnly: false
       hostNetwork: true
       tolerations:
@@ -83,3 +90,6 @@ spec:
       - name: var-lock
         hostPath:
           path: /var/lock
+      - name: var-log
+        hostPath:
+          path: /var/log


chancez:
cc @ericchiang and @Quentin-M for review

s-urbaniak:
`installer/assets/self-hosted/kube-apiserver.yaml.tmpl` is being removed in https://github.com/coreos/tectonic-installer/pull/325/files#diff-c3edf5721bd83571b7d1538d17355da3, hence I don't think it is necessary to be included here.

s-urbaniak:
Generally this LGTM modulo the obsolete change to `installer/assets/self-hosted/kube-apiserver.yaml.tmpl`. `--audit-log-path` is the canonical way for doing audit logging of the api-server, so we should leverage that.

I agree with the reasoning of @chancez of not using a side-car.

chancez:
Alright, I'll remove that change and rebase to fix conflicts, and we can have this depend on #325

chancez:
Alright, removed the modification to the obsolete file + rebased.

Quentin-M:
Have we verified that the rotation works as expected?

Quentin-M:
Shouldn't the bootstrapped API server log too? It could be exploited.

chancez:
@Quentin-M rotation works as expected. I asked @sym3tri and @aaronlevy about the bootstrap API being logged, but they seemed to think we don't really need it. I'd be fine with adding it, but I'm unclear whether or not the file locking they do is sufficient to ensure they're not both trying to write to the audit log file at the same time. It might mean the bootstrap API pod needs to write it's audit log entries to a different file, and I could update the config to look at that file as well.

Quentin-M:
It's fine, we will see if customers request that, probably not! Thanks for your answer.

aaronlevy:
> I asked @sym3tri and @aaronlevy about the bootstrap API being logged, but they seemed to think we don't really need it.

I wasn't approached about this -- but I also don't have any strong feelings if it is enabled or not. 

`flock` should be sufficiently safe - but there's always a chance the implementation changes at a later date.

PR #324, opened 2017-04-24
scripts: Don't require S3 to make a tarball.

ggreer:
- Don't check AWS creds at the start of make_release_tarball.sh.
- Copy local installer binaries to release dir instead of grabbing them from s3.
- Move terraform-specific env vars to get_terraform_bins.sh. They're only used in this one place.
- Use straightforward string interpolation instead of eval() to construct terraform binary URLs.

### DIFF
--- a/installer/scripts/release/get_terraform_bins.sh
+++ b/installer/scripts/release/get_terraform_bins.sh
@@ -8,8 +8,8 @@ mkdir -p "${TMP_DIR}" "${TERRAFORM_BIN_TMP_DIR}"
 set -x
 
 for os in "linux" "darwin" "windows"; do
-    curl -L $(eval echo ${TERRAFORM_BIN_URL}) -o ${TERRAFORM_BIN_TMP_DIR}/terraform_${os}.zip
-    mkdir -p ${INSTALLER_RELEASE_DIR}/${os}/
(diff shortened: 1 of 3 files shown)


installer/scripts/release/get_terraform_bins.sh
```diff
@@ -3,13 +3,17 @@
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 source "$DIR/common.env.sh"
 
+TERRAFORM_BIN_TMP_DIR="$TMP_DIR/terraform-bin"
+TERRAFORM_BIN_VERSION=0.8.8
+TERRAFORM_BIN_BASE_URL="https://releases.hashicorp.com/terraform/${TERRAFORM_BIN_VERSION}/terraform_${TERRAFORM_BIN_VERSION}"
```
Quentin-M: I think that regardless of whether they are used once or not, we were having them in common.env.sh to ease bumping.

Quentin-M:
@dghubble There used to be a reason we were uploading the build-artifacts binaries to S3, right?

sym3tri:
> There used to be a reason we were uploading the build-artifacts binaries to S3, right? Is it still the case?

@Quentin-M no, that's legacy from a long time ago. No longer needed.

ggreer:
Alright, vars moved back to common.

Quentin-M:
@ggreer did you make a fake release to test your changes?

ggreer:
I did. It works.

Quentin-M:
THANKS!

PR #326, opened 2017-04-24
frontend: Upgrade JS dependencies.

ggreer:
- Use 1.x notation instead of ^1.0 notation (which lots of people seem to get confused about).
- Bump patch-level of most outdated dependencies.
- Run `yarn upgrade`

There's no need to pin the version in package.json now that we're using yarn.lock. `yarn install` uses the yarn.lock file. `yarn upgrade` looks in package.json and updates yarn.lock with the latest available version.


ggreer:
And yes, this does result in a working installer GUI :)

sym3tri:
While you're in there... mind updating the README with whatever is required to get it running? NPM/Node versions etc.

ggreer:
Already in: https://github.com/coreos/tectonic-installer/tree/master/installer#frontend

sym3tri:
oops. thanks!

PR #328, opened 2017-04-25
Move make destroy to post in Jenkinsfile

ethernetdan:
This will ensure that clusters are always destroyed.


ethernetdan:
Need to make terraform state accessible to the `post` step

ethernetdan:
Moving to fork PR

PR #333, opened 2017-04-25
Update testing environment for AWS

ethernetdan:
Update configuration for new testing environment

### DIFF
--- a/test/aws.tfvars
+++ b/test/aws.tfvars
@@ -6,7 +6,7 @@ tectonic_etcd_count = "3"
 
 tectonic_etcd_servers = [""]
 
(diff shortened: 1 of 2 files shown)


test/aws.tfvars
```diff
@@ -6,7 +6,7 @@ tectonic_etcd_count = "3"
 
 tectonic_etcd_servers = [""]
 
-tectonic_base_domain = "staging.core-os.net"
+tectonic_base_domain = "tectonic.dev.coreos.systems"
```
alexsomesan: This account is going to be a fun ride to clean up :)
But I guess it's the best we have so far.

PR #294, opened 2017-04-21
Bump monitoring setup manifests

fabxc:
@s-urbaniak

PR #336, opened 2017-04-25
Bump monitoring setup manifests

brancz:
Super-seeds #294 

With these changes everything should work except for the number of Pods graph on the nodes pages, as this is a change that has to be done in console/bridge.

@s-urbaniak @alexsomesan @fabxc


s-urbaniak:
No objections from my side if we manage to bump the console until the release ;)

sym3tri:
@brancz what needs to be changed in Console?

s-urbaniak:
@sym3tri a hint was given in https://github.com/coreos-inc/tectonic-issues/issues/232#issuecomment-297080919

modules/tectonic/resources/manifests/monitoring/prometheus-k8s.yaml
```diff
@@ -0,0 +1,25 @@
...
+  namespace: tectonic-system
+  labels:
+    prometheus: k8s
+spec:
+  replicas: 1
+  version: ${prometheus_version}
+  serviceAccountName: prometheus-k8s
+  serviceMonitorSelector:
+    matchExpression:
+    - {key: k8s-apps, operator: Exists}
```
chancez: So this means that if someone else wanted promtheus to monitor their app, they would need to have a label `k8s-apps`, even if its not a k8s app?

brancz:
Thanks for the quick and smooth collaboration everyone! :tada:

PR #340, opened 2017-04-25
Expose cluster ID to all smoke tests

ethernetdan:
Fixes smoke test not knowing CLUSTER during `post`


ethernetdan:
Looks like smoke didn't pass, the platform isn't being exposed. I'll send a fix.

PR #342, opened 2017-04-25
[WIP] Fix teardown

ethernetdan:
Fixes issues from #340

PR #343, opened 2017-04-25
Revert using post block until cleanup script is created

ethernetdan:
Until #342 is ready and merged lets go back to master passing.

I'm going to add protection to the master branch after a few stable runs.


ethernetdan:
Since failure is resource related and cleanup logic worked merging on bad build.

PR #344, opened 2017-04-25
installer/scripts/release: bump matchbox to v0.6.0

Quentin-M:
As documented, the minimum required version of matchbox for our Tectonic usage is v0.6.0.

Ref: https://github.com/coreos/matchbox/releases/tag/v0.6.0


Quentin-M:
@dghubble

dghubble:
How about we stop re-packaging matchbox inside Tectonic releases? The official releases contain the same contents, plus contents being stripped by tectonic and they provide a more uniform entrypoint to support. I want it to be clear that matchbox is an open-source, infrastructure service/tool. 

https://github.com/coreos/matchbox/releases
cc @sym3tri

dghubble:
For now, let's merge just so Tectonic isn't including the **wrong** version. But I'm in favor of pointing folks at the upstream releases.

PR #359, opened 2017-04-25
baremetal-tf GUI

ggreer:
TODO:
~~- terraform doesn't support multiple ssh keys. need to change the GUI to disallow that~~
~~- lots of copy-pasted code that can be generalized~~
~~- doesn't actually work yet. appears to be a backend bug~~

### DIFF
--- a/installer/frontend/cluster-config.js
+++ b/installer/frontend/cluster-config.js
@@ -1,6 +1,6 @@
 import _ from 'lodash';
 
-import { BARE_METAL } from './platforms';
+import { BARE_METAL, BARE_METAL_TF } from './platforms';
 import { keyToAlg } from './utils';
 
 let defaultPlatformType = '';
@@ -101,7 +101,7 @@ export const toVPCSubnet = (region, subnets, deselected) => {
 };
 
 const getZoneDomain = (cc) => {
-  if (cc[PLATFORM_TYPE] === BARE_METAL) {
+  if (_.includes([BARE_METAL, BARE_METAL_TF], cc[PLATFORM_TYPE])) {
     throw new Error("Can't get base domain for bare metal!");
   }
   // TODO: if we ever change toExtraData()'s key, this breaks
@@ -109,14 +109,14 @@ const getZoneDomain = (cc) => {
 };
 
 export const getControllerDomain = (cc) => {
-  if (cc[PLATFORM_TYPE] === BARE_METAL) {
+  if (_.includes([BARE_METAL, BARE_METAL_TF], cc[PLATFORM_TYPE])) {
     return cc[CONTROLLER_DOMAIN];
   }
   return `${cc[CLUSTER_SUBDOMAIN]}-k8s.${getZoneDomain(cc)}`;
 };
 
 export const getTectonicDomain = (cc) => {
-  if (cc[PLATFORM_TYPE] === BARE_METAL) {
+  if (_.includes([BARE_METAL, BARE_METAL_TF], cc[PLATFORM_TYPE])) {
     return cc[BM_TECTONIC_DOMAIN];
   }
   const tectonicDomain = cc[CLUSTER_SUBDOMAIN] + (cc[CLUSTER_SUBDOMAIN].endsWith('.') ? '' : '.') + getZoneDomain(cc);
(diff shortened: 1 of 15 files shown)


dghubble:
Trying out the updated bare-metal GUI. Looks nice.

Few things:

* matchbox_http_endpoint should have an `http://` protocol (matchbox_rpc_endpoint does not). Machines cannot boot currently
* UI is defaulting service CIDR to 10.3.0.0/16, we'd like it to be 10.3.0.0/16 (all platforms, minor, see config.tf)
* Removing the Power On screen, there is no indication the user should power on hardware
  * Quick fix for now: We could edit the text "Kubernetes is starting up, we're committing..." text
* Start over should probably prompt, but out of scope for this PR

installer/frontend/cluster-config.js
```diff
@@ -436,3 +436,51 @@ export const toAWS_TF = (cc, FORMS) => {
...
+      tectonic_metal_cl_version: cc[BM_OS_TO_USE],
+      tectonic_metal_ingress_domain: getTectonicDomain(cc),
+      tectonic_metal_controller_domain: getControllerDomain(cc),
+      tectonic_metal_controller_domains: cc[BM_MASTERS].map(({name}) => name),
+      tectonic_metal_controller_names: cc[BM_MASTERS].map(({name}) => name.split('.')[0]),
+      tectonic_metal_controller_macs: cc[BM_MASTERS].map(({mac}) => mac),
+      tectonic_metal_worker_domains: cc[BM_WORKERS].map(({name}) => name),
+      tectonic_metal_worker_names: cc[BM_WORKERS].map(({name}) => name.split('.')[0]),
+      tectonic_metal_worker_macs: cc[BM_WORKERS].map(({mac}) => mac),
+      tectonic_metal_matchbox_http_endpoint: cc[BM_MATCHBOX_HTTP],
```
dghubble: Before, I think the backend was probably adding http:// + endpoint

ggreer:
Alright. I'm pretty sure I've fixed the blockers to booting a cluster and the rest are just UI tweaks.

dghubble:
Hm, still doesn't seem to work. The GUI-based apply can't seem to SSH to the nodes, though I can directly or by running terraform apply myself.

ggreer:
I think the fix is to pass ssh-related environment variables along when executing terraform. eg: `SSH_AUTH_SOCK`

ggreer:
After reading the ssh manpage, we should pass the following environment vars through: DISPLAY, PATH, SSH_ASKPASS, SSH_AUTH_SOCK, and maybe SSH_ORIGINAL_COMMAND, SSH_CONNECTION, and SSH_TTY.

It's probably best if we pass DISPLAY & PATH through, and anything that starts with "SSH_".

dghubble:
Yeah, I had this working with SSH_AUTH_SOCK passthrough.

installer/server/terraform/executor.go
```diff
@@ -188,6 +188,14 @@ func (ex *Executor) Execute(args ...string) (int, chan struct{}, error) {
 	// the right place), extra environment variables and outputs.
 	cmd := exec.Command(ex.binaryPath, args...)
 	cmd.Env = append(cmd.Env, fmt.Sprintf("TERRAFORM_CONFIG=%s", ex.configPath))
+	// ssh changes its behavior based on these. pass them through so ssh-agent & stuff works
```
Quentin-M: I would really replace all these by:

```
cmd.Env = os.Environ()
cmd.Env = append(cmd.Env, fmt.Sprintf("TERRAFORM_CONFIG=%s", ex.configPath))
for k, v := range ex.envVariables {
  cmd.Env = append(cmd.Env, fmt.Sprintf("%s=%s", strings.ToUpper(k), v))
}
```

Seems way saner.

ggreer: It's not clear whether we want to import all env vars. Eg: My personal AWS creds or region, or terraform-related environment vars that the user might happen to have set.

I erred on the side of safety and only passed-through variables that I needed.

Quentin-M: They might be set, it should not really matter if they are in our use-case- they will be overwritten by `ex.envVariables` if necessary. I don't want to end up with 15 lines of obscure/barely documented env passthroughs that we will never be able to get rid of. I'd rather either have them all passed, or at least have a non-exported global variable that will whitelist them, with the exact documentation of which provider / use-case they are for.

dghubble: For now, I think we should pass just what is needed to fix terraform SSH execution and get this PR to work.

dghubble:
Successful run through the GUI
Caveats:
* User must know to provide a local SSH pubkey and power on machines after terraform apply starts
* Must choose create etcd for me, bring your own etcd isn't supported yet

screenshot from 2017-04-28 11-45-32
screenshot from 2017-04-28 16-39-16
screenshot from 2017-04-28 16-42-12

installer/server/terraform/executor.go
ggreer: Passing all environment variables would practically guarantee breakage for some users in the future. I just tried your patch with TF_LOG and TF_LOG_PATH set in my zshrc, and I didn't see some logs in the installer GUI.

Even if we blacklisted those variables today, Terraform can add new vars that change its behavior without us knowing. We'd have to play whack-a-mole blacklisting them whenever we updated terraform. Likely, we'd only notice when some user complained about the installer not working for them.

PR #361, opened 2017-04-25
platforms/metal: Template config.tf versions into CL configs

dghubble:
cc @aaronlevy might simplify your v.1.6.2 bump.


Quentin-M:
go for it

PR #366, opened 2017-04-25
resources: add container-linux-updater

euank:
Mask locksmithd as well.

Marked do-not-merge for now because it uses my personal test build of CLUO, pending a merge+release there.

### DIFF
--- a/modules/aws/ignition/ignition.tf
+++ b/modules/aws/ignition/ignition.tf
@@ -30,13 +30,7 @@ data "ignition_systemd_unit" "docker" {
 
 data "ignition_systemd_unit" "locksmithd" {
   name = "locksmithd.service"
-
-  dropin = [
-    {
-      name    = "40-etcd-lock.conf"
-      content = "[Service]\nEnvironment=REBOOT_STRATEGY=${var.locksmithd_disabled ? "off" : "etcd-lock"}\n"
-    },
-  ]
+  mask = true
 }
 
 data "template_file" "kubelet" {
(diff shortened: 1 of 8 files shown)


euank:
I've updated the tag to the actual release tag

I've also masked locksmithd in a bunch more places.

The intent is that it's only run on the etcd nodes now (since they're not part of k8s, there's not much I can do about them).

euank:
cc @dghubble @ethernetdan as possible reviewers I guess?

dghubble:
I think you missed these https://github.com/coreos/tectonic-installer/tree/master/platforms/metal/cl if it was a goal to do this for bare-metal.

euank:
@dghubble added it there too and rebased

s-urbaniak:
General question: Is this feature supposed to be gated by `tectonic_experimental` or is the update operator to be on by default?

euank:
@s-urbaniak That is intentional. The reasoning basically goes "we need either locksmithd or cluo to be in any way safe, and migrating between them is a little tricky (they're mutually exclusive) so making it optional means more work for us later".

If there's good reason to make it experimental, we can, but that means that we'll have to do extra work to be able to migrate it to stable in the future.

I personally am fine with either and willing to do the "more work" if it's necessary, so I'm leaving the call up to @sym3tri

dghubble:
Are you sure this works? Looking at that image, it seems to be private and doesn't include licensed Tectonic users so I'd expect pulls to fail across all platforms.

euank:
@dghubble the repo's public?

`docker pull quay.io/coreos/container-linux-update-operator:v0.2.0` works fine for me on a non-logged-in-machine.

dghubble:
Thanks, testing for bare-metal.

PR #367, opened 2017-04-25
Ignore AMI updates on etcd/master nodes to prevent destroying nodes accidently

chancez:
Currently if a new version of CoreOS comes out on a user's configured
channel after they've already deployed, and they re-run terraform apply,
Terraform will detect the new AMI, and when reconciling desired state,
it will attempt to destroy existing nodes to update the AMI.

In order to avoid this, I added a ignore_changes to both the etcd node
resource, and the master launch configuration resource, to avoid
updating masters/etcd nodes if the AMI changes.

Longer term, I would expect the Container Linux Update Operator, or a
terraform operator could resolve this, but this seems like a reasonable
change to prevent accidental destruction of a cluster.

Previously this change also set prevent_destroy, but that cannot be
parameterized with vars, so it was removed.

### DIFF
--- a/modules/aws/etcd/nodes.tf
+++ b/modules/aws/etcd/nodes.tf
@@ -32,6 +32,10 @@ resource "aws_instance" "etcd_node" {
   user_data              = "${ignition_config.etcd.*.rendered[count.index]}"
   vpc_security_group_ids = ["${var.sg_ids}"]
 
+  lifecycle {
+    ignore_changes = ["ami"]
+  }
+
   tags = "${merge(map(
       "Name", "${var.cluster_name}-etcd-${count.index}",
       "KubernetesCluster", "${var.cluster_name}"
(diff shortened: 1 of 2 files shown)


chancez:
I was going to make changes to parameterize the `force_destroy`, but ran into issues and found this:

https://github.com/hashicorp/terraform/issues/10730

It seems there's no way to easily prevent destruction of a resource during apply, but allow for it during `destroy` today. I'll remove the `force_destroy` bit, as the `ignore_changes` should handle the case I was seeing, but it would be good if we could find a way to prevent accidental deletion of etcd nodes on a `terraform apply`, as I've done so a few times now by accident, even after changing nothing in my vars.

alexsomesan:
I agree with @s-urbaniak. It's a good fix.
Users aren't left out in the dark past anyway if they keep CL updates enabled.
Let keep doing this until we have a more solid story around it based on Container Linux Update Operator or the like.

LGTM.

sym3tri:
Why not add to workers as well?

s-urbaniak:
@sym3tri good catch! @chancez do you mind to submit another PR which includes workers?

chancez:
I figured I'd keep the scope small since it's a no brainer for etcd and masters, but wasn't sure about workers. I'd be glad to make another PR with the same change.

PR #369, opened 2017-04-26
config.tf - Update Tectonic Update Server Variable

alekssaul:
Replace CoreOS Container Linux Update server default variable `https://public.update.core-os.net` with Tectonic Update server default variable `https://tectonic.update.core-os.net`. This fixes with the issue where updates were not available in Tectonic Console

### DIFF
--- a/config.tf
+++ b/config.tf
@@ -154,7 +154,7 @@ EOF
 
 variable "tectonic_update_server" {
   type        = "string"
-  default     = "https://public.update.core-os.net"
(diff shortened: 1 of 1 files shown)


alekssaul:
As far as I can see all platform modules set locksmithd to use etcd-lock without making any changes to CL coreupdate server. Variable in the PR is used by the Tectonic Operators.

alekssaul:
@sym3tri can't merge due to `continuous-integration/jenkins/pr-merge — This commit cannot be built`

PR #372, opened 2017-04-26
Make target to import cluster from UI generated assets

alexsomesan:
This enables the CLI build system to consume assets produced by the UI installer.

Users will be able to boot a cluster using the UI installer, then download the assets.zip and continue to manage their cluster using the CLI flows (make apply / make destroy).

Among other things, this allows for an easy way to destroy clusters created with the UI after the installer has been shut down.

To import assets created by the UI, a user would use the following command:
```
make init FROM_ASSETS=~/Downloads/assets-test.zip
```
Which then outputs the necessary environment for CLI operations:
```
...
  inflating: /Users/alex/go/src/github.com/coreos/tectonic-installer/build/tectonic212910865/pull_secret.json
  inflating: /Users/alex/go/src/github.com/coreos/tectonic-installer/build/tectonic212910865/terraform.tfstate
  inflating: /Users/alex/go/src/github.com/coreos/tectonic-installer/build/tectonic212910865/terraform.tfvars
...

Done!
Your assets for cluster tectonic212910865 are imported.
Now just use the cluster name variable and you are good to go.

CLUSTER=tectonic212910865 make plan / apply / destroy

Happy applying!
```

I will follow up with a document describing in more details the interaction flows a user would take through the UI and then CLI for the most common cluster lifecycle operation.

NOTE: the new `init` target replaces the old `localconfig` target, but can still create a empty build location if not passed an assets bundle.

/cc: @sym3tri @Quentin-M @s-urbaniak

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,8 +4,8 @@
 config.tfvars
 id_rsa*
 pull-secret.json
-ssh-key.pem
 tectonic-*.tar.gz
 tectonic-license.txt
 terraform.tfstate*
 terraform.tfvars
+build/**
--- a/Makefile
+++ b/Makefile
@@ -7,19 +7,32 @@ INSTALLER_BIN = $(TOP_DIR)/installer/bin/$(shell uname | tr '[:upper:]' '[:lower
 TF_DOCS := $(shell which terraform-docs 2> /dev/null)
 TF_CMD = TERRAFORM_CONFIG=$(TOP_DIR)/.terraformrc terraform
 
-$(info Using build directory [${BUILD_DIR}])
-
 all: apply
 
 $(INSTALLER_BIN):
 	make build -C $(TOP_DIR)/installer
 
-installer-env: $(INSTALLER_BIN) terraformrc.example	
-	sed "s|<PATH_TO_INSTALLER>|$(INSTALLER_BIN)|g" terraformrc.example > .terraformrc
+installer-env: $(INSTALLER_BIN) terraformrc.example
+	$(info Using build directory [${BUILD_DIR}])
+	@sed "s|<PATH_TO_INSTALLER>|$(INSTALLER_BIN)|g" terraformrc.example > .terraformrc
 
-localconfig:
+init:
+ifdef FROM_ASSETS
+	$(eval CLUSTER = $(shell unzip -Z1 $(FROM_ASSETS) | grep -Eo '^[^/\\]+' | sort -u))
+	$(eval BUILD_DIR = $(TOP_DIR)/build/$(CLUSTER))
+	@unzip -d $(TOP_DIR)/build $(FROM_ASSETS)
+	@echo 
+	@echo Done!
+	@echo Your assets for cluster $(CLUSTER) are imported.
+	@echo Now just use the cluster name variable and you are good to go.
+	@echo 
+	@echo CLUSTER=$(CLUSTER) make plan / apply / destroy
+	@echo 
+	@echo Happy applying!
+else
 	mkdir -p $(BUILD_DIR)
 	touch $(BUILD_DIR)/terraform.tfvars
+endif
 
 $(BUILD_DIR)/.terraform:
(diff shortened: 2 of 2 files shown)


dghubble:
I'd expect users to download installer release binaries and place them somewhere locally - its unlikely they'll have this repo setup in a Go/Yarn environment (implied by your target dependency). You might think about this as a developer flow. Still useful, but not necessarily the user entrypoint.

robszumski:
This is great for folks developing/contributing to the repo. UX wise, I'd like to move away from requiring `make` for end-users/customers.

For two reasons:
1. You need make, yarn, npm, whatever. Greatly increases friction, especially across platforms.
2. Using the raw terraform commands teaches you how they work, and increases confidence that your existing knowledge of terraform applies to this product. We're using upstream as much as possible, follow community best practices, etc.

I'm going to work on documenting the desired workflows for devs vs end users.

Makefile
```diff
@@ -7,19 +7,32 @@ INSTALLER_BIN = $(TOP_DIR)/installer/bin/$(shell uname | tr '[:upper:]' '[:lower
...
-localconfig:
+installer-env: $(INSTALLER_BIN) terraformrc.example
+	$(info Using build directory [${BUILD_DIR}])
+	@sed "s|<PATH_TO_INSTALLER>|$(INSTALLER_BIN)|g" terraformrc.example > .terraformrc
+
+init:
+ifdef FROM_ASSETS
+	$(eval CLUSTER = $(shell unzip -Z1 $(FROM_ASSETS) | grep -Eo '^[^/\\]+' | sort -u))
+	$(eval BUILD_DIR = $(TOP_DIR)/build/$(CLUSTER))
+	@unzip -d $(TOP_DIR)/build $(FROM_ASSETS)
```
Quentin-M: Do you think you could make it so we only unzip once? Can we use the `cmd1 | tee >(cmd2) >(cmd3) | cat` trick maybe? Not so sure because Make makes things weird.

.gitignore
```diff
@@ -4,8 +4,8 @@
 config.tfvars
 id_rsa*
 pull-secret.json
-ssh-key.pem
 tectonic-*.tar.gz
 tectonic-license.txt
 terraform.tfstate*
 terraform.tfvars
+build/**
```
ggreer: You don't need to use a globbing rule here. `/build/` will ignore the build directory. `build/**` will ignore things you might not want to. eg: `vendor/whatever/build/something`

alexsomesan:
Rob, Dalton,

I agree with you both. Make is not the best user experience we can provide.

This change was mearly intended to fill the gap in the workflow between was is the current CLI experience (make based) and the UI flow and the artifacts that need to be presisted between the two.

Looking forward to contributing to the doc Rob proposed, document the UX flow. It's what I mentioned to Ed earlier today that I wanted to do myself. Rob has the better skills to do that.

Ideally we should provide an experience where users are not shielded away from Terrafom itself, like Rob explained, but are also not supposed to do housekeeping like writing the terrafomrc.

All proposals are welcome, as long as we can produce a simpler experience.

Makefile
alexsomesan: It's pretty ugly as it is. Actually `unzip -Z1` only outputs the archive index (zipinfo) so it's not a full-weight unzip operation. Doesn't really feel like a bottleneck. But maybe we can get rid of this altogether (like discussed here).

.gitignore
alexsomesan: good point! i'll fix that.

Quentin-M: yes, we had so much trouble with vendors and that.

alexsomesan:
I'm going to drop this PR since there seems to be consensus (to which I subscribe) that a `make` based CLI is not the best UX we can provide to users. We should strive for the better option.

We still need to bridge the gap between UI and CLI for common cluster lifecycle operations, not just for destroy. Issue #552 is tracking that conversation, so no need to follow up here.

PR #374, opened 2017-04-26
WIP: Fixes existing resource group and existing dns zone.

coresolve:
- In testing we found that setting `tectonic_azure_dns_resource_group`
had no effect.
- In this existing code `tectonic_base_domain` is used as the zone name
when associating host records with the zone. This was also broken.
- It seems that when things are set here dns would be resolved the last
item in the zone list from `az resource resource list -g tectonic-dns-group
--resource-type Microsoft.Network/dnszones --query [].name`
After this change it works as expected.

- TODO: introduce logic to set lifecycle to prevent deletion if the dns
zone already exists. Not sure how to test for that without a data
source.
https://www.terraform.io/docs/configuration/resources.html#prevent_destroy

### DIFF
--- a/modules/azure/dns/etcd.tf
+++ b/modules/azure/dns/etcd.tf
@@ -5,6 +5,4 @@ resource "azurerm_dns_a_record" "etcd" {
   name    = "${var.cluster_name}-etcd"
   ttl     = "60"
   records = ["${var.etcd_ip_addresses}"]
-
-  count = "${var.create_dns_zone == "true" ? 1 : 0}"
 }
--- a/modules/azure/dns/main.tf
+++ b/modules/azure/dns/main.tf
@@ -1,5 +1,4 @@
 resource "azurerm_dns_zone" "tectonic_azure_dns_zone" {
   name                = "${var.base_domain}"
   resource_group_name = "${var.resource_group_name}"
-  count               = "${var.create_dns_zone == "true" ? 1 : 0}"
 }
--- a/modules/azure/dns/master.tf
+++ b/modules/azure/dns/master.tf
@@ -5,8 +5,6 @@ resource "azurerm_dns_a_record" "tectonic-api" {
   name    = "${var.cluster_name}-k8s"
   ttl     = "60"
   records = ["${var.master_ip_addresses}"]
-
-  count = "${var.create_dns_zone == "true" ? 1 : 0}"
 }
 
 resource "azurerm_dns_a_record" "tectonic-console" {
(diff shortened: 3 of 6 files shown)


metral:
To add some more context as to what we were trying to achieve that this fix was needed:

- We wanted to make use of a custom, existing DNS zone in Azure's DNS resource group to hold our record sets
- And we wanted to make use of the custom base domain for the DNS resource group, as the current options in `tfvars` were not working. Specifically, the custom base domain set was not being used, and instead the conditional was using the FQDN's (e.g. `module.masters.ingress_internal_fqdn`, `module.masters.api_internal_fqdn` etc.) which the output seems to be setting to the last item in the `az resource resource list -g tectonic-dns-group --resource-type Microsoft.Network/dnszones --query [].name` as Duff noted, not the base domain we want it to be.

coresolve:
Just found that the state file still addresses things in the fallthru zone... 
```
dcooley@lynx ~/go/src/github.com/coreos/tectonic-installer
$ grep eastus.cloudapp terraform.tfstate
                            "fqdn": "tectonic-cluster-dcooley-etcd.eastus.cloudapp.azure.com",
                    "value": "tectonic-cluster-dcooley-k8s.eastus.cloudapp.azure.com"
                    "value": "tectonic-cluster-dcooley-k8s.eastus.cloudapp.azure.com"
                    "value": "tectonic-cluster-dcooley.eastus.cloudapp.azure.com"
                    "value": "tectonic-cluster-dcooley.eastus.cloudapp.azure.com"
                            "fqdn": "tectonic-cluster-dcooley-k8s.eastus.cloudapp.azure.com",
                            "fqdn": "tectonic-cluster-dcooley.eastus.cloudapp.azure.com",
git: fix_dns_module [?]
```

s-urbaniak:
/cc @sozercan 

I will piggy-pack on this PR to fix the currently broken integration with the etcd operator.

s-urbaniak:
Oh I just realized, this has the `do-not-merge` label set, please therefore ignore my previous comment.

sozercan:
Talked with @coresolve offline yesterday, his issue was leaving `tectonic_azure_create_dns_zone` as `false` which ignores `tectonic_base_domain` and just uses built-in Azure FQDN as intended. If you set `tectonic_azure_create_dns_zone`  as `true` it should use `tectonic_base_domain`.

metral:
I opened a new PR: https://github.com/coreos/tectonic-installer/pull/417 to further address the issues discussed here.

This PR can probably be closed, but I defer to @coresolve

metral:
Closing this PR, now that https://github.com/coreos/tectonic-installer/pull/417 has been merged.

PR #375, opened 2017-04-26
Ignore AMI changing on worker launch config

chancez:
Prevents terraform from recreating the resource when a new CoreOS release
comes out.

This an extension of the same changes done to etcd and masters in #367


chancez:
cc @s-urbaniak and @alexsomesan

PR #377, opened 2017-04-26
installer/server/terraform: add an option to retry `terraform apply`

Quentin-M:
This PR introduces a new `retry` boolean to the `terraform/apply` API endpoint. When specified, the backend will try to restore the executor from the session rather than creating a new one. If there was no executor in the session, an error will be thrown.

### DIFF
--- a/installer/server/terraform.go
+++ b/installer/server/terraform.go
@@ -147,6 +147,7 @@ type TerraformApplyHandlerInput struct {
 	License       string                 `json:"license"`
 	PullSecret    string                 `json:"pullSecret"`
 	DryRun        bool                   `json:"dryRun"`
+	Retry         bool                   `json:"retry"`
 }
 
 func terraformApplyHandler(sessionProvider sessions.Store) ctxh.ContextHandler {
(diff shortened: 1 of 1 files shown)


installer/server/terraform.go
```diff
@@ -162,10 +163,20 @@ func terraformApplyHandler(sessionProvider sessions.Store) ctxh.ContextHandler {
 		}
 		defer req.Body.Close()
 
-		// Create a new TerraForm Executor with the TF variables.
-		ex, errCtx := newExecutorFromApplyHandlerInput(&input)
-		if errCtx != nil {
-			return errCtx
+		var ex *terraform.Executor
+		var errCtx *ctxh.AppError
+		if !input.Retry {
```
ggreer: Nitpick: If you have an else, why invert the conditional? Just swap the bodies.

installer/server/terraform.go
```diff
@@ -162,10 +163,20 @@ func terraformApplyHandler(sessionProvider sessions.Store) ctxh.ContextHandler {
...
+			// Create a new TerraForm Executor with the TF variables.
+			ex, errCtx = newExecutorFromApplyHandlerInput(&input)
+			if errCtx != nil {
+				return errCtx
+			}
+		} else {
+			// Restore the execution environment from the session.
+			_, ex, _, errCtx = restoreExecutionFromSession(req, sessionProvider, &input.Credentials)
+			if errCtx != nil {
```
ggreer: I think errCtx check can be moved out of the if/else, reducing repetition. Code golf! ⛳️

Quentin-M: Yeap but typically had the exact opposite review on projects like Clair for readability reasons.

installer/server/terraform.go
Quentin-M: Just to have the usual case first. Can swap if you want.

PR #378, opened 2017-04-26
platforms/metal: Hack to allow a single terraform apply

dghubble:
* Disallow SSH as user core during the install to disk
* Increase SSH timeout to 60 minutes to allow leeway
* This prevents terraform from SSH'ing to machines during the install phase so that a single terraform apply is needed to complete the install process.


dghubble:
Testing...

PR #381, opened 2017-04-26
ingress config: Disable http2.

ggreer:
nginx sends "transfer-encoding: chunked" header, which is blacklisted in http2.
This causes spdy errors in Chrome.

### DIFF
--- a/installer/assets/ingress/default-backend-configmap.yaml
+++ b/installer/assets/ingress/default-backend-configmap.yaml
@@ -4,3 +4,4 @@ metadata:
   name: tectonic-custom-error
 data:
   custom-http-errors: "400,401,403,404,500,503,504"
+  use-http2: "false"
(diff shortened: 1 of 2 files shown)


ggreer:
I'm booting a cluster to test this....

chancez:
@ggreer can confirm this causes the "Restricted access page" to show back up again:
screen shot 2017-04-26 at 2 52 46 pm

installer/assets/ingress/default-backend-configmap.yaml
```diff
@@ -4,3 +4,4 @@ metadata:
   name: tectonic-custom-error
 data:
   custom-http-errors: "400,401,403,404,500,503,504"
+  use-http2: false
```
chancez: i think these need to be `"false"` with quotes.

ggreer: Thanks. Fixed.

ggreer:
Alright. Boots a cluster and no more SPDY errors. 🎉

PR #393, opened 2017-04-26
terraform apply retry button

ggreer:


ggreer:
Screenshots: 
screen shot 2017-04-26 at 16 27 11
screen shot 2017-04-26 at 16 32 19
screen shot 2017-04-26 at 16 37 33

meghanschofield:
@ggreer Does a user need to destroy AND retry Terraform? Before when we talked about issues with the apply that was the way to fix it up...and depending on the answer it may change the UX suggestion I have!

meghanschofield:
If you need to do both, I recommend this:
artboard 3

We have so many CTAs and icons it's really overwhelming. Here I think we add the buttons to the error message. Also. we should remove:

- Icon for Save Logs
- Icon for Hide logs
- Aren't we getting rid of Save Progress? That's confusing here...and I don't think you actually can. 
- Icon for Start Over

Other notes:
- Shouldn't "Next Step" be gray for this state?
- It'd really help to have more whitespace (10 px) between statuses. I added that to the mock. 

Hope that helps!

meghanschofield:
@ggreer Can I get an updated screenshot? It's nice to know what got in or didn't...

PR #401, opened 2017-04-27
change terraform link to 0.9.4

rmenn:
noticed that the link was still pointing to 0.8.8, this just corrects that


squat:
Thanks @rmenn!

PR #403, opened 2017-04-27
glide.yaml: Fix unpinned dependencies in glide.yaml

dghubble:
Running `make vendor` (without any glide.yaml changes) was causing vendor diffs. It should be a no-op unless something in glide.yaml is bumped. This should pin the unpinned dependencies based on the current glide.lock.

Run `make vendor` and verify the diff is empty.

```
make vendor
git status
```

### DIFF
--- a/installer/glide.yaml
+++ b/installer/glide.yaml
@@ -131,6 +131,135 @@ import:
   - pkg/util/validation
   - pkg/util/yaml
   - third_party/forked/reflect
+- package: k8s.io/client-go
+  version: d81cb85237595f720d83eda492bae8f6162fc5c0
+  subpackages:
+  - discovery
+  - kubernetes
+  - kubernetes/typed/apps/v1beta1
+  - kubernetes/typed/authentication/v1beta1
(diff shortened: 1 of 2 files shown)


installer/glide.yaml
```diff
@@ -131,6 +131,135 @@ import:
   - pkg/util/validation
   - pkg/util/yaml
   - third_party/forked/reflect
+- package: k8s.io/client-go
+  version: d81cb85237595f720d83eda492bae8f6162fc5c0
+  subpackages:
```
ericchiang: Do we actually need to list all the subpackages?

Quentin-M:
curious, what version of glide did you use?

installer/glide.yaml
dghubble: Just copied from glide.lock, perhaps not

dghubble:
@Quentin-M 0.13.0

Quentin-M:
no diff on 0.12.x either 👍

PR #406, opened 2017-04-27
tectonic: stop capturing 400x errors in default ingress backend

sym3tri:
This was causing a number of bugs in console.

- 400 errors messages were getting captured and un-viewable
- I suspect 401/403 errors were breaking the "you need to login" logic.

/cc @rithujohn191


ggreer:
Does https://github.com/coreos/tectonic-installer/blob/master/installer/assets/ingress/default-backend-configmap.yaml need to change as well? I don't know which one matters (or if both do).

squat:
@ggreer only the terraform one matters now

sym3tri:
@ggreer yeah, that one will be deleted shortly

PR #408, opened 2017-04-27
modules/aws/vpc: lock down security groups

squat:
This commit locks down the master and worker security groups for AWS
since we know that etcd pods and control plane pods that need access to
etcd will only be scheduled on master nodes. For that reason, the etcd
ports, 2379-2380 and 12379-12380, only need to accept connections from
other master nodes.

See
https://github.com/kubernetes-incubator/bootkube/blob/master/pkg/util/etcdutil/migrate.go#L132.

cc @sym3tri @Quentin-M


squat:
Tested working.

PR #414, opened 2017-04-28
Use cloudformation to manage the autoscaling groups

chancez:
The overall purpose of this PR is to enable users to modify parameters related to their ASG/LaunchConfig and have those modifications take effect.

Currently, if you change a variable on the launch_config, and re-run terraform apply, it will create a new launch_config, update the ASG to use the launch_config, destroy the old one, but the instances do not get updated.

The simplest use-case that I was testing is changing the instance_type of `worker` nodes, after realizing I didn't have enough resources to deploy my apps.

RollingUpdates of AutoScalingGroups are not a built in feature of
LaunchConfigs or ASGs, but is a feature of cloud formation. To handle
this, autoscaling group resources can be replaced with a cloudformation
resources which manages 1 thing, the autoscaling group and update policy
of that autoscaling group, as recommended by terraform upstream:
https://github.com/hashicorp/terraform/issues/1552#issuecomment-290315313

This allows preforming rolling updates, and actually makes it possible
to modify instance related parameters such as instance_type, and have
that take effect without having to manually the autoscaling groups.

This also gets us closer to a point where we can use `cfn-signal` to
indicate that a node is ready (ie: kubelet is healthy). I've also got a
WIP branch for implementing basic cfn signalling on top of this which
waits for kubelet to be healthy before proceeding with the ASG rolling
update.

## TODO

- Merge tectonic_autoscaling_group_extra_tags with the tags defined
on the ASG within the cloudformation. I was unable to get this to work. It's hard because `join` doesn't work on non-flat lists, and `encodejson` wasn't getting me what I needed. I'd love help on this @s-urbaniak or @alexsomesan.

- Currently this works fairly well overall. I can easily update my worker nodes without any trouble now. Updating masters hasn't been so successful. Without setting up `cfn-init` and `cfn-signal` to signal when a node is ready, it may be necessary to just increase the pause time on the master ASG UpdatePolicy. The new masters come up, but as far as I can tell, they're not getting all the pods they need before the other masters die off.

The lifecycle `create_before_destroy` changes are necessary because the Terraform document mentions any `create_before_destroy` resource can only depend on other `create_before_destroy` resources. Our launch_config is `create_before_destroy`, but the SGs it depends on are not. This fixes that. Without those changes, I was seeing some resources getting leaked and not cleaned up.

### DIFF
--- a/modules/aws/master-asg/master.tf
+++ b/modules/aws/master-asg/master.tf
@@ -22,29 +22,76 @@ data "aws_ami" "coreos_ami" {
   }
 }
 
-resource "aws_autoscaling_group" "masters" {
-  name                 = "${var.cluster_name}-masters"
-  desired_capacity     = "${var.instance_count}"
-  max_size             = "${var.instance_count * 3}"
-  min_size             = "1"
-  launch_configuration = "${aws_launch_configuration.master_conf.id}"
-  vpc_zone_identifier  = ["${var.subnet_ids}"]
+resource "aws_cloudformation_stack" "master_autoscaling_group" {
+  name = "${var.cluster_name}-master"
 
-  load_balancers = ["${aws_elb.api-internal.id}", "${join("",aws_elb.api-external.*.id)}", "${aws_elb.console.id}"]
+  # Health check grace period is 5 minutes.
+  # Batch size is 1, and we pause 3 minutes between batches.
+  # Plus 1 minute extra
+  timeout_in_minutes = "${(var.instance_count * 5 * 3 ) + 1}"
 
-  tag {
-    key                 = "Name"
-    value               = "${var.cluster_name}-master"
-    propagate_at_launch = true
+  # Rolling back the master ASG likely won't improve things, so disable it.
+  disable_rollback = true
+
+  tags = "${merge(map(
+      "Name", "${var.cluster_name}-cloudformation-master-asg",
+      "KubernetesCluster", "${var.cluster_name}"
+    ), var.extra_tags)}"
+
+  template_body = <<EOF
+{
+  "Resources": {
+    "AutoScalingGroup": {
+      "Type": "AWS::AutoScaling::AutoScalingGroup",
+      "Properties": {
+        "VPCZoneIdentifier": ["${join("\",\"", var.subnet_ids)}"],
+        "LaunchConfigurationName": "${aws_launch_configuration.master_conf.id}",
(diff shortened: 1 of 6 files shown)


dghubble:
What in the world? We've already written Tectonic AWS in CloudFormation once before :)

alexsomesan:
Let me first subscribe to what Dalton said: What in the word?!

On a more serious note, this looks like a funky approach. I know this approached has been discussed in Terraform forums for a while now.
But to be honest, it looks a bit overly complex for what it achieves.
Similar results can be obtained by just using Terraform alone. Albeit not 100% the same behaviour, the result is also updating the instances managed by the auto-scaling groups in a rolling manner.

The key is to trigger the replacement of the ASG and let Terraform do the draining on the old one and scale up the new one. You can achieve this by causing a dependency on the launch configuration in an ASG attribute that causes it to be replaced. Name is a good option there. You would also set `create_before_destroy` so that the there are always some workers online. Finally, using [Scaling Policies](http://docs.aws.amazon.com/autoscaling/latest/userguide/as-scale-based-on-demand.html#as-scaling-types) you can control the rate at which instances are terminated / created and also how many should be kept running while doing this.
In the end, this approach gets you the same result.

See POC [here](https://github.com/alexsomesan/tectonic-installer/blob/rolling-autoscaling/modules/aws/worker-asg/worker.tf#L25-L75)

chancez:
@alexsomesan I tried to take that approach but couldn't get it working, so I tried this based on the linked terraform issue, and it worked. I agree this solution is less than idea, I really dislike having to use CF, so alternatives welcome. I don't see any way to cause an ASG to be replaced by terraform, so I'd like to see an example of how to do this.

As far as using scaling policies, this means we would probably need to use SQS/SNS to signal, because the only other scaling policies are based on resources/monitoring as far as I can tell. This also seems complex, so I'm also hesitant of that.

That POC you link has no changes, did you forget to push something?

alexsomesan:
I did forget to push the changes (I'm retarded).

I'll get back to my computer in half an hour and push the example. Sorry
about that!

sym3tri:
@chancez This brings up a good topic of discussion & illustrates your point, but we need to try solving it other ways first. This is too big of an architectural shift from the overall design. This should be discussed with the team before we consider any changes like this. We've deliberately moved away from CloudFormation for various reasons I won't go into here.

PR #415, opened 2017-04-28
Only create minimum number of NAT gateways needed

alexsomesan:
This fixes an issue @sym3tri noticed. Before this change, the number of NAT gateways created was based on the number of master subnets. This doesn't make sense if the number of worker subnets is lower, since not all NAT gateways will be attached to worker subnets in that case.

With this change, we create as little NAT gateways as needed to support the number of worker subnets attached or to fit in the number of master subnets available (hence the min(..) expression).

@sym3tri @Quentin-M @s-urbaniak

### DIFF
--- a/modules/aws/vpc/vpc-private.tf
+++ b/modules/aws/vpc/vpc-private.tf
@@ -12,7 +12,7 @@ resource "aws_route" "to_nat_gw" {
   count                  = "${var.external_vpc_id == "" ? var.worker_az_count : 0}"
   route_table_id         = "${aws_route_table.private_routes.*.id[count.index]}"
   destination_cidr_block = "0.0.0.0/0"
-  nat_gateway_id         = "${aws_nat_gateway.nat_gw.*.id[count.index]}"
+  nat_gateway_id         = "${element(aws_nat_gateway.nat_gw.*.id, count.index)}"
   depends_on             = ["aws_route_table.private_routes"]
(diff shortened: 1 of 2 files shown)


modules/aws/vpc/vpc-private.tf
```diff
@@ -12,7 +12,7 @@ resource "aws_route" "to_nat_gw" {
   count                  = "${var.external_vpc_id == "" ? var.worker_az_count : 0}"
   route_table_id         = "${aws_route_table.private_routes.*.id[count.index]}"
   destination_cidr_block = "0.0.0.0/0"
-  nat_gateway_id         = "${aws_nat_gateway.nat_gw.*.id[count.index]}"
+  nat_gateway_id         = "${element(aws_nat_gateway.nat_gw.*.id, count.index)}"
```
sym3tri: Won't this break if there are fewer AZs used for masters than for workers? Is this case possible?

alexsomesan:
That's actually why I added the min() call there.
It builds the least number of nat-gw of the two.

I did test it manually in both scenarios.

sym3tri:
Ah... the `element()` call threw me off. Did not realize:

"If the index is greater than the number of elements, this function will wrap using a standard mod algorithm."

sym3tri:
going to merge this b/c we need it for the release.

PR #440, opened 2017-05-01
Extend timeout for terraform apply test

ethernetdan:
The test `TestExecutorSimple` was frequently timing out when the Jenkins executors are overloaded. Extending the timeout to ensure this passes.

/cc @Quentin-M


Quentin-M:
Thank you!

PR #442, opened 2017-05-02
server/terraform: create the executor next to the installer

### LINKED ISSUE #446: Force user to save assets in backend
i.e. stop using temp dir

Quentin-M:
Rather than tmp, so users always have the assets somewhere,
regardless of whether they download them or not, and
regardless of potential system reboots.

We should probably have a flag / env variable that controls that,
but we have little time right now. Must switch to doing self-hosted etcd
cleanup for the release now.

Fixes #446

### DIFF
--- a/installer/server/terraform/executor_test.go
+++ b/installer/server/terraform/executor_test.go
@@ -52,8 +52,14 @@ func TestMain(m *testing.M) {
 // worked (State/Status), and then create a new executor at the path of the
 // existing one and verify the state is shared.
 func TestExecutorSimple(t *testing.T) {
+	tmpDir, err := ioutil.TempDir("", "tectonic")
+	if err != nil {
+		t.Logf("Failed to create temporary directory: %s", err)
+		t.FailNow()
+	}
+
 	// Create an executor.
-	ex, err := NewExecutor()
+	ex, err := NewExecutor(tmpDir)
 	if err == ErrBinaryNotFound {
 		t.Skip("TerraForm not found, skipping")
 		return
@@ -98,7 +104,7 @@ func TestExecutorSimple(t *testing.T) {
 	assert.NotZero(t, len(outputBytes))
 
 	// Creates a new executor at the same existing one.
-	ex2, err := NewExecutorFromPath(ex.WorkingDirectory())
+	ex2, err := NewExecutor(ex.WorkingDirectory())
 	assert.Nil(t, err)
 	assert.NotNil(t, ex2)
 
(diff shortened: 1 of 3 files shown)


installer/server/terraform/executor_test.go
```diff
@@ -135,7 +147,7 @@ func TestExecutorMissingVar(t *testing.T) {
 	// Wait for its termination.
 	select {
 	case <-done:
-	case <-time.After(1 * time.Second):
+	case <-time.After(10 * time.Second):
```
ggreer: Just curious, but why was this timeout increased?

Quentin-M: Jenkins executors are waaaaayyy to slooowwww sometimes and fails to execute TerraForm within a second, which usually takes about 100ms.

Quentin-M: It is actually on master already, we made the change at the same time with Dan. Just need to rebase for it to be gone.

Quentin-M:
I am a little afraid with this change because calling apply again with the same cluster name but different variables will overwrite these variables and try to reconcile.. it might wipe an existing cluster. Should we still add a random number after the cluster name?

Quentin-M:
Just tested. Worked just fine.

ggreer:
Tested on mac & it works. 👍

ggreer:
Ideally I'd like clustername + datetime as the directory name. eg: `ggreer-test_2017-05-02_184520` for a cluster named "ggreer-test" booted today at 18:45:20. That way we avoid conflicts when a user picks the same name for a cluster.

Quentin-M:
@ggreer Just changed. Hopefully this will not cause too much trouble to users when they retry on failed deployments. `bin/darwin/clusters/qmachu_2017-05-02_13-35-40/`

ggreer:
Just FYI, user experience with a conflicting cluster is not ideal:

screen shot 2017-05-02 at 13 23 57
screen shot 2017-05-02 at 13 25 38
screen shot 2017-05-02 at 13 35 21

I created a cluster named ggtest, then started over and created another one with the same name but different network settings.

Quentin-M:
The timeout does not seem so related but yeah, TerraForm is trying to reconcile, and thus is deleting the networks to re-create them. Having separate folders will just make TerraForm fail in this case because things like entries in the public Route53 zone will already exist.

Quentin-M:
Tests passed.

ggreer:
Yep. New directory name works for me too.

PR #473, opened 2017-05-02
installer/assets: bump dex to v2.4.1

rithujohn191:


Quentin-M:
(waiting for test but no reason it'd break, talked offline)

rithujohn191:
Tested it with the latest master. Works as expected. No issue loading clients on the my accounts page.

PR #474, opened 2017-05-02
platform/metal: Add kubelet-env.service to workers

dghubble:
* Problem: In the case a cluster is deployed, various Kubernetes updates are applied over time, and then workers are added by re-using the terraform config to scale workers, the Kubelet version would be out-of-date. This could be done by extending `tectonic_metal_worker_names`, `tectonic_metal_worker_domains`, and `tectonic_metal_worker_macs` and re-applying.
* Start the `kubelet-env.service` to detect the Kubelet version from the bootstrapped control plane to create the kubelet.env file if missing

Note: Adding new controllers in tfvars to try to scale controllers is not supported to begin with on bare-metal (controllers are etcd peers), so launching controllers with version mismatch is not a problem to solve. Controller scaling will come once etcd is run via self-hosted etcd.

### DIFF
--- a/platforms/metal/cl/bootkube-worker.yaml.tmpl
+++ b/platforms/metal/cl/bootkube-worker.yaml.tmpl
@@ -38,11 +38,26 @@ systemd:
         ExecStart=/bin/sh -c 'while ! /usr/bin/grep '^[^#[:space:]]' /etc/resolv.conf > /dev/null; do sleep 1; done'
         [Install]
         RequiredBy=kubelet.service
+    - name: kubelet-env.service
+      enable: true
+      contents: |
+        [Unit]
+        Description=Determine the Kubelet Image Version
+        ConditionPathExists=!/etc/kubernetes/kubelet.env
+        [Service]
+        ExecStartPre=/bin/mkdir -p /etc/kubernetes
+        ExecStartPre=/usr/bin/bash -c "docker run --rm -v /etc/kubernetes/kubeconfig:/kubeconfig {{.kube_version_image}} --kubeconfig=/kubeconfig > /etc/kubernetes/kube.version"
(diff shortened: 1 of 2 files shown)


platforms/metal/cl/bootkube-worker.yaml.tmpl
```diff
@@ -38,19 +38,33 @@ systemd:
...
+        Restart=on-failure
+        RestartSec=10
+        [Install]
+        WantedBy=multi-user.target
     - name: kubelet.service
       contents: |
         [Unit]
         Description=Kubelet via Hyperkube ACI
         [Service]
+        EnvironmentFile=/etc/kubernetes/kubelet.env
```
dghubble: Kubelet will start once `kubeconfig` is copied to the machine and crash loop until `kubelet-env.service` creates `kubelet.env`

platforms/metal/cl/bootkube-worker.yaml.tmpl
```diff
@@ -38,19 +38,33 @@ systemd:
         ExecStart=/bin/sh -c 'while ! /usr/bin/grep '^[^#[:space:]]' /etc/resolv.conf > /dev/null; do sleep 1; done'
         [Install]
         RequiredBy=kubelet.service
+    - name: kubelet-env.service
+      enable: true
```
dghubble: Starts on boot and crash loops until the control plane is available.

yifan-gu:
@dghubble What's our plan for the master nodes?

dghubble:
I mentioned it in the PR description. Bare-Metal controllers run etcd peers (e.g. 3 controllers = 3 node etcd) rather than having special bare-metal machines just for etcd. Once self-hosted etcd graduates from experimental, etcd can be run as pods instead of on-host.

yifan-gu:
@dghubble Thank you for the details, /cc @aaronlevy

PR #475, opened 2017-05-02
Documentation/install/aws: specify go version 1.8

rithujohn191:
Specify which version of go should be used. `make build` will fail if not at version 1.8


Quentin-M:
It's in other files too ~

Quentin-M:
Oh there we go.

PR #477, opened 2017-05-02
Docs: adding AWS S3 credentials info.

zbwright:
I think this info should be added to the Tectonic tutorials as well, but I'm not sure where.

### DIFF
--- a/Documentation/install/aws/requirements.md
+++ b/Documentation/install/aws/requirements.md
@@ -92,9 +92,11 @@ Use the `SECRET_ACCESS_KEY`, `ACCESS_KEY_ID`, and `SESSION_TOKEN` to authenticat
 
 The final step of the Tectonic install requires an SSH key and access to standard utilities like `ssh` and `scp`. Setting up a new key on AWS should take less than 5 minutes.
 
+Tectonic uses AWS S3 to store all credentials, using server-side AES encryption for storage, and TLS encryption for upload/download. Any pod run in the system can query the AWS metadata, get node AWS credentials, and pull down cluster credentials from AWS S3.
+
 First, create a key.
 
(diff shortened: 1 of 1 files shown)


Documentation/install/aws/requirements.md
```diff
@@ -92,9 +92,11 @@ Use the `SECRET_ACCESS_KEY`, `ACCESS_KEY_ID`, and `SESSION_TOKEN` to authenticat
 
 The final step of the Tectonic install requires an SSH key and access to standard utilities like `ssh` and `scp`. Setting up a new key on AWS should take less than 5 minutes.
 
+Tectonic uses AWS S3 to store all credentials, using server-side AES encryption for storage, and TLS encryption for upload/download. Any pod run in the system can query the AWS metadata, get node AWS credentials, and pull down cluster credentials from AWS S3.
```
ericchiang: Maybe note that the long term fix will be preventing pods from talking to the AWS metadata service through network policies? cc @Quentin-M

Quentin-M: I agree that this should be fixed ASAP. Pods should probably not be able to talk to the AWS API using the creds.

zbwright: I'm not clear here. Should I add a (temporary) note that users should prevent this access? add:

... pull down cluster credentials from AWS S3. To prevent pods from talking to the AWS metadata service, create network policies which deny them access.

ericchiang: @zbwright we should add a note that we plan to address this in a later release since it's a pretty serious issue. I don't believe there's a fix that user's can apply today.

zbwright: ok. I will leave the planned fix vague. Is this a serious enough issue that it should be bumped to the top of the page?

joshix:
ok to test

PR #481, opened 2017-05-02
frontend: Scrolls to the bottom of the logs when show logs is selected

amrutac:


installer/frontend/components/tf-poweron.jsx
```diff
@@ -47,17 +47,22 @@ class TF_PowerOn extends React.Component {
...
     }
 
-    this.shouldScroll = node.scrollHeight - node.clientHeight <= node.scrollTop;
+    this.shouldScroll = node.scrollHeight - node.clientHeight <= node.scrollTop + 10;
   }
 
   componentDidUpdate () {
     if (this.shouldScroll && this.outputNode) {
-      this.outputNode.scrollTop = this.outputNode.scrollHeight - this.outputNode.clientHeight;
+      this.outputNode.scrollTop = this.outputNode.scrollHeight;
```
ggreer: From https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTop

> If set to a value greater than the maximum available for the element, scrollTop settles itself to the maximum value.

Looks like browsers obey that, so this is safe. 👍

amrutac: I reverted back this change, it was working fine.

Issue #456, opened 2017-05-02
GUI: remove 1.5 options from dropdowns


sym3tri:
Only allow users to install 1.6

PR #486, opened 2017-05-02
server: Only list terraform platforms.

ggreer:
Fixes https://github.com/coreos/tectonic-installer/issues/456


ggreer:
This creates a decent amount of unused code, but it's better to make a small change before release. Users don't notice unused code. They notice bugs. :)

Quentin-M:
@ggreer Can you create a follow-up issue then to cleanup these? :)

PR #495, opened 2017-05-02
Temporarily remove windows from release binaries.

ggreer:
The installer doesn't work on windows, so we shouldn't get users' hopes up by giving them a windows binary.


amrutac:
Do the docs need to be updated too?

PR #502, opened 2017-05-03
Use /var/log/kubernetes for apiserver audit logs hostVolume

### LINKED ISSUE #499: use /var/log/kubernetes/api-server
I am concerned about having a bind mount to /var/log into the API server. Did we consider using /var/log/kubernetes/api-server or something instead?

https://github.com/coreos/tectonic-installer/blob/c10fb33624ce0aad08b63cd3618e801234b1db2e/modules/bootkube/resources/manifests/kube-apiserver.yaml#L57

chancez:
Closes #499


chancez:
cc @robszumski @philips

s-urbaniak:
Feel free to self-merge once all parties agree that this is the correct setting.

PR #503, opened 2017-05-03
platform/metal: bootkube module etcd_endpoint takes no port now

dghubble:


Quentin-M:
MRW I realize the file I selected was 2.5GB not the 300MB and it just finished downloading to my phone (without WiFi)

PR #507, opened 2017-05-03
test: Run make {examples,docs} and check for changes

mxinden:
This PR adds an example and a docs stage to our Jenkins build pipeline. It first runs the respective targets and then checks if anything changed.

@s-urbaniak Let me know what you think.

Is there a way to force make to run a target to prevent `make: Nothing to be done for 'examples'.`?

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -25,6 +25,36 @@ pipeline {
       }
     }
 
+    stage('Generate docs') {
+      steps {
+        sh """#!/bin/bash -ex
+
+        # Prevent "fatal: You don't exist. Go away!" git error
+        git config --global user.name "jenkins tectonic installer"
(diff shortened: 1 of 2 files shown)


s-urbaniak:
ok to test

s-urbaniak:
> Is there a way to force make to run a target to prevent make: Nothing to be done for 'examples'.?

I think this is because phony targets must be explicit. Let me see if I can come up with something.

sym3tri:
@ethernetdan bump

mxinden:
@ethernetdan bumping this again. Tests are failing because docs are out of sync on master.

sym3tri:
@mxinden can you please just include the latest generated docs with this PR?

Also in the future please use a fork, we don't create branches on origin.

mxinden:
@ethernetdan @sym3tri Updated the PR. CI is running green. Any further comments?

Issue #489, opened 2017-05-02
Standardize labels for k8s/tectonic components


aaronlevy:
There are quite a few variations of how we are identifying applications by label: "app", "k8s-app", "k8s-addon", "operator", "component". This could make inspecting cluster state difficult later on -- we should try and standardize how we identify applications we are managing.

It looks like upstream somewhat standardizes on two sets of identifiers, a "control-plane" component, and addon applications.

**For control-plane:**

```
tier=control-plane
component=kube-scheduler
```
[example](https://github.com/kubernetes/kubernetes/blob/master/cluster/saltbase/salt/kube-apiserver/kube-apiserver.manifest#L199-L202)

**For addons:**

```
k8s-app=kube-dns
```
[example](https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/kubedns-controller.yaml.base#L26)
Note: the other labels are for addon-manager which we do not use (one of which is [deprecated](https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/addon-manager/kube-addons.sh#L44-L48) anyway)

## Proposal:

We standardize on the same labels as upstream.

For the core k8s components (etcd, api, scheduler, controller-manager, proxy) we use the `tier`/`component` label keys.

For all other applications we use `k8s-app`.

philips:
There has been some talk of creating a k8s.io/labels project. If someone wants to take that on upstream.

philips:
Overall this makes sense to me.

A lingering concern in my mind is using k8s-app in such a way that would conflict with a change upstream.

aaronlevy:
Couple clarifying points:

kube-proxy would get `tier=node` (all other core components are `tier=control-plane`)

We could also add the `tier` to tectonic components (`tier=monitoring` for example).

@philips we could do k8s-app for upstream addons (dns, heapster, HPA, etc.) but use `tectonic-app` for all others?

Ref: upstream change: https://github.com/kubernetes/kubernetes/pull/24293

sym3tri:
SGTM
Need to validate with @xiang90 and @fabxc that labels are not relied on by etcd or Prometheus operators first.

sym3tri:
@aaronlevy do you also want to label ConfigMaps & Secrets etc? Or just DaemonSets and Deployments and Services? Or all of the things?

aaronlevy:
[...]

PR #517, opened 2017-05-03
*: standardize on k8s-app label

pbx0:
This standardizes using the `k8s-app` as the de-facto label for identifying components. Discussed in https://github.com/coreos/tectonic-installer/issues/489.

My approach was for anything using 1 label to identify a component I changed that label to `k8s-app`. Mostly this consists of substitutions from `app` -> `k8s-app`. Control plane components were using the `component` label.

In cases in which there were two labels identifying a component as the same value I delete one of them and made sure the other was `k8s-app`. Mostly this consists of removing an `app` label and keeping the `k8s-app` one around. 

There are some outstanding questions I have about certain labels:
 * I see a lot of component labels floating around that differ from the `app` label. Are these redundant? Delete? For example:
     * `installer/assets/stats-emitter-deployment.yaml`
     * `modules/tectonic/resources/manifests/console/service.yaml`
 * Here are some files I was unsure what to do with:
     * ~~`installer/assets/monitoring/prometheus-k8s.json`~~
     * ~~`installer/assets/monitoring/prometheus-operator.yaml`~~
     * `Documentation/files/logging/monitoring/prometheus-fluentd.yaml`
 * Its not safe to change the label selector for `modules/bootkube/resources/experimental/manifests/etcd-service.yaml`
until the etcd-operator updates to use the k8s-app label as well. So we won't change this for now and etcd pods will sort of be the odd one out for now.
 * Not familiar with jobLabel field in `Documentation/files/logging/monitoring/fluentd-prometheus-service-monitor.yaml`
    * Presumably should also just change to k8s-app

### DIFF
--- a/Documentation/admin/ingress.md
+++ b/Documentation/admin/ingress.md
@@ -85,13 +85,13 @@ metadata:
   name: production-ingress
   namespace: production
   labels:
-    app: ingress
+    k8s-app: ingress
 spec:
   replicas: 2
   template:
     metadata:
       labels:
-        app: ingress
+        k8s-app: ingress
     spec:
       containers:
         - name: nginx-ingress-lb
--- a/Documentation/admin/onboard-service-account.md
+++ b/Documentation/admin/onboard-service-account.md
@@ -47,7 +47,7 @@ If multiple pods running in the same namespace require different levels of acces
       template:
         metadata:
           labels:
-            app: nginx
+            k8s-app: nginx
         spec:
           containers:
           - name: nginx
--- a/Documentation/admin/user-management.md
+++ b/Documentation/admin/user-management.md
@@ -213,7 +213,7 @@ spec:
   template:
     metadata:
       labels:
-        app: nginx
+        k8s-app: nginx
     spec:
       containers:
       - name: nginx
--- a/Documentation/files/logging/fluentd-aggregator-deployment.yaml
+++ b/Documentation/files/logging/fluentd-aggregator-deployment.yaml
@@ -4,7 +4,7 @@ metadata:
   name: fluentd-aggregator
   namespace: logging
   labels:
-    app: fluentd-aggregator
+    k8s-app: fluentd-aggregator
     component: log-aggregator
 spec:
   replicas: 2
@@ -18,7 +18,7 @@ spec:
   template:
     metadata:
       labels:
-        app: fluentd-aggregator
+        k8s-app: fluentd-aggregator
     spec:
       containers:
       - name: fluentd-aggregator
--- a/Documentation/files/logging/fluentd-aggregator-svc.yaml
+++ b/Documentation/files/logging/fluentd-aggregator-svc.yaml
@@ -4,11 +4,11 @@ metadata:
   name: fluentd-aggregator
   namespace: logging
   labels:
-    app: fluentd-aggregator
+    k8s-app: fluentd-aggregator
 spec:
   type: ClusterIP
   selector:
-    app: fluentd-aggregator
+    k8s-app: fluentd-aggregator
   ports:
   - name: fluentd-input
     port: 24224
--- a/Documentation/files/logging/fluentd-configmap.yaml
+++ b/Documentation/files/logging/fluentd-configmap.yaml
@@ -4,7 +4,7 @@ metadata:
   name: fluentd-config
   namespace: logging
   labels:
-    app: fluentd
+    k8s-app: fluentd
 data:
   fluentd.conf: |
     # Use the config specified by the FLUENTD_CONFIG environment variable, or
--- a/Documentation/files/logging/fluentd-ds.yaml
+++ b/Documentation/files/logging/fluentd-ds.yaml
@@ -4,7 +4,7 @@ metadata:
   name: fluentd
   namespace: logging
   labels:
-    app: fluentd
+    k8s-app: fluentd
     component: logging-agent
 spec:
   minReadySeconds: 10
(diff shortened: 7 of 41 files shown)


installer/assets/identity-api-svc.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
   namespace: tectonic-system
 spec:
   selector:
-    app: tectonic-identity
+    k8s-app: tectonic-identity
     component: identity
```
aaronlevy: @ericchiang these seem a bit redundant -- do you want to keep both?

ericchiang: I think we're fine removing the second. cc @rithujohn191

pbx0:
This is more or less ready to go. Etcd pods won't have updated labels because etcd-operator would need updating. There is an outstanding question of if we want to delete the mostly redundant component labels floating around the tectonic components, but if they aren't used, I think they could be deleted at a later time. No preference. I also have some outstanding questions about some of the Prometheus yamls, but its all documentation only.

So I just need some advice help on testing this then.

installer/assets/identity-api-svc.yaml
rithujohn191: :+1:

pbx0:
cc @alexsomesan @s-urbaniak

aaronlevy:
lgtm -- but I'd wait for one of the maintainers to sign off too.

installer/assets/console-deployment.yaml
```diff
@@ -21,7 +21,7 @@ spec:
     metadata:
       creationTimestamp: null
       labels:
-        app: tectonic-console
+        k8s-app: tectonic-console
```
sym3tri: @pbx0 nothing under `/installer/assets` is used anymore and will all be deleted soon.

sym3tri:
Doesn't hurt anything, but can omit all changes to `/installer/assets/*` that's all dead code we'll be deleting soon.

pbx0:
@sym3tri Sure, I'll remove the dead code. Should installer/examples go as well?

Quentin-M:
@pbx0 yes.

pbx0:
Removed changes to dead code

Quentin-M:
@pbx0 Could you remove the CLUO conflict too? I'll test your PR <3 Thanks!

pbx0:
@Quentin-M rebased

Quentin-M:
Trying~

pbx0:
Sorry, just pushed some updates. Found some usage of `k8s-apps` instead of `k8s-app`

Quentin-M:
Tests broke:
```
k8s_test.go:232: get apiserver pod: no pods matched the label selector "component=kube-apiserver" in the kube-system namespace
```

Quentin-M:
https://github.com/coreos/tectonic-installer/blob/master/installer/tests/sanity/k8s_test.go#L315

pbx0:
Updated the test to use the new label.

brancz:
all the label changes in the monitoring manifests will be overridden if self driving is enabled

brancz:
@sym3tri basically to solve my previous comment I see two possibilities:

* release a patch version of the tectonic prometheus operator that uses the labels in the same way as in this PR
* accept the "inconsistency" for now and fix it with the next tectonic-monitoring release

Both possible, just let us know what is preferred.

Maybe next time tag us when there are changes to the manifests we're responsible for. Generally our workflow for the monitoring manifests is that we do the changes in the TPO repository and then sync them in here so technically the manifests should never be touched in this repo unless it's syncing from the TPO repo. Maybe this is not the best approach, let us know if you think we can improve something here. Off the top of my head maybe we can add a comment at the top of each manifest stating that they are synced from the other repo. /cc @fabxc @mxinden

fabxc:
Yea, this case here sums up my doubt about the approach of replicating the setup a given tectonic-*-operator is responsible for with a bunch of manifests+kubectl in the installer itself.
The installer just creating tectonic-prometheus-operator + AppVersion and waiting for the AppVersion status indicating completion should be sufficient IMO. If the TPO doesn't converge, the install process failed.  
Throwing a bunch of (hopefully in-sync) manifests in there with kubectl doesn't give the same kind of feedback as far as I can tell.

brancz:
Agreed. Installer should just create TPO + AppVersion and leave the rest to it.

sym3tri:
@brancz @fabxc 
We tagged you guys in https://github.com/coreos/tectonic-installer/issues/489#issuecomment-298790146

Based on https://github.com/coreos/tectonic-installer/issues/489#issuecomment-298822756 we thought the change were ok to make?

:confused: 

Getting a patched operator into the installer is preferred, but would need to happen by Monday.

Let's discuss best strategy for maintaining/deploying manifests going forward.

aaronlevy:
We can back out the prometheus changes too. This isn't a hard requirement -- it's just trying to get as much (log-hanging) consistency as possible. But if this introduces more complexity elsewhere, we can revisit later.

aaronlevy:
@brancz @fabxc is this a change we should roll back for the monitoring components?

brancz:
reverting the changes for all the manifests in the monitoring directory would be the easiest and quickest solution for now, then we can make sure we do these changes for the next tectonic monitoring release

aaronlevy:
@brancz it can be a little risky to change labels that would affect selectors -- has a chance to leave orphaned pods (or at least it has in the past - not sure if handling has gotten better). Just a side-note (the "app" label isn't a huge deal if it ends up being painful to change after-the-fact).

@pbx0 would you be able to open a new PR which reverts the changes to the monitoring directory?

aaronlevy:
Actually opened https://github.com/coreos/tectonic-installer/issues/576 to track this (rather than in a closed PR)

PR #325, opened 2017-04-24
installer: Remove dead code

Quentin-M:
Following our documented process of migrating out of the CloudFormation and Go-powered bare-metal provisioning methods, towards using exclusively TerraForm, this PR aims at removing all the code that is exclusively used for the provisioning - and refactoring/consolidating the rest.

PR #521, opened 2017-05-03
Add Bare-Metal smoke tests

ethernetdan:
This setup will be changed eventually but throwing together some basic smoke tests from existing configs for baremetal.


Jenkinsfile
```diff
@@ -103,6 +103,38 @@ pipeline {
             installer/bin/sanity -test.v -test.parallel=1
             '''
             }
+          },
+          "Bare Metal": {
+            agent none
+            node('bare-metal') {
```
dghubble: fedora and bare-metal

Jenkinsfile
```diff
@@ -40,65 +46,45 @@ pipeline {
...
+                          ]) {
+                checkout scm
+                unstash 'installer'
+                unstash 'sanity'
+                sh '''#!/bin/bash -e
 
-            # Create local config
-            make localconfig
+                  export PLATFORM=metal
+                  export CLUSTER="tf-${PLATFORM}-${BRANCH_NAME}-${BUILD_ID}"
```
dghubble: CLUSTER could just be the default "demo" here. The executor creates one cluster and deletes it at the end, there are not simultaneous uses of the same executor.

Jenkinsfile
```diff
@@ -40,65 +46,45 @@ pipeline {
...
 
-            export NODE_COUNT=$(( ${MASTER_COUNT} + ${WORKER_COUNT} ))
+                  # lowercase cluster names
+                  export TF_VAR_tectonic_cluster_name=$(echo ${CLUSTER} | awk '{print tolower($0)}')
+                  cd installer
 
-            export TEST_KUBECONFIG=${WORKSPACE}/build/${CLUSTER}/generated/auth/kubeconfig
-            installer/bin/sanity -test.v -test.parallel=1
-            '''
+                  ./tests/scripts/aws/up-down.sh
```
dghubble: You're calling aws, not bare-metal. Also, you need to set ASSETS_DIR. https://github.com/coreos/matchbox/blob/master/Jenkinsfile#L43

Jenkinsfile
```diff
@@ -4,20 +4,21 @@
...
-    docker {
-      image 'quay.io/coreos/tectonic-builder:v1.7'
-      label 'worker'
-    }
-  }
+  agent none
 
   options {
-    timeout(time:35, unit:'MINUTES')
+    timeout(time:60, unit:'MINUTES')
```
dghubble: This seems far too long

installer/tests/scripts/bare-metal/up-down.sh
```diff
@@ -27,8 +24,14 @@ main() {
   git checkout $MATCHBOX_SHA
   chmod 600 tests/smoke/fake_rsa
   popd
+
+  echo "Copying matchbook test creds"
   cp examples/fake-creds/{ca.crt,server.crt,server.key} matchbox/examples/etc/matchbox
 
+  echo "Adding test SSH credentials to ssh-agent"
+  ssh-add matchbox/tests/smoke/fake_rsa
```
dghubble: Can you add this to setupSSH please

Jenkinsfile
```diff
@@ -95,35 +108,71 @@ pipeline {
...
+                echo "Destroying ${CLUSTER}..."
+                make destroy || true
+              done
+            '''
+          }
+        }
+      }
+    }
+
+    stage("Smoke Tests - Bare Metal") {
```
dghubble: Please parallelize this with AWS before merge.

Jenkinsfile
```diff
@@ -47,6 +54,12 @@ pipeline {
         }
       }
       stage("Smoke Tests") {
+      agent {
```
dghubble: Indent the block like the other blocks

dghubble:
Can you squash this and show it running successfully? Also, this doesn't seem to be limited to master branch as one of your intermediate commits mentions.

dghubble:
Check the old Jenkinsfile for how to run AWS and bare-metal tests in parallel https://github.com/coreos-inc/tectonic/blob/2926480bf057f732651dcc47f0496325f4e1eb75/Jenkinsfile#L89

Quentin-M:
About to follow-up to make these run on our Packet.net instances.

Quentin-M:
Refreshed the PR and rebase on origin/master ... turned out that Jenkins was just buggy because one of the executor ran out of memory.

Quentin-M:
@ethernetdan @sym3tri Good to go. I am not modifying this further. Each build+bare-metal smoke test takes ~15min. Can expect this to be way faster once we have https://github.com/coreos/tectonic-installer/pull/721 + K8S executors (maybe 10min total?). Thank you so much for your help during the past days - I learned a lot. Next step would be to get rid of the EC2 executors make the smoke tests run in parallel.

Quentin-M:
How is it going? :)

sym3tri:
The test that ran passed, but there are still errors in the logs. Is this expected?

```
+ echo 'Cleanup testing environment'
Cleanup testing environment
+ cleanup
+ set +e
+ cd /jenkins/workspace/r_tectonic-installer_PR-521-AJHV5TPWIFOY3L3ZRVNU57JINLM2WRLUZK4E7XHCM6QRFVRANFAA/installer/tests/scripts/bare-metal/../../../matchbox
+ sudo ./scripts/libvirt destroy
error: failed to get domain 'node1'
error: Domain not found: no domain with matching name 'node1'

error: failed to get domain 'node2'
error: Domain not found: no domain with matching name 'node2'

error: failed to get domain 'node3'
error: Domain not found: no domain with matching name 'node3'

error: failed to get domain 'node4'
error: Domain not found: no domain with matching name 'node4'

error: failed to get domain 'node1'
error: Domain not found: no domain with matching name 'node1'

error: failed to get domain 'node2'
error: Domain not found: no domain with matching name 'node2'

error: failed to get domain 'node3'
error: Domain not found: no domain with matching name 'node3'

error: failed to get domain 'node4'
error: Domain not found: no domain with matching name 'node4'

Pool default refreshed

error: failed to get vol 'node1.qcow2'
error: Storage volume not found: no storage vol with matching path 'node1.qcow2'

error: failed to get vol 'node2.qcow2'
error: Storage volume not found: no storage vol with matching path 'node2.qcow2'

error: failed to get vol 'node3.qcow2'
error: Storage volume not found: no storage vol with matching path 'node3.qcow2'

error: failed to get vol 'node4.qcow2'
error: Storage volume not found: no storage vol with matching path 'node4.qcow2'
```

Quentin-M:
It is. There is a bunch of safety cleanup before hand. These tell us that there was no leftover to cleanup. I don't want to hide these whatsoever as a failure to cleanup any leftover VMs (which should not be there in the first place) would lead in a test failure that would be difficult to debug. We have had these logs for a long time :)

Quentin-M:
@ethernetdan Hi, do you think you will be able to make any progress on this?

sym3tri:
@Quentin-M as discussed, please take this over or close & reopen.

Quentin-M:
@sym3tri I am actively working on this. Fixing a bunch of bad practices that were introduced in the meantime, and wrapping up the parallelism on multiple nodes.

ggreer:
I realize this is still a work in progress, but I think the parallelization changes to Jenkins should be in a separate PR. They're not necessary for baremetal tests and they're much harder to vet for correctness.

Quentin-M:
@ggreer I am about to push the final commit. It works now.

Quentin-M:
Squashed the 20 commits as they were not making any sense. Looots of different changes and git is not able to understand properly nesting/brackets/indentation that is core to the Jenkinsfile. Will now split the big commit a bit.

Quentin-M:
ok to test

Quentin-M:
- Fixed Bare-Metal smoke test script

- Squashed and then re-splitted the commits properly
- Squashed a few Jenkins stages for speedups and maintainability
- Cleaned up the Jenkinsfile a bit (added var, removed logic, ...)

- Build / Test / AWS smoke tests still to be scheduled on EC2 instances, BM on Packet.net, as planned initial
- Adjusted caps/time-outs in Jenkins to avoid spawning too many instances of each kind, and avoid having a high turnover / wait time

Quentin-M:
These tests already helped me find: https://github.com/coreos/tectonic-installer/pull/1010
Now, it appears that the API server is not being brought back up after being killed. The checkpointed manfiest is in place though, but kubelet does not start the container.

Quentin-M:
Thanksss! Sure!

ggreer:
> Now having parallelism means that tests will run for at least an hour or more.

If that's true, then I think the best thing to do would be:

1. Split out the parallelism changes into a separate PR.
2. Don't merge the baremetal tests PR until after parallelism PR.

This pull request contains two new features (BM tests & parallel tests), some refactoring, and some tangentially-related cleanup... all rolled into one PR. In short: It gives me vibes of #325.

Matt & I spent weeks fixing Jenkins. I don't want to do that again.

Quentin-M:
I believe that we should add tests regardless if they break a bit or not. If they do, then we can fix them, and it will likely lead to having less support work to do, and increasing our confidence in what we are doing. I would rather have broken tests, than shipping broken. These tests already revealed a [regression](https://github.com/coreos/tectonic-installer/pull/1010). If Jenkins breaks or there is any new flakiness, I will be happy to fix it immediately.

I think we are all in the same team. We all work on improving Jenkins, as well as our test coverage and this is our common responsibility to handle that continuous task. I have spent days setting it up, and evenings fixing configuration issues introduced later on.

It is critical to add back bare-metal smoke testing and has been a top goal for a while. This is not a net new feature, we are merely porting and adapting what we had before - but on Packet.net rather than in the lab machines (concurrency and stability). This has been an on-going work for the 4-6 weeks, that involved a lot of collaborative work between Dan and I, including brainstorms in the office and other meetings (i.e. EC2 vs Kubernetes pod executors). 

Parallel testing is not strictly a new feature either, we are simply switching from one way to do it, to another, that is compatible with cross-node pipelines. Since we use the declarative method for writing Jenkins Pipelines, this part of the change should be fairly deterministic. I think we agreed at the beginning of this work ([hinted in this thread here](https://github.com/coreos/tectonic-installer/pull/521#issuecomment-300345899)) that alongside with porting the bare-metal tests, we should make them parallel immediately. This will avoid the frustration of having 45min long tests, that might be flaky at the very end during the newly introduced bare-metal tests. It's better to have them fail after 10min, so we don't loose significant amounts of time. Additionally, the parallelism work, due to the way Jenkins was working before that change, we would have to run all stages on the bare-metal machines, which only have a single executor. Running both build, test, AWS smoke tests, and bare-metal smoke tests, in sequence, would lead to extreme contention on the bare-metal machines and might have forced us to spawn a lot of them.

I agree that they come together and it can be scary. However, I believe that the changes are well separated in commits, and are fairly contained and this PR will improve significantly our maintainability and confidence. The Jenkinsfile is modified for it to work with bare-metal testing, the Makefile is modified accordingly, old AWS scripts are deleted, and bare-metal scripts are moved/adapted. Again, I will be more than happy to fix Jenkins!

PR #525, opened 2017-05-03
platform/metal: Add terraform.tfvars.example back

dghubble:
The bare-metal example `terraform.tfvars.example` was deleted in #370 and the auto-generated examples are quite unhelpful. Restore this example.

cc @s-urbaniak @robszumski

### DIFF
--- a/platforms/metal/terraform.tfvars.example
+++ b/platforms/metal/terraform.tfvars.example
@@ -1,0 +1,144 @@
+tectonic_metal_matchbox_http_endpoint = "http://matchbox.example.com:8080"
+tectonic_metal_matchbox_rpc_endpoint = "matchbox.example.com:8081"
+
+tectonic_metal_matchbox_client_cert = <<EOF
+-----BEGIN CERTIFICATE-----
+MIIEYDCCAkigAwIBAgICEAEwDQYJKoZIhvcNAQELBQAwEjEQMA4GA1UEAwwHZmFr
+ZS1jYTAeFw0xNzAyMDEyMjEzMjVaFw0xODAyMDEyMjEzMjVaMBYxFDASBgNVBAMM
+C2Zha2UtY2xpZW50MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAr8S7
+x/tAS6W+aRW3X833OvNfxXjUJAiRkUV85Raln7tqVcTG/9iyRhWgNpUn/WU1/3qV
+obto4ZCURIwoQh0kWk8io1lafZJ+S6Znm3+0TKo7u6QMavolJyetsOQkT/bIoZ73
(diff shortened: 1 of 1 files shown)


robszumski:
I agree that this is useful, but also agree that auto-generation is most sustainable going forward...

dghubble:
Hm, I took the time to write that example to be what users will actually need so its concerning it was discarded without notice. Perhaps try following the bare-metal CLI docs using only the auto-generated https://github.com/coreos/tectonic-installer/blob/master/examples/terraform.tfvars.metal as your guide and assess.

Several steps need to happen for the autogen to be suitable:

1) At the moment, config.tf "common" variables are fairly cloud-centric. Those need to be moved into the cloud platforms themselves for auto-generation to produce useful output. For example, there is no indication that `tectonic_admin_email` needs to be filled in, but the following aren't used:

```
tectonic_worker_count
tectonic_etcd_count
tectonic_cluster_name
tectonic_base_domain
tectonic_master_count
```

2) Autogeneration also needs to faithfully include an example for every field, like the original example. Users can inspect examples and deduce how they can edit to suit their needs far better than they can guess a reasonable value based on a description. This is all the more important because the terraform-based installer now has little validation for CLI users.

dghubble:
Basically, I worry the bare-metal CLI instructions are unusable without this.

s-urbaniak:
@dghubble I am sorry I deleted your examples file without notice, I should have /cc'ed you back then, the change was made in a batch-style mode.

I agree with your assessment regarding baremetal. We could omit auto-generation for this for now, as it indeed differs vastly from the other (cloud-centric) platforms. I can file a follow-up PR with enhanced documentation including more examples as well.

Let me brainstorm how we can split the more general cloud-centric configs (and thus examples/documentation) which are not needed by terraform, we have all the facilities to do so, that is actually a very good idea.

dghubble:
No worries and thanks. You're right, it looks like the generated examples are part of the Makefile process https://github.com/coreos/tectonic-installer/blob/master/Makefile#L82 so I suppose metal would need to be omitted for now or we reference a different example. Just adding back this terraform.tfvars.example as I proposed here won't help the Makefile user.

I think this can be closed if you have your proposed fix.

s-urbaniak:
Yep, we just need to remove the line https://github.com/coreos/tectonic-installer/blob/9c0ed04/Makefile#L77 for now.

Just as an initial thought-experiment:

I think what we need next to `config.tf` is a `cloud.tf` which distills the variables baremetal doesn't need. `cloud.tf` would be linked into openstack, aws, azure, et al but omitted in baremetal. This would document and generate examples for dedicated variables.

This still leaves out dedicated configuration examples in `config.tf` for baremetal, but we I believe we can simply have extensive comments there, as HEREDOC is supported for the description.

dghubble:
That sounds good to me. I'll be on vacation Thursday/Friday so feel free to merge any improvements to examples based on terraform.tfvars.example and @rithujohn191 can sign off on variable changes / validation if that happens before Monday.

s-urbaniak:
Superseded by #536, hence closing

PR #526, opened 2017-05-03
platforms/aws/route53: delete records before deleting the zone

Quentin-M:
Fixes https://github.com/coreos-inc/tectonic-issues/issues/234.

- Should fix eventual consistency related issues & co.
- Should also fix when LB Services have been created: the records will get removed - but probably not the ELBs tho!


Quentin-M:
Tests are failing because master is broken..

sym3tri:
i poked jenkins again

Issue #528, opened 2017-05-03
installer 'make build' doesn't include newly built frontend static assets


sym3tri:

ggreer:
Fixed in #529 (along with a lot of other bugs in the Makefiles).

PR #529, opened 2017-05-03
Improve Makefiles

ggreer:
- Add front-end js as a dependency for binassets.
- Add `make dist` rule to download & create a tarball. (but not upload a release)
- Fix path of frontend assets in `make clean`
- Make all go files a dependency for installer binaries.
- Make all js files a dependency for `yarn build`
- Make `package.json` & `yarn.lock` dependencies for `yarn install`

Fixes multiple issues, such as:
  - `make` rebuilt js even if no js code changed
  - `make` didn't bundle changed js (had to run `make clean && make` to get new js bundled)
  - `make clean` didn't `rm app-bundle.js`
  - `make clean` didn't `rm .workspace` (used for making release)

Fixes issue #528, where make didn't bundle changed js. 

This still rebuilds some go stuff if you run `make`, but it's much better than the current behavior (which rebuilds everything but doesn't bundle it correctly unless you `make clean`).

### DIFF
--- a/installer/Makefile
+++ b/installer/Makefile
@@ -83,3 +83,4 @@ clean:
 	rm -rf bin/
 	rm -f binassets/assets.go
 	rm -rf assets/frontend/scripts
+	rm -rf .workspace/
--- a/installer/Makefile
+++ b/installer/Makefile
@@ -6,23 +6,25 @@ LD_FLAGS="-w -X github.com/coreos/tectonic-installer/installer/server/version.Ve
 
 REPO=github.com/coreos/tectonic-installer/installer
 
+GO_FILES=$(shell find server -type f -name '*.go')
+
 .PHONY: all
 all: build
 
 .PHONY: build
 build: bin/$(OS)/installer
 
-bin/windows/installer.exe: frontend binassets/assets.go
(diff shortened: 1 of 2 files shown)


installer/Makefile
```diff
@@ -6,47 +6,53 @@ LD_FLAGS="-w -X github.com/coreos/tectonic-installer/installer/server/version.Ve
...
+binassets/assets.go: frontend bin/go-bindata $(shell find assets -type f)
+	./bin/go-bindata -pkg binassets -o binassets/assets.go -prefix assets assets/...
 
 .PHONY: frontend
 frontend:
 	$(MAKE) -C frontend
 
 .PHONY: test
 test: build bin/golint
-	@./scripts/test
```
chancez: So you want to print the script being executed here? not just the output of it?

chancez:
This all correct to me. I'd have someone else take a look though.

installer/Makefile
ggreer: In general, it's handy to see exactly what commands `make` is running, and in what order.

chancez: Completely reasonable, just making sure that's why you're doing it.

aaronlevy:
lgtm

ggreer:
OK, can someone approve?

Issue #461, opened 2017-05-02
Bare-Metal UI Cleanup


sym3tri:
Overall usability.

@ggreer please fill out details here.

ggreer:
There's still lots of low-hanging fruit in the baremetal UI flow, but I think it'd be best to create specific issues for those things. eg: making masters/workers lists in the UI instead of "enter number of masters" then clicking next and getting that many inputs.

ggreer:
#603 fixes a lot of this.

sym3tri:
@kyoto can handle this one

Quentin-M:
@ggreer Is there any other UI cleanup we want to do here? Can we close that one?

Quentin-M:
I haven't heard anything from either the UI team or from the product team (asked offline) regarding to further cleanup. Closing. Feel free to open specific issues if needed.

PR #532, opened 2017-05-03
Metal UI cleanup

ggreer:
Related to #461.

- Remove all code related to multiple ssh keys. metal-tf only supports one ssh key
- Warn user that the ssh key must be in the local ssh-agent, otherwise terraform apply won't work.
- Port baremetal ssh key form over to new form class.
- Use SSH_AUTHORIZED_KEY in clusterConfig instead of SSH_AUTHORIZED_KEYS.
- Update tests & example progress file for new ssh key format.


ggreer:
Before: 
screen shot 2017-05-04 at 15 34 16

After:
screen shot 2017-05-04 at 15 30 44

installer/frontend/components/bm-sshkeys.jsx
```diff
@@ -15,104 +14,39 @@ ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQDxL
...
-      <div className="form-group">Access to the nodes is intended for use by admins.
-        End users can run applications using the API, CLI and Tectonic
-        Console and typically don’t need SSH access.</div>
-      <div className="form-group">The public keys below will be added to all machines in this cluster.</div>
-      {fields}
+export const BM_SSHKeys = () => <div>
+  <div className="form-group">
+    Access to the nodes is intended for use by admins.
+    End users can run applications using the API, CLI and Tectonic
+    Console and typically don’t need SSH access.
```
amrutac: nit: Extra <code>and</code>

ggreer: I copy-pasted that from the old baremetal poweron page. It's not the best sentence, but I think it's grammatically correct. eg, with parens to differentiate the clauses:

> End users can run applications using (the API, CLI and Tectonic Console) and typically don’t need SSH access.

PR #539, opened 2017-05-04
Define master_count so that baremetal works again. Broken by PR #468.

ggreer:
Fixes https://github.com/coreos-inc/tectonic-issues/issues/240

Regression introduced in #468

### DIFF
--- a/platforms/metal/tectonic.tf
+++ b/platforms/metal/tectonic.tf
@@ -67,6 +67,7 @@ module "tectonic" {
   kubectl_client_id = "tectonic-kubectl"
   ingress_kind      = "HostPort"
   experimental      = "${var.tectonic_experimental}"
+  master_count      = "${length(var.tectonic_metal_controller_names)}"
 }
 
 data "archive_file" "assets" {


s-urbaniak:
great catch, LGTM!

platforms/metal/tectonic.tf
```diff
@@ -67,6 +67,7 @@ module "tectonic" {
   kubectl_client_id = "tectonic-kubectl"
   ingress_kind      = "HostPort"
   experimental      = "${var.tectonic_experimental}"
+  master_count      = "${var.tectonic_master_count}"
```
s-urbaniak: oh, just a second, afaik baremetal does not recognize `tectonic_master_count`, hence what value should we put here?

s-urbaniak: I vote for defaulting this to `1` for the metal platform, `tectonic_master_count` is not used there, and in fact https://github.com/coreos/tectonic-installer/pull/536 does not even configure this value anymore for baremetal.

ggreer:
OK it's 1.

platforms/metal/tectonic.tf
```diff
@@ -67,6 +67,7 @@ module "tectonic" {
   kubectl_client_id = "tectonic-kubectl"
   ingress_kind      = "HostPort"
   experimental      = "${var.tectonic_experimental}"
+  master_count      = "1"
```
Quentin-M: Should use the length of var.tectonic_metal_controller_names instead.

ggreer:
OK now it's "${len(var.tectonic_metal_controller_names)}"

PR #544, opened 2017-05-04
Clean workspace after build

ethernetdan:
Jenkins executors were filling up with files in build workspaces causing jobs to fail. This cleans the workspace after each execution regardless of success or failure.

Since don't currently perform any caching this should not affect build time.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -105,4 +105,9 @@ pipeline {
       }
     }
   }
+  post {
+    always {
+      deleteDir()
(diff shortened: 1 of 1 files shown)


Quentin-M:
If deleteDir is builtin and if it does not break parallel tests.

ethernetdan:
It is and it gets run always at the [end of pipeline execution](https://jenkins.io/doc/book/pipeline/syntax/#post).

PR #550, opened 2017-05-05
modules/azure: Enable the use of external master & worker subnets

metral:


platforms/azure/variables.tf
```diff
@@ -86,3 +86,15 @@ variable "tectonic_azure_create_dns_zone" {
   description = "If set to true, create an Azure DNS zone"
   default     = "true"
 }
+
+variable "tectonic_azure_external_master_subnet_id" {
+  type        = "string"
+  description = "Subnet ID within an existing VNet to deploy master nodes into. Required to use an existing VNet. Example: the subnet ID starts with `\"/subscriptions/{subscriptionId}\"` or `\"/providers/{resourceProviderNamespace}\"'`. "
```
s-urbaniak: Suggestion: let's use HEREDOC syntax, so this does not end up in one long huge line, escaping of quotes is not necessary any more, and double new-lines will be translated to `<br>` lines in the generated markdown:

```
description = <<EOF
Subnet ID within an existing VNet to deploy master nodes into.
Required to use an existing VNet.

Example: the subnet ID starts with `"/subscriptions/{subscriptionId}"` or `"/providers/{resourceProviderNamespace}"'`. "
EOF
```

s-urbaniak: Additionally I suggest to add (optional) like so:
```
description = <<EOF
(optional) Subnet ID within an existing VNet to deploy master nodes into.
...
EOF
```

The terraform-examples tool will auto-detect the prefix `(optional)` and will comment out the variable in the generated examples tfvars file.

metral:
@s-urbaniak Thanks for the feedback! I've gone ahead and addressed your changes

s-urbaniak:
ok to test

metral:
ok to test

metral:
@s-urbaniak is there anything else needed in this PR?

PR #551, opened 2017-05-05
Always destroy clusters after smoke

ethernetdan:
Moved the cluster destruction invocation the `post` section of the Jenkinsfile which ensures that cluster cleanup regardless of build status.


ethernetdan:
Started destroying successfully after build is aborted

PR #437, opened 2017-05-01
Docs/install: clone repo -> download release

ElijahCaine:
I'm pretty sure I got some of this wrong. Please let me know what needs to be changed.

The purpose of this PR is to direct people to download the release archive instead of cloning the repo to install Tectonic.

PR #553, opened 2017-05-05
Documentation: fix INSTALLER_PATH for release

Quentin-M:
We changed to using a release tarball rather than the compiled version. But forgot to change the path to the installer - which is different in that case.


joshix:
fixed in #437

PR #554, opened 2017-05-05
config.tf: fix tectonic version

Quentin-M:
We are not releasing 1.6.2-tectonic.0 but 1.6.2-tectonic.1 afaik.


Quentin-M:
Heard there is already a PR for it.

PR #556, opened 2017-05-05
added generated directory to .gitignore

sozercan:


s-urbaniak:
@sozercan usually we use the `build` folder to build clusters, each cluster in a separate `build/<cluster>` folder. That directory is already ignored.

sym3tri:
@sozercan I'm going to close this out for now. Please let us know if Serg's comment doesn't make sense or doesn't suite your use-case.

PR #557, opened 2017-05-05
platforms/azure: added ssh key to etcd

### LINKED ISSUE #464: azure: etcd machines are on the Internet and the VMs have a static password
If I'm reading the Terraform correctly, this is putting etcd on the Internet. Is this intentional? Why public over an internal LB for etcd?

https://github.com/coreos/tectonic-installer/blob/master/modules/azure/etcd/lb.tf#L14

Also, is it intentional that the admin password is hardcoded? Is that just for ease of access since SSH isn't exposed to the internet for those machines?

https://github.com/coreos/tectonic-installer/blob/9bdb83ee488c798ee1760dd3ff3854fd3be86179/modules/azure/etcd/etcd.tf#L33

sozercan:
Fixes 1/2 of #464 

Also, added `tectonic_master_count` since that seems like a required variable now.

### DIFF
--- a/modules/azure/etcd/etcd.tf
+++ b/modules/azure/etcd/etcd.tf
@@ -30,12 +30,17 @@ resource "azurerm_virtual_machine" "etcd_node" {
   os_profile {
     computer_name  = "etcd"
     admin_username = "core"
-    admin_password = "Microsoft123!"
+    admin_password = ""
     custom_data    = "${base64encode("${data.ignition_config.etcd.*.rendered[count.index]}")}"
   }
 
   os_profile_linux_config {
-    disable_password_authentication = false
+    disable_password_authentication = true
+
+    ssh_keys {
+      path     = "/home/core/.ssh/authorized_keys"
+      key_data = "${file(var.public_ssh_key)}"
+    }
   }
 }
 
--- a/modules/azure/etcd/ignition.tf
+++ b/modules/azure/etcd/ignition.tf
@@ -15,7 +15,7 @@ data "ignition_user" "core" {
   name = "core"
 
   ssh_authorized_keys = [
-    "${file(var.ssh_key)}",
+    "${file(var.public_ssh_key)}",
   ]
 }
 
--- a/modules/azure/etcd/variables.tf
+++ b/modules/azure/etcd/variables.tf
@@ -33,7 +33,7 @@ variable "cluster_name" {
   type = "string"
 }
 
-variable "ssh_key" {
+variable "public_ssh_key" {
   type = "string"
 }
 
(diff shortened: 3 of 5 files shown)


modules/azure/etcd/etcd.tf
```diff
@@ -30,12 +30,17 @@ resource "azurerm_virtual_machine" "etcd_node" {
...
   }
 
   os_profile_linux_config {
-    disable_password_authentication = false
+    disable_password_authentication = true
+
+    ssh_keys {
+      path     = "/home/core/.ssh/authorized_keys"
+      key_data = "${file(var.public_ssh_key)}"
+    }
```
s-urbaniak: I am wondering, is this not redundant with the ignition config as per [1]?

[1] https://github.com/coreos/tectonic-installer/blob/1827f92aaf5537bc57e66ea48278fdb969c6e37f/modules/azure/etcd/ignition.tf#L14-L20

modules/azure/etcd/ignition.tf
```diff
@@ -15,7 +15,7 @@ data "ignition_user" "core" {
   name = "core"
 
   ssh_authorized_keys = [
-    "${file(var.ssh_key)}",
+    "${file(var.public_ssh_key)}",
```
s-urbaniak: see comment above

modules/azure/etcd/etcd.tf
```diff
@@ -30,12 +30,17 @@ resource "azurerm_virtual_machine" "etcd_node" {
   os_profile {
     computer_name  = "etcd"
     admin_username = "core"
-    admin_password = "Microsoft123!"
```
metral: This PR can be reduced down to just this line as the SSH key is already placed in by Ignition.

And `master_count = "${var.tectonic_master_count}"` has already been addressed in https://github.com/coreos/tectonic-installer/commit/e0c2d67a559aca2d0514c537985b9ce7c0072304#diff-0106e47f11400f0434e4340687831ab7

metral:
@sozercan can you PTAL at the comments above?

@s-urbaniak It'd be great to get this PR in to remove the admin password before the release.

modules/azure/etcd/etcd.tf
s-urbaniak: ok, let's get this in then, we can tackle ssh key redundancy in a follow-up PR.

modules/azure/etcd/etcd.tf
metral: We need this in because by setting `disable_password_authentication = true`, Azure requires SSH to be specified. If not you hit this error like I just did when testing (https://github.com/coreos/tectonic-installer/pull/624):

`* azurerm_virtual_machine.etcd_node: compute.VirtualMachinesClient#CreateOrUpdate: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="InvalidParameter" Message="Authentication using either SSH or by user name and password must be enabled in Linux profile."`

PR #569, opened 2017-05-06
Bump Tectonic Update Channel

alekssaul:
Currently Tectonic Update channel manifests refer to Tectonic-1.5


alekssaul:
Just to note apart from the documentation, PR also updates the Tectonic Update Channel variable that gets used by the Tectonic Update Operator. Currently, tectonic clusters spun off of HEAD will be set to use "Tectonic 1.5" update channel.

PR #573, opened 2017-05-06
Update ELB SG port

snsumner:
Add port 22 to API security group to be able to SSH to master nodes. Please note the ELB already allows SSH but the security group is missing SSH rule.

### DIFF
--- a/modules/aws/vpc/sg-elb.tf
+++ b/modules/aws/vpc/sg-elb.tf
@@ -20,6 +20,14 @@ resource "aws_security_group" "api" {
     from_port   = 443
     to_port     = 443
   }
+
+  ingress {
+    protocol    = "tcp"
+    cidr_blocks = ["0.0.0.0/0"]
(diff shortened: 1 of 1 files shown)


s-urbaniak:
@snsumner thanks for the contribution! Nevertheless we already declare port 22 on the SG for master nodes already, see [1].

The only scenario where I think this might make sense is, when you used in conjunction with `tectonic_aws_external_vpc_id`. But in this case the recommended method would be to spawn a dedicated bastion login host, which is connected to this VPC rather than ssh'ing via the ELB which will end up at an arbitrary master node?!

[1] https://github.com/coreos/tectonic-installer/blob/45dfc2b/modules/aws/vpc/sg-master.tf#L30-L38

sym3tri:
@snsumner Closing this out for now b/c we haven't heard back. Please let us know if the bastion host solution doesn't work for your use-case.

PR #574, opened 2017-05-06
ingress tls_cert_request: add subjectAltName

coyled:
if a cert doesn’t contain subjectAltNames Chrome >=58 and future
versions of Firefox will give a `NET::ERR_CERT_COMMON_NAME_INVALID`
error even if the CA is trusted and the CN matches.

currently the generated CSR only contain a subject CN.  this change
adds a matching subjectAltName to the ingress cert signing request.

### DIFF
--- a/modules/tectonic/crypto.tf
+++ b/modules/tectonic/crypto.tf
@@ -26,6 +26,12 @@ resource "tls_cert_request" "ingress" {
   subject {
     common_name = "${element(split(":", var.base_address), 0)}"
   }
+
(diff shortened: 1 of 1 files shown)


s-urbaniak:
ok to test

s-urbaniak:
@coyled This patch sounds reasonable to me, thanks for the contribution! (/cc @alexsomesan for additional thoughts?)

Do you mind to fix the offending file pointed out by the CI?
```
misformatted files (run 'terraform fmt .' to fix): modules/tectonic/crypto.tf
```

PR #578, opened 2017-05-08
frontend: Re-populate FileArea if file is the same.

### LINKED ISSUE #470: GUI: Improve license upload
The installer refers to "coreos-license.txt" when it is downloaded from the account site as "tectonic-license.txt".

The placeholder text is very confusing, as it looks like the required info is there. This text color needs to be lightened to make it more clear. I think we're using browser defaults, but you can also [customize it](https://stackoverflow.com/questions/2610497/change-an-html5-inputs-placeholder-color-with-css) to #AAAAAA per mocks.

ggreer:
Fixes #470.

### DIFF
--- a/installer/frontend/components/tectonic-license.jsx
+++ b/installer/frontend/components/tectonic-license.jsx
@@ -65,7 +65,7 @@ export const TectonicLicense = () =>
       </div>
       <div className="col-xs-9">
         <Connect field={TECTONIC_LICENSE}>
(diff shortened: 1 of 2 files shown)


amrutac:
Are you sure you are referencing the right issue? Seems like the suggestion was to change the placeholder text color?

ggreer:
Yes. The placeholder color was already fixed. The other issue was that uploading the same file wouldn't re-populate the license field. To reproduce (on master):

- Upload a license file.
- Select the text in the license field & delete it.
- Upload the same license file again.

The field won't be populated with anything. This is because onchange doesn't fire when the same file is loaded again. This PR fixes that.

ggreer:
I did fix the license filename though. Thanks for pointing that out.

Issue #576, opened 2017-05-08
Revert "app" --> "k8s-app" label change for monitoring components


aaronlevy:
ref: https://github.com/coreos/tectonic-installer/pull/517#issuecomment-299404338

PR #580, opened 2017-05-08
modules/.../monitoring: revert label changes

pbx0:
Revert any labels changes to the monitoring assets made in PR:  https://github.com/coreos/tectonic-installer/pull/517

Fixes: https://github.com/coreos/tectonic-installer/issues/576

@brancz @fabxc @aaronlevy

### DIFF
--- a/modules/tectonic/resources/manifests/monitoring/node-exporter-ds.yaml
+++ b/modules/tectonic/resources/manifests/monitoring/node-exporter-ds.yaml
@@ -12,7 +12,7 @@ spec:
     metadata:
       name: node-exporter
       labels:
-        k8s-app: node-exporter
+        app: node-exporter
     spec:
       hostNetwork: true
       hostPID: true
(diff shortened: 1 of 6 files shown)


Documentation/files/logging/monitoring/fluentd-prometheus-service-monitor.yaml
```diff
@@ -4,12 +4,12 @@ metadata:
   name: fluentd
   namespace: logging
   labels:
-    k8s-app: fluentd
```
brancz: These are actually not maintained by team monitoring (I believe they were contributed by @chancez), so they don't need to be reverted, they can have the consistent labeling already.

brancz:
lgtm

Documentation/files/logging/monitoring/fluentd-prometheus-service-monitor.yaml
pbx0: Sounds good, I un-reverted this doc change.

PR #587, opened 2017-05-08
frontend: Disable experimental option for baremetal.

ggreer:
It doesn't work yet.


ggreer:
Apparently we don't want this anymore and the plan is to add the new feature to baremetal and delay the release.

dghubble:
Apparently. I think the safest option is to ship without since experimental self-hosted etcd has never been tested or stabilized on bare-metal and constitutes a new feature introduced in a release candidate. Defer to product to make the right call.

Quentin-M:
I soon have the PR.

PR #590, opened 2017-05-08
Fix license and pull secret / reenable Pod logging test

ethernetdan:
The pull secret and license had been swapped in the Jenkinsfile causing Tectonic resources to not be created since the pull secret was missing.

Also enabled Pod logging test since the cause of it's failures should be this issue.


Quentin-M:
Fixed IRL :D Auto-approving!

Issue #571, opened 2017-05-06
AWS Terraform config doesn't work due to invalid aws_autoscaling_group argument


aknuds1:
The Terraform config for AWS doesn't work since an invalid [aws_autoscaling_group](https://www.terraform.io/docs/providers/aws/r/autoscaling_group.html) is in use: `tags`:

    terraform plan ../../platforms/aws
    2 error(s) occurred:

    * module.masters.aws_autoscaling_group.masters: : invalid or unknown key: tags
    * module.workers.aws_autoscaling_group.workers: : invalid or unknown key: tags

s-urbaniak:
Thanks for the report! Nevertheless, this is not a bug, but a feature ;-) We rely on a not-yet-merged AWS tagging feature, see [1] and include a custom AWS plugin in this repository.

Please follow the top-level README [2] to build an installer binary which includes the custom AWS plugin. More concretely it should be sufficient to invoke `make build` in the `installer` directory.

I am closing this issue then for now.

[1] https://github.com/hashicorp/terraform/pull/13574
[2] https://github.com/coreos/tectonic-installer/blob/master/README.md

aknuds1:
@s-urbaniak I'm wondering if something changed, since the build fails differently from before: 

    $ PLATFORM=aws make plan
    Using build directory [/Users/arve/go/src/github.com/coreos/tectonic-installer/build/demo]
    sed "s|<PATH_TO_INSTALLER>|/Users/arve/go/src/github.com/coreos/tectonic-installer/installer/bin/darwin/installer|g" terraformrc.example > .terraformrc
    cd /Users/arve/go/src/github.com/coreos/tectonic-installer/build/demo && TERRAFORM_CONFIG=/Users/arve/go/src/github.com/coreos/tectonic-installer/.terraformrc terraform plan /Users/arve/go/src/github.com/coreos/tectonic-installer/platforms/aws
    Failed to load root config module: Error loading modules: module vpc: not found, may need to be downloaded using 'terraform get'
    make: *** [plan] Error 1

What do you think?

aknuds1:
@s-urbaniak I found out that the reason for the module error was that the directory build/<cluster>/.terraform already existed, which in turn means that make won't invoke `terraform get`. Might it be a good idea to always invoke `terraform get` to ensure that modules are up to date?

aknuds1:
@s-urbaniak I made a PR for always running `terraform get` instead of checking that the .terraform dir exists.

PR #591, opened 2017-05-08
Always invoke terraform get

aknuds1:
Always invoke [terraform get](https://www.terraform.io/docs/commands/get.html) in order to ensure that necessary modules are installed, to avoid build errors, rather than relying on whether the .terraform directory exists, since it might be stale.

See #571 for reference.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,16 +21,16 @@ installer-env: $(INSTALLER_BIN) terraformrc.example
 localconfig:
 	mkdir -p $(BUILD_DIR)
 
-$(BUILD_DIR)/.terraform:
+terraform-get:
 	cd $(BUILD_DIR) && $(TF_CMD) get $(TOP_DIR)/platforms/$(PLATFORM)
 
-plan: installer-env $(BUILD_DIR)/.terraform
+plan: installer-env terraform-get
 	cd $(BUILD_DIR) && $(TF_CMD) plan $(TOP_DIR)/platforms/$(PLATFORM)
 
-apply: installer-env $(BUILD_DIR)/.terraform
+apply: installer-env terraform-get
 	cd $(BUILD_DIR) && $(TF_CMD) apply $(TOP_DIR)/platforms/$(PLATFORM)
 
-destroy: installer-env ${BUILD_DIR}/.terraform
+destroy: installer-env terraform-get
 	cd $(BUILD_DIR) && $(TF_CMD) destroy -force $(TOP_DIR)/platforms/$(PLATFORM)
 
 payload:
@@ -113,4 +113,4 @@ structure-check:
 canonical-syntax:
 	terraform fmt -list .
 
-.PHONY: make clean terraform terraform-dev structure-check canonical-syntax docs examples
+.PHONY: make clean terraform terraform-dev structure-check canonical-syntax docs examples terraform-get


Makefile
```diff
@@ -21,16 +21,17 @@ installer-env: $(INSTALLER_BIN) terraformrc.example
...
-plan: installer-env $(BUILD_DIR)/.terraform
+plan: installer-env terraform-get
 	cd $(BUILD_DIR) && $(TF_CMD) plan $(TOP_DIR)/platforms/$(PLATFORM)
 
-apply: installer-env $(BUILD_DIR)/.terraform
+apply: installer-env terraform-get
 	cd $(BUILD_DIR) && $(TF_CMD) apply $(TOP_DIR)/platforms/$(PLATFORM)
 
-destroy: installer-env ${BUILD_DIR}/.terraform
+destroy: installer-env terraform-get
```
ggreer: Are destroy, apply, & plan supposed to be phony? In other words: do you want them to run if a file named `destroy` is in the current dir?

aknuds1: Will my change make `destroy`, `apply` and `plan` phony? I'm not a Make expert, but I figured that only the `terraform-get` target would become phony?

aknuds1: I just tested creating such files FWIW (destroy/apply/plan), it seems the targets are executed no matter what, even if they exist. I would expect Make to skip them since corresponding files exist. This behaviour is the same without my patch.

ggreer: It's always building because its dependencies also have missing outputs. I think your PR is better than what's currently in master, so I don't think it should be blocked on this.

I just didn't realize the Makefile was so hacky.

aknuds1: Haha thanks for clarifying this @ggreer :) I guess these targets *should* be phony.

alexsomesan: @aknuds1 

Can you please give a bit of background on why this change is needed?
What is your use-case where you need to run get before every apply?

The current behaviour is to invoke get only when module cache is missing.

aknuds1: The case is when the cache is missing one or more modules. I don't know how it came about, but I imagine it's because modules were added in the source, so that the cache became stale. It's only logical that the cache can become out of sync with the source, we can't expect that it will stay relevant over time.

alexsomesan: You're right that a new module being added would not trigger a get. In that sense, your change is brings more convenience. There is also a downside because this is invisible to the user and could "bite" unsuspecting users into applying changes that they weren't expecting. 

However, since the cache is nothing more than a bunch of symlinks to the actual module files, changes to those could also bite in the same way with the current state of things.

Things being the way they are, I think your change is useful and won't surprise with unexpected side-effects.

Please rebase and it should be good to merge.

aknuds1:
Thanks @alexsomesan, I've rebased.

Makefile
```diff
@@ -21,16 +21,17 @@ installer-env: $(INSTALLER_BIN) terraformrc.example
 localconfig:
 	mkdir -p $(BUILD_DIR)
 
-$(BUILD_DIR)/.terraform:
+.PHONY: terraform-get
```
alexsomesan: One more thing: there already is a .PHONY section (at the end of the file).
Could you please move this over there?
It makes spotting target blocks easier when reading the Makefile.

alexsomesan:
Thanks! I forgot one thing: see comment about the .PHONY location.

aknuds1:
@alexsomesan Moved the phony declaration!

alexsomesan:
ok to test

Issue #552, opened 2017-05-05
docs: guide for jumping from GUI to CLI for AWS


robszumski:
Guide users on how to use the GUI as a config generator, but then use the CLI to complete the installation. We will link to this guide on the "manual boot" screen that you can select instead of submitting your config.

Topics:
 - download assets
 - set up env vars, path, etc
 - terraform plan
 - terraform apply
 - access the cluster
 - link to destroy docs

zbwright:
The ask is to generate a doc that users can access after having walked through Tectonic Installer to the Submit page. Using the GUI as a config generator is just an opening sentence. This doc will explain how to do the things listed above.
Link out from here: 
![installer-submit](https://cloud.githubusercontent.com/assets/9601502/25758951/5a641aa0-3185-11e7-879b-281ca6f5a706.png)

robszumski:
Quick correction, you actually get linked out from this screen, which is where you land after clicking the link:

![image](https://cloud.githubusercontent.com/assets/755540/25818321/1a45534c-33df-11e7-87b2-62438472b225.png)

sym3tri:
Need to include a blurb about what type of changes we support in the product:
Only changes to underlying infrastructure only... changes to Kubernetes manifests or Tectonic components run in the cluster are not supported.

ggreer:
Now that #633 has been merged, it looks like this:
![screen shot 2017-05-10 at 11 14 45](https://cloud.githubusercontent.com/assets/200121/25914100/ef1d6288-3571-11e7-8b8f-7771f9443034.png)

The link is to https://coreos.com/tectonic/docs/latest/install/aws/manual-boot

robszumski:
@ggreer with a `.html` on the end?

robszumski:
This doc merged in https://github.com/coreos/tectonic-installer/issues/594

PR #594, opened 2017-05-08
docs: adding a page for manual boot.

zbwright:
Adding this as a PR, in the interest of speed. Please flesh this out. I am especially missing instruction on how to customize the files contained in the assets folder for use with Tectonic.


Documentation/install/aws/manual-boot.md
```diff
@@ -0,0 +1,49 @@
...
+## Use assets.zip to customize your deployment
+
+I would add how to 1) navigate to the folder 2) make two or three edits, to address commonly requested changes, then 3) save the work out for use in deployment.
+
+## Deploy the cluster
+
+Once customized, use `terraform plan` to test your deployment, then `terraform apply` to deploy your cluster.
+
+```
+$ terraform plan -var-file=build/${CLUSTER}/terraform.tfvars platforms/aws
```
ElijahCaine: @sym3tri @robszumski I get an error when running a version of this command:
```
$ terraform plan -var-file=terraform.tfvars platforms/aws
2 error(s) occurred:

* module.masters.aws_autoscaling_group.masters: : invalid or unknown key: tags
* module.workers.aws_autoscaling_group.workers: : invalid or unknown key: tags
```

sym3tri: This happens if you're not using the correct `.terraformrc` file pointing to the installer binary as a TF plugin

sym3tri: be sure you follow the same steps, prior to the `terraform destroy` command.

https://github.com/coreos/tectonic-installer/blob/master/Documentation/install/aws/uninstall.md#destroy-the-cluster

Documentation/install/aws/manual-boot.md
```diff
@@ -0,0 +1,49 @@
...
+
+First, use Tectonic Installer to generate and download cluster assets. Walk through the Installer to configure the cluster, then click *Download assets* from the *Start Installation* screen.
+
+## Initialize Terraform
+
+Follow the instructions to [initialize and configure Terraform][initialize-tf].
+
+## Use assets.zip to customize your deployment
+
+I would add how to 1) navigate to the folder 2) make two or three edits, to address commonly requested changes, then 3) save the work out for use in deployment.
```
robszumski: Let's use the example of making two clusters based off the same config, but in different regions. Basically, edit the tfvars to have a new region, domain and cluster name. Note that region is defined as an env var right now I believe.

robszumski:
Fixes https://github.com/coreos/tectonic-installer/issues/552

Documentation/install/aws/manual-boot.md
```diff
@@ -0,0 +1,68 @@
...
+Assets generated by Tectonic Installer allow you to make changes to the underlying infrastructure for a cluster. They do not provide support for changes to Kubernetes manifests, nor to Tectonic components run in the cluster (such as etcd or clair).
+
+These downloaded assets are also necessary to destroy existing clusters, once they have been deployed.
+
+## Generate and download cluster assets
+
+First, use Tectonic Installer to generate and download cluster assets.
+
+* Walk through the Installer in your browser, entering credentials and cluster information as usual.
+* On the *Start Installation* screen **do not start the installation**
```
robszumski: where does "start installation" appear on the screen?

Documentation/install/aws/manual-boot.md
```diff
@@ -0,0 +1,68 @@
...
+* Instead, click the *Download assets* link. This downlods the Terraform files into a file called `assets.zip`.
```
robszumski: `downloads`

Documentation/install/aws/manual-boot.md
```diff
@@ -0,0 +1,68 @@
...
+Unzip the `assets.zip` archive and navigate to the folder created by unzip.
+
+```sh
+$ unzip assets.zip # creates `my-cluster_2017_05_15-13-31`
+$ cd my-cluster_2017_05_15-13-31
+$ ls
+config.tf        license.txt      logs             modules          platforms        pull_secret.json terraform.tfvars
+```
+
+You may edit the file `terraform.tfvars` before moving on to change things like worker count, region, ssh key, and more.
```
robszumski: I'd take region out of here, as its not in the vars file.

robszumski: Nevermind, I see that region is in the sample aws file

Documentation/install/aws/manual-boot.md
```diff
@@ -13,33 +13,30 @@ These downloaded assets are also necessary to destroy existing clusters, once th
...
-```sh
+```bash
 $ unzip assets.zip # creates `my-cluster_2017_05_15-13-31`
 $ cd my-cluster_2017_05_15-13-31
 $ ls
 config.tf        license.txt      logs             modules          platforms        pull_secret.json terraform.tfvars
 ```
 
-You may edit the file `terraform.tfvars` before moving on to change things like worker count, region, ssh key, and more.
+Then, edit `terraform.tfvars` to customize your deployment. For example, to create two matching clusters in two different regions, edit one node's `tectonic_aws_region` value to set it to a different region, then edit the `tectonic_worker_count` value appropriately.
```
robszumski: I think you'll need to add "then edit the subnets to reflect the new region"

Documentation/install/aws/manual-boot.md
```diff
@@ -0,0 +1,65 @@
+# Manually boot clusters with Terraform
+
+The Tectonic Installer can be used to easily create Terraform assets and configuration files used to provision a Tectonic cluster via the terminal.
+
+This process may be used to create multiple clusters with similar configurations, to deploy a single cluster multiple times, or to quickly generate cluster templates for distribution to your team.
+
+Assets generated by Tectonic Installer allow you to make changes to the underlying infrastructure for a cluster. They do not provide support for changes to Kubernetes manifests, nor to Tectonic components run in the cluster (such as etcd or clair).
```
alexsomesan: This statement is a bit confusing. Technically you can make changes to the Kubernetes manifests that the installer comes with (modules/bootkube/resources/manifests), before creating a cluster.

Do you mean the installer doesn't allow to make changes to the Kubernetes components already deployed on a running cluster?

zbwright: @robszumski or @sym3tri for the answer to that one.

robszumski: @alexsomesan We're trying to steer folks away from modifying things in `kube-system` or `tectonic-system`, as we won't support those changes.

@zbwright We should probably tweak that to say that changes to the Kubernetes manifests voids your ability to auto-update, with more docs to come soon that expand that out

alexsomesan: @robszumski That approach sound good.

Documentation/install/aws/manual-boot.md
```diff
@@ -0,0 +1,65 @@
...
+In the `tectonic` installer assets directory you configured in the previous step, unzip the `assets.zip` archive and navigate to the folder created.
+
+```bash
+$ unzip assets.zip # creates `my-cluster_2017_05_15-13-31`
+$ cd my-cluster_2017_05_15-13-31
+$ ls
+config.tf        license.txt      logs             modules          platforms        pull_secret.json terraform.tfvars
+```
+
+Then, edit `terraform.tfvars` to customize your deployment. For example, to create two matching clusters in two different regions, edit one node's `tectonic_aws_region` value to set it to a different region, edit the subnets to reflect the new region, then edit the `tectonic_worker_count` value appropriately.
```
sym3tri: If your mentioning multiple clusters, you'll need to mention that each cluster should have its own directory and own state file. So you'd need to make a copy of the assets.zip for each cluster you want to create.

zbwright: so - copy / paste to create multiple 'assets' folders, then edit one terraform.tfvars per folder to customize the cluster that will be created, then run terraform plan / apply on each in sequence, correct? (If this is how you use the one download to create multiple clusters, it's important to call it out correctly. I don't want to get it wrong.)

ElijahCaine: This is a relevant doc. https://www.terraform.io/docs/state/
Ultimately the reason you need multiple copies of the assets is so different states don't step on eachother. Each assets instance doesn't need to be different necessarily, just needs a different directory so each instance has it's own state.

zbwright: ok. try this:

Documentation/install/aws/manual-boot.md
sym3tri: +1

Documentation/install/aws/manual-boot.md
sym3tri: @zbwright what you described is correct.

Documentation/install/aws/manual-boot.md
```diff
@@ -0,0 +1,74 @@
...
+
+When complete, your Tectonic cluster will be ready for access.
+
+## Use assets.zip to create multiple deployments
+
+If you wish to use downloaded assets to deploy multiple clusters, first make a copy of the directory for each planned instance. Then edit each directory's `terraform.tfvars`, and deploy individually.
+
+For example, to create two matching clusters in two different regions, first create two copies of the assets folder. Edit `terraform.tfvars` in one directory to change one node's `tectonic_aws_region` value to a different region, and edit the subnets to reflect the new region. Leave the second directory's `terraform.tfvars` file in its default state.
+
+Then, run `terraform plan` and `terraform apply` on each `terraform.tfvars` file in turn to deploy your Tectonic clusters.
```
sym3tri: I just realized the paths in the `.terraformrc` might prevent this from working. @Quentin-M can you verify?

If not, let's just skip all mention of multiple clusters for now.

Quentin-M: It will work, the only thing that `.terraformrc` has is an absolute path to the installer binary. Therefore multiple asset folders will work properly, as long as they copy the whole folder and not only the visible files.

zbwright: so, we're good? or should I make this more emphatic?

Quentin-M: We good 👍

PR #598, opened 2017-05-08
Export CLUSTER variable from Makefile as Terraform variable

aknuds1:
The `tectonic_cluster_name` Terraform variable currently has to be defined by the user, but there is a corresponding Makefile variable `CLUSTER` (with a default of `demo`). I figure it might make sense to link the two variables, so that Make configures Terraform's `tectonic_cluster_name` variable via `CLUSTER`.

I'm not 100% sure if it's a good idea to make the two variables correspond to each other, but it seems reasonable enough to me at least, to not have to configure them separately from each other. Creating this PR as a proposal to introduce this behaviour, looking forward to hearing your thoughts on it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,7 @@ TF_DOCS := $(shell which terraform-docs 2> /dev/null)
 TF_EXAMPLES := $(shell which terraform-examples 2> /dev/null)
 TF_RC := $(TOP_DIR)/.terraformrc
 TF_CMD = TERRAFORM_CONFIG=$(TF_RC) terraform
+export TF_VAR_tectonic_cluster_name = $(CLUSTER)
 
 $(info Using build directory [${BUILD_DIR}])
 


s-urbaniak:
This change sounds reasonable to me, @alexsomesan any additional thoughts/doubts?

alexsomesan:
@aknuds1 @s-urbaniak The change itself is reasonable. I do think this is a nice convenience feature.

Unfortunately it doesn't work as expected when used in combination with assets downloaded from the UI installer, because in that case the folder name is different that the actual value for `tectonic_cluster_name` (it has a timestamp appended). We would need to alter that behaviour before we can incorporate this change.

s-urbaniak:
ok to test

alexsomesan:
This is blocked on a behaviour quirk of the UI installer.
Since there are discussions going on about overhauling and unifying the UX between the CLI and UI experiences, I will close this but keep in mind this feature for the future overhaul.

PR #602, opened 2017-05-09
modules/azure: refactor FQDN conditionals and their usage

metral:
This PR:
- cleans up the conditional construction of the FQDN's to only occur in the module outputs.
- ensures that the outputs won't use the `base_domain` if `use_custom_fqdn` is not requested, and instead rely on the FQDN Azure sets up.
- removes duplicate `master_count = "${var.tectonic_master_count}"` introduced in https://github.com/coreos/tectonic-installer/pull/557

### DIFF
--- a/modules/azure/dns/variables.tf
+++ b/modules/azure/dns/variables.tf
@@ -31,6 +31,5 @@ variable "etcd_ip_addresses" {
 }
 
 variable "use_custom_fqdn" {
-  type    = "string"
-  default = "false"
+  default = true
 }
--- a/modules/azure/master/output.tf
+++ b/modules/azure/master/output.tf
@@ -7,17 +7,17 @@ output "console_ip_address" {
 }
 
 output "ingress_external_fqdn" {
-  value = "${azurerm_public_ip.tectonic_console_ip.domain_name_label}.${var.base_domain}"
+  value = "${var.use_custom_fqdn ? "${azurerm_public_ip.tectonic_console_ip.domain_name_label}.${var.base_domain}" : azurerm_public_ip.tectonic_console_ip.fqdn}"
 }
 
 output "ingress_internal_fqdn" {
-  value = "${azurerm_public_ip.tectonic_console_ip.domain_name_label}.${var.base_domain}"
+  value = "${var.use_custom_fqdn ? "${azurerm_public_ip.tectonic_console_ip.domain_name_label}.${var.base_domain}" : azurerm_public_ip.tectonic_console_ip.fqdn}"
 }
 
 output "api_external_fqdn" {
-  value = "${azurerm_public_ip.tectonic_api_ip.domain_name_label}.${var.base_domain}"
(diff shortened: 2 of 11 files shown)


metral:
ok to test

modules/azure/dns/variables.tf
```diff
@@ -32,5 +32,5 @@ variable "etcd_ip_addresses" {
 
 variable "use_custom_fqdn" {
```
alexsomesan: Terraform supports boolean variables.
https://www.terraform.io/docs/configuration/variables.html#booleans

Any particular limitation on not using boolean here?

modules/azure/master/output.tf
```diff
@@ -7,17 +7,17 @@ output "console_ip_address" {
 }
 
 output "ingress_external_fqdn" {
-  value = "${azurerm_public_ip.tectonic_console_ip.domain_name_label}.${var.base_domain}"
+  value = "${var.use_custom_fqdn == "true" ? "${azurerm_public_ip.tectonic_console_ip.domain_name_label}.${var.base_domain}" : azurerm_public_ip.tectonic_console_ip.fqdn}"
```
alexsomesan: If using a boolean, you can omit the comparison and clean up the expression a bit more.

modules/azure/master/variables.tf
```diff
@@ -88,3 +88,8 @@ variable "tectonic_service_disabled" {
   description = "Specifies whether the tectonic installer systemd unit will be disabled. If true, no tectonic assets will be deployed"
   default     = false
 }
+
+variable "use_custom_fqdn" {
```
alexsomesan: Same point about booleans (as above).

metral:
Thanks for the review - all feedback has been addressed.

metral:
rebased

metral:
ok to test

modules/azure/master/output.tf
```diff
@@ -7,17 +7,17 @@ output "console_ip_address" {
...
 }
 
 output "api_external_fqdn" {
-  value = "${azurerm_public_ip.tectonic_api_ip.domain_name_label}.${var.base_domain}"
+  value = "${var.use_custom_fqdn ?  "${azurerm_public_ip.tectonic_api_ip.domain_name_label}.${var.base_domain}" : azurerm_public_ip.tectonic_api_ip.fqdn}"
 }
 
 output "api_internal_fqdn" {
-  value = "${azurerm_public_ip.tectonic_api_ip.domain_name_label}.${var.base_domain}"
+  value = "${var.use_custom_fqdn ?  "${azurerm_public_ip.tectonic_api_ip.domain_name_label}.${var.base_domain}" : azurerm_public_ip.tectonic_api_ip.fqdn}"
```
cehoffman: Should the internal version not always be the azure domain? That would route all traffic over the private network instead of hitting the public ip.

metral: This PR is aimed at refactoring the usage of the right FQDN's for setups that operate over public networks. Private networks would need to disable public IP's on the LB's, and leverage either private IP's or DNS instead, but that is not in scope for this PR.

metral:
@s-urbaniak Is it possible to get this merged? This is a fix for an existing bug in setups that don't need/want to use `use_custom_fqdn`

PR #604, opened 2017-05-09
Add setup for deploying to DigitalOcean

aknuds1:
This is a port of Tectonic Installer to DigitalOcean.


modules/digitalocean/etcd/ignition.tf
```diff
@@ -0,0 +1,61 @@
...
+Environment="ETCD_IMAGE=${var.container_image}"
+ExecStart=
+ExecStart=/usr/lib/coreos/etcd-wrapper \
+  --name=etcd \
+  --discovery-srv=${var.base_domain} \
+  --advertise-client-urls=http://${var.cluster_name}-etcd-${count.index}.${var.base_domain}:2379 \
+  --initial-advertise-peer-urls=http://${var.cluster_name}-etcd-${count.index}.${var.base_domain}:2380 \
+  --listen-client-urls=http://0.0.0.0:2379 \
+  --listen-peer-urls=http://0.0.0.0:2380
+EOF
```
glevand: Seems like you should just use the etcd environment variables here and not redefine ExecStart, so something like: 

    Environment="ETCD_ADVERTISE_CLIENT_URLS=http://${var.cluster_name}-etcd-${count.index}.${var.base_domain}:2379"

aknuds1: I copied this configuration from modules/aws/etcd/ignition.tf. I guess the same suggestion holds for that implementation?

glevand: If the ExecStart command in the CoreOS provided etcd-member.service ever changes you won't get those updates if you redefine ExecStart in the drop-in.

aknuds1: @glevand I mean, should we make the same change in the AWS module, since my configuration was copied from there?

glevand: @crawford What do you think?

modules/digitalocean/etcd/ignition.tf
```diff
@@ -0,0 +1,61 @@
...
+  dropin = [
+    {
+      name = "40-etcd-cluster.conf"
+
+      content = <<EOF
+[Service]
+Environment="ETCD_IMAGE=${var.container_image}"
+ExecStart=
+ExecStart=/usr/lib/coreos/etcd-wrapper \
+  --name=etcd \
```
alekssaul: I believe you'll need to set `--name=etcd` to hostname. Otherwise all members of the cluster will attempt to register with their name etcd

s-urbaniak:
@aknuds1 sorry, that this fell under the radar. I actually thing this is a really great effort and you made quite an impressive progress on that. Thanks a lot for the contribution! I picked up this PR locally and got pretty far with it, finalizing the last bits. Are you interested in finalizing this PR?

modules/digitalocean/etcd/ignition.tf
```diff
@@ -0,0 +1,156 @@
...
+${var.tls_enabled ? "Environment=\"LOCKSMITHD_ETCD_KEYFILE=/etc/ssl/etcd/client.key\"" : ""}
+${var.tls_enabled ? "Environment=\"LOCKSMITHD_ETCD_CERTFILE=/etc/ssl/etcd/client.crt\"" : ""}
+Environment="LOCKSMITHD_ENDPOINT=${var.tls_enabled ? "https" : "http"}://etcd-${count.index}.etcd.${var.cluster_domain}:2379"
+EOF
+    },
+  ]
+}
+
+data "template_file" "etcd-cluster" {
+  template = "${file("${path.module}/resources/etcd-cluster.tpl")}"
```
s-urbaniak: @aknuds1 👍 That is a great idea, using a generic template for this. I actually went a step further and the the same for `40-etcd-cluster.conf`

aknuds1:
Yes I am interested 🙂 I would've worked more in it if it weren't for the
fact that I got a full time project in the meantime, also doing Kubernetes
but with kops. I'm currently getting into more high level stuff, i.e.
setting up ELK for logging and Prometheus for monitoring.

I will see if I can find the time to make this port work again! Last time I
synced with upstream and tried it was broken unfortunately...

modules/digitalocean/etcd/ignition.tf
aknuds1: I would like to do more "generification" of the whole project in order to minimize code duplication between the different platforms. An idea I have in mind is to generate Terraform files from templates, through e.g. a Python script (I do this with all my Kubernetes manifests). Making generic configurations with Terraform gets pretty painful at some point I find.

s-urbaniak: agreed, doing this in TF is not adorable, but unfortunately we can't sprinkle in another language in here, because it is used in GUI based installer. I have been thinking of introducing custom Terraform data sources for other reasons, and it could solve some of the generification problems. At least we could then use Go code then to set up more complex content data structures.

s-urbaniak:
@aknuds1 great! let me finalize my local branch to a working state. I squashed everything into one commit to be able to move forward. You could cherry-pick it and re-author it under your name, such that the credit won't get lost. Hang on, while I am finalizing this.

aknuds1:
@s-urbaniak Thanks! Let me know when it is ready please :)

s-urbaniak:
@aknuds1 I have a working cluster using the following (squashed) commit based on your PR: https://github.com/s-urbaniak/tectonic-installer/commit/d79cd1cf0921d79d1f0768f881dddeb03c39a07c

Admittedly it leaves out a few details you had, i.e. enabling swap, the sshguard, but this is to align it more with the current setup we have at the other cloud providers.

Regarding DNS my commit also assumes you configured a domain already yourself, that is also what we assume in the other platforms.

There's a few tweaks I would like to make, namely exposing masters/workers behind LBs and factor the DNS setup out in their own package. Once done, we can finalize this PR.

s-urbaniak:
having said that, a LB costs quite some money, so if we are targeting DO for simple try-out setups, we can certainly also live with the current solution ;-)

aknuds1:
@s-urbaniak Could you base your branch on mine? I can't just cherry-pick it since there are tons of conflicts between the two.

aknuds1:
@s-urbaniak @alekssaul Could you help me in figuring out how to write acceptance tests for this platform, according to the current model, so that I can finish it and submit for review?

cpanato:
hello @aknuds1, thanks for the PR, can you please rebase with our current master, then I can run  our smoke tests for other platforms just to validate if this change do not break others.

thanks

cpanato:
Thanks, @aknuds1, we will prioritize this PR for our next sprint :)

aknuds1:
@cpanato Thanks, appreciate you guys looking into this!

sym3tri:
Hey @aknuds1 the first thing we'll need here is to get this branch rebased on the latest master and remove all the merge commits. Also please follow our commit message guidelines here: https://github.com/coreos/tectonic-installer/blob/master/CONTRIBUTING.md#format-of-the-commit-message

aknuds1:
Thanks @sym3tri, will get on it when time permits, hopefully this weekend :)

aknuds1:
@sym3tri I rolled everything into one commit, trying to follow your commit message guidelines. I didn't add any subsystem prefix to the commit message though, since I'm really adding a subsystem. If you think I should add digitalocean: as prefix to the commit message though I can do so.

Where should we go from here? I think I will need to sync to the changes that've been made to the master branch in the meantime and make things work again, as I'm pretty sure my branch is currently broken.

Are there tests I should write?

modules/digitalocean/etcd/dns.tf
```diff
@@ -0,0 +1,14 @@
...
+  name = "etcd.${var.cluster_domain}"
+  ip_address = "${digitalocean_droplet.etcd_node.*.ipv4_address[0]}"
+}
+
+resource "digitalocean_record" "etcd_nodes" {
+  count = "${var.droplet_count}"
+  domain = "${digitalocean_domain.etcd.id}"
+  type = "A"
+  name = "etcd-${count.index}"
+  value = "${digitalocean_droplet.etcd_node.*.ipv4_address[count.index]}"
```
squat: The IPv4 address should not be indexed by count. Instead, the selected IP address should always be `0`, no?

modules/sshguard/resources/services/sshguard.service
```diff
@@ -0,0 +1,32 @@
+[Unit]
```
squat: let's remove this sshguard service for now. It is non-standard across platforms, uses an image that is not maintained, and cutting it out for now will simplify this PR.

modules/digitalocean/swap/ignition.tf
```diff
@@ -0,0 +1,12 @@
+data "template_file" "swap" {
```
squat: Similarly, can we remove this module for now? This will greatly simplify this PR and the number of variables to maintain. We do not enable a swapfile on any other platform currently, so this behavior is a departure from that. It is very reasonable for a pre-alpha platform to simplify code at the expense of suggesting that user's deploy larger machines in order to accommodate memory needs.

config.tf
```diff
@@ -467,3 +467,9 @@ variable "tectonic_bootstrap_upgrade_cl" {
   default     = "true"
   description = "(internal) Whether to trigger a ContainerLinux upgrade on node bootstrap."
 }
+
+variable "tectonic_do_ssh_key_path" {
```
squat: this variable should not be declared here but rather in `platforms/digitalocean/variables.tf`. This file is for common variables only. Otherwise, this change would cause ALL clusters, even those deployed to AWS, Azure, etc to declare a DO ssh key path.

.gitignore
```diff
@@ -14,3 +14,5 @@ bin/
 bin_test/
 matchbox/
 modules/update-payload/generated/
+# Allow for user configuration
```
squat: Let's remove this for now so we can merge this PR.

Makefile
```diff
@@ -12,7 +12,7 @@ TF_DOCS := $(shell which terraform-docs 2> /dev/null)
 TF_EXAMPLES := $(shell which terraform-examples 2> /dev/null)
 TF_CMD = terraform
 TEST_COMMAND = /bin/bash -c "bundler exec rspec spec/${TEST}"
-
+export TF_VAR_tectonic_cluster_name = $(CLUSTER)
```
squat: Please remove this line. This PR does not need to introduce this coupling.

modules/digitalocean/etcd/dns.tf
```diff
@@ -0,0 +1,14 @@
+# Create separate domain for etcd nodes, because depending on main cluster domain would cause
+# a dependency cycle, as the latter must be instantiated with the API server IP address
```
squat: This diverges from how the other platforms are created. All other platforms expect your cloud provider account to already have the `tectonic_base_domain`. Then we create etcd DNS entries like: `${var.cluster_name}-etcd-${count.index}.${var.tectonic_base_domain}`. This avoids any dependency cycle, eliminates the need for an extra domain, `etcd.${tectonic_base_domain}`, and puts this platform in line with the other platforms.

squat:
@aknuds1 let's add tests in a follow up as this PR is large enough as is.

sym3tri:
> If you think I should add digitalocean: as prefix to the commit message though I can do so.

Yes, please.

aknuds1:
Thanks guys, will look at your requests tonight!!

modules/digitalocean/etcd/dns.tf
aknuds1: @squat Why should the IP address not be indexed by count? The purpose here is to create a DNS record per etcd cluster member isn't it, pointing to its IP address? I mean, as far as I can tell it perfectly correlates with [the AWS implementation](https://github.com/coreos/tectonic-installer/blob/master/modules/aws/etcd/dns.tf#L7).

aknuds1:
@squat I tried creating a cluster now, my branch contains errors according to Terraform though. How about I try to fix the breakage in the meantime before considering the DNS change you want me to make?

modules/digitalocean/etcd/dns.tf
squat: Yes sorry, please disregard. I read it incorrectly.

aknuds1:
@squat I'm trying to bring my branch back up to date and into working order, in a second commit. Is that an OK approach with you, that we first make it work before you merge?

squat:
@aknuds1 if you estimate that it wont be too much work to bring it back to a working state then I would prefer to have you fix the branch; it'll let us iterate faster once its merged. If you estimate it will be a lot of work we should make the changes I requested and merge and then fix.

aknuds1:
@squat Let me try to bring it back to a working state over the weekend, OK? I don't know if it'll work 100% by then, but I'm pretty sure I'll make it processable by Terraform again. In my opinion it should at least be in a better state for merging. There's a lot that has changed in master since I last had this branch working, so I'm making some radical changes.

aknuds1:
@squat I fixed the basic problems so that Terraform is now able to parse my branch (and pushed these changes). It doesn't yet work of course...

aknuds1:
@squat Could you try creating a cluster with the current revision of my branch? I have aligned it to the current AWS implementation as best I can, and am using DigitalOcean's Spaces instead of S3 in order to distribute kubeconfig to masters and workers and tectonic.zip to the first master. Since Terraform doesn't yet have DO Spaces support, I had to upload to the bucket from a local script...

I don't know yet how well my implementation works, but am at least able to create a cluster. I'll have to do some more inspection of the cluster to see how well it works, but I was hoping you could help out here :)

aknuds1:
@squat Since my last comment, I've aligned my implementation even more closely to the AWS implementation by creating a master load balancer and by not using a separate etcd domain, so that the etcd DNS records look like those of AWS.

I have the problem though that the API server doesn't start, or at least I can't connect to it. I don't know what's wrong. I also suspect my etcd cluster isn't working, although I can't tell for sure.

Might you be able to create a cluster on DigitalOcean with my branch and see what's going wrong?

squat:
Yes, I'll look into it Tuesday. Thanks for the effort this far.

aknuds1:
Thanks @squat :) I removed the master load balancer for now, since I can't figure out how to configure the health check. I don't think the load balancer should make any difference wrt. whether things work or not, so we might as well figure out why the API server doesn't work first.

aknuds1:
@squat Are you able to help me finish this PR this week? The current problem I'm having is that cluster bootstrapping fails because certificates aren't in /etc/kubernetes/bootstrap-secrets on the master node, as far as I can tell. I can't see logic either to populate /etc/kubernetes/bootstrap-secrets, so I'm really at a loss as to what's wrong with my implementation.

aknuds1:
@squat I have the impression now that the issue is that the apiserver fails to boot because it can't connect to the etcd cluster. I can't connect to etcd from the master via etcdctl nor curl, looks as if the etcd client CA could be bad (the same thing works from etcd members):
```
$ curl -vvv --cacert /opt/tectonic/tls/etcd-client-ca.crt --cert /opt/tectonic/tls/etcd-client.crt https://coreos-testing-etcd-0.socialfoodie.club:2379 --key /opt/tectonic/tls/etcd-client.key
* Rebuilt URL to: https://coreos-testing-etcd-0.socialfoodie.club:2379/
*   Trying 138.197.188.10...
* TCP_NODELAY set
* Connected to coreos-testing-etcd-0.socialfoodie.club (138.197.188.10) port 2379 (#0)
* ALPN, offering http/1.1
* Cipher selection: ALL:!EXPORT:!EXPORT40:!EXPORT56:!aNULL:!LOW:!RC4:@STRENGTH
* successfully set certificate verify locations:
*   CAfile: /opt/tectonic/tls/etcd-client-ca.crt
  CApath: /etc/ssl/certs
* TLSv1.2 (OUT), TLS header, Certificate Status (22):
* TLSv1.2 (OUT), TLS handshake, Client hello (1):
* TLSv1.2 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (OUT), TLS alert, Server hello (2):
* SSL certificate problem: unable to get local issuer certificate
* Closing connection 0
* TLSv1.2 (OUT), TLS alert, Client hello (1):
curl: (60) SSL certificate problem: unable to get local issuer certificate
```

aknuds1:
@squat I figured out my problems and am now able to successfully bootstrap a cluster. However, out-of-the-box `kube-system` pods fail to start in the cluster, due to mysterious Docker errors such as:
```
Failed to pull image "quay.io/coreos/flannel:v0.8.0-amd64": rpc error: code = 2 desc = failed to register layer: lstat /var/lib/docker/overlay/9f90e9b8b77923af85f0859f38177fd52ceac0637b06e0d8389a718141175d6c: no such file or directory
```

Any idea what would cause such overlay corruption?

squat:
That’s great news! I have not seen that error before. Are you seeing it consistently and on every node?

aknuds1:
@squat I'm not seeing it consistently, but on several nodes. It wouldn't have anything to do with `rkt image gc` would it?

squat:
Can you point me to the line you’re referring to?

aknuds1:
@squat https://github.com/coreos/tectonic-installer/pull/604/files#diff-73db5aaec419c0e6817dcdc7c36f5584R5

aknuds1:
BTW, the Terraform error I'm seeing is due to a bug in `digitalocean_floating_ip` I think:

```
$ terraform refresh
[…]
Error: Error refreshing state: 1 error(s) occurred:

* module.masters.digitalocean_floating_ip.master: 1 error(s) occurred:

* module.masters.digitalocean_floating_ip.master: digitalocean_floating_ip.master: Error retrieving FloatingIP: GET https://api.digitalocean.com/v2/floating_ips/165.227.244.25: 404 The resource you were accessing could not be found.
```

squat:
No, this is unrelated. rkt does not store images in the same place as docker, and flannel is never run in rkt.

aknuds1:
@squat I see. I'll try to create a Tectonic cluster in AWS to compare when I have the time.

aknuds1:
@squat Are you able to investigate why this Docker overlay error might be happening? Do you eventually know where I should ask?

aknuds1:
@squat I think I might have found the solution to the Docker overlay corruption I was experiencing: Configure Docker to use the _overlay2_ storage driver instead of _overlay_. I made [a PR](https://github.com/coreos/tectonic-installer/pull/2371) for this purpose.

After testing by creating clusters several times, I'm inclined to think my fix works as they have all come up without problems.

seocam:
What's the current status on this?

aknuds1:
@seocam Hi there! The current status is that it's working, I currently have two clusters up and running on DigitalOcean, but it needs to be synced and aligned again with the master branch. My thinking is to do this during the holidays, hopefully it won't be too much work. There is still the question when the Tectonic Installer maintainers will be able to merge the PR though, I cannot answer to that.

seocam:
thanks @aknuds1! Let me know if you need help testing. I'm be able to spin some droplets to create test clusters here.

aknuds1:
Feel free to test it in its current state 🙂 It should work.

aknuds1:
@seocam I have re-synced [my branch](https://github.com/aknuds1/tectonic-installer/tree/digitalocean) with CoreOS master. I _think_ it should be good, would you mind testing?

aknuds1:
@seocam Never mind, I've found it's broken in that the API server won't come up during bootstrapping. Investigating why.

seocam:
@aknuds1 tried to run the `plan` and `apply` but it always breaks after adding DO API token. 

```
Error: module.tectonic.template_dir.tectonic: 1 error(s) occurred:

* module.tectonic.template_dir.tectonic: file: open : no such file or directory in:

${base64encode(file(var.pull_secret_path))}
```

Tried with k8s true and a valid license and with k8s false and no license  and both have the same outcome.

aknuds1:
@seocam Do you want to connect on Slack? It's a bit easier to discuss there. I'm aknudsen on the Kubernetes workspace.

My terraform.tfvars looks as follows:

```
tectonic_vanilla_k8s = false
tectonic_admin_email = "<example>"
tectonic_admin_password = "<example>"
tectonic_do_spaces_access_key_id = "<example>"
tectonic_do_spaces_secret_access_key = "<example>"
tectonic_do_spaces_region = "ams1"
tectonic_do_worker_droplet_size = "2gb"
tectonic_do_master_droplet_size  = "2gb"
tectonic_worker_count = 2
tectonic_base_domain = "<example>"
tectonic_cluster_name = "<example>"
tectonic_license_path = "./tectonic-license.txt"
tectonic_pull_secret_path = "./config.json"
tectonic_do_ssh_keys = ["<example>"]
tectonic_do_droplet_region = "fra1"
tectonic_do_token = "<example>"
tectonic_networking = "flannel"
tectonic_etcd_count = 1
tectonic_do_etcd_droplet_size = "512mb"
```

seocam:
I was editing the example file instead of the `terraform.tfvars`. After edited the proper file I managed to get the machines up but couldn't access anything. Looks like master isn't replying in `32000`, `32001` e `32002`. 

I'll connect in slack but it might take an hour until I can! Thanks

aknuds1:
@seocam Have you got the very latest revision of my branch? I found out that I had to configure the API server to listen on port 443, since it had changed to 6443 since last I synced with master. I need 443 since I don't have a load balancer in front of the API server.

seocam:
I couldn't manege here.

Here are my settings:
```
tectonic_admin_email = "<my email>"
tectonic_admin_password = "<my pass>"
tectonic_do_token = "<my token>"
tectonic_do_spaces_region = "nyc3"
tectonic_do_spaces_access_key_id = "<my access key>"
tectonic_do_spaces_secret_access_key = "<my private key>"
tectonic_do_etcd_droplet_size = "2gb"
tectonic_do_master_droplet_size = "2gb"
tectonic_do_worker_droplet_size = "2gb"
tectonic_do_master_droplet_size = "2gb"
tectonic_do_worker_droplet_size = "2gb"
tectonic_do_ssh_keys = [<my ssh-key id>]
tectonic_do_droplet_region = "nyc1"
tectonic_base_domain = "<my domain>"
tectonic_cluster_name = "k8s-sandbox"
tectonic_container_linux_version = "latest"
tectonic_etcd_count = "0"
tectonic_license_path =  "./tectonic-license.txt"
tectonic_master_count = "1"
tectonic_pull_secret_path = "./config.json"
tectonic_tls_validity_period = "26280"
tectonic_vanilla_k8s = false
tectonic_worker_count = "3"
```

aknuds1:
@seocam I can't see why it would fail for you, please give me more details via Slack :)

modules/ignition/resources/bin/do-puller.sh
```diff
@@ -0,0 +1,19 @@
...
+if [ "$#" -ne "2" ]; then
+    echo "Usage: $0 location destination"
+    exit 1
+fi
+
+export ACCESS_KEY_ID=${do_access_key_id}
+export SECRET_ACCESS_KEY=${do_secret_access_key}
+export REGION=${do_region}
+
+docker pull aknudsen/do-spaces-tool:0.2.0 > /dev/null
```
sym3tri: What is this image exactly? Can add a comment to link to docs or Dockerfile or something?

modules/ignition/variables.tf
```diff
@@ -161,3 +161,18 @@ variable "custom_ca_cert_pem_list" {
...
+}
+
+variable "do_spaces_secret_access_key" {
+  type = "string"
+  default = ""
+}
+
+variable "do_spaces_region" {
+  type = "string"
+  default = ""
```
sym3tri: Please add descriptions for all of these vars.

modules/tls/etcd/signed/main.tf
```diff
@@ -14,7 +14,7 @@ resource "tls_self_signed_cert" "etcd_ca" {
   private_key_pem = "${tls_private_key.etcd_ca.private_key_pem}"
 
   subject {
-    common_name  = "etcd-ca"
+    common_name = "etcd-ca"
```
sym3tri: Why was this changed? Seems unnecessary, if so please revert this.

platforms/digitalocean/resources/do-pusher.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
sym3tri: Please add docs or comments on what this is used for.

platforms/digitalocean/resources/do-pusher.sh
```diff
@@ -0,0 +1,20 @@
...
+export SECRET_ACCESS_KEY=${do_secret_access_key}
+export REGION=${do_region}
+
+# shellcheck disable=SC2034,SC1083
+filename=$(basename $${1})
+# shellcheck disable=SC2034,SC1083
+cp $${1} /tmp
+docker pull aknudsen/do-spaces-tool:0.2.0 > /dev/null
+# shellcheck disable=SC2034,SC1083
+docker run -t --net=host -e ACCESS_KEY_ID -e SECRET_ACCESS_KEY -v /tmp:/spaces aknudsen/do-spaces-tool:0.2.0 upload /spaces/$${filename} $${2}
```
sym3tri: It seems this is running as a `local-exec` meaning docker is required to be installed on the user's system. No other platforms require this. If absolutely necessary then it's worth adding some "prerequisite" docs and detecting if docker is on the system in this script to show the user a friendly message if it's not found.

platforms/digitalocean/resources/etc/resolved.conf
```diff
@@ -0,0 +1,22 @@
...
+# Entries in this file show the compile time defaults.
+# You can change settings by editing this file.
+# Defaults can be restored by simply deleting this file.
+#
+# See resolved.conf(5) for details
+
+[Resolve]
+#DNS=
+#FallbackDNS=
+Domains=${domain}
```
sym3tri: Some comments/docs on why this is required would be nice.

Makefile
```diff
@@ -12,7 +12,6 @@ TF_DOCS := $(shell which terraform-docs 2> /dev/null)
 TF_EXAMPLES := $(shell which terraform-examples 2> /dev/null)
 TF_CMD = terraform
 TEST_COMMAND = /bin/bash -c "bundler exec rspec spec/${TEST}"
-
```
sym3tri: revert if not required

modules/bootkube/assets.tf
```diff
@@ -44,11 +44,12 @@ resource "template_dir" "bootkube" {
     oidc_groups_claim   = "${var.oidc_groups_claim}"
     oidc_ca_cert        = "${base64encode(var.oidc_ca_cert)}"
 
-    kube_ca_cert       = "${base64encode(var.kube_ca_cert_pem)}"
-    apiserver_key      = "${base64encode(var.apiserver_key_pem)}"
-    apiserver_cert     = "${base64encode(var.apiserver_cert_pem)}"
-    serviceaccount_pub = "${base64encode(tls_private_key.service_account.public_key_pem)}"
-    serviceaccount_key = "${base64encode(tls_private_key.service_account.private_key_pem)}"
+    kube_ca_cert          = "${base64encode(var.kube_ca_cert_pem)}"
+    apiserver_secure_port = "${var.apiserver_secure_port}"
```
sym3tri: This addition seems like an isolated change that deserves its own separate commit.

modules/ignition/variables.tf
```diff
@@ -161,3 +161,18 @@ variable "custom_ca_cert_pem_list" {
...
+}
```
sym3tri: I don't really like that platform-specific vars are being added to this top-level common module. Have you explored any other ways to plumb these values through?

modules/ignition/resources/bin/do-puller.sh
aknuds1: @sym3tri Thanks for asking, it's a tool for uploading to and downloading from DigitalOcean Spaces object storage (equivalent to S3). I'll add a link in a comment in this file and in do-pusher.sh, OK?

platforms/digitalocean/resources/do-pusher.sh
aknuds1: Should I add documentation to the top of do-pusher.sh, even if this isn't the case for do-puller.sh (or s3-puller.sh/gcs-puller.sh)?

platforms/digitalocean/resources/do-pusher.sh
aknuds1: @sym3tri You're right. I had to solve it through local-exec because Terraform doesn't support DigitalOcean Spaces. I didn't think too much about the Docker dependency.

What do you suggest we do? Where should I put the documentation, if we stick with this approach?

platforms/digitalocean/resources/etc/resolved.conf
aknuds1: Where would you like to see these docs?

Makefile
aknuds1: @sym3tri Done, thanks for noticing.

modules/bootkube/assets.tf
aknuds1: @sym3tri I included the addition of the `apiserver_secure_port` variable in the DigitalOcean port commit since it's necessary for this purpose, but I'll make a separate commit for it.

modules/ignition/variables.tf
aknuds1: @sym3tri I agree it's a bit strange, but I did it this way because `data.ignition_file.s3_puller` is also defined in modules/ignition. It doesn't quite make sense to me that an S3 utility is defined in a common module (`ignition`), but I thought there might be technical reasons why this was done. Do you think the S3 puller utility should also be removed from modules/ignition?

platforms/digitalocean/README.md
```diff
@@ -0,0 +1,31 @@
+# Tectonic Installer for DigitalOcean
```
sym3tri: Also, please move this doc to 
https://github.com/coreos/tectonic-installer/tree/master/Documentation/dev
like the gcp ones.

platforms/digitalocean/resources/do-pusher.sh
sym3tri: The more comments the better

platforms/digitalocean/resources/etc/resolved.conf
sym3tri: https://github.com/coreos/tectonic-installer/tree/master/Documentation/dev

in a new directory: `digital-ocean`

modules/ignition/variables.tf
sym3tri: Well the s3 puller is there too, but there are no variables plumbed through to it. This is b/c the AWS EC2 nodes have their own IAM roles and don't need credentials to access the AWS API. It seems DO doesn't have anything like this. 

There may be no workaround, but it's worth thinking through. 

Maybe @squat has some ideas?

platforms/digitalocean/resources/do-pusher.sh
aknuds1: @sym3tri Done, I also added a comment to document do-puller.sh.

platforms/digitalocean/resources/etc/resolved.conf
aknuds1: @sym3tri Does it make sense to call the directory `digitalocean` instead, to be consistent with `platforms/digitalocean` and `modules/digitalocean`?

modules/ignition/variables.tf
aknuds1: @sym3tri I was able to move my DO specifics from modules/ignition to platforms/digitalocean, seemingly without any negative consequences (I tried `terraform plan` after). I'm still wondering though why one would add AWS and GCP logic to modules/ignition. There shouldn't be any technical reason for it? It doesn't new aws/gcs variables directly, but there are references to container images as a result (awscli and gcloudsdk).

aknuds1:
@sym3tri I've revised according to all your comments now, except I named the documentation directory Documentation/dev/digitalocean instead of Documentation/dev/digital-ocean, for reasons of consistency.

platforms/digitalocean/resources/etc/resolved.conf
sym3tri: yes thanks

aknuds1: @sym3tri Alright, I added Documentation/dev/digitalocean with a README that also documents the use of resolved.

enxebre:
Hey @aknuds1 this is great, thanks for this huge effort!
Some functional notes:
I managed to deploy a cluster but the console is not available.
Also as mentioned in you comment above when running terraform destroy I always get: "digitalocean_floating_ip.master: Error Unassigning FloatingIP (159.203.159.11) from the droplet: POST https://api.digitalocean.com/v2/floating_ips/159.203.159.11/actions: 500 Server was unable to give you a response."

I dropped a few minor comments, most of them can be addressed in follow up PRs but I’d try to slightly modify the readme to illustrate how to get the cluster up and running as suggested in the review.

We'll need basic test coverage for this.
- Create a digitalocean folder here https://github.com/coreos/tectonic-installer/tree/master/tests/smoke and the corresponding sample vars file
- Under /tests/rspec/spec/digitalocean we'll need the analogous to https://github.com/coreos/tectonic-installer/blob/master/tests/rspec/spec/aws/basic_spec.rb
- Finally we’ll need the specific cluster implementation similar to https://github.com/coreos/tectonic-installer/blob/master/tests/rspec/lib/aws_cluster.rb

Documentation/dev/digitalocean/README.md
```diff
@@ -0,0 +1,24 @@
+# Tectonic Installer for DigitalOcean
```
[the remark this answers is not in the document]
aknuds1: @enxebre Thanks, on it.

modules/digitalocean/etcd/dns.tf
```diff
@@ -0,0 +1,7 @@
+resource "digitalocean_record" "etcd_nodes" {
```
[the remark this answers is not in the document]
aknuds1: @enxebre Just out of curiosity, do you mind explaining why DNS logic is put beneath a module separate from the platform module?

platforms/digitalocean/spaces.tf
```diff
@@ -0,0 +1,39 @@
...
+  }
+}
+
+resource "null_resource" "kubeconfig" {
+  # Changes to kubeconfig requires re-provisioning
+  triggers {
+    assets = "${module.bootkube.kubeconfig}"
+  }
+
+  provisioner "local-exec" {
```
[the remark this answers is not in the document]
aknuds1: Yes, we depend on Docker locally unfortunately @enxebre. Do you mean in Documentation/dev/digitalocean/README.md?

platforms/digitalocean/tectonic.tf
```diff
@@ -0,0 +1,148 @@
...
+  etcd_server_key_pem       = "${module.etcd_certs.etcd_server_key_pem}"
+  kube_ca_cert_pem          = "${module.kube_certs.ca_cert_pem}"
+  kubelet_cert_pem          = "${module.kube_certs.kubelet_cert_pem}"
+  kubelet_key_pem           = "${module.kube_certs.kubelet_key_pem}"
+  etcd_backup_size          = "${var.tectonic_etcd_backup_size}"
+  etcd_backup_storage_class = "${var.tectonic_etcd_backup_storage_class}"
+  etcd_endpoints            = "${module.etcd.endpoints}"
+  master_count              = 1
+
+  # The default behavior of Kubernetes's controller manager is to mark a node
```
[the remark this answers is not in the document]
aknuds1: @enxebre Thanks, I just copied everything beneath `# Platform-independent variables wiring, do not modify.`, but apparently, not everything following said comment is platform independent :/

aknuds1:
@enxebre Thanks, I've revised according to most of your comments now and am currently revising the README.

modules/digitalocean/master/resources/init-assets.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+set -e
+set -o pipefail
+
+# Download the assets from Spaces
+# shellcheck disable=SC2154
```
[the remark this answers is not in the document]
aknuds1: @enxebre Done, in do-puller.sh.

aknuds1:
@enxebre Can you have another look, I think it should be pretty much ready? I didn't do the DNS re-grouping yet, but maybe that could be done in a follow-up PR anyhow.

modules/digitalocean/master/resources/register-master.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+set -ex
+set -o pipefail
+
+register_floating_ip() {
+  curl -X POST -H "Content-Type: application/json" -H "Authorization: Bearer ${1}" -d \
```
enxebre: the curl request will almost always return 0, unless you use --fail, you can check the exit code with ```echo $?``` so the until loop won't work as expected. Also after deploying a cluster the load balancer show no droplets so I can't access to the console

modules/digitalocean/etcd/dns.tf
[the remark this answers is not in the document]
enxebre: Part of the reasoning behind it is to have ability to easily swap dns approach

enxebre:
Hey @aknuds1 thanks. I dropped a few comments but I'm in favour of getting this in as it is, just please make sure everything worthy is included in the readme, e.g as per my last deployment, load balancer has no droplets.

aknuds1:
@enxebre Hey, sorry I couldn't respond until now, I've been busy moving stuff between apartments :/ I also see that load balancer and floating IP creation are now broken. Sorry, I'll try to fix it.

aknuds1:
Maybe I see now how my register-master.sh script is broken, in the middle of testing my fixes...

modules/digitalocean/master/resources/register-master.sh
aknuds1: Thanks @enxebre, I have missed the `--fail` option.

aknuds1:
@enxebre I think I fixed the issue with registering the master droplet with the load balancer and the floating IP, it was just a stupid bash programming mistake. Can you please test again? I've just created a cluster, and I can't tell yet if it's working or not except that the LB and the floating IP both have the droplet assigned to them.

aknuds1:
@enxebre Good news! After my shell script fixes, I could successfully bootstrap a cluster!

enxebre:
lgtm cc @squat

aknuds1:
Thanks @enxebre!!!

nreisb:
> I managed to deploy a cluster but the console is not available.

[This was fixed in a recent PR](https://github.com/aknuds1/tectonic-installer/issues/9)

> Also as mentioned in you comment above when running terraform destroy I always get: "digitalocean_floating_ip.master: Error Unassigning FloatingIP (159.203.159.11) from the droplet: POST https://api.digitalocean.com/v2/floating_ips/159.203.159.11/actions: 500 Server was unable to give you a response."

This is a known issue, I'm working on a fix for this now, but will require some rework of the bootstrapping process due to dependency cycles.

tsal:
This thread is a nail-biter.  Any updates on when this might make it to mainline?

aknuds1:
@tsal According to CoreOS, it's undergoing testing at the moment with the goal of integrating the PR. I'm also waiting for them to let me know if I should keep this branch in sync with coreos/master or not (and also integrate PRs from others).

sym3tri:
@aknuds1 Sorry for the delay. I think this is good to merge. Can you just do one last rebase to fix the merge conflicts?

aknuds1:
@sym3tri Sorry for the delayed reply, I've got the flu :( I've rebased, but there are a couple of changes since my last rebase that I can't reconcile. A couple of files that I've modified in order to control the secure port of the API server have disappeared, _modules/bootkube/resources/bootstrap-manifests/bootstrap-apiserver.yaml_ and _modules/bootkube/resources/manifests/kube-apiserver.yaml_, where do I configure the API server secure port now?

aknuds1:
@enxebre Do you know where to find these two API server manifests that have disappeared since my last rebase: modules/bootkube/resources/bootstrap-manifests/bootstrap-apiserver.yaml and modules/bootkube/resources/manifests/kube-apiserver.yaml? At some point I needed to modify these to configure the secure port for the API server to use.

sym3tri:
@aknuds1 yeah we're changing up the way things work in master. Can you close this PR and re-open it against the `track-1` branch instead of `master`?

klausenbusk:
> @enxebre Do you know where to find these two API server manifests that have disappeared since my last rebase: modules/bootkube/resources/bootstrap-manifests/bootstrap-apiserver.yaml and modules/bootkube/resources/manifests/kube-apiserver.yaml? At some point I needed to modify these to configure the secure port for the API server to use.

They was removed in: https://github.com/coreos/tectonic-installer/pull/2702

aknuds1:
@sym3tri Thanks for the heads up, will try.

sym3tri:
For those following along, this PR was moved to https://github.com/coreos/tectonic-installer/pull/2862

modules/digitalocean/master/master.tf
```diff
@@ -0,0 +1,44 @@
...
+resource "digitalocean_droplet" "master_node" {
+  count      = 1
+  depends_on = ["null_resource.tectonic_assets", "null_resource.kubeconfig"]
+  name       = "${var.cluster_name}-master-${count.index}"
+  image      = "${var.droplet_image}"
+  region     = "${var.droplet_region}"
+  size       = "${var.droplet_size}"
+  ssh_keys   = ["${var.ssh_keys}"]
+  tags       = ["${var.extra_tags}"]
+  user_data  = "${data.ignition_config.main.rendered}"
```
ditansu: maybe add `private_networking = true ` (and for node.tf too) to create private network?

aknuds1: @ditansu It's a good point, I wasn't aware of DigitalOcean's support for this until recently. Maybe it should be an option whether to create private nodes or not?

ditansu: > Maybe it should be an option whether to create private nodes or not?

Why not, but I think a privet net need be always, because cluster it is always two or more nodes  and  it's much "cheaper" and secure use privet net. Especially if we take in account  privet network is don't exclude public IP e.g. node always will be have public IP in any case  - enable or disable privet IP.

aknuds1: @ditansu Pardon my ignorance, but what's the benefit of creating nodes with both private and public IPs, will it not be as insecure so long as nodes have public IPs? I do see though that Typhoon creates nodes with both private and public IPs, and I was wondering what the reasoning is behind this. Is the benefit that cluster nodes will communicate with each other over the private net?

I was wondering if it might be an idea to introduce an option to create nodes with private IPs only, for security purposes? Kops has an option on AWS to create nodes with private IPs and a single bastion host to allow SSH-ing into the cluster. In the case of DigitalOcean I'm not sure if there's a need for a bastion host as it doesn't have VPCs, and I would think one might as well just use any other node in one's account as the bastion. Thoughts?

ditansu: @aknuds1, I guess the benefit of public IP is only for deploy stage without bastion. So you could rollout a cluster from your laptop directly. After deploy you could wrap the cluster in firewall. By the way https://github.com/kubicorn/kubicorn/  can up cluster with DO's firewall out-of-box.   And sure you could use any DO's server as bastion in deploy time and usage time. But I don't see benefits from bastion it self, because there is a very small opportunity of hacker attack during cluster rollout time, and we have firewall   (for tagging droplet group) to protect whole our cluster.

aknuds1: @ditansu Sounds like the right approach to me with firewalling the cluster! Do you know if [Typhoon](https://github.com/poseidon/typhoon) firewalls the cluster as well_

ditansu: I know about Typhoon but haven't tried it yet.  But judging this https://github.com/poseidon/terraform-digitalocean-kubernetes/blob/master/network.tf  they support firewall from out-of-box.

ditansu: @aknuds1  you are welcome ))) buy the way, what do you think regarding tectonic's future under RedHat ?

aknuds1: @ditansu I am also unsure! But in my opinion it's good to have Typhoon to as a second project in this space :)

ditansu: @aknuds1 absolutely agree)

PR #605, opened 2017-05-09
Tolerate API errors while waiting for Pods to be ready

ethernetdan:
Smoke tests have been failing because of API server errors listing Pods while waiting for them all to be ready. This seemed to happen after fixing deploying Tectonic components, since these components get deployed during the test potentially the extra load increases the number of failures. 

This change makes the Pod status check tolerant of Pod listing failure during the check. Any failures will be logged in the test.

If we see frequent API server failures logged we should determine if there is a regression.


s-urbaniak:
side-stepping here giving a tentative LGTM, we do actually do the same in `tectonic.sh`.

PR #606, opened 2017-05-09
modules/bootkube: specify the current context

everett-toews:
It's helpful to name the context and specify the current-context for
tooling that keys off of the current-context in a kubeconfig.

In my case I have a PS1 that inspects the current-context to let me
know what Kubernetes cluster I'm working with at the moment.
https://gist.github.com/everett-toews/e5a51e54ec7355a2d6716bfbf6c08fe2

### DIFF
--- a/modules/bootkube/resources/kubeconfig
+++ b/modules/bootkube/resources/kubeconfig
@@ -11,6 +11,8 @@ users:
     client-certificate-data: ${kubelet_cert}
     client-key-data: ${kubelet_key}
 contexts:
-- context:
+- name: local
+  context:
     cluster: local
     user: kubelet
+current-context: local


chancez:
Setting `current-context` within the file would make it difficult to use within the `KUBECONFIG` env var, when using multiple kubeconfig files. I think naming it makes sense however. Although I wish the names that were being used weren't all just `local`.

everett-toews:
@chancez Can you please provide an example of how that would make it difficult?

I use the `KUBECONFIG` env var all the time and it typically points to a kubeconfig file that has the  `current-context` set. I haven't had any issues. I'm curious if there's a use case I'm unaware of that makes it difficult.

FWIW, I'm agreed that `local` isn't a great name for all the things in this kubeconfig file. I think that would be best addressed in another PR though.

everett-toews:
Looking a bit more closely at `KUBECONFIG`

```bash
$ kubectl help config
Modify kubeconfig files using subcommands like "kubectl config set current-context my-context"

The loading order follows these rules:

  1. If the --kubeconfig flag is set, then only that file is loaded.  The flag may only be set once and no merging takes
place.
  2. If $KUBECONFIG environment variable is set, then it is used a list of paths (normal path delimitting rules for your
system).  These paths are merged.  When a value is modified, it is modified in the file that defines the stanza.  When a
value is created, it is created in the first file that exists.  If no files in the chain exist, then it creates the last
file in the list.
  3. Otherwise, ${HOME}/.kube/config is used and no merging takes place.
```

Do you typically use option 2 and `KUBECONFIG` is a list of paths?

In which case, potentially having multiple `current-context`s results in a bit of ambiguity about which is the real `current-context`?

chancez:
@everett-toews correct. That's basically the issue.

everett-toews:
I see. I'll close this then. Naming the context or changing the name would be best left to another PR. Thanks for taking a look.

PR #607, opened 2017-05-09
metal: implement tectonic_vanilla_k8s

coreypobrien:
Setting `tectonic_vanilla_k8s` to `true` now correctly omits the tectonic systemd unit.

Follows https://github.com/coreos/tectonic-installer/pull/194

### DIFF
--- a/platforms/metal/remote.tf
+++ b/platforms/metal/remote.tf
@@ -44,7 +44,7 @@ resource "null_resource" "bootstrap" {
       "sudo mkdir -p /opt",
       "sudo rm -rf /opt/tectonic",
       "sudo mv /home/core/tectonic /opt/",
-      "sudo systemctl start tectonic",
+      "sudo systemctl start ${var.tectonic_vanilla_k8s ? "bootkube" : "tectonic"}",
     ]
   }
 }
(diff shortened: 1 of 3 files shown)


s-urbaniak:
Thanks a lot for the contribution! Generally this LGTM, but I would like to retest this myself as we don't have CI set up yet for metal.

platforms/metal/remote.tf
```diff
@@ -44,7 +44,7 @@ resource "null_resource" "bootstrap" {
       "sudo mkdir -p /opt",
       "sudo rm -rf /opt/tectonic",
       "sudo mv /home/core/tectonic /opt/",
-      "sudo systemctl start tectonic",
+      "sudo systemctl start ${element(list("tectonic", "bootkube"), var.tectonic_vanilla_k8s)}",
```
alexsomesan: This expression is relying on the fact that booleans (which `tectonic_vanilla_k8s` is) also evaluate to numerical `0` or `1`. It is harder to understand how this expression evaluates, if you're not aware of that fact.

Would you mind turning this into a conditional expression, for better readability?
Thanks!

s-urbaniak:
Tested this with various combination of `tectonic_experimental` and `tectonic_vanilla_k8s`, hence LGTM.

dghubble:
With this "vanilla k8s", you'll still get Tectonic identity flags on the api server and possibly other bits that are somewhat Tectonic specific.

PR #610, opened 2017-05-09
Fix idempotency for plan / apply: don't render optional manifests into template_dir outputs

alexsomesan:
Currently, attempting a `plan` immediately after a successfull `apply` falsely reports that changes are still necessary. A subsequent `apply` for these supposed changes leads to other changes being reported by further `plans`. A vicious circle is created.

This is caused by a previous change to support optional manifests which renders additional manifests into the output location of a `template_dir` resource. Because `template_dir` checksums the contents of this location to determine if changes are necessary, it leads to above mentioned behaviour.

This fix avoids rendering the optional manifests into the location checksummed by `template_dir`. Consequently, calling `plan` or `apply` after a successful build reports no changes necessary.

/cc: @sym3tri

### DIFF
--- a/modules/aws/ignition/resources/init-assets.sh
+++ b/modules/aws/ignition/resources/init-assets.sh
@@ -49,10 +49,6 @@ fi
 unzip -o -d /opt/tectonic/ /opt/tectonic/tectonic.zip
 rm /opt/tectonic/tectonic.zip
 
-# Move optional experimental manifests into bootkube friendly locations
-[ -d /opt/tectonic/experimental ] && mv /opt/tectonic/experimental/* /opt/tectonic/manifests/ && rm -r /opt/tectonic/experimental
-[ -d /opt/tectonic/bootstrap-experimental ] && mv /opt/tectonic/bootstrap-experimental/* /opt/tectonic/bootstrap-manifests/ && rm -r /opt/tectonic/bootstrap-experimental
-
 # Populate the kubelet.env file.
 mkdir -p /etc/kubernetes
 echo "KUBELET_IMAGE_URL=${kubelet_image_url}" > /etc/kubernetes/kubelet.env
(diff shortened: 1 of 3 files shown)


s-urbaniak:
The moment you pair with someone considering complex solution, and then comes this :-) LGTM!

s-urbaniak:
One question though: do we need to change anything for the metak

s-urbaniak:
(Sorry for closing && reopening, I hit the wrong buttons)

dghubble:
Yes. As-is, this only works for AWS.

alexsomesan:
ACK, I have to follow up with the corresponding metal changes.
Wanted to get feedback first whether this is acceptable as a solution.
I will follow up with metal chances.

dghubble:
I ran into this issue today with plan and apply reporting that changes were needed - it makes adding configs to an infrastructure repo impractical. I do like this approach and can't think of much better without switching away from template_dir (undesired) so :+1:

dghubble:
If you also added the mv lines in https://github.com/coreos/tectonic-installer/blob/master/modules/bootkube/resources/bootkube.sh it should work for bare-metal since that's called here: https://github.com/coreos/tectonic-installer/blob/master/platforms/metal/cl/bootkube-controller.yaml.tmpl#L95

This strategy is working well enough in bare-metal bootkube tests https://github.com/coreos/matchbox/pull/522

alexsomesan:
@dghubble Moved the mv statements to the bootkube service script as suggested. Tested on AWS -  works fine there.

alexsomesan:
ok to test

dghubble:
Tested on bare-metal with the experimental flag turned on and this bootstraps successfully. Its unclear whether terraform re-plan/apply is clean when using the GUI.

dghubble:
Tested again, but with the manual terraform plan and apply.

```
```

alexsomesan:
rebased

PR #622, opened 2017-05-10
modules/azure: remove kubernetes cluster tag

metral:
Fixes: https://github.com/coreos/tectonic-installer/issues/596


robszumski:
This is an easy one, so it would be great to get in before the release.

robszumski:
Plus, it's blocking master from working.

PR #626, opened 2017-05-10
platforms/aws: Support additional security groups

bison:
This allows users to add additional security groups to nodes on AWS. My specific use-case is that I have a `base` security group managed outside the Tectonic config which allows SSH from bastion hosts among other things.

### DIFF
--- a/platforms/aws/variables.tf
+++ b/platforms/aws/variables.tf
@@ -30,6 +30,39 @@ variable "tectonic_aws_etcd_ec2_type" {
   default     = "t2.medium"
 }
 
+variable "tectonic_aws_etcd_extra_sg_ids" {
+  description = <<EOF
+(optional) List of additional security group IDs for etcd nodes.
+
+Example: `["sg-51530134", "sg-b253d7cc"]`
+EOF
+
+  type    = "list"
+  default = []
+}
+
+variable "tectonic_aws_master_extra_sg_ids" {
+  description = <<EOF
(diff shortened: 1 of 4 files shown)


platforms/aws/variables.tf
```diff
@@ -30,6 +30,24 @@ variable "tectonic_aws_etcd_ec2_type" {
   default     = "t2.medium"
 }
 
+variable "tectonic_aws_etcd_extra_sg_ids" {
+  type        = "list"
+  description = "List of additional security group IDs for etcd nodes."
```
s-urbaniak: Do you mind to give some examples in the documentation using heredoc format? Additionally I suggest to mark this as `(optional)` so the `terraform-examples` binary will mark it as such in the generated tfvars examples.

Here is a suggested desription snippet:
```
description = <<EOF
(optional) List of additional security group IDs for etcd nodes.

Example: `["123", "456"]`
EOF
```

Also please install https://github.com/segmentio/terraform-docs and https://github.com/s-urbaniak/terraform-examples, and execute `make docs examples` and add them to this PR.

alexsomesan:
This totally makes sense. We even had it in early versions of the tf installer but somehow got lost among all the refactorings.

LGTM

platforms/aws/variables.tf
bison: Done.

bison:
Can we merge this? I've generated docs and examples as requested.

PR #634, opened 2017-05-10
frontend: More button styling.

ggreer:
Add btn-info which is blue.
Better styling of hover for buttons.


ggreer:
screen shot 2017-05-10 at 11 30 00

PR #639, opened 2017-05-10
Initial VMware Support

alekssaul:
# Initial VMware support for Tectonic 1.6.2 

## Scenarios tested
- Provisioning with self-hosted / experimental flags
- Provisioning with standalone etcd cluster.
- Scaling up the cluster by increasing variables such as master_count or worker_count
- Destroying a Virtual Machine by hand and recovering it via `terraform apply`
## Highlights
- End user brings in VMware CoreOS container image that's supported by the OS team. Currently Terraform does not have support for manipulating VMware virtual machine images/templates. It's left up to the operator to import this image (documented) and keep it up to date
- End user created DNS records required (documented). There is no universal DNS api for on-premise deployments.
- IP addresses are assigned statically using terraform map variables
- All machines are assumed to be deployed into the same Network subnet.

## Modules and Platform additions

Platforms follow similar syntax to existing supported platforms:
- tectonic.tf : contains use of bootkube and tectonic modules based mostly on AWS and Metal as examples.
- provider.tf : sets up vSphere provider configuration and creates a folder to deploy resources to. 
- remote.tf : scp assets generated similar to Baremetal flow. However copying of kubeconfig has been moved to individual node creation within the modules
- main.tf : uses node modules and etcd modules to create etcd and master-worker nodes. Master and worker generation are both handled by node module.

Modules/vmware contains two modules:
- etcd : Sets up an etcd cluster. Platform specific changes are:
  * Enable VMware Guest Tools, which provides hypervisor integrations. Guest tools are shipped with CoreOS Container Linux VMware image
  * Setup static networking
  * Setup etcd cluster by setting `--initial-cluster` flag to rendering of template_file. This produces static list of nodes based on end users input of etcd node count and fully qualified hostnames
- nodes: Sets up worker and master nodes. Platform Specific changes are:
  * Enable VMware Guest Tools, which provides hypervisor integrations. Guest tools are shipped with CoreOS Container Linux VMware image
  * Setup static networking
  * Uses `remote-exec` and `file` provisioners to copy kubeconfig similar to Baremetal setup


Documentation/variables/vmware.md
```diff
@@ -0,0 +1,39 @@
+<!-- DO NOT EDIT. THIS FILE IS GENERATED BY THE MAKEFILE. -->
+# Terraform variables
+This document gives an overview of the variables used in the different platforms of the Tectonic SDK.
+
+## Inputs
+
+| Name | Description | Default | Required |
```
s-urbaniak: We don't render the `Required` column anymore, as its semantics are dubious. Do you mind to regenerate this with a newer version of `terraform-docs` as outlined in https://github.com/coreos/tectonic-installer/blob/master/Makefile#L40-L45 ?

Documentation/variables/vmware.md
```diff
@@ -0,0 +1,39 @@
...
+|------|-------------|:-----:|:-----:|
+| tectonic_ssh_authorized_key | SSH public key to use as an authorized key. Example: `"ssh-rsa AAAB3N..."` | - | yes |
```
s-urbaniak: I suggest to call this `tectonic_vmware_ssh_authorized_key`

examples/terraform.tfvars.vmware
```diff
@@ -0,0 +1,178 @@
...
+// (Optional) Memory size to assign Worker nodes in MB
+// tectonic_vmware_worker_vm_memory = 4096
+
+// Hostnames of ETCD Nodes 
+// ex: tectonic_vmware_etcd_hostnames = {
+//  "0" = "etcd0"
+//  "1" = "etcd1"
+//  "2" = "etcd2"
+//  "3" = "etcd3"
+//  "4" = "etcd5"
```
s-urbaniak: I believe you added this manually. Since we are auto-generating this, I suggest to add a HEREDOC in the variable declaration below.

modules/vmware/etcd/ignition.tf
```diff
@@ -0,0 +1,116 @@
...
+data "ignition_user" "core" {
+  name                = "core"
+  ssh_authorized_keys = ["${var.core_public_keys}"]
+}
+
+data "ignition_systemd_unit" "locksmithd" {
+  count = "${length(var.external_endpoints) == 0 ? 1 : 0}"
+
+  name   = "locksmithd.service"
+  enable = true
```
s-urbaniak: We mask locksmith in favor of CLUO.

modules/vmware/etcd/nodes.tf
```diff
@@ -0,0 +1,27 @@
+resource "vsphere_virtual_machine" "etcd_node" {
+  count           = "${length(var.external_endpoints) == 0 ? var.instance_count : 0}"
```
s-urbaniak: Additionally if `experimental` is enabled, we deploy self-hosted etcd, hence etcd nodes don't need to be created in this case.

modules/vmware/etcd/nodes.tf
```diff
@@ -0,0 +1,27 @@
...
+  domain          = "${var.base_domain}"  
+
+  network_interface {
+    label         = "${var.vm_network_label}"
+  }
+
+  disk {
+    datastore     = "${var.vm_disk_datastore}"  
+    template      = "${var.vm_disk_template_folder}/${var.vm_disk_template}"
+    type	        = "thin"
```
s-urbaniak: This is misaligned, I suggest to invoke `terraform fmt` on all files.

modules/vmware/etcd/ignition.tf
alekssaul: Thanks @s-urbaniak , this is also enabled in AWS; https://github.com/coreos/tectonic-installer/blob/master/modules/aws/etcd/ignition.tf#L25-L37 . Also I believe CLUO is deployed to master& worker nodes, not etcd nodes?

Documentation/variables/vmware.md
alekssaul: I've used the variable without "vmware" prefix because of the usage in metal; https://github.com/coreos/tectonic-installer/blob/master/platforms/metal/variables.tf#L179-L187 . Happy to make the change if there are no plans to roll `tectonic_ssh_authorized_key` into global variables.

modules/vmware/etcd/ignition.tf
s-urbaniak: ah, you are right and I was blind ;-) Indeed, etcd nodes are not under control of CLUO.

modules/vmware/etcd/nodes.tf
alekssaul: `experimental` flag is handled via; https://github.com/alekssaul/tectonic-installer/blob/vmware-1.6/platforms/vmware/main.tf#L3 within the platform instead of the module. This I believe is similar to AWS; https://github.com/coreos/tectonic-installer/blob/master/modules/aws/etcd/nodes.tf#L26 and https://github.com/coreos/tectonic-installer/blob/master/platforms/aws/main.tf#L56

s-urbaniak: correct, my blindness is quite shocking at this point ;-)

alekssaul:
Thanks @s-urbaniak for your review. I've made following changes based on your feedback.
- modify the Makefile to add vmware docs and examples
  * broke Makefile during the process and fixed (sorry about multiple commits)
- Added HEREDOC to variables file instead
- cleaned up indentation by running `terraform fmt`
- auto-generated docs and examples via `make docs ; make examples`
- change `tectonic_ssh_authorized_key` to `tectonic_vmware_ssh_authorized_key`

modules/vmware/etcd/nodes.tf
alekssaul: you call it blindness, I call it Mondays :)

alekssaul:
+ Bootkube content_hash variable changed related to; https://github.com/coreos/tectonic-installer/pull/575

alekssaul:
Hi @s-urbaniak , 

Whenever you are available, I believe this is ready for review :)

platforms/vmware/variables.tf
```diff
@@ -0,0 +1,212 @@
...
+  description = "vCenter Server IP/FQDN"
+}
+
+variable "tectonic_vmware_username" {
+  type        = "string"
+  default     = "Administrator@vsphere.local"
+  description = "Username to Use to connect to vCenter"
+}
+
+variable "tectonic_vmware_password" {
```
s-urbaniak: Is there any possibility to reference this and potentially the username outside of terraform? Currently we generally reference these sorts of credentials outside of terraform itself.

The discussion per se is in flux (see https://github.com/coreos/tectonic-installer/issues/710 is the umbrella issue) but as long as the discussion is going we would be more flexible referencing outside credentials first.

alekssaul:
Thanks @s-urbaniak 
- removed credentials from variables.tf
- Re-ran `make examples` and `make docs`
- Updated the documentation noting that end user will have to provide vmware credentials at runtime.

alekssaul:
@MattMencel was testing this in his environment and noticed that Master node provisioning was using CPU/memory size variables meant for Workers. https://github.com/coreos/tectonic-installer/pull/639/commits/b662c7140f8db0ec3b6f5354195b4fc4d6843063 fixes this.

alekssaul:
Hi @s-urbaniak, 

I removed the credentials from variables files. Users will be prompted during runtime (or they set `VSPHERE_USER` & `VSPHERE_PASSWORD` env variables).

Please let me know if there are any other concerns.

@alexsomesan

s-urbaniak:
@alekssaul from my side this is looking good! please re-execute `make docs examples` as these are out of sync causing the CI to fail.

alekssaul:
@s-urbaniak broke it first then fixed it :) thanks for help

alekssaul:
with merging of master into VMware branch the PR was broken

- https://github.com/coreos/tectonic-installer/pull/639/commits/1b18905b3d23ef20121092c3e615afa7ccc70863 adds changes required to support https://github.com/coreos/tectonic-installer/pull/767 
- https://github.com/coreos/tectonic-installer/pull/639/commits/c3e5ff305168e10bbfbaecb062e5707e08bd512c adds changes required for https://github.com/coreos/tectonic-installer/pull/829
- https://github.com/coreos/tectonic-installer/pull/639/commits/508a1db0fc6d93472ae09bc6ce00a6d066e30d93 add changes required for https://github.com/coreos/tectonic-installer/pull/805

alekssaul:
Master had 3 breaking changes ^^^ that needed to be applied to VMware platform.

s-urbaniak:
@alekssaul cool, thanks!

just a very last nit, CI tells me this failure:
```
misformatted files (run 'terraform fmt .' to fix): platforms/vmware/tectonic.tf
```

I guess you will have to invoke `terraform fmt` on the offending file.

s-urbaniak:
and once CI is green we shall merge this finally ;)

alekssaul:
@s-urbaniak all green

alexsomesan:
Way cool!

PR #646, opened 2017-05-11
Filter logs for credentials

ethernetdan:
This uses a very rough heurestic (string length) to remove credentials from log output. This is a requirement for opening log output up to the public.

cc/ @philips

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -78,8 +78,10 @@ pipeline {
             # Use smoke test configuration for deployment
             ln -sf ${WORKSPACE}/test/aws.tfvars ${WORKSPACE}/build/${CLUSTER}/terraform.tfvars
 
-            make plan
-            make apply
+            alias filter=${WORKSPACE}/installer/scripts/filter.sh
+
+            make plan | filter
+            make apply | filter
 
             # TODO: replace in Go
             CONFIG=${WORKSPACE}/build/${CLUSTER}/terraform.tfvars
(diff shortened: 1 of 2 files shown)


ethernetdan:
ok to test

ethernetdan:
@sym3tri @Quentin-M @alexsomesan @s-urbaniak @philips are we happy with [this](https://jenkins-tectonic-installer.prod.coreos.systems/job/coreos%20-%20tectonic-installer/job/tectonic-installer/job/PR-646/11/consoleFull) level of filtering?

Quentin-M:
This is omitting a bunch of paths. Considering we are not building keypairs that are less than 1024 chars, we might increase that threshold a little?

ethernetdan:
Upped the threshold to 500, looks like it only got keys this time :1st_place_medal:

Quentin-M:
Cool! LGTM!

ggreer:
This seems to lack both sensitivity & specificity. It filters public certs in addition to keys. Also, it doesn't hide console_secret, kubectl_secret, and pull_secret.

I suggest a twofold method:

1. Filter private keys with a regex that matches private keys: `-----BEGIN [A-Z]{2,10} PRIVATE KEY-----[^]*-----END [A-Z]{2,10} PRIVATE KEY-----` (at least, that's what it would look like in JavaScript)
2. Have a blacklist containing the names of other secrets. Filter any lines containing the blacklisted names.

Quentin-M:
Yeah that's what I suggested IRL too. Did not realize we didn't hide the pull_secret.

PR #661, opened 2017-05-11
docs: reformatting controller doc for consistency.

zbwright:


zbwright:
question: is there a missing . at line 22?

Documentation/troubleshooting/controller-recovery.md
```diff
@@ -1,164 +1,195 @@
...
+ .metadata.name = $name | \
+ .spec.containers[0].name = $name | \
+ .spec.nodeName = $node | \
+ .spec.serviceAccount = "default" | \
+ .spec.serviceAccountName = "default" ' | kubectl convert -f-
+```
+
+(Or, manually copy the`spec` section under `template`.)
+
+Executing this command creates the following rescue pod:
```
robszumski: define rescue pod maybe? 

"Executing this command creates the following rescue pod, **which is our temporary instance of the scheduler**:"

Documentation/troubleshooting/controller-recovery.md
```diff
@@ -1,164 +1,195 @@
...
+    - scheduler
+    - --leader-elect=true
+    image: quay.io/coreos/hyperkube:v1.6.2_coreos.0
+    imagePullPolicy: IfNotPresent
+    name: kube-scheduler
+```
+
+Then, get the name of the master node:
+```bash
+kubectl get nodes -l master=true
```
robszumski: can we show the output of this? and then use it below?

Documentation/troubleshooting/controller-recovery.md
```diff
@@ -1,164 +1,195 @@
...
+```
+Wrap the `spec` in a pod header and specify the name of the master node in `nodeName`:
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: kube-scheduler-rescue
+spec:
+  nodeName: <name-of-the-tectonic-master-node>
```
robszumski: can this match the above?

Documentation/troubleshooting/controller-recovery.md
zbwright: sure - if we get a cluster up and running. is there a default master-node name for clusters created with Tectonic Installer? if not, it's not too useful to just use some random name here.

robszumski: nope, but im thinking that a real value will at least show you what format, that something like `ip-10-0-117-76.us-west-1.compute.internal` is the "name" and not something like "node1"

zbwright: ah, ok. you provide, and i will add and plug in below. I may have to get some redbeard time to get it working on this machine.

zbwright:
@robszumski this should be good now. approve, and we can push it. (I thought it went out a week ago. sorry I lost track.)

Issue #660, opened 2017-05-11
terraform get step needed, not in documentation


colhom:
```sh
$> terraform plan -var-file=./build/chom-test-metal/terraform.tfvars platforms/metal
Failed to load root config module: Error loading modules: module bootkube: not found, may need to be downloaded using 'terraform get'
```
In order to make this work, you have to do `terraform get ./platforms/metal` before running `terraform plan`. We should put this in the documentation.

colhom:
I have verified this is the case with `platforms/aws` as well, I'm guessing we need to do this for all platforms.

robszumski:
[AWS contains this](https://github.com/coreos/tectonic-installer/blob/master/Documentation/install/aws/aws-terraform.md#initialize-and-configure-terraform), can you add the same language to metal?

zbwright:
@robszumski what about azure and openstack?

$ terraform get platforms/azure
$ terraform get platforms/openstack

?

zbwright:
https://github.com/coreos/tectonic-installer/pull/669

s-urbaniak:
fixed in #669, hence closing

PR #669, opened 2017-05-12
docs: adding terraform get command.

zbwright:
for https://github.com/coreos/tectonic-installer/issues/660


robszumski:
Fixes https://github.com/coreos/tectonic-installer/issues/660#

PR #677, opened 2017-05-12
Move bcrypt(password) logic from backend to frontend

ggreer:


ggreer:
You can view all the changes except the re-vendoring here: https://github.com/coreos/tectonic-installer/compare/master...cf7f2ac565eeeea8f5d97a5e8865d21c8edee909

ggreer:
TODO: check which browsers support webcrypto and if one of the important ones doesn't, set a fallback function in bcrypt.

ggreer:
I've booted a cluster with this code and checked that the password worked.

PR #684, opened 2017-05-12
frontend: move bm clusterinfo to framework

kans:
...which requires sharing some Fields between AWS and BM.  This PR should introduce no new behavior.


kans:
test cluster

PR #686, opened 2017-05-14
Add editorconfig

aknuds1:
Add [.editorconfig](http://editorconfig.org/) file in order to be able to control editing parameters. I found it necessary to use EditorConfig in order to ensure that newlines weren't added to the end of Terraform template files automatically, since said newlines will make it into the rendered Terraform config and can lead to confusing errors if newlines shouldn't be in the rendered output.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,10 @@
+[*]
+indent_style = space
+indent_size = 2
+insert_final_newline = true
+
+[*.tpl]
+insert_final_newline = false
+
+[Makefile]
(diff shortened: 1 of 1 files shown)


alexsomesan:
ok to test

s-urbaniak:
I believe we can skip tests for this ;-)

ggreer:
This makes editors save .go files with spaces. :(

aknuds1:
@ggreer Oops, good fix!

PR #697, opened 2017-05-15
update-payload: add signing script and require sig upload

marineam:


yifan-gu:
Where is the yubikey? How can I test it?

marineam:
@yifan-gu For testing purposes that don't get checked in you can sign with your own gpg key (no need for the script, just `gpg --detach-sign`) and add your public key to the `publicKeyOverride` option. For real updates @dghubble is listed as being in charge of the Tectonic yubikey so updating payload.json will need to be coordinated with him.

yifan-gu:
@marineam yeah, but I mean test this script... But I think it's ok to merge this

marineam:
@yifan-gu I tested the script by temporarily adding my own yubikey to it.

PR #708, opened 2017-05-16
Remove the tectonic_aws_az_count input variable

alexsomesan:
This change follows up on cleanup due after moving to Terraform 0.9.x
In 0.8.x we had a limitation of not being able to reference data-sources as values for `count` attributes.
The `tectonic_aws_az_count` was working around this by asking the user to fill in the number of AZs.

This was fixed in 0.9.x and we can now pick those up from a datasource.

With this change we make it simpler for users to choose between auto-computed subnets or explicitly set ones. We also have one less input variable to care for.

/cc: @sym3tri

### DIFF
--- a/platforms/aws/main.tf
+++ b/platforms/aws/main.tf
@@ -23,17 +23,16 @@ module "vpc" {
   # A. Explicitly configure a list of AZs + associated subnet CIDRs
   # B. Let the module calculate subnets accross a set number of AZs
   #
-  # To enable mode A, make sure "tectonic_aws_az_count" variable IS NOT SET to any value 
-  # and instead configure a set of AZs + CIDRs for masters and workers using the
+  # To enable mode A, configure a set of AZs + CIDRs for masters and workers using the
   # "tectonic_aws_master_custom_subnets" and "tectonic_aws_worker_custom_subnets" variables.
   #
   # To enable mode B, make sure that "tectonic_aws_master_custom_subnets" and "tectonic_aws_worker_custom_subnets" 
-  # ARE NOT SET. Instead, set the desired number of VPC AZs using "tectonic_aws_az_count" variable.
+  # ARE NOT SET.
 
   # These counts could be deducted by length(keys(var.tectonic_aws_master_custom_subnets)) 
   # but there is a restriction on passing computed values as counts. This approach works around that.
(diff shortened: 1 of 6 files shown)


platforms/aws/main.tf
```diff
@@ -23,17 +23,16 @@ module "vpc" {
...
   #
   # To enable mode B, make sure that "tectonic_aws_master_custom_subnets" and "tectonic_aws_worker_custom_subnets" 
-  # ARE NOT SET. Instead, set the desired number of VPC AZs using "tectonic_aws_az_count" variable.
+  # ARE NOT SET.
 
   # These counts could be deducted by length(keys(var.tectonic_aws_master_custom_subnets)) 
   # but there is a restriction on passing computed values as counts. This approach works around that.
-  master_az_count = "${var.tectonic_aws_az_count == "" ? "${length(keys(var.tectonic_aws_master_custom_subnets))}" : var.tectonic_aws_az_count}"
-  worker_az_count = "${var.tectonic_aws_az_count == "" ? "${length(keys(var.tectonic_aws_worker_custom_subnets))}" : var.tectonic_aws_az_count}"
+  master_az_count = "${length(keys(var.tectonic_aws_master_custom_subnets)) > 0 ? "${length(keys(var.tectonic_aws_master_custom_subnets))}" : "${length(data.aws_availability_zones.azs.names)}"}"
```
Quentin-M: do we still need to pass these counts if we already give the list of custom subnets? we can still do the exact same (new) conditional you are doing here but in the module.

alexsomesan:
I asked myself the same question initially.

I've noticed though that those *_az_count vars are then used in a few places inside the module as part of other HIL expressions that are quite verbose. Moving those input conditionals into the module would only make them even uglier. That's why I chose to keep them as inputs.

What do you think? Makes sense? @Quentin-M

robszumski:
Can you check to see if there are guides that need to be updated in addition to the variable descriptions?

alexsomesan:
Filled https://github.com/coreos-inc/coreos-pages/pull/1602 to update the website.

alexsomesan:
@Quentin-M 
I know about that approach, but personally I try to think of the `null_data_source` as a last resort, when nothing else fits. As you know, the resource is not officially supposed to be there and if they suddenly break it we would be refactoring a bit :)

PR #716, opened 2017-05-16
openstack/neutron: Fix security groups

coreypobrien:
When assigning neutron ports, the `security_groups` attribute is ignored on
`openstack_compute_instance_v2`. Apply all the security groups by id on the
`openstack_networking_port_v2` instead. Also create a specific port for etcd
nodes for consistency with the controllers and nodes.


coreypobrien:
Yep. In my local stuff I removed all the DNS work in favor of only using IPs.

PR #721, opened 2017-05-16
Always cache go builds with -i

ethernetdan:
This PR for updates the Makefile for the installer to always pass `-i` when performing a Go build. This will cause the build to be cached in `$GOPATH/pkg` and will speed up future builds.

### DIFF
--- a/installer/Makefile
+++ b/installer/Makefile
@@ -15,14 +15,14 @@ all: build
 build: bin/$(OS)/installer
 
 bin/windows/installer.exe: $(GO_FILES) assets/bindata.go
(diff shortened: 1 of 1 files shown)


Quentin-M:
Let's try :)

dghubble:
Starting with this commit:

```
GOOS=linux go build -i -o bin/linux/installer -ldflags "-w -X github.com/coreos/tectonic-installer/installer/cmd/installer.Version=fe5c5bfd2baa7c12b9d9af7138e39f754ff42dcb" github.com/coreos/tectonic-installer/installer/cmd/installer
go install os/user: open /usr/local/go/pkg/linux_amd64/os/user.a: permission denied
go install net: open /usr/local/go/pkg/linux_amd64/net.a: permission denied
Makefile:21: recipe for target 'bin/linux/installer' failed
```

PR #722, opened 2017-05-17
platforms/metal: Fix race condition adding workers

dghubble:
* kubelet-env starts and docker creates missing directories so
it could create a directory /etc/kubernetes/kubeconfig
* Terraform races with kubelet-env can if it runs after docker
has run once, the kubeconfig will erroneously be placed in a
directory /etc/kubernetes/kuebconfig/kubeconfig

Impact is that under the right (wrong) circumstances, no workers will join the cluster.

cc @sym3tri

### DIFF
--- a/platforms/metal/cl/bootkube-worker.yaml.tmpl
+++ b/platforms/metal/cl/bootkube-worker.yaml.tmpl
@@ -34,8 +34,7 @@ systemd:
         Description=Determine the Kubelet Image Version
         ConditionPathExists=!/etc/kubernetes/kubelet.env
         [Service]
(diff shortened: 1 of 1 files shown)


dghubble:
Testing re-provisioning a cluster which experienced this issue.

dghubble:
Better :+1:

rithujohn191:
Were you able to force the race condition to occur and test out the fix as well?

dghubble:
@rithujohn191 no, probably the easiest way would be to disable network connectivity while Terraform is running to force the desired ordering, but I haven't done that.

PR #727, opened 2017-05-17
frontend: Set tectonic_aws_external_vpc_public in output tfvars.

ggreer:
Depends on #664.

### DIFF
--- a/installer/frontend/cluster-config.js
+++ b/installer/frontend/cluster-config.js
@@ -2,7 +2,7 @@ import _ from 'lodash';
 import bcrypt from 'bcryptjs';
 
 import { BARE_METAL_TF } from './platforms';
-import { keyToAlg } from './utils';
+import { keyToAlg, toExtraData } from './utils';
 
 const bcryptCost = 12;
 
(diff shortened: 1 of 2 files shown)


s-urbaniak:
@evfurman is it possible for you to test it this PR in conjunction with #664, i.e. using a local build?

sym3tri:
UI changes:
- Add a checkbox that appears _only_ if user selects a _public_ hosted zone.
- Defaults to checked (value is "")
- When checked off value is same public hosted zone ID from dropdown.
- Text is: "Optimize VPC networking by associating a new private hosted zone ([see AWS documentation]())."

@alexsomesan to provide link to docs.

ggreer:
screen shot 2017-05-18 at 14 19 16

Just need the doc URL and this is ready.

ggreer:
For now I've made the doc URL https://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/vpc-dns.html#vpc-private-hosted-zones

alexsomesan:
I think this one is a bit more comprehensive and it also explains the
split-view technique that we use and also touches on the exact resolution
steps when using both private and public.

http://docs.aws.amazon.com/Route53/latest/DeveloperGuide/hosted-zones-private.html

What do you think?

I can also write something up on how this approach allows us to avoid
public DNS outages affecting intra-cluster communications.

ggreer:
OK, link updated.

meghanschofield:
@ggreer We're doing a quick UX round! We will update soon...

alexsomesan:
Merged #664 

There's still some UX work happening here, right?

robszumski:
@alexsomesan Yes, hoping to finish that up today

meghanschofield:
@kans here are the updated mocks! IRL discussion:

- create a dropdown for Optimize DNS
- swap order of screen elements

#### Default Screen, Optimize DNS
default screen - don t optimize

#### Additional drop down choice:
optimize option

^^^ @tlwu2013 check it out!

ggreer:
Continued in #788.

PR #728, opened 2017-05-17
server: Log HTTP requests.

ggreer:
Before #325, debug logging logged HTTP requests as they came in. Restore this behavior.


ggreer:
Tests failed because we ran hit AWS marklar limits. It actually works and logs stuff.

PR #729, opened 2017-05-17
modules/aws: add route53 to policies

squat:
We need to add permissions to perform route53 actions to both worker and
master policies to ensure that federation controller managers that get
scheduled to Tectonic nodes can list, create, and destroy DNS entries
for federated services. Without this, the service controller in the
federation controller manager cannot start and service federation will
not work at all.

### DIFF
--- a/modules/aws/master-asg/master.tf
+++ b/modules/aws/master-asg/master.tf
@@ -122,6 +122,11 @@ resource "aws_iam_role_policy" "master_policy" {
       "Effect": "Allow"
     },
     {
+      "Action": "route53:*",
+      "Resource": "*",
+      "Effect": "Allow"
+    },
+    {
       "Action": [
         "ecr:GetAuthorizationToken",
         "ecr:BatchCheckLayerAvailability",


sym3tri:
@squat why is this needed by both Master and Worker policies? Forcing the the Fed Control Plane to run on masters only would be desirable so we don't have to open this up on workers.

squat:
I can add the nodeselector to make federation controller managers only schedule on master nodes. It's just not the default anywhere at the moment

robszumski:
It will need to tolerate the taint as well.

alexsomesan:
One more thing: is there a really a need for * premissions? Can we narrow that down?

squat:
It needs to read, create, and destroy entries in a zone. We should really narrow down ALL of the policies; right now the workers and master have * on ec2, * on elasticloadbalancers etc

sym3tri:
Let's defer this until we have a clearer picture on how we plan to integrate the federation stuff. For internal testing let's just pin to master nodes for now.

PR #730, opened 2017-05-17
installer/scripts/release: drop Terraform binary on Windows

### LINKED ISSUE #704: Remove windows folder from release tarball
Currently the installer release tarball bundles a windows folder under ./tectonic-installer.
This location only contains the windows terraform binary but not an installer binary, so it's a bit confusing.

I suggest to remove it from the tarball until we can officially support windows.

Quentin-M:
Fixes #704

### DIFF
--- a/installer/scripts/release/get_terraform_bins.sh
+++ b/installer/scripts/release/get_terraform_bins.sh
@@ -7,7 +7,8 @@ mkdir -p "${TMP_DIR}" "${TERRAFORM_BIN_TMP_DIR}"
 
 set -x
 
-for os in "linux" "darwin" "windows"; do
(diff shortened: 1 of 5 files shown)


installer/scripts/release/get_terraform_bins.sh
```diff
@@ -7,7 +7,8 @@ mkdir -p "${TMP_DIR}" "${TERRAFORM_BIN_TMP_DIR}"
 
 set -x
 
-for os in "linux" "darwin" "windows"; do
+#for os in "linux" "darwin" "windows"; do
+for os in "linux" "darwin"; do
```
ggreer: Shouldn't the entire PR just be these two lines?

PR #737, opened 2017-05-17
Cache x-compilation objects (incl. std) to GOPATH

Quentin-M:
Caching x-compilation objects effectively reduces Go build time from >2mins to ~6secs on every platforms (timed on a Mid-2014 MacBook Pro, assuming executor has built at least once). Given the fact that we are currently compiling twice for each CI run, this is significant. This is due to the fact that we were recompiling both the standard library, our dependencies and our binary every time - `go build` thrashes most of the results by default.

When building releases, we clean the environment entirely, to be on the safe side, even though the Go compiler can be trusted.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -37,7 +37,6 @@ pipeline {
         go version
         cd $GO_PROJECT/installer
 
-        make clean
         make tools
         make build
         make test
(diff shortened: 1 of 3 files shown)


Quentin-M:
This needs a little modification as we are currently executing builds in a Docker agent without shared volumes.

PR #747, opened 2017-05-18
Jenkinsfile: fail test if terraform destroy failed

ggreer:
Terraform destroy was failing in jenkins for days and we didn't notice.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -112,7 +112,10 @@ pipeline {
                          passwordVariable: 'AWS_SECRET_ACCESS_KEY'
                        ]
                        ]) {
-        // Destroy all clusters within workspace
+        /* Destroy all clusters within workspace
+         * Try 3 times before failing because tf destroy is flaky.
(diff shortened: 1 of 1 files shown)


Quentin-M:
Can you add a loup around this? Try a few times before failing. Will help when a gateway cannot detached etc. Less chance of leaking and doing false positive. It's pretty much accepted that users would have to retry sometimes on destroy due time AWS eventual consistency vs Terraform internal state.

ggreer:
I guess, but it'll fail 100% of the time right now because `terraform destroy` never works when `tectonic_etcd_count` is > 0.

Quentin-M:
For sure, this one issue here is going to fail the _n_ allowed times :)

ggreer:
It looks like the first `terraform destroy` failure was https://jenkins-tectonic-installer.prod.coreos.systems/job/coreos%20-%20tectonic-installer/job/tectonic-installer/job/PR-708/1/console which is for #708. The error-squelching (the other part of this bug) was introduced in #551.

PR #756, opened 2017-05-18
[WIP]: IAM policy with permissions required by installer

alexsomesan:
This is the role policy I was able to use to successfully apply / destroy a cluster.

This is WIP as I need to add documentation on how to use this policy.

/cc: @sym3tri


examples/installer-iam-policy.json
```diff
@@ -0,0 +1,169 @@
...
+        },
+        {
+            "Action": [
+                "sts:GetCallerIdentity"
+            ],
+            "Effect": "Allow",
+            "Resource": "*"
+        }
+    ],
+    "Version": "2012-10-17"
```
s-urbaniak: silly question: what is this magic date about?

robszumski: Amazon's terrible versioning scheme...

alexsomesan: What Rob said.
It's the date that particular IAM API version was released.

alexsomesan:
I will update that one as well, or better yet we can have the website point to this one.

I would prefer to have the reference copy in here so I can reference it in instructions on how to use and also have CLI examples or a Terraform template that users can run to create the said role.

@sym3tri What do you think?

sym3tri:
@alexsomesan I'd prefer we only keep 1 file to ease maintenance. 

Docs can't point to this one b/c all website docs are rooted in `/Documentation` and can't reference files not in the dir.

Having the examples reference files in `Documentation` or the website URL seems easiest to me.

alexsomesan:
I agree on the 1 file approach.

How about if we move this under /Documentation?

robszumski:
Can we make sure this lands in the same spot as the old example? Same filename, etc? This way we don't have to update any links?

sym3tri:
`/Documentation/files/aws-policy.json`

sym3tri:
@alexsomesan status on this? Needs to merge before we can release.

alexsomesan:
Pending writing usage instructions.
I'll update before EOD today.

sym3tri:
Thanks. I believe we can rely on the existing usage instructions in our docs.
https://coreos.com/tectonic/docs/latest/install/aws/requirements.html#temporary-credentials

alexsomesan:
@sym3tri 
* moved the policy in the right location
* updated the example assume role policy document to reference users rather than a service
* included instructions for using the role directly from terraform

sym3tri:
@alexsomesan safe to remove the `do-not-merge` label and merge?

sym3tri:
@alexsomesan I'm going to merge this for now since it's a release blocker and LGTM and is better than what we have in master today. Please amend later if needed.

alexsomesan:
Sorry for forgetting the label on this. It was actually good to merge from my POV.
Thanks for taking care of that!

Issue #579, opened 2017-05-08
Ship tectonic-monitoring 1.2.0


robszumski:
https://github.com/kubernetes/kube-state-metrics

In the 0.5 release, the ability to track labels on pods and other objects was added:

```
Add label metrics for Pods, Nodes and Services.
```

This will help us power use-cases like chargeback.

fabxc:
kube-state-metrics 0.5 will be part of the iteration of the monitoring subsystem.

brancz:
yes already in progress https://github.com/coreos-inc/tectonic-prometheus-operator/pull/49 should go out in the upcoming days

sym3tri:
Is there explicit work required for the installer, or just bump the prom operator?

brancz:
Once we finalize the tectonic-monitoring 1.2.0 we'll sync the manifests in this repo, so no explicit work needed.

sym3tri:
OK, I'll rename the issue to reflect that.

sym3tri:
@fabxc @brancz How close is this? Will it be ready for us to test on Thursday? Friday is code freeze.

brancz:
@sym3tri we sadly had a few set-backs namely https://github.com/coreos/prometheus-operator/pull/367 and https://github.com/prometheus/prometheus/pull/2731. We plan on releasing both today and then hopefully by the time you get to the office in SF we have the PR out to the tectonic-installer :wink: .

brancz:
@sym3tri see https://github.com/coreos-inc/tectonic-prometheus-operator/pull/49.

PR #757, opened 2017-05-18
Prometheus requires security group rule to access etcd from workers

alexsomesan:
This changes allows Prometheus running on a worker to scrape etcd.
/cc: @brancz 

required by:
https://github.com/coreos-inc/tectonic-prometheus-operator/pull/49
https://github.com/coreos/tectonic-installer/issues/579

### DIFF
--- a/modules/aws/vpc/sg-master.tf
+++ b/modules/aws/vpc/sg-master.tf
@@ -167,6 +167,16 @@ resource "aws_security_group_rule" "master_ingress_etcd" {
   self      = true
 }
 
+resource "aws_security_group_rule" "master_ingress_etcd_from_workers" {
+  type              = "ingress"
+  security_group_id = "${aws_security_group.master.id}"
+
+  protocol                 = "tcp"
+  from_port                = 2379
+  to_port                  = 2379
+  source_security_group_id = "${aws_security_group.worker.id}"
+}
+
 resource "aws_security_group_rule" "master_ingress_bootstrap_etcd" {
   type              = "ingress"
   security_group_id = "${aws_security_group.master.id}"


philips:
Agreed, we should't do this. What is the use case? What alternatives were considered or are available?

alexsomesan:
We shouldn't run Prometheus on masters because of it's high demand for
resources, so that's not really an option.

Also not an option to create a ClusterIP service for etcd, which falls
under the same security concerns raised here.

I will have to check with @brancz and brainstorm on alternatives.

alexsomesan:
A possible option would be enabling client cert auth on etcd to secure it against unauthorized access and still go with allowing access via ClusterIP.

@philips @sym3tri Thoughts?

brancz:
Use case is that Prometheus can scrape etcd metrics. Client cert authentication could be a solution, but we would also want to make sure that only the metrics are accessible with that client cert, not actually using etcd. Does etcd have an authz model of some sort that we could leverage for this? If not then we have to consider whether monitoring and alerting or security outweighs the other.

Until we figure out how to do this we won't be able to collect etcd metrics.

/cc @fabxc

brancz:
We can also consider whether putting metrics on different port and only allow access to that is isolation enough. It might be a compromize, but I'm not sure whether etcd allows this today.

sym3tri:
I'm open to other suggestions, but we should not try to rush this in.

Design constraint is: customers need to completely block network access from worker nodes to etcd (whether authenticated or not).

alexsomesan:
I just realized another option here:

We can deploy a statically configured Prometheus on the master nodes, that
just scrapes etcd and exports it's metrics onwards for the main Prometheus
to scrape.
On this Prometheus instance we could have tight resource constrains to make
sure it won't brown out the control plane components running on the same
master.

@brancz Is this feasible?

brancz:
Funnily enough that is exactly what @mxinden just suggested.

Yes that would work in theory, but currently the Prometheus Operator does not support tolerations and node affinity, so we'll have to implement that support before we can do that. Other than that this would also require federation support in the Prometheus Operator which we have conceptually thought about, but have not implemented yet either.

fabxc:
Federation technically works with the current operator as we support the `honorLabels` paramters now, which is the only thing differentiating a federation target from a regular one. 

With some feature additions to prometheus-operator, we can place a proxy on the master node that Prometheus can scrape etcd through.
It would mean that we still expose an endpoints object for etcd. So the IPs of etcd nodes are still visible but cannot be routed to from worker nodes. Are we okay with that? The setup proxy would then just let `GET /metrics` through. If there are DDoS concerns, we can also rate-limit that to only match roughly Prometheus' scrape interval.

sym3tri:
Closing for now. Please open another issue that details the requirements and we can agree on a solution there.

PR #765, opened 2017-05-18
platforms/aws: use element() instead of indexing

kans:
This allows us to destroy clusters, but I'm not sure how safe it is :-/


Quentin-M:
While this should not theoretically cause no issue during `apply` as the count are equals, this may result in an un-healthy etcd cluster, as the count is extremely important to the establishment of a proper quorum. And because this will definitely mask any potential error there, it will be a nightmare to debug then.

kans:
Closing for a future PR via @Quentin-M who will bump terrorform

thanks for digging up https://github.com/hashicorp/terraform/issues/13828#issuecomment-296205568

PR #771, opened 2017-05-19
Tectonic Monitoring 1.2

### LINKED ISSUE #579: Ship tectonic-monitoring 1.2.0

brancz:
Closes #579 

I've tried out these three scenarios:

- deploy from this branch without experimental on
- deploy from this branch with experimental on
- deploy from master and upgrade the TPO and the tectonic-monitoring appversion

All of the above worked. I am now attempting to deploy a cluster from the last release and then attemp the upgrade but am not expecting any issues. I'll report back here once that is finished.

We have decided to not ship the Alertmanager yet as the new UI is about to be released and we don't want the first thing customers see to be a drastic UI change from one tectonic monitoring version to the next, as well as our set of alerting rules still needs some work.

What we are shipping with this release:

- latest node-exporter (this is just an upgrade)
- latest kube-state-metrics (new component)
- latest Prometheus and Prometheus Operator (upgrades as well)
- we enabled collecting metrics from the following components: node-exporters, kube-state-metrics, Prometheus itself, kubelets and apiserver

For testing rolling upgrades we have pushed a preliminary version of the Tectonic Prometheus Operator to `quay.io/coreos/tectonic-prometheus-operator:v1.2.0`, so if you want to test this, just change the version you are running (which should be the same repo with the `v1.1.0` tag) and change the `desiredVersion` in your `tectonic-monitoring` named `AppVersion`.

@alexsomesan @s-urbaniak @sym3tri

/cc @fabxc @mxinden @pst @mfburnett


s-urbaniak:
ok to test

brancz:
Not sure why jenkins reported a failure, because when looking at the build it was green.

alexsomesan:
@brancz Under the "Smoke Tests" bullet (top of page) there is a failed step. Gotta love all the green Jenkins covers it under.

PR #773, opened 2017-05-19
Comment out `tectonic_azure_image_reference`

### LINKED ISSUE #772: azure: Terraform plans fail on `tectonic_azure_image_reference` type error
By default, the var `tectonic_azure_image_reference` is set to an empty string in the terraform.tfvars.azure example file.
If not commented out, TF will pass this empty string into `terraform plan` instead of using the default map in `platforms/azure/variables.tf`, resulting in the following error:

`variable "tectonic_azure_image_reference" should be type map, got string`

justaugustus:
Fixes #772


coresolve:
Thanks s-urbaniak I didn't know that either.

justaugustus:
w00t!

PR #779, opened 2017-05-19
tectonic: parametrize missing images

colhom:
From work in the field on offline installs, we've found these four images were not configurable via terraform.

\cc @alekssaul @coresolve

### DIFF
--- a/config.tf
+++ b/config.tf
@@ -60,6 +60,7 @@ variable "tectonic_container_images" {
     awscli                          = "quay.io/coreos/awscli:025a357f05242fdad6a81e8a6b520098aa65a600"
     kube_version                    = "quay.io/coreos/kube-version:0.1.0"
     tectonic_etcd_operator          = "quay.io/coreos/tectonic-etcd-operator:v0.0.1"
+    pod_infra_image                 = "gcr.io/google_containers/pause-amd64:3.0"
   }
 }
 
--- a/modules/tectonic/resources/manifests/monitoring/prometheus-k8s.yaml
+++ b/modules/tectonic/resources/manifests/monitoring/prometheus-k8s.yaml
@@ -8,7 +8,7 @@ metadata:
     prometheus: k8s
 spec:
   replicas: 1
-  baseImage: ${element(split(":",prometheus_image),0)}
+  baseImage: ${replace(prometheus_image,image_re,"$1")}
   version: ${prometheus_version}
   externalUrl: ${prometheus_external_url}
   serviceAccountName: prometheus-k8s
--- a/platforms/metal/cl/bootkube-worker.yaml.tmpl
+++ b/platforms/metal/cl/bootkube-worker.yaml.tmpl
@@ -74,7 +74,8 @@ systemd:
           --hostname-override={{.domain_name}} \
           --node-labels=node-role.kubernetes.io/node \
           --cluster_dns={{.k8s_dns_service_ip}} \
-          --cluster_domain=cluster.local
+          --cluster_domain=cluster.local \
+          --pod-infra-container-image={{.pod_infra_image}}
         ExecStop=-/usr/bin/rkt stop --uuid-file=/var/run/kubelet-pod.uuid
         Restart=always
         RestartSec=5
--- a/modules/aws/ignition/ignition.tf
+++ b/modules/aws/ignition/ignition.tf
@@ -42,6 +42,7 @@ data "template_file" "kubelet" {
     node_label             = "${var.kubelet_node_label}"
     node_taints_param      = "${var.kubelet_node_taints != "" ? "--register-with-taints=${var.kubelet_node_taints}" : ""}"
     kubeconfig_s3_location = "${var.kubeconfig_s3_location}"
+    pod_infra_image        = "${var.container_images["pod_infra_image"]}"
   }
 }
 
--- a/modules/aws/ignition/resources/services/kubelet.service
+++ b/modules/aws/ignition/resources/services/kubelet.service
@@ -34,7 +34,8 @@ ExecStart=/usr/lib/coreos/kubelet-wrapper \
   --cluster-domain=cluster.local \
   --client-ca-file=/etc/kubernetes/ca.crt \
   --anonymous-auth=false \
-  --cloud-provider=aws
+  --cloud-provider=aws \
+  --pod-infra-container-image={{.pod_infra_image}}
 ExecStop=-/usr/bin/rkt stop --uuid-file=/var/cache/kubelet-pod.uuid
 
 Restart=always
--- a/modules/tectonic/assets.tf
+++ b/modules/tectonic/assets.tf
@@ -38,7 +38,7 @@ resource "template_dir" "tectonic" {
     alertmanager_version           = "${var.versions["alertmanager"]}"
     tectonic_version               = "${var.versions["tectonic"]}"
     etcd_version                   = "${var.versions["etcd"]}"
-    tectonic_etcd_operator_version = "${element(split(":", var.container_images["tectonic_etcd_operator"]), 1)}"
+    tectonic_etcd_operator_version = "${replace(var.container_images["tectonic_etcd_operator"],var.image_re,"$2")}"
 
     etcd_cluster_size = "${var.master_count > 2 ? 3 : 1}"
 
@@ -89,6 +89,8 @@ resource "template_dir" "tectonic" {
     certificates_strategy    = "${var.ca_generated == "true" ? "installerGeneratedCA" : "userProvidedCA"}"
     identity_api_service     = "${var.identity_api_service}"
     tectonic_updater_enabled = "${var.experimental ? "true" : "false"}"
+
+    image_re = "${var.image_re}"
   }
 }
 
--- a/modules/tectonic/resources/manifests/monitoring/alertmanager.yaml
+++ b/modules/tectonic/resources/manifests/monitoring/alertmanager.yaml
@@ -8,7 +8,7 @@ metadata:
     alertmanager: main
 spec:
   replicas: 2
-  baseImage: ${element(split(":",alertmanager_image),0)}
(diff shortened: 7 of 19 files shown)


config.tf
```diff
@@ -25,7 +25,9 @@ variable "tectonic_container_images" {
     kube_version_operator           = "quay.io/coreos/kube-version-operator:v1.6.2"
     tectonic_channel_operator       = "quay.io/coreos/tectonic-channel-operator:0.3.2"
     node_agent                      = "quay.io/coreos/node-agent:787844277099e8c10d617c3c807244fc9f873e46"
+    prometheus                      = "quay.io/prometheus/prometheus:v1.6.1"
```
brancz: On latest master of the tectonic-installer this is `v1.6.3`, but judging from how this is used later in this PR I think you can just omit the tag and use the variable without it and name it `prometheus_base` or something like that (naming is @s-urbaniak's call here).

modules/tectonic/resources/manifests/monitoring/prometheus-k8s.yaml
```diff
@@ -13,6 +13,7 @@ spec:
   serviceMonitorSelector:
     matchExpression:
     - {key: k8s-app, operator: Exists}
+  baseImage: ${element(split(":",prometheus_image),0)}
```
brancz: This change would be removed by the tectonic prometheus operator on the next reconcile loop so we can't do this today, the next release of the tectonic prometheus operator will allow this to be configurable though, then we can introduce this.

colhom: in my observations, this commit stacked on 1.6.1 release did have the end effect of allowing the prometheus pods to come up successfully and remain so for over an hour.

config.tf
colhom: @brancz we are attempting to use this image list in the field as a canonical manifest of all image dependencies required for an install. Either in `tectonic_versions` or here, all the image tags needs to be state explicitly in the tf file.

modules/tectonic/resources/manifests/monitoring/prometheus-k8s.yaml
fabxc: @colhom I cannot 100% infer which setup you are referring to as k8s and Prometheus are in the same major/minor right now. But it doesn't really matter because this is touching a manifest which is automatically synced from a tectonic-prometheus-operator release and should not be edited manually. 

This brings us back to the question we had on why we vendor everything deployed by tectonic-x-operators. Every week someone inevitably comes in and makes changes across those vendored YAMLs, which are overwritten again either at runtime or through our vendoring process.
If lucky, one of us sees it before it gets merged, but that often doesn't happen with time zone differences.

@aaronlevy @yifan-gu I think we should revisit this decision and rather rely on the AppVersion task status fields to determine the initial setup rollout of an tectonic-x-operator-controlled sub-system succeeded.

fabxc: In general, we cannot have both – either we give control over sub-systems to tectonic-x-operators XOR we have tectonic-installer as the central registry where we pin versions and images for everything.
The vendoring approach we are taking right now somewhat works, but not as a two-way-binding that is often assumed. 

That's not to say we don't want to allow customization of things such as the base image. That's actually one of the next things we'll work on but it has to happen through tectonic-prometheus-operator and not tectonic-installer. At least if we are committed to that model.

fabxc:
> From work in the field on offline installs, we've found these four images were not configurable via terraform.

Commented in the sub-thread. TL;DR everything below tectonic-prometheus-operator is not meant to be directly modified via terraform variables as defined by the overall operator design.

modules/tectonic/resources/manifests/monitoring/prometheus-k8s.yaml
colhom: @fabxc that seems like a good plan- I'm going to back out the monitoring stuff for now then. Do we have an issue to track adding the flags to the operator and templating in the terraform image strings?

modules/tectonic/resources/manifests/monitoring/prometheus-operator.yaml
```diff
@@ -21,6 +21,7 @@ spec:
          args:
          - "--kubelet-service=kube-system/kubelet"
          - "--config-reloader-image=${config_reload_image}"
+         - "--prometheus-config-reloader=${prometheus_config_reloader_image}"
```
colhom: @fabxc I'm assuming that this flag is OK to leave in place?

fabxc: Unfortunately no =/ The prometheus-operator is also deployed by tectonic-prometheus-operator. So this has to go into its configuration as well.

fabxc:
Okay, so from a comment by @robszumski elsewhere, I inferred that this vendoring actually is meant to be removed already. And really just temporary until the tectonic-x-operators ship by default. That was not quite clear to me before.

Just talked to @brancz. If temporary, we are okay with the extra effort when updating the vendoring. But everybody must be 100% aware that any templating to those vendored files will only ever work on setups without tectonic-prometheus-operator and always need a symmetric feature in that one. Otherwise things will drift over time and we cannot seamlessly enable operators by default.

robszumski:
@fabxc can you write up a migration plan/path so we're all on the same page? curious about rough timing as well as how to proceed in the interim.

fabxc:
The plan would be that team-monitoring carries over changes made to the vendored manifests for now. Further changes must only be made after syncing with us, so that we can add symmetric customizations features in tectonic-prometheus-operator. (We'll add a notice to the vendored files about that.)

Once we ship tectonic-x-operators by default, I'd expect all those vendored manifests to be removed from tectonic-installer. As we ensured to keep feature parity, this should then be seamless.

@colhom this would mean we'll go through with all your changes and soon we add them to TPO through a configmap as described above. Until that is in, we have to be aware that TPO may overwrite the customizations.

modules/tectonic/resources/manifests/monitoring/prometheus-k8s.yaml
aaronlevy: @fabxc

> This brings us back to the question we had on why we vendor everything deployed by tectonic-x-operators

I think revisiting this assumption is reasonable. Likely can still be on a per-operator basis, but in the case of monitoring I don't see any big issue that we assume the operator will roll out all sub-components - and the only thing that is "vendored" is the top-level operator.

>  I think we should revisit this decision and rather rely on the AppVersion task status fields to determine the initial rollout of a tectonic-x-operator-controlled sub-system succeeded

We could do this - or we could punt on this as well and assume if the top-level operator has been installed, that is "success". I don't have immediate strong feelings here.

colhom:
@s-urbaniak could you take a look?

config.tf
```diff
@@ -25,7 +25,9 @@ variable "tectonic_container_images" {
     kube_version_operator           = "quay.io/coreos/kube-version-operator:v1.6.2"
     tectonic_channel_operator       = "quay.io/coreos/tectonic-channel-operator:0.3.3"
     node_agent                      = "quay.io/coreos/node-agent:787844277099e8c10d617c3c807244fc9f873e46"
+    prometheus                      = "quay.io/prometheus/prometheus:v1.6.1"
     prometheus_operator             = "quay.io/coreos/prometheus-operator:v0.9.1"
+    tectonic_prom_cfg_reloader      = "quay.io/coreos/prometheus-config-reloader:v0.0.1"
```
colhom: @s-urbaniak I abbreviated this simply so the `terraform fmt` step did not pollute the entire diff for this block by re-indenting all the images. Makes rebase-ing to keep up with master more difficult, can un-abbreviate immediately prior to merge if desired.

platforms/metal/cl/bootkube-worker.yaml.tmpl
```diff
@@ -74,7 +74,8 @@ systemd:
           --hostname-override={{.domain_name}} \
           --node-labels=node-role.kubernetes.io/node \
           --cluster_dns={{.k8s_dns_service_ip}} \
-          --cluster_domain=cluster.local
+          --cluster_domain=cluster.local \
+          --pod-infra-container-image={{.pod_infra_image}}
```
s-urbaniak: I suppose all the other plarforms need to be updated as well?

config.tf
s-urbaniak: I think that is ok, the abbreviated variables does not read badly.

platforms/metal/cl/bootkube-worker.yaml.tmpl
colhom: My plan initially was to do AWS and baremetal for now, with eventual support planned for the rest of the platforms at a subsequent PR.

s-urbaniak:
@colhom sorry about that, but this PR needs a rebase

platforms/metal/cl/bootkube-worker.yaml.tmpl
s-urbaniak: ack, ok. Please let's only ensure that the changes here are compatible with the other platforms, i.e. they still can be deployed.

colhom: ack

s-urbaniak:
closing in favor of #1042

s-urbaniak:
Actually reopening after OOB with @alekssaul because #1042 fixed the monitoring bits only. @colhom I guess we can move forward with this PR, but without the prometheus changes?

brancz:
The monitoring bits should now all be configurable, if not let us know.

colhom:
Just a heads up- waiting for https://github.com/coreos/tectonic-installer/pull/1065 to merge, then I will rebase this

robszumski:
Why was this closed? Taking another approach?

alekssaul:
@robszumski : Colin is OOO.
I believe he is breaking it apart to couple of PRs.

Issue #694, opened 2017-05-15
Add Calico network policy support


aaronlevy:
This should be "policy-only" support (leaving flannel as the pod-network provider).

See: Calico policy-only documentation here:

http://docs.projectcalico.org/v2.2/getting-started/kubernetes/installation/hosted/kubernetes-datastore/

@abhinavdahiya is going to pick this up. I'll assign once he's part of the org.

/cc @joshrosso @yifan-gu

xref: https://github.com/kubernetes-incubator/bootkube/issues/513

joshrosso:
@aaronlevy @abhinavdahiya 

Will this implementation be etcd-less?

philips:
Yes.

philips:
@joshrosso yes, this will be etcd-less. That is a hard requirement.

joshrosso:
@caseydavenport Is there a rough estimate on when [etcd-less calico will support advanced policy features](https://github.com/projectcalico/canal/issues/70#issuecomment-299597923)?

Per @philips comment, our usage in tectonic will be quite limited without all the egress, etc support we get with the etcd-based calico today.

Thanks in advanced!

lxpollitt:
If you want all the advance features of Calico policy then you will need to use Calico in etcd mode.  Either sharing the k8s etcd cluster (which is what most users choose to do today) or using a second etcd cluster (typically self hosted, using host networking).  

We do plan to add full Calico policy via TPRs at some point, but that is likely not until Q3 timeframe.

rithujohn191:
Please add a smoke test to verify that the calico network policy works using apply. You can modify one of the existing tests to do this. CC @squat

squat:
fixed by #780. closing

PR #780, opened 2017-05-19
Calico policy support 

abhinavdahiya:
#694 
Notes:

Uses Calico's Kubernetes Datastore [installation](http://docs.projectcalico.org/v2.2/getting-started/kubernetes/installation/hosted/kubernetes-datastore/) to provide policy only support
Needs to be discussed:
hyperkube image already ships with cni binaries in /opt/cni/bin and kube-calico DaemonSet also installs its cni binaries in /opt/cni/bin.

So there would be chances of version mismatch as kubelet doesn't mount this directory from host. To prevent this I have mounted an empty /opt/cni/bin from the host to the kubelet.


s-urbaniak:
Thanks a lot for the PR! A couple of concerns/thoughts from my side:

1. Using this PR we will always deploy those Callico policies, independent of whether we use them or not.
1. Exchanging binaries between pods using the host sounds smelly. I am not aware of ordering guarantees in k8s so we could have potential races when those binaries actually become visible?! I'd rather lean against one supported version of callico and ship it in the hyperkube image
1. This still deploys flannel.

Generally I'd like to see a modularized structure for networking setup. Something like:
```
modules/net/flannel
modules/net/callico
modules/net/weave
...
```

/cc @alexsomesan

aaronlevy:
I haven't had a chance to review the PR yet, but wanted to respond to some of this feedback:

> Using this PR we will always deploy those Callico policies, independent of whether we use them or not.

We can feature-flage this as well. Part of me just thinks we enable by default, but I can see it as an option as well. If we're unsure of stability / updatability I'd be more inclined to have this start out behind a feature flag. cc @robszumski for some product background.

> Exchanging binaries between pods using the host sounds smelly. I am not aware of ordering guarantees in k8s so we could have potential races when those binaries actually become visible?! I'd rather lean against one supported version of callico and ship it in the hyperkube image

There has been some prior discussion around this -- and the general view was that distributing the needed CNI binaries with the networking tool is actually an improvement (over the tool being deployed in isolation of the binaries which are shipped as part of the hyperkube container. See: https://github.com/kubernetes-incubator/bootkube/issues/405 for some prior discussion.

The issues with shipping as part of hyperkube are:

1) It's divergent from upstream (we carry a patch to add calico). The only reason flannel is naturally in the hyperkube image is that it was the first implementation of CNI and is therefor part of the official releases - but it's a little bit of a misbalanced artifact.

2) if we need to bump a version of any CNI binary that means it is a full hyperkube release. In practice this hasn't been needed very often, but it also has lent to stale CNI binaries.

cc @philips as he has also been part of these "cni distributed with network tool" discussions.

> This still deploys flannel

That is the expected behavior. Flannel is used for the networking layer, and calico is used as the network-policy agent. You could hypothetically deploy flannel in isolation and add calico later.

robszumski:
> We can feature-flag this as well. Part of me just thinks we enable by default, but I can see it as an option as well. If we're unsure of stability / updatability I'd be more inclined to have this start out behind a feature flag. cc @robszumski for some product background.

Starting out behind a flag but migrating to the default option sounds great to me.

abhinavdahiya:
Flannel is deployed as default and no longer depends on calico to install cni binaries.

>Starting out behind a flag but migrating to the default option sounds great to me

The current flag used is `tectonic_enable_calico_network_policy`. Would `tectonic_experimental` flag be better option?

> the general view was that distributing the needed CNI binaries with the networking tool is actually an improvement

Looks like kubernetes follows this order to find corresponding cni binaries `/opt/<type set in config>/bin` then `/opt/cni/bin`. Moved calico cni binaries to /opt/calico/bin to prevent override of cni binaries present in hyperkube image

> Generally I'd like to see a modularized structure for networking setup

I have chosen this hierarchy..
```
modules
   ├── aws
   ├── azure
   ├── bootkube
   ├── net
   │   ├── calico-network-policy
   │   │   └── resources
   │   │       └── manifests
   │   └── flannel
   │       └── resources
   │           └── manifests
   ├── openstack
   ├── tectonic
   └── update-payload
```
@aaronlevy @s-urbaniak

robszumski:
I can't comment on the rest, but pretend like `tectonic_experimental` doesn't exist for now. We were bitten hard by lumping in too many things behind this flag, and it reduced the ability to graduate somethings from alpha to beta. Using your own flag is great and gives us more flexibility in the future.

modules/bootkube/assets.tf
```diff
@@ -59,7 +59,8 @@ resource "template_dir" "bootkube" {
     kubedns_image          = "${var.container_images["kubedns"]}"
     kubednsmasq_image      = "${var.container_images["kubednsmasq"]}"
     kubedns_sidecar_image  = "${var.container_images["kubedns_sidecar"]}"
-    flannel_image          = "${var.container_images["flannel"]}"
+    etcd_operator_image    = "${var.container_images["etcd_operator"]}"
```
yifan-gu: Why the change appears in this PR?

abhinavdahiya: I removed merge commits on this branch similar to one on bootkube (I wasn't rebasing
 from upstream)

test/aws-exp-np.tfvars
```diff
@@ -0,0 +1,31 @@
+tectonic_worker_count = "2"
```
yifan-gu: nit: might want to make the file name more explicit, e.g. `aws-experimental-networkpolicy.tfvars`

modules/bootkube/assets.tf
yifan-gu: @abhinavdahiya We usually do rebase on the upstream master to fix the conflicts instead of merging, that will keep your commit history clean.

yifan-gu:
Adding experimental flag for this LGTM. 
Ideally the calico plugin should only install calico binaries, and flannel installs flannel + cni binaries. In that case the hyperkube image doesn't need to carry the calico and flannel patch anymore.
But this will require a small patch to calico upstream first, so I am ok with the PR at the mean time.

modules/aws/ignition/ignition.tf
```diff
@@ -34,6 +34,15 @@ data "ignition_systemd_unit" "locksmithd" {
   mask = true
 }
 
+data "null_data_source" "kubelet-cni-binaries" {
```
s-urbaniak: what about the other platforms (metal, openstack, azure, vmware)?

abhinavdahiya: Shouldn't be any different for other platforms too. I will have to add similar options to other platforms too as this support requires change to kubelet service.

modules/bootkube/assets.tf
sym3tri: yes. please rebase from master

modules/net/calico-network-policy/resources/manifests/kube-calico.yaml
```diff
@@ -0,0 +1,214 @@
...
+  name: kube-calico-cfg
+  namespace: kube-system
+data:
+  # The CNI network configuration to install on each node.
+  cni_network_config: |-
+    {
+        "name": "k8s-pod-network",
+        "cniVersion": "0.1.0",
+        "type": "calico",
+        "log_level": "debug",
```
sym3tri: should probably not be "debug" for production.

modules/net/calico-network-policy/resources/manifests/kube-calico.yaml
```diff
@@ -0,0 +1,214 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kube-calico-cfg
+  namespace: kube-system
+data:
+  # The CNI network configuration to install on each node.
+  cni_network_config: |-
```
sym3tri: In comments can you please explain what this is doing & link to docs about the config file?

modules/net/calico-network-policy/resources/manifests/kube-calico.yaml
```diff
@@ -0,0 +1,214 @@
...
+    {
+        "name": "k8s-pod-network",
+        "cniVersion": "0.1.0",
```
sym3tri: Also comment on when this should change.

sym3tri: Should probably be templated in.

modules/net/calico-network-policy/resources/manifests/kube-calico.yaml
```diff
@@ -0,0 +1,214 @@
...
+      serviceAccountName: kube-calico
+      tolerations:
+        - key: node-role.kubernetes.io/master
+          effect: NoSchedule
+        - key: "CriticalAddonsOnly"
+          operator: "Exists"
+      containers:
+        - name: kube-calico
+          image: ${calico_image}
+          env:
```
sym3tri: Add comment that links to docs on these flags

modules/net/flannel/resources/manifests/kube-flannel.yaml
```diff
@@ -70,13 +70,38 @@ spec:
         - name: flannel-cfg
           mountPath: /etc/kube-flannel/
       - name: install-cni
-        image: busybox
-        command: [ "/bin/sh", "-c", "set -e -x; TMP=/etc/cni/net.d/.tmp-flannel-cfg; cp /etc/kube-flannel/cni-conf.json $TMP; mv $TMP /etc/cni/net.d/10-flannel.conf; while :; do sleep 3600; done" ]
+        image: alpine:3.4
```
sym3tri: While you're in here, this needs to be templateable. And it should defafult to a mirror on quay.io/coreos

abhinavdahiya: > And it should defafult to a mirror on quay.io/coreos
Couldn't find an alpine mirror in quay.io/coreos ???

abhinavdahiya:
@sym3tri I have updates the manifests files to template some of the things you asked and also added some comments to explain flags and variables.

modules/net/flannel/resources/manifests/kube-flannel.yaml
```diff
@@ -70,13 +70,38 @@ spec:
         - name: flannel-cfg
           mountPath: /etc/kube-flannel/
       - name: install-cni
-        image: busybox
-        command: [ "/bin/sh", "-c", "set -e -x; TMP=/etc/cni/net.d/.tmp-flannel-cfg; cp /etc/kube-flannel/cni-conf.json $TMP; mv $TMP /etc/cni/net.d/10-flannel.conf; while :; do sleep 3600; done" ]
+        image: ${flannel_install_image}
+        command:
```
aaronlevy: Also commented on bootkube PR - but this makes me nervous and I would think we need to ship a container image with the release binaries instead (and this side-car simply copies files from container image to host mount)

Quentin-M:
@aaronlevy @s-urbaniak Thoughts on this one?

aaronlevy:
A couple discussion points:

- Should we call the "enable" flag out as experimental/alpha?
- I'd prefer that there was a way to track that the user made this decision as part of the installation. Technically we won't support upgrades of calico (we might be able to, but we have no tests and this is not an immediate requirement) - so in the future when/if we do support upgrades, we need some reasonable way to know that calico was installed while it was experimental - don't upgrade.

cc @diegs @squat as this might relate to our "cluster config" that we've all vaguely talked about.

The closest we have right now is: https://github.com/coreos/tectonic-installer/blob/master/modules/tectonic/resources/manifests/config.yaml and config.tf file itself (the former is a install-time options, and latter is fixed install-time assets).

What I'd imagine is something like a field for alpha features:

```
{
    kind: TectonicConfig
    version: v1
    data: {
        "clusterID": "foo",
        ...
        "alphaFeatures": {
           calico: true,
       }
    }
}
```

aaronlevy:
Also fwiw I'm not suggesting we create a new object, but rather we move https://github.com/coreos/tectonic-installer/blob/master/modules/tectonic/resources/manifests/config.yaml to be a more structured (and versioned) object (also makes it easier to read as a configMap file on disk).

robszumski:
> Should we call the "enable" flag out as experimental/alpha?

As long as we can come up with a clean upgrade path for when it graduates. For example in TF vars calling it alpha then renaming is kind of a pain for end users.

Quentin-M:
Tentatively adding release-blocker to this guy.

Quentin-M:
ok to test

Quentin-M:
My take would be to mention experimental in the description of the variable. Therefore, there will be no friction moving forward when it graduates.

Quentin-M:
My take would be to mention experimental in the description of the variable. Therefore, there will be no friction moving forward when it graduates. @abhinavdahiya Could you do that?

robszumski:
Let's go with "alpha" over experimental. We're trying to use the same terminology with the platforms and other features.

abhinavdahiya:
@Quentin-M do you mean add ALPHA to description in config.tf like this ?

```
variable "tectonic_calico_network_policy" {
  description = "[ALPHA] If set to true, calico network policy support will be deployed"
  default     = false
}

```

Quentin-M:
@abhinavdahiya Absolutely! Can even be something like:
```
   description = "If set to true, alpha support for calico networking will be deployed"
```

aaronlevy:
What actually uses: https://github.com/coreos/tectonic-installer/blob/master/modules/tectonic/resources/manifests/config.yaml ?

I'd like to propose that we get that cleaned up / versioned, then add in the "enabled-alpha" features section so we can track for later upgrade safety.

If nothing uses it directly, it would just be a structure change and could be done in this PR

Quentin-M:
This needs rebase.

Quentin-M:
@aaronlevy The stats emitter [uses](https://github.com/coreos/tectonic-installer/blob/068cdfc1da05809139b7fc69d497387efd1cb263/modules/tectonic/resources/manifests/stats-emitter.yaml#L33-L61) it, and the syntax will not allow us to have it in a better format.

abhinavdahiya:
@Quentin-M rebased the PR

aaronlevy:
Opened https://github.com/coreos/tectonic-installer/issues/1148 to track discussion of the cluster-config object.

We can punt on this for now - I'm just worried these untracked install-time decisions will come back to bite us.

Quentin-M:
ok to test

Quentin-M:
ok to test

Quentin-M:
@coresolve Are you comfortable with Calico to give this a review by any chance?

Quentin-M:
Removing release-blocker as discussed offline as it seems that no one is comfortable reviewing it in such a short period of time.

philips:
I feel like we need to strongly evaluate using /opt/cni/bin. Why do we need the new path on the host? How does it differ from /var/lib/cni? Are there docs on that?

Quentin-M:
@philips This is one of the things I was not sure about when reading the PR. How does this mount affect updates?

philips:
Not only the updates but I think we should just put this stuff in /var/lib. It is ephemeral.

abhinavdahiya:
@philips https://kubernetes.io/docs/admin/kubelet/
```
      --cni-bin-dir string   <Warning: Alpha feature> The full path of the directory in which to search for CNI plugin binaries. Default: /opt/cni/bin
      --cni-conf-dir string <Warning: Alpha feature> The full path of the directory in which to search for CNI config files. Default: /etc/cni/net.d
```

aaronlevy:
Discussing using /var/lib/cni with @philips:

Because /var/lib/cni is already mounted into the kubelet, that reduces one piece of skew between versions. Then thinking this through:

flannel needs to be updated to also mount /var/lib/cni so that it can install the binaries there. This would mean that our update payload needs to do the same thing (so we're not maintaining two copies - we have no real way to determine "use copy B because alpha feature canal is turned on").

So our update payload contains a new flannel manifest that has a hostmount of /var/lib/cni, so that it can install CNI binaries there. But that is effectively a no-op for existing clusters because the kubelet won't look there anyway. Existing flannel installations keep using CNI binaries shipped with hyperkube image.

New clusters get a kubelet.service with `--cni-bin=/var/lib/cni/bin` which will ultimately be populated by the flannel/canal daemonsets.

In my mental walk-through this seems like it should work. @abhinavdahiya what do you think?

aaronlevy:
We also might want to feature-flag the kubelet.service flag change until we have a plan for full upgrade support of all clusters (and bringing existing cluster flags into sync)

abhinavdahiya:
@aaronlevy @philips 
* updated the PR to use `/var/lib/cni/bin` to store cni binaries on host.
* also added the `--cni-bin-dir=/var/lib/cni/bin` flag in kubelet.service behind the `tectonic_calico_network_policy` feature-flag

yifan-gu:
@abhinavdahiya Need a rebase now.

sstarcher:
@abhinavdahiya could we get a rebase on this

sstarcher:
Is anything additionally needed for this?

aaronlevy:
My one question would be about upcoming changes in v1.7. I believe the networkPolicy object was promoted to stable -- were there any other changes that need to be taken into account?

config.tf
```diff
@@ -331,3 +333,8 @@ variable "tectonic_stats_url" {
   default     = "https://stats-collector.tectonic.com"
   description = "The Tectonic statistics collection URL to which to report."
 }
+
+variable "tectonic_calico_network_policy" {
+  description = "[ALPHA] If set to true, calico network policy support will be deployed"
```
aaronlevy: Maybe we add something like:

> Enabling an alpha feature means that future updates may become unsupported. This should only be enabled on clusters that are meant to be short-lived to begin validating the alpha feature.

/cc @robszumski

modules/aws/ignition/ignition.tf
```diff
@@ -41,6 +41,7 @@ data "template_file" "kubelet" {
     cluster_dns_ip         = "${var.kube_dns_service_ip}"
     node_label             = "${var.kubelet_node_label}"
     node_taints_param      = "${var.kubelet_node_taints != "" ? "--register-with-taints=${var.kubelet_node_taints}" : ""}"
+    cni_bin_dir            = "${var.kubelet_cni_bin_dir != "" ? "--cni-bin-dir=${var.kubelet_cni_bin_dir}" : ""}"
```
aaronlevy: I'm a little unclear on the use of `cni_bin_dir` vs `kubelet_cni_bin_dir`.

It looks like `cni_bin_dir` is actually the full flag+ path (a night might be to rename to something like `cni_bin_dir_flag` =)

Then if I'm reading correctly, `kubelet_cni_bin_dir` is a bit inconsistent (mostly just path, but sometimes flag+path).

modules/bootkube/net/flannel/variables.tf
```diff
@@ -0,0 +1,29 @@
...
+  description = "Container image for flanneld"
+  type        = "string"
+}
+
+variable "flannel_cni_image" {
+  description = "Container image for flannel cni"
+  type        = "string"
+}
+
+variable "host_cni_bin" {
```
abhinavdahiya: Sure

modules/vmware/node/resources/services/kubelet.service
```diff
@@ -28,6 +28,7 @@ ExecStart=/usr/lib/coreos/kubelet-wrapper \
   --allow-privileged \
   --node-labels=${node_label} \
   ${node_taints_param} \
+  ${cni_bin_dir} \
```
abhinavdahiya: just followed how `node_taints_param` gets handled in master and worker kubelet.sevice file.

modules/aws/ignition/ignition.tf
abhinavdahiya: I was just trying to keep the convention used for `node_taints` and `kubelet_node_taints`.

platforms/metal/cl/bootkube-controller.yaml.tmpl
```diff
@@ -83,6 +83,7 @@ systemd:
           --allow-privileged \
           --hostname-override={{.domain_name}} \
           --node-labels=node-role.kubernetes.io/master \
+          {{.kubelet_cni_bin_dir}} \
```
abhinavdahiya: all the platforms eg. `platform/aws` use `kubelet_cni_bin_dir` to define the path and the corresponding module eg. `modules/aws/ignition` use `cni_bin_dir` to parse dir path to flag for kubelet.service file.

but `platform/metal` has aal the code in `plaftorm/` dir, hence this inconsistency ...

sstarcher:
@abhinavdahiya I ran tectonic-installer from your PR and wanted to verify that these are expected messages.  They stopped after a few seconds.

```
2017-07-06 13:02:35.329 [WARNING][45] ipip_mgr.go 104: Failed to add IPIP tunnel device error=exit status 1
2017-07-06 13:02:35.329 [WARNING][45] ipip_mgr.go 81: Failed configure IPIP tunnel device, retrying... error=exit status 1
2017-07-06 13:02:55.701 [WARNING][45] k8s.go 852: Node ip-10-0-9-22.us-east-2.compute.internal does not have podCIDR for HostConfig
2017-07-06 13:02:55.717 [WARNING][45] k8s.go 852: Node ip-10-0-9-22.us-east-2.compute.internal does not have podCIDR for HostConfig
2017-07-06 13:02:55.766 [WARNING][45] k8s.go 852: Node ip-10-0-9-22.us-east-2.compute.internal does not have podCIDR for HostConfig
```

abhinavdahiya:
@sstarcher In this configuration calico expects flannel to handle networking. So after flannel has setup the networking these messages would have stopped.

robszumski:
@aaronlevy Think you can get a sec to review this today?

config.tf
robszumski: sgtm

platforms/metal/matchers.tf
```diff
@@ -32,6 +32,7 @@ resource "matchbox_group" "controller" {
   metadata {
     domain_name        = "${element(var.tectonic_metal_controller_domains, count.index)}"
     k8s_dns_service_ip = "${module.bootkube.kube_dns_service_ip}"
+    cni_bin_dir_flag   = "${var.tectonic_calico_network_policy ? "--cni-bin-dir=/var/lib/cni/bin" : "" }"
```
squeed: Hm, I'm curious why this has to be specified now. We already deploy cni binaries via the Flannel daemonset, and that expects a specific path...

abhinavdahiya: @squeed  The flannel daemonset https://github.com/coreos/tectonic-installer/commit/db6839a0b5cebff1b04fc3ce152e1130ab5e43da installs the cni binaries to /opt/cni/bin on host (which @philips suggested we change to /var/lib/cni/bin)

But the kubelet by default expects the cni binaries to be present in /opt/cni/bin unless you provide --cni-bin-dir flag.

squeed: So, will we be in trouble if this is set? Or is the flannel daemonset consuming this flag somehow? (Sorry, not super familiar with the whole terraform setup). If it works, I'm happy to believe you. Just making sure.

squeed: Of course, the kubelet will successfully start even if it can't find the CNI binaries - it will just fail to launch pods that don't use host networking.

aaronlevy: flannel doesn't actually care where the CNI binaries end up (to my knowledge). Only the kubelet does. So we're essentially supporting two modes here:

- Existing setup where flannel CNI binaries are expected to exist in the hyperkube filesystem. The kubelet will look for them in the default location (/opt/cni/bin) in the hyperkube filesystem. The flannel daemonset will also copy CNI binaries into /var/lib/cni/bin on the host (exposed to the daemonset via /host/opt/bin/cni mountpoint). However these CNI binaries just sit there and are unused.

- "new" / with calico setup: kubelet will set cni bin dir to /var/lib/cni/bin (this dir was used because it is already mounted from the host into kubelet via kubelet-wrapper). Then the flannel/calico daemonsets will copy their CNI binaries to that location and the kubelet will actually make use of them.

In both cases the flannel daemonset can be deployed the same. The only difference should be the addition of a flag on kubelet, and the addition of the calico assets.

@abhinavdahiya does all of the above sound right?

config.tf
aaronlevy: @abhinavdahiya can we add that description (or something like it) here?

aaronlevy: Also not sure if it's better to have that here, or in the tfvars examples, or both

modules/vmware/node/resources/services/kubelet.service
aaronlevy: ack

platforms/metal/cl/bootkube-controller.yaml.tmpl
aaronlevy: ack

tests/smoke/aws/vars/aws-exp-np.tfvars
```diff
@@ -0,0 +1,31 @@
...
+
+tectonic_master_count = "1"
+
+tectonic_base_domain = "tectonic.dev.coreos.systems"
+
+tectonic_cl_channel = "stable"
+
+tectonic_admin_email = "example@coreos.com"
+
+tectonic_admin_password_hash = "$2a$12$T8hTe.NlOPDP0SS3DxNeDuVhHSFbdGXZEhGps/W.BG4QC7.1/nDaG"
```
aaronlevy: I'd really prefer this wasn't necessary. This should be auto generated during CI runs.

aaronlevy:
I have one outstanding comment about adding the alpha warning blurb. Then also the test with hard-coded password is pretty unfortunate - not sure if that's something to be resolved in this PR or is a side effect of the CI needs.

Then finally I'd still like someone with terraform experience to give the thumbs up

aaronlevy:
@Quentin-M for the terraform review, maybe?

platforms/metal/matchers.tf
abhinavdahiya: Yup

aaronlevy:
@s-urbaniak Would you mind taking a pass at this to make sure it conforms to how you want the terraform stuff is structured.

platforms/vmware/tectonic.tf
```diff
@@ -93,5 +94,5 @@ data "archive_file" "assets" {
   # Additionally, data sources do not support managing any lifecycle whatsoever,
   # and therefore, the archive is never deleted. To avoid cluttering the module
   # folder, we write it in the TerraForm managed hidden folder `.terraform`.
-  output_path = "${path.cwd}/.terraform/generated_${sha1("${module.tectonic.id} ${module.bootkube.id}")}.zip"
+  output_path = "./.terraform/generated_${sha1("${module.tectonic.id} ${module.bootkube.id})}.zip"
```
s-urbaniak: why is this change necessary?

modules/bootkube/net/flannel/resources/manifests/kube-flannel.yaml
```diff
@@ -95,12 +95,12 @@ spec:
         - name: cni
           hostPath:
             path: /etc/kubernetes/cni/net.d
-        - name: flannel-cfg
-          configMap:
+        - name: flannel-cfg		
+          configMap:
```
s-urbaniak: these seem to have trailing tabs.

modules/bootkube/net/flannel/assets.tf
```diff
@@ -0,0 +1,33 @@
...
+    bootkube_dir = "${var.bootkube_dir}"
+  }
+}
+
+resource "local_file" "flannel" {
+  count = "${
+    var.enabled
+      ? 1
+      : 0
+    }"
```
s-urbaniak: This can be one line.

modules/bootkube/assets.tf
```diff
@@ -238,3 +236,26 @@ resource "local_file" "etcd_peer_key" {
   content  = "${join("", tls_private_key.etcd_peer.*.private_key_pem)}"
   filename = "./generated/tls/etcd/peer.key"
 }
+
+module "flannel" {
```
s-urbaniak: Nesting modules is something we avoided as of now due to the inability of terraform to optionally switch on/off modules and thus due to composability restrictions.

I suggest we create top-level `modules/net/flannel-vxlan` and `modules/net/calico-network-policy` and leave the composition of those modules to the platforms (that's exactly their purpose).

Additionally by renaming `net/flannel` to `net/flannel-vxlan` we could introduce more backends like `net/flannel-hostgw` et al.

modules/bootkube/net/calico-network-policy/assets.tf
```diff
@@ -0,0 +1,36 @@
+data "null_data_source" "calico-network-policy" {
+  inputs = {
+    output_id = "${
+      var.enabled
+        ? "${sha1("${join(" ", local_file.calico-network-policy.*.id)}")}"
+        : "# calico policy disabled"
+      }"
```
s-urbaniak: `null_data_source`s are an undocumented feature of Terraform so we try to avoid it as much as we can, as we have been bitten by this. I see that this interpolation function is only used once, in the output. So I suggest to get rid of `null_data_source` here and use it in outputs directly.

modules/bootkube/net/calico-network-policy/assets.tf
```diff
@@ -0,0 +1,36 @@
...
+    bootkube_dir = "${var.bootkube_dir}"
+  }
+}
+
+resource "local_file" "calico-network-policy" {
+  count = "${
+    var.enabled
+      ? 1
+      : 0
+    }"
```
s-urbaniak: same here, I suggest to squash those lines.

modules/bootkube/net/flannel/assets.tf
```diff
@@ -0,0 +1,33 @@
+data "null_data_source" "flannel" {
```
s-urbaniak: see comment above regarding `null_data_source`.

abhinavdahiya:
@s-urbaniak fixed modules structure, waiting for tests to be green

abhinavdahiya:
ok-to-test

abhinavdahiya:
ok-to-test

abhinavdahiya:
ok-to-test

Quentin-M:
ok to test

tests/smoke/aws/vars/aws-exp-np.tfvars
s-urbaniak: Agreed, we'll follow up with this across all platforms.
/cc @mxinden

mxinden: @aaronlevy Good catch! This is tackled for azure once PR #1154 is merged, but is still missing on all other platforms.

mxinden: Created an issue for this. #1367

yifan-gu:
@abhinavdahiya Need a rebase again.

platforms/openstack/neutron/main.tf
```diff
@@ -169,3 +171,25 @@ module "secgroups" {
   cluster_name          = "${var.tectonic_cluster_name}"
   tectonic_experimental = "${var.tectonic_experimental}"
 }
+
+module "flannel-vxlan" {
+  source = "../../modules/net/flannel-vxlan"
```
coreypobrien: This needs one more `../` since the platform for openstack has `neutron` in the path as well.

platforms/openstack/neutron/main.tf
```diff
@@ -169,3 +171,25 @@ module "secgroups" {
...
+
+  flannel_image     = "${var.tectonic_container_images["flannel"]}"
+  flannel_cni_image = "${var.tectonic_container_images["flannel_cni"]}"
+  cluster_cidr      = "${var.tectonic_cluster_cidr}"
+
+  bootkube_id = "${module.bootkube.id}"
+}
+
+module "calico-network-policy" {
+  source = "../../modules/net/calico-network-policy"
```
coreypobrien: This needs one more ../ since the platform for openstack has neutron in the path as well.

coreypobrien:
Nice work! I ran a quick NetworkPolicy demo with `openstack/neutron` (after tweaking those module source paths) and it worked great there.

platforms/openstack/neutron/main.tf
abhinavdahiya: @coreypobrien Fixed it. thanks for pointing that out!

robszumski:
> I would like to priorize Azure though, before we merge this into master.

@s-urbaniak What do you mean by this?

s-urbaniak:
@robszumski sorry for being imprecise. I suggested to merge https://github.com/coreos/tectonic-installer/pull/1154 first. So from my side we are good to go with this PR and merge it once green 👍

PR #763, opened 2017-05-18
AWS: Support bringing your own IAM policy

alekssaul:
New variables `tectonic_aws_external_worker_arn` and `tectonic_aws_external_master_arn` allow users to bring in their own IAM Roles. 
Fixes: https://github.com/coreos/tectonic-installer/issues/666

/cc: @metral , @joshrosso

PR #785, opened 2017-05-19
Implement BYO node IAM roles

### LINKED ISSUE #666: AWS: Support bringing your own IAM policy
## Story

Users may desire to specify the exact AWS policy associated with their worker and controller roles. This would provide an alternative to having Tectonic create them for each new cluster.

User's may prefer an externally managed policy for:
- Ability to reuse a universal controller/worker policy across clusters
  -- Has its trade-offs
- Ability to restrict worker/controller right beyond what tectonic sets
- Ability to allow for workers/controllers to have more rights. e.g. if they want pods to be able to assume the default role and expect more privileges than tectonic would add to its policy by default.

alexsomesan:
This change allows users to specify their own external IAM role names for use as instance profiles on master and worker nodes.

Fixes #666 

/cc: @robszumski

### DIFF
--- a/platforms/aws/variables.tf
+++ b/platforms/aws/variables.tf
@@ -239,3 +239,15 @@ variable "tectonic_aws_region" {
   default     = "eu-west-1"
   description = "The target AWS region for the cluster."
 }
+
+variable "tectonic_aws_master_iam_role" {
+  type        = "string"
+  default     = ""
+  description = "IAM role to use for the instance profiles of master nodes."
+}
+
+variable "tectonic_aws_worker_iam_role" {
+  type        = "string"
+  default     = ""
+  description = "IAM role to use for the instance profiles of worker nodes."
+}
--- a/platforms/aws/variables.tf
+++ b/platforms/aws/variables.tf
@@ -240,14 +240,30 @@ variable "tectonic_aws_region" {
   description = "The target AWS region for the cluster."
 }
 
-variable "tectonic_aws_master_iam_role" {
-  type        = "string"
-  default     = ""
-  description = "IAM role to use for the instance profiles of master nodes."
+variable "tectonic_aws_master_iam_role_name" {
+  type    = "string"
+  default = ""
+
+  description = <<EOF
+(optional) Name of IAM role to use for the instance profiles of master nodes.
+The name is also the last part of a role's ARN.
+
+Example:
+ * Role ARN  = arn:aws:iam::123456789012:role/tectonic-installer
+ * Role Name = tectonic-installer
+EOF
 }
 
-variable "tectonic_aws_worker_iam_role" {
-  type        = "string"
-  default     = ""
-  description = "IAM role to use for the instance profiles of worker nodes."
+variable "tectonic_aws_worker_iam_role_name" {
+  type    = "string"
+  default = ""
+
+  description = <<EOF
+(optional) Name of IAM role to use for the instance profiles of worker nodes.
+The name is also the last part of a role's ARN.
+
+Example:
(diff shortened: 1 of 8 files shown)


alekssaul:
@alexsomesan ah... was working with a customer that had the need so I created something similar in; https://github.com/coreos/tectonic-installer/pull/763 . Let me know if you'd like me to drop mine

platforms/aws/variables.tf
```diff
@@ -206,3 +206,15 @@ variable "tectonic_aws_region" {
   default     = "eu-west-1"
   description = "The target AWS region for the cluster."
 }
+
+variable "tectonic_aws_master_iam_role" {
+  type        = "string"
+  default     = ""
+  description = "IAM role to use for the instance profiles of master nodes."
```
robszumski: Just so we're super clear with customers, is this an IAM role name? or an ID?

sym3tri:
@robszumski thoughts on adding GUI support?

platforms/aws/variables.tf
everydaynerd: We use the policy ID when webhave to customize our installer: 
where as mine; doesn't create either.. and just requires the 
```
arn:aws:iam::982404467199:instance-profile/demo-cluster-master-profile
```

alexsomesan: @everydaynerd Appreciate your input here.

I've been torn about what is the easiest input to request from users in this case. 
Seemed to me like *role* is the concept users might be most accustomed to, whereas instance profiles are immutable and usually just a "shadow copy" of said role in EC2.
So in that sense, asking the user for a role made the most sense to me. That way they don't have to struggle with the difference between a role and instance profile and we just create the instance profile for them.

I'd appreciate your POV on this convention, from the perspective of a user.

alexsomesan: @robszumski 
That is certainly a **role name**, as per the IAM API's apparent preference for using "names" as identifiers (as opposed to ARNs or IDs).
See my comment above about the convention we're proposing to users with this change.

alexsomesan:
@robszumski Here's an example tfvars input for the master's role:
```
// Use arn:aws:iam::123456789012:role/test-master-role for master instances
//
tectonic_aws_master_iam_role = "test-master-role"

// Use arn:aws:iam::123456789012:role/test-worker-role for worker instances
//
tectonic_aws_worker_iam_role = "test-worker-role"
```

robszumski:
> thoughts on adding GUI support?

@sym3tri we can skip for now

platforms/aws/variables.tf
s-urbaniak: My suggestion is to emphasize this is being a role name in the description and also give an example.

platforms/aws/variables.tf
```diff
@@ -239,3 +239,15 @@ variable "tectonic_aws_region" {
...
+}
+
+variable "tectonic_aws_worker_iam_role" {
+  type        = "string"
+  default     = ""
+  description = "IAM role to use for the instance profiles of worker nodes."
```
s-urbaniak: Same comment as above: I would emphasize this is a role name including an example.

sym3tri:
@alexsomesan bump. Ready to merge once rebased & Jenkins is happy?

alexsomesan:
I rebased. Feel free to merge once Jenkins is happy.

Quentin-M:
ok to test

Quentin-M:
The docs are outdated as revealed by the tests. Can you run `make docs` and commit the changes? Thanks.

alexsomesan:
@Quentin-M updated docs. Feel free to merge if it looks good now.

everydaynerd:
THANKS!  This is a HUGE help!

PR #792, opened 2017-05-19
Bump bootkube v0.4.4 & hyperkube v1.6.4_coreos.0

aaronlevy:
There was a change in bootkube where it now generates additional self-hosted etcd assets as part of the rendering step - so those assets have also been added here.

### DIFF
--- a/modules/bootkube/assets.tf
+++ b/modules/bootkube/assets.tf
@@ -47,7 +47,8 @@ resource "template_dir" "bootkube" {
     etcd_cert_flag = "${data.null_data_source.etcd.outputs.cert_flag}"
     etcd_key_flag  = "${data.null_data_source.etcd.outputs.key_flag}"
 
-    etcd_service_ip = "${cidrhost(var.service_cidr, 15)}"
+    etcd_service_ip           = "${cidrhost(var.service_cidr, 15)}"
+    bootstrap_etcd_service_ip = "${cidrhost(var.service_cidr, 200)}"
 
     cloud_provider = "${var.cloud_provider}"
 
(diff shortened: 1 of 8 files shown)


modules/bootkube/assets.tf
```diff
@@ -139,10 +140,28 @@ data "template_file" "bootstrap-etcd" {
   template = "${file("${path.module}/resources/experimental/bootstrap-manifests/bootstrap-etcd.yaml")}"
 
   vars {
+    bootstrap_etcd_service_ip = "${cidrhost(var.service_cidr, 200)}"
     etcd_image = "${var.container_images["etcd"]}"
```
Quentin-M: alignment for the structure-check test

Quentin-M: can run `terraform fmt`

modules/bootkube/assets.tf
```diff
@@ -139,10 +140,28 @@ data "template_file" "bootstrap-etcd" {
...
+    bootstrap_etcd_service_ip = "${cidrhost(var.service_cidr, 200)}"
+  }
+}
+
+data "template_file" "migrate-etcd-cluster" {
+  template = "${file("${path.module}/resources/experimental/etcd/migrate-etcd-cluster.json")}"
+
+  vars {
+    etcd_version = "${var.etcd_version}"
+    bootstrap_etcd_service_ip = "${cidrhost(var.service_cidr, 200)}"
```
Quentin-M: same here

modules/bootkube/assets.tf
```diff
@@ -139,10 +140,28 @@ data "template_file" "bootstrap-etcd" {
   template = "${file("${path.module}/resources/experimental/bootstrap-manifests/bootstrap-etcd.yaml")}"
 
   vars {
+    bootstrap_etcd_service_ip = "${cidrhost(var.service_cidr, 200)}"
     etcd_image = "${var.container_images["etcd"]}"
   }
 }
 
+data "template_file" "bootstrap-etcd-service" {
```
Quentin-M: is there a way we can squash by using `template_dir` instead? In other words, templating the whole folder (recursively if needed) rather than individual files.

Quentin-M: we have something in bootkube.sh that moves manifests for them to be used or not. i'd like to see the less template_file/local_file as possible.

s-urbaniak:
This generally is looking good, but I agree with @Quentin-M that we should squash all those single templates into one `template_dir` resource declaration.

I created #805 as a follow-up, hence closing here.

PR #796, opened 2017-05-20
Provide masters access to s3:HeadObject and s3:ListBucket

chancez:
Give masters access to s3:HeadObject to avoid 403 when key doesn't exist
within bucket (due to s3 propagation delays, details below).

Reasons for this are: @diegs and I were troubleshooting his cluster not coming up. We found the masters had a failed `init-assets.service`, logs gave us a 403 Permission Denied error on `aws s3 cp` for `HeadObject`. It turns out you get a 403 if a object's key (which is a file or directory) doesn't exist, and you don't have permissions to ListBucket/HeadObject, as mentioned in  mentioned in https://github.com/aws/aws-cli/issues/1689 and documented in https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectHEAD.html#rest-object-head-permissions

It's pretty hard to reproduce, as it seems to be due to the eventual consistency model in S3. This specific case is mentioned here here: https://docs.aws.amazon.com/AmazonS3/latest/dev/Introduction.html#ConsistencyModel

I believe this is the relevant case:

> A process writes a new object to Amazon S3 and immediately lists keys within its bucket. Until the change is fully propagated, the object might not appear in the list.

We were able to restart the service and it downloaded all the assets fine.

### DIFF
--- a/modules/aws/master-asg/master.tf
+++ b/modules/aws/master-asg/master.tf
@@ -142,7 +142,8 @@ resource "aws_iam_role_policy" "master_policy" {
     {
       "Action" : [
         "s3:GetObject",
-        "s3:HeadObject"
+        "s3:HeadObject",
+        "s3:ListBucket"
       ],
       "Resource": "arn:aws:s3:::*",
       "Effect": "Allow"
--- a/modules/aws/master-asg/master.tf
+++ b/modules/aws/master-asg/master.tf
@@ -141,7 +141,8 @@ resource "aws_iam_role_policy" "master_policy" {
     },
     {
       "Action" : [
-        "s3:GetObject"
+        "s3:GetObject",
+        "s3:HeadObject"
       ],
       "Resource": "arn:aws:s3:::*",
       "Effect": "Allow"


chancez:
I was able to reproduce the s3 error locally. Here's it using the just an s3 policy that allows `s3:GetObject`, using the awscli docker image used by the `init-assets.service` (via `s3-puller.sh`).

```
bash-4.3# aws --region=us-east-2 s3 cp s3://chancez-test-bucket/testfile.txt /tmp
download: s3://chancez-test-bucket/testfile.txt to tmp/testfile.txt
bash-4.3# aws --region=us-east-2 s3 cp s3://chancez-test-bucket/testfile.txt1231231 /tmp
fatal error: An error occurred (403) when calling the HeadObject operation: Forbidden
```

After adding `s3:HeadObject` I was still getting the same error. Once I added `s3:ListBucket` to the policy, I am now getting a proper 404:

```
bash-4.3# aws --region=us-east-2 s3 cp s3://chancez-test-bucket/testfile.txt1231231 /tmp
fatal error: An error occurred (404) when calling the HeadObject operation: Key "testfile.txt1231231" does not exist
```

I'll add ListBucket to this my PR, which is used to list the objects within a bucket.

alexsomesan:
Hey Chance, 
I also struggled a bit with that strange mapping of S3 permissions to HTTP operations.
This doc helped a lot: http://docs.aws.amazon.com/AmazonS3/latest/dev/using-with-s3-actions.html
See the table where they map IAM permissions to S3 Operations.

chancez:
I actually mostly get the permissions, it's more that I'm unsure which operations `aws s3 cp` end's up calling. I think I've got most of that covered in this PR now however.

kans:
@chancez your tests timed out.  Would you rebase on master to pick up better tests please?

chancez:
@kans rebased.

kans:
@chancez according to the docs, It looks like you need [partOf](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#PartOf=) too...

chancez:
Looks like everything took too long (37 min) so it got killed.

kans:
@chancez yeah, timed out... which most likely means your cluster did not come up.  Could be related to this changes, but probably not...

modules/aws/ignition/resources/services/init-assets.service
```diff
@@ -8,6 +8,9 @@ Type=oneshot
 RemainAfterExit=true
 WorkingDirectory=/opt/tectonic
 
+Restart=on-failure
```
squat: Services of type `oneshot` cannot restart. This is only valid for `simple`, `forking` etc

chancez:
I was being told this by @kans too. I can update it to only include the policy changes, which I think are still valid. What do you think @squat ?

kans:
@chancez please do that

chancez:
@squat @kans I have removed the commit containing the restart changes, and updated the PR title/message to reflect these changes.

PR #417, opened 2017-04-28
platforms/azure: fix the usage of an existing DNS zone resource group

metral:
This PR amends https://github.com/coreos/tectonic-installer/pull/374 to:

- Cleanup the public_ips & master FQDN's from using long and incorrect FQDN's e.g. when using an existing DNS zone resource group
- Rename `tectonic_azure_create_dns_zone` to `tectonic_azure_use_custom_fqdn` to convey its functionality better around which FQDN's is used: either 1) custom FQDN that uses `tectonic_base_domain` (default) or 2) the [FQDN](https://www.terraform.io/docs/providers/azurerm/r/public_ip.html#fqdn) set up by Azure.
- Update Azure's `etcd_service_ip` in the bootkube module (similar to recent changes for AWS: https://github.com/coreos/tectonic-installer/pull/411)

This has been tested against Terraform v0.9.4 (since it's now the [default](https://github.com/coreos/tectonic-installer/pull/219) in master).

/cc @coresolve

PR #803, opened 2017-05-22
azure: restore create_dns_zone and make use_custom_fqdn optional

metral:
This PR fixes the following issues introduced in https://github.com/coreos/tectonic-installer/pull/417 and https://github.com/coreos/tectonic-installer/pull/602, which replaced `create_dns_zone` with `use_custom_fqdn` and changed the default value of `use_custom_fqdn`:

- If `use_custom_fqdn` was `false` a DNS Zone would not be created
- If `user_custom_fqdn` was `true`, which was the default, it would destroy an existing DNS zone whether the user intended it or not, as it took the place of `create_dns_zone` in the logic of the DNS module.

This PR restores `create_dns_zone` and limits its use to the DNS module, and limits `use_custom_fqdn` to the Master module, creating the ability to configure both if desired, and separating the use cases out as independent variables.


s-urbaniak:
ok to test

PR #808, opened 2017-05-22
Existing resources

jmeisner3:


.gitignore
```diff
@@ -10,3 +10,6 @@ tectonic-license.txt
 terraform.tfstate*
 terraform.tfvars
 /build/
+generated
```
ggreer: Note that this will ignore all files or directories named "generated", including vendored stuff. You probably want to use `/generated/`.

PR #811, opened 2017-05-22
docs: add terraform version to install docs.

zbwright:


Quentin-M:
Yeap, it's already the case. But it's not a final v0.9.6 version of Terraform - it is built on top of their master branch.

Quentin-M:
Yeap, it's already the case. But it's not a final v0.9.6 version of Terraform - it is built on top of their master branch.

sym3tri:
@robszumski @zbwright please update to reflect what Quentin said. We have our own special build since 0.9.6 is not out yet. It can be found inside the tarball. _only_ that terraform binary should be used (this will change in a future release).

zbwright:
ok. tell me, please: I imagine this means that ONLY that version of Terraform can be used for all interactions with the generated deploy, including destroying the cluster down the line. correct?

zbwright:
I will change the bullet points to read:

**Terraform:** Tectonic Installer includes and requires a custom version of Terraform, built from the Terraform pre-release version 0.9.6. This build is included in the Tectonic Installer tarball, and may also be downloaded directly from [GitHub](https://github.com/coreos/terraform/releases/tag/v0.9.6-fcdf494). Tectonic Installer will be updated to Terraform v0.9.6 when it is released.

Will that work?

sym3tri:
> ok. tell me, please: I imagine this means that ONLY that version of Terraform can be used for all interactions with the generated deploy, including destroying the cluster down the line. correct?

correct.

What you said is good but it will create a docs maintenance burden for you as the version will probably change with each release. Thoughts on making it more general, like:

**Terraform:** Tectonic Installer includes and requires a specific version of Terraform. This is included in the Tectonic Installer tarball. See the [Tectonic Installer release notes](https://coreos.com/tectonic/releases/) for information about which Terraform versions are compatible.

zbwright:
Works for me. I'll suggest adding a Terraform version note at the top of the Release Notes list, along with the Kubernetes version.

sym3tri:
:+1: thanks.
Yeah we should have those versions in there anyway.

zbwright:
I believe this just needs a LGTM @robszumski @sym3tri

Issue #819, opened 2017-05-23
smoke tests continue to run even if 'terraform apply' fails


sym3tri:
They should not. If `terraform apply` fails the entire test pipeline should exit early (but still cleanup via destroy).

squat:
Fixed via #893

PR #823, opened 2017-05-23
Smoke test bail apply

squat:
Ensure smoke tests exit if `terraform apply` fails.

Addresses #819

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -61,7 +61,10 @@ pipeline {
                              ]) {
             unstash 'installer'
             unstash 'sanity'
-            sh '''
+            sh '''#!/bin/bash -ex
+            set -o pipefail
+            shopt -s expand_aliases
+
             # Set required configuration
             export PLATFORM=aws
             export CLUSTER="tf-${PLATFORM}-${BRANCH_NAME}-${BUILD_ID}"


Jenkinsfile
```diff
@@ -62,6 +62,8 @@ pipeline {
             unstash 'installer'
             unstash 'sanity'
             sh '''
+            # Exit if any command fails
+            set -e
```
ggreer: Just curious, but do we know what shell `sh` runs? Some of these scripts (like on line 36) have shebangs for `/bin/bash`. They might be using a different shell than steps without shebangs.

squat: pretty sure it's straight `sh`

ggreer:
Looks like the regression was introduced in #646.

squat:
mhm.. the issue is that by doing `| filter` we mask the error exit code of the `make apply` so we need `pipefail`

Jenkinsfile
```diff
@@ -61,7 +61,8 @@ pipeline {
                              ]) {
             unstash 'installer'
             unstash 'sanity'
-            sh '''
+            sh """#!/bin/bash -ex -o pipefail
+
```
ggreer: You need single quotes here, not double quotes.

squat: grr we need to be more consistent across this file

ggreer:
Fails because filter is broken. We should have a separate PR that reverts the filter stuff.

kans:
@squat 
```+ alias filter=/home/ubuntu/workspace/r_tectonic-installer_PR-823-77SS223TD2KXKHELHY34AGVWHYRRI5CZ7N4XGKR33Y5UJ6PQ36FQ/installer/scripts/filter.sh

+ make plan

+ filter

/home/ubuntu/workspace/r_tectonic-installer_PR-823-77SS223TD2KXKHELHY34AGVWHYRRI5CZ7N4XGKR33Y5UJ6PQ36FQ@tmp/durable-93663a4c/script.sh: line 22: filter: command not found
```

At least it didn't run the smoke tests :P

squat:
haha true, at least it stops now! It is failing on filter because we need to explicitly allow alias expansion. Should be good now I think

squat:
ooofta

PR #830, opened 2017-05-23
jenkins - lets trace terrorform

kans:


alexsomesan:
Turning on tracing like that is EXTREMELY verbose.

This is going to result in very unusable logs, which will make it miserable for anyone trying to inspect ANY issue. Most build problems however are debuggable with the standard error messages.

I also use tracing regularly to debug, but I think we should only turn it on as a last resort and only for specific runs, not as the default.

It would be really useful to have a simple way of enabling one-off traced builds. I would focus effort on that approach.

kans:
@alexsomesan Sorry, we were just trying to get Jenkins up :-/

PR #854, opened 2017-05-24
Change console to serve on port 8080.

diegs:
Also make it run as non-root.

Fixes #847

PR #855, opened 2017-05-24
modules/tectonic/resources/manifests: change console port to 8080

### LINKED ISSUE #847: Console should not run on port 80 if possible
Currently the console listens on port 80 (https://github.com/coreos/tectonic-installer/blob/master/modules/tectonic/resources/manifests/console/deployment.yaml#L42). It is behind a service (https://github.com/coreos/tectonic-installer/blob/master/modules/tectonic/resources/manifests/console/service.yaml) also exposing 80.

Listening on a low port requires running the pod as root or granting it `CAP_NET_BIND_SERVICE`. Would it be possible to run the console pod on a non-privileged port such as 8080 and then just set that `targetPort` in the service?

cc @amrutac @aaronlevy @sym3tri

squat:
fixes #847 

cc @diegs

### DIFF
--- a/modules/tectonic/resources/manifests/ingress/ingress.yaml
+++ b/modules/tectonic/resources/manifests/ingress/ingress.yaml
@@ -19,7 +19,7 @@ spec:
       - path: /
         backend:
           serviceName: tectonic-console
-          servicePort: 80
+          servicePort: 8080
       - path: /identity
         backend:
           serviceName: tectonic-identity
--- a/modules/tectonic/resources/manifests/console/deployment.yaml
+++ b/modules/tectonic/resources/manifests/console/deployment.yaml
@@ -39,7 +39,7 @@ spec:
               name: tectonic-config
               key: kubeAPIServerURL
         - name: BRIDGE_LISTEN
-          value: http://0.0.0.0:80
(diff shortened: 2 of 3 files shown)


diegs:
I just put up https://github.com/coreos/tectonic-installer/pull/854

modules/tectonic/resources/manifests/ingress/ingress.yaml
```diff
@@ -19,8 +19,8 @@ spec:
       - path: /
         backend:
           serviceName: tectonic-console
-          servicePort: 80
+          servicePort: 8080
```
diegs: Is this an intrusive change for existing users? I have to write a migration for this in the KVO.

sym3tri:
Let's hold on this for now and address in the _next_ release.

modules/tectonic/resources/manifests/ingress/ingress.yaml
squat: not an intrusive change. access to the console is handled by the ingress resource, which forwards requests to the console service.

squat: closing in favor of #854

PR #856, opened 2017-05-24
modules/tectonic/resources: updating the identity deployment to be more resilient

### LINKED ISSUE #849: Tectonic identity liveness probe timeout limit is too low
BUG REPORT 

What happened? https://github.com/coreos/tectonic-forum/issues/126

What you expected to happen? tectonic identity currently has its liveliness probe timeout set to 1s https://github.com/coreos/tectonic-installer/blob/master/modules/tectonic/resources/manifests/identity/deployment.yaml#L57. We should increase this value to avoid false health check failure. Also since Dex is expected to be HA we should increase replica count to 2

rithujohn191:
Fixes #849


rithujohn191:
Punting to next release

Issue #485, opened 2017-05-02
Move all the scripting from the jenkinsfile into a bash script or Makefile


sym3tri:
Right now it is a pain to reproduce Jenkins builds locally.

All of this logic should be executable locally:
https://github.com/coreos/tectonic-installer/blob/master/Jenkinsfile

squat:
Almost all of this was addressed by https://github.com/coreos/tectonic-installer/pull/930. Will look into cleaning more today.

Quentin-M:
I feel like this is pretty much done as of today. https://github.com/coreos/tectonic-installer/pull/1085 will help further with the initial goal of this issue.

PR #862, opened 2017-05-25
Jenkinsfile: test experimental features

squat:
Our smoke tests failed to catch a regression with one of the
experimental features because the smoke tests do not enable
`tectonic_experimental`. This PR does two things:
1. it adds a parallel step to the tests that runs with the experimental features enabled.
2. it moves the inline jenkins test code to its own file.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -91,48 +91,22 @@ pipeline {
                              ]) {
             unstash 'installer'
             unstash 'sanity'
-            sh '''#!/bin/bash -ex
-            set -o pipefail
-            shopt -s expand_aliases
-
-            # Set required configuration
-            export PLATFORM=aws
-            export CLUSTER="tf-${PLATFORM}-${BRANCH_NAME}-${BUILD_ID}"
-
-            # s3 buckets require lowercase names
-            export TF_VAR_tectonic_cluster_name=$(echo ${CLUSTER} | awk '{print tolower($0)}')
-
-            # randomly select region
-            REGIONS=(us-east-1 us-east-2 us-west-1 us-west-2)
-            export CHANGE_ID=${CHANGE_ID:-${BUILD_ID}}
-            i=$(( ${CHANGE_ID} % ${#REGIONS[@]} ))
-            export TF_VAR_tectonic_aws_region="${REGIONS[$i]}"
-            export AWS_REGION="${REGIONS[$i]}"
-            echo "selected region: ${TF_VAR_tectonic_aws_region}"
-            # make core utils accessible to make
-            export PATH=/bin:${PATH}
-
-            # Create local config
-            make localconfig
-
-            # Use smoke test configuration for deployment
-            ln -sf ${WORKSPACE}/test/aws.tfvars ${WORKSPACE}/build/${CLUSTER}/terraform.tfvars
-
-            alias filter=${WORKSPACE}/installer/scripts/filter.sh
-
-            make plan | filter
-            make apply | filter
-
-            # TODO: replace in Go
-            CONFIG=${WORKSPACE}/build/${CLUSTER}/terraform.tfvars
-            MASTER_COUNT=$(grep tectonic_master_count ${CONFIG} | awk -F "=" '{gsub(/"/, "", $2); print $2}')
-            WORKER_COUNT=$(grep tectonic_worker_count ${CONFIG} | awk -F "=" '{gsub(/"/, "", $2); print $2}')
-
-            export NODE_COUNT=$(( ${MASTER_COUNT} + ${WORKER_COUNT} ))
-
-            export TEST_KUBECONFIG=${WORKSPACE}/build/${CLUSTER}/generated/auth/kubeconfig
-            installer/bin/sanity -test.v -test.parallel=1
-            '''
+            sh '${WORKSPACE}/test/scripts/aws.sh aws.tfvars'
+            }
+          },
+          "TerraForm: AWS-experimental": {
+            withCredentials([file(credentialsId: 'tectonic-pull', variable: 'TF_VAR_tectonic_pull_secret_path'),
+                             file(credentialsId: 'tectonic-license', variable: 'TF_VAR_tectonic_license_path'),
+                             [
+                               $class: 'UsernamePasswordMultiBinding',
+                               credentialsId: 'tectonic-aws',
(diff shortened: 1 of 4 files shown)


Jenkinsfile
```diff
@@ -157,28 +131,8 @@ pipeline {
...
-          i=$(( ${CHANGE_ID} % ${#REGIONS[@]} ))
-          export TF_VAR_tectonic_aws_region="${REGIONS[$i]}"
-          export AWS_REGION="${REGIONS[$i]}"
-          echo "selected region: ${TF_VAR_tectonic_aws_region}"
-
-          echo "Destroying ${CLUSTER}..."
-          make destroy || make destroy || make destroy
-        '''
+        sh '${WORKSPACE}/test/scripts/aws-destroy aws.tfvars'
+        sh '${WORKSPACE}/test/scripts/aws-destroy aws-exp.tfvars'
```
kans: What happens if the first sh throws?

Jenkinsfile
```diff
@@ -91,48 +91,22 @@ pipeline {
...
-
-            export NODE_COUNT=$(( ${MASTER_COUNT} + ${WORKER_COUNT} ))
-
-            export TEST_KUBECONFIG=${WORKSPACE}/build/${CLUSTER}/generated/auth/kubeconfig
-            installer/bin/sanity -test.v -test.parallel=1
-            '''
+            sh '${WORKSPACE}/test/scripts/aws aws.tfvars'
+            }
+          },
+          "TerraForm: AWS-experimental": {
```
kans: Is  it cromulent to run multiple tfs in the same dir at the same time?

test/aws-exp.tfvars
```diff
@@ -0,0 +1,33 @@
+tectonic_worker_count = "4"
```
sym3tri: You can scale this down to 1 or 2. There's no need for 4 nodes.

test/aws-exp.tfvars
```diff
@@ -0,0 +1,33 @@
...
+
+tectonic_master_count = "3"
+
+tectonic_etcd_count = "3"
```
sym3tri: This has no effect when using self-hosted etcd right? If not please remove.

test/aws-exp.tfvars
```diff
@@ -0,0 +1,33 @@
...
+
+tectonic_etcd_servers = [""]
```
sym3tri: necessary?

test/aws-exp.tfvars
```diff
@@ -0,0 +1,33 @@
...
+
+tectonic_cl_channel = "stable"
+
+tectonic_admin_email = "example@coreos.com"
+
+tectonic_admin_password_hash = "$2a$12$k9wa31uE/4uD9aVtT/vNtOZwxXyEJ/9DwXXEYB/eUpb9fvEPsH/kO" # PASSWORD
+
+tectonic_ca_cert = ""
+
+tectonic_ca_key = ""
```
sym3tri: While your at it can you please provide a test CA to ensure the "provide your own CA" path works too?

Jenkinsfile
sym3tri: Seems weird having 2 side-by-side. Only 1 will ever be valid.

sym3tri: Maybe we should exec the destroy in each block above after the apply steps? Then this would be the "catch all" in case that one fails. If the first one succeeds I think it'll just be a noop.

squat: Having two destroys is required in order to delete the different clusters that would be launched by the different `applys`. the destroy script uses the var file name to generate the build dir name so each invocation will destroy a different cluster.

test/aws-exp.tfvars
squat: I can add a CA though, though this should really be a separate test in the matrix.

squat:
Tests failing because of resource limits on instances

test/aws-exp.tfvars
sym3tri: In an ideal world yes, but we can't boot test clusters for ever possible permutation. I just want to increase our coverage with the fewest number of test files.

test/scripts/aws
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash -ex
+set -o pipefail
+shopt -s expand_aliases
+
+# Set the specified vars file
+TF_VARS_FILE=$1
+PLATFORM=$(echo ${TF_VARS_FILE} | cut -d "." -f 1)
+
+# Set required configuration
+export CLUSTER="tf-${PLATFORM}-${BRANCH_NAME}-${BUILD_ID}"
```
sym3tri: platform is special & used elsehwere

sym3tri: BRANCH_NAME should just be PR number due to char constraints

Jenkinsfile
kans: +1

test/scripts/aws
squat: Just checked and apparently `BRANCH_NAME` evaluates to `pr-862` in this case, so it seems to be the pull request name. To shorten this cluster name, I'm setting it to be `${TEST_NAME}-${BRANCH_NAME}-{BUILD_ID}` so it may be `aws-exp-pr-862-10`. This should work for the time being.

kans:
I removed Singapore because its too flaky - https://github.com/coreos/tectonic-installer/pull/877/files

squat:
ok. well tests are passing here! I'll rebase

sym3tri:
NOTE: this is related https://github.com/coreos/tectonic-installer/issues/485

test/scripts/aws
```diff
@@ -0,0 +1,45 @@
...
+# Create local config
+make localconfig
+
+# Use smoke test configuration for deployment
+ln -sf ${WORKSPACE}/test/${TF_VARS_FILE} ${WORKSPACE}/build/${CLUSTER}/terraform.tfvars
+
+alias filter=${WORKSPACE}/installer/scripts/filter.sh
+
+make plan | filter
+make apply | filter
```
sym3tri: Don't forget to optionally bail after plan. I don't want to run this on every PR until we're more confident our resource limits can handle it.

squat:
added bail option. PTAL @sym3tri @kans

kans:
@squat would you rename the scripts to *.sh and merge?

squat:
Will do @kans

Issue #463, opened 2017-05-02
Consolidate all images to quay.io


sym3tri:
Images we don't manage should be mirrored to quay.io. Currently not on quay.io:

- [ ] busybox (via flannel: https://github.com/coreos/tectonic-installer/blob/3ac7f7519ee071f04e621a4ae38ca0e88b8f82c3/modules/bootkube/resources/manifests/kube-flannel.yaml)
- [ ] gcr.io/google_containers/pause
- [ ] gcr.io/google_containers/heapster
- [ ] gcr.io/google_containers/addon-resizer
- [ ] gcr.io/google_containers/nginx-ingress-controller
- [ ] gcr.io/google_containers/k8s-dns-kube-dns-amd64
- [ ] gcr.io/google_containers/k8s-dns-dnsmasq-nanny-amd64
- [ ] gcr.io/google_containers/k8s-dns-sidecar-amd64

philips:
cc @colhom

colhom:
related #779

Quentin-M:
We now have an intern working on integrating Quay Security Scanning / [Clair](https://github.com/coreos/clair) with Tectonic. Every images that come from Quay will be supported.

Quentin-M:
We will need an automated job to do that.

sym3tri:
moved: https://coreosdev.atlassian.net/browse/INST-192

PR #873, opened 2017-05-25
modules/tectonic/resources: add security context to ingress controller

squat:
Ensure ingress controller runs as non-root. This is trickier than some other pods in Tectonic since we cannot make ingress controller run on ports other than 80 and 443 because these values are hardcoded into the ingress controller source [0] [1]. This means that we cannot make ingress controller run as non-root without adding additional CAPs to allow it to listen on privileged ports.

[0] https://github.com/kubernetes/ingress/blob/master/controllers/nginx/pkg/cmd/controller/nginx.go#L94
[1] https://github.com/kubernetes/ingress/blob/master/controllers/nginx/pkg/cmd/controller/nginx.go#L210

### DIFF
--- a/modules/tectonic/resources/manifests/ingress/hostport/daemonset.yaml
+++ b/modules/tectonic/resources/manifests/ingress/hostport/daemonset.yaml
@@ -69,4 +69,10 @@ spec:
         node-role.kubernetes.io/node: ""
       dnsPolicy: ClusterFirst
       restartPolicy: Always
+      securityContext:
+        capabilities:
+          add:
+          - NET_BIND_SERVICE
+        runAsNonRoot: true
+        runAsUser: 65534
       terminationGracePeriodSeconds: 60
--- a/modules/tectonic/resources/manifests/ingress/nodeport/deployment.yaml
+++ b/modules/tectonic/resources/manifests/ingress/nodeport/deployment.yaml
@@ -54,6 +54,13 @@ spec:
               path: /healthz
               port: 10254
               scheme: HTTP
+          securityContext:
+            capabilities:
+              add:
+              - NET_BIND_SERVICE
       dnsPolicy: ClusterFirst
       restartPolicy: Always
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 65534
       terminationGracePeriodSeconds: 60


diegs:
I wasn't able to get `CAP_NET_BIND_SERVICE` to work when trying to run the console as non-root; maybe you'll have more luck though.

Quentin-M:
Testing! 💯

Quentin-M:
- The `/nginx-ingress-controller` binary does not have the required caps, nor does the `nginx` binary. Both binaries try to bind.
- The `nobody` user does not have the rights for `/run/nginx.pid`, `/ingress-controller/ssl` (created), `/var/lib/nginx/` (created) nor `/etc/nginx`. 
- The `0.9.0-beta.4`'s apt repositories are broken, and therefore we can't install `libcap2-bin`. Thus a bump to `0.9.0-beta.4` is necessary.

The following Dockerfile appears to work:
```
FROM gcr.io/google_containers/nginx-ingress-controller:0.9.0-beta.7
RUN apt-get update && apt-get install -y libcap2-bin && rm -rf /var/lib/apt/lists/*
RUN setcap 'cap_net_bind_service=+ep' /nginx-ingress-controller
RUN setcap 'cap_net_bind_service=+ep' /usr/sbin/nginx
RUN mkdir -p /ingress-controller/ssl /var/lib/nginx/
RUN chmod -R 777 /run /ingress-controller /var/lib/nginx/ /etc/nginx
```

I pushed it to [quay.io/quentin_m/nginx-ingress-controller:0.9.0-beta.7.cap](quay.io/quentin_m/nginx-ingress-controller:0.9.0-beta.7.cap) and tested it.

If we add some kind of support for patching when we do https://github.com/coreos/tectonic-installer/issues/463, then it'd be usable.

cpanato:
@squat @s-urbaniak this PR still makes sense? if yes lets review and merge otherwise I will close and then in the future if needed someone can open a new PR

thanks

s-urbaniak:
I still endorse this PR, it would be great to have console be non-root. Re-ping @squat.

s-urbaniak:
having said that, once this is green, we need another upgrade path for the KVO.

s-urbaniak:
ok to test

diegs:
One thing we found when adding the capabilities to hyperkube was that it would add new layers and blow up the image size. I know very little about docker but I would check to see if that's the case here (apologies if it's obvious that it won't in this case).

lucab:
Are those privileged ports inside a pod netns or directly on the host? If it is inside a namespace, lowering `/proc/sys/net/ipv4/ip_unprivileged_port_start` may be enough. Despite the pathname, it affects both ipv4 and ipv6 sockets.

squeed:
I am going to fix this with the tectonic-ingress-controller.

PR #876, opened 2017-05-25
examples: note max k8s service cidr size

colhom:
Came up during a customer install. Haven't been able to find this limit in upstream docs, but validation snippet is [here](https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-apiserver/app/options/validation.go#L29-L32)

PR #878, opened 2017-05-25
config.tf: add a note about max size for the service CIDR

sym3tri:
moved from https://github.com/coreos/tectonic-installer/pull/876


colhom:
Thanks @sym3tri

sym3tri:
only docs. does not affect any functionality.

PR #885, opened 2017-05-25
frontend: Use Alert component for styling errors.

ggreer:
Before:
screen shot 2017-05-25 at 14 58 48
screen shot 2017-05-25 at 14 59 00

After:
screen shot 2017-05-25 at 14 57 35
screen shot 2017-05-25 at 14 59 21


tlwu2013:
Align the alert component with other screens, so looks good to me.

meghanschofield:
Hello! The only thing here is that we should start removing the little triangle "!" icons. 

If it's easy to do all at once, we should. If it's difficult, then we should do it in smaller steps. But they really aren't necessary, and lose meaning when used too much.

Thanks!

PR #886, opened 2017-05-25
jenkins: echo github release stuff

kans:


kans:
@sym3tri we just echo the length of the creds ```${#GITHUB_CREDENTIALS}```

PR #892, opened 2017-05-26
WIP- no review yet: Parametrize on-host rkt images

colhom:
Allows rkt to pull from registries other than quay.

### DIFF
--- a/modules/aws/ignition/resources/services/kubelet.service
+++ b/modules/aws/ignition/resources/services/kubelet.service
@@ -9,7 +9,8 @@ Environment="RKT_RUN_ARGS=--uuid-file-save=/var/cache/kubelet-pod.uuid \
   --volume var-lib-cni,kind=host,source=/var/lib/cni \
   --mount volume=var-lib-cni,target=/var/lib/cni \
   --volume var-log,kind=host,source=/var/log \
-  --mount volume=var-log,target=/var/log"
+  --mount volume=var-log,target=/var/log \
+  --insecure-options={{.rkt_insecure_options}}"
 
 ExecStartPre=/bin/mkdir -p /etc/kubernetes/manifests \
   /srv/kubernetes/manifests /etc/kubernetes/checkpoint-secrets \
--- a/modules/bootkube/variables.tf
+++ b/modules/bootkube/variables.tf
@@ -94,3 +94,15 @@ variable "oidc_groups_claim" {
   description = "The OpenID claim to use for specifying user groups (string or array of strings)"
   type        = "string"
 }
+
+variable "rkt_insecure_options" {
+  description = "Comma-separted values to provide for --insecure-options rkt flag"
+  type        = "string"
+  default     = "none"
+}
+
+variable "rkt_image_protocol" {
+  description = "Image protocol to use when fetching rkt images (eg: `docker://`)"
+  type        = "string"
+  default     = ""
+}
--- a/platforms/metal/cl/bootkube-controller.yaml.tmpl
+++ b/platforms/metal/cl/bootkube-controller.yaml.tmpl
@@ -7,6 +7,9 @@ systemd:
         - name: 40-etcd-cluster.conf
           contents: |
             [Service]
+            Environment="RKT_RUN_ARGS=--uuid-file-save=/var/lib/coreos/etcd-member-wrapper.uuid \
+              --insecure-options={{.rkt_insecure_options}}"
+            Environment="ETCD_IMAGE_URL={{.rkt_image_protocol}}{{.etcd_image_url}}"
             Environment="ETCD_IMAGE_TAG={{.etcd_image_tag}}"
             Environment="ETCD_NAME={{.etcd_name}}"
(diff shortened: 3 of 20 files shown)


platforms/metal/cl/bootkube-controller.yaml.tmpl
```diff
@@ -7,6 +7,8 @@ systemd:
         - name: 40-etcd-cluster.conf
           contents: |
             [Service]
+            Environment="RKT_RUN_ARGS=--uuid-file-save=/var/lib/coreos/etcd-member-wrapper.uuid --insecure-options={{.rkt_insecure_opts}}"
```
dghubble: Can you try using if conditions to ensure these lines don't even get added unless the user chooses to use them? I know the effect is the same, but this seems cluttered atm and I worry the complexity will lead to mistakes.

modules/bootkube/variables.tf
```diff
@@ -94,3 +94,15 @@ variable "oidc_groups_claim" {
...
   type        = "string"
 }
+
+variable "rkt_insecure_opts" {
+  description = "Comma-separted values to provide for --insecure-options rkt flag"
+  type        = "string"
+  default     = "none"
+}
+
+variable "rkt_image_proto" {
```
dghubble: proto -> protocol

modules/bootkube/variables.tf
```diff
@@ -94,3 +94,15 @@ variable "oidc_groups_claim" {
   description = "The OpenID claim to use for specifying user groups (string or array of strings)"
   type        = "string"
 }
+
+variable "rkt_insecure_opts" {
```
dghubble: opts -> options

modules/bootkube/variables.tf
colhom: done

modules/bootkube/variables.tf
colhom: done

platforms/metal/cl/bootkube-controller.yaml.tmpl
colhom: done

colhom:
@dghubble good call, those two variables are now commented out in the example tfvars files- very nice.

platforms/metal/cl/bootkube-controller.yaml.tmpl
colhom: I've noticed this complicates things in a lot of places. Between line continuations and terraform templating not supporting conditional blocks, it's really a lot cleaner to just have the default value for `rkt_insecure_options=none` and template it in regardless.

colhom: Imagine implementing these conditionals in /modules/aws/etcd/ignition.tf to set `RKT_RUN_ARGS` for `etcd.service` - I don't even think it would be possible. On side note, that platform is going to be next to support these configuration directives.

s-urbaniak:
Just a general comment from my side, as I have been experimenting with this too. I assume you are changing image names in `config.tf` in tandem with the changes included in this PR?

If so, this will work only during bootstrap time. If operators reference canonical image names (i.e. without additional prefixes for custom registries) these settings will be overridden.

colhom:
> changing image names in config.tf

@s-urbaniak I just override `var.tectonic_container_images` in `build/${CLUSTER}>/terraform.tfvars`, but yes, same effect.

> If operators reference canonical image names (i.e. without additional prefixes for custom registries) these settings will be overridden.

are you referring to issue at hand in #779 ? or something beyond that?

s-urbaniak:
@colhom when I was experimenting locally I configured local repository proxies and then set up image URLs like this:

```
tectonic_container_images = {
  hyperkube                       = "quay-proxy.k8s/coreos/hyperkube:v1.6.2_coreos.0"
...
  heapster                        = "gcr-proxy.k8s/google_containers/heapster:v1.3.0"
...
```

This works fine for the initial bootstrap, but if any operator resets the image name in the manifest to the canonical name i.e. `quay.io/coreos/hyperkube:v1.6.4_coreos.0` then the above configuration is aborted.

I don't know if this applies to your use case at all. If not, ignore my hint ;-)

modules/aws/ignition/resources/services/kubelet.service
```diff
@@ -3,13 +3,15 @@ Description=Kubelet via Hyperkube ACI
 
 [Service]
 EnvironmentFile=/etc/kubernetes/kubelet.env
+Environment="KUBELET_IMAGE={{.rkt_image_protocol}}{{.kubelet_image_url}}:{{.kubelet_image_tag}}"
```
colhom: @s-urbaniak I'll trade you another hint then! This will unfortunately break with kube-version-operator updates :( the `/etc/kubernetes/kubelet.env` file is the coordination point for updating the kubelet version).

cpanato:
@colhom can you please rebase or see if still makes sense. Otherwise, I will close this POR and then you can open another in the future. Thanks!!

s-urbaniak:
Let's close this one due to staleness. There is a user story PROD-14 covering the proposed solution. In order to implement this properly, I suggest to have a proper design discussion/proposal and priorization of PROD-14.

PR #865, opened 2017-05-25
azure/network: Conform Azure Network Security Groups to Tectonic generic platform standards

justaugustus:
Initial work to conform Azure NSGs (Network Security Groups) to [Tectonic Generic Platform guidelines](https://github.com/coreos/tectonic-installer/blob/master/Documentation/generic-platform.md)

This PR introduces NSGs (and supporting vars) for:
* etcd Availability Set
* master nodes
* worker nodes
* load balancers

Fixes: #337

Issue #875, opened 2017-05-25
installations break if CLUSTER is too many characters


sym3tri:
b/c things like ELB names are derived from the CLUSTER name, and have strict character limits. We need a smarter way to handle this.

sym3tri:
related https://github.com/coreos/tectonic-installer/issues/413

s-urbaniak:
Thought experiment: We have all the infrastructure in place to include custom TF plugins.

We could implement custom validation data sources invoked in `main.tf` of i.e. the AWS platform. Such naming validation data source would indeed be platform specific considering (alpha)numeric constraints and maximum lengths. Its input would mainly be i.e. `tectonic_cluster_name` and its output could be all the names needed for resources to be created, ranging from subnets, autoscaling groups et al.

This would also centralize the naming logic for a given platform, i.e. AWS. If the validation fails, none of the resources would be created in the first place, failing very early in the creation process.

I am hesitant to reflect these sorts of validations/computations using HCL. Nuances in evaluation patterns (see the ternary operator friction) bite us more than often, also we could easily unit test those validation data sources, moving quite some logic into a qualified programming language environment (Go).

/cc @alexsomesan for the verification of that idea.

sym3tri:
I actually had the same thought. Maybe there's a way to make it general enough so that it can be up-streamed eventually.

Worth trying out a POC once we get through some of the other higher-priority issues.

robszumski:
@sym3tri would love your thoughts on if this makes sense to tackle or punt for now.

sym3tri:
Closing in favor of a new issue to implement better overall validation
https://coreosdev.atlassian.net/browse/INST-201

Issue #435, opened 2017-05-01
AWS: Better naming for resources


sym3tri:
We need better naming for all the resources like: subnets, security groups, etc... 

Today it's really hard to manage and know which resource belongs to which cluster (i.e. use `mycluster-worker-subnet-1` instead of them all having the same name `mycluster`).

s-urbaniak:
while naming is an orthogonal issue, I believe that https://github.com/coreos/tectonic-installer/pull/776 already helps.

I will iterate through all created resources and apply the following scheme:

```
<clustername>-<resource_type>[-<instance_count>]
```

alexsomesan:
May I argue that adding the type of the resource to the name could be a bit redundant?

Type is usually inferred by the way the resource is accessed:
- in terraform fully qualified resource names include type: `module.vpc.aws_subnet.worker_subnet`
- in the AWS cli, the operation encodes the type to be accessed: `aws ec2 describe-subnets`
- AWS client libraries encode resource type in the method name: `ec2_client.DescribeSubnets()`

s-urbaniak:
@alexsomesan the referencing PR #807 doesn't change the naming scheme, just adds some missing tags on created resources. I agree that the resource type doesn't need to be part of the name.

Issue #413, opened 2017-04-27
AWS: better handling of long ELB names


robszumski:
My apply run failed because my internal API ELB's name was longer than 32 characters. Seems like we should either:

1. truncate the name automatically, at risk of collisions
2. validate and warn in the UI

I favor #1, since the name I entered wasn't particularly long and will probably be hit often-ish.

ggreer:
The frontend limits cluster names to 28 chars max. Just FYI, this is what the user sees when they use a cluster name with >19 characters:

![screen shot 2017-05-03 at 13 22 10](https://cloud.githubusercontent.com/assets/200121/25679872/c13d7c18-3003-11e7-9986-44dc338b4d4d.png)


    3 error(s) occurred:

    * module.masters.aws_elb.api-external: "name" cannot be longer than 32 characters: "testing-long-cluster-name-api-external"
    * module.masters.aws_elb.api-internal: "name" cannot be longer than 32 characters: "testing-long-cluster-name-api-internal"
    * module.masters.aws_elb.console: "name" cannot be longer than 32 characters: "testing-long-cluster-name-console"

I think the real fix is to not create super-long elb names.

Quentin-M:
Can truncate the cluster name using something like:

```
${replace(var.tectonic_cluster_name, "/(.{0,20})(.*)/", "$1")}
```

ggreer:
Alright, now master supports names up to 28 chars and we correctly validate the length & restrictions in the UI.

PR #895, opened 2017-05-26
[WIP] tectonic/terraform: Enable custom cluster names & enforce cluster name string length

justaugustus:
WIP - DO NOT MERGE

Adds `tectonic_cluster_prefix`, which allows for a prefix to the cluster name.
Customer request.

Some more things need to be done here, namely:
- [ ] Extend to support all platforms
- [ ] Update resources to use shorter names
- [ ] Enforce `tectonic_cluster_prefix` and `tectonic_cluster_name` string lengths
- [ ] Update variable descriptions
- [ ] Merge in #865 and change references from `"tectonic-cluster-${var.tectonic_cluster_name}"` to `"${var.tectonic_cluster_name}"`
- [ ] fixup (regenerate docs & `terraform fmt`)

### DIFF
--- a/config.tf
+++ b/config.tf
@@ -168,6 +168,12 @@ This applies only to cloud platforms.
 EOF
 }
 
+variable "tectonic_cluster_prefix" {
+  type = "string"
+  description = ""
+  default = ""
+}
+
 variable "tectonic_cluster_name" {
   type = "string"
 
--- a/modules/azure/master/lb.tf
+++ b/modules/azure/master/lb.tf
@@ -1,5 +1,5 @@
 resource "azurerm_lb" "tectonic_lb" {
-  name                = "api-lb"
+  name                = "${var.cluster_name}-api-lb"
   location            = "${var.location}"
   resource_group_name = "${var.resource_group_name}"
 
--- a/modules/azure/master/master.tf
+++ b/modules/azure/master/master.tf
@@ -66,7 +66,7 @@ resource "azurerm_virtual_machine_scale_set" "tectonic_masters" {
   }
 
   os_profile {
-    computer_name_prefix = "tectonic-master-"
+    computer_name_prefix = "${var.cluster_name}-master-"
     admin_username       = "core"
     admin_password       = ""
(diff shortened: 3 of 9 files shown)


robszumski:
Seems like this is applicable to all platforms, can you explore making the change for AWS and Openstack?

What happens when folks provide long prefixes? Can we be defensive about that?

robszumski:
Related: https://github.com/coreos/tectonic-installer/issues/875

robszumski:
Related: https://github.com/coreos/tectonic-installer/issues/435

robszumski:
Related: https://github.com/coreos/tectonic-installer/issues/413

config.tf
```diff
@@ -168,6 +168,12 @@ This applies only to cloud platforms.
 EOF
 }
 
+variable "tectonic_cluster_prefix" {
```
sym3tri: What's the purpose of a prefix? Why not just name the cluster with the prefix you want?

justaugustus: @sym3tri I think this be useful for instances where a customer decides to use the prefix for some alternate purpose, like establishing cost centers.

Say a cluster is named `team-x-tec-prod01` with the cost center being tied to `team-x`.
`tectonic_cluster_name = "team-x-tec-prod01"` would not be sufficient to uniquely identify the cost center, without some variable magic that would be specific to a customer's env / naming conventions.

Having `tectonic_cluster_prefix = "team-x"`, `tectonic_cluster_name = "tec-prod01"` would allow us to use the prefix as a tag for resources and thus uniquely define that cost center.

justaugustus: @sym3tri thoughts?

cpanato:
@justaugustus @sym3tri this PR still makes sense? if makes can you please rebase? Otherwise lets close it.

justaugustus:
@cpanato I'll take a look next week and maybe close this out!

cpanato:
Closing due inactivity. Please reopen if you want this PR merge :) thanks for all @justaugustus

PR #896, opened 2017-05-26
Idempotency fix: don't set IOPS value unless device type is IO1

alexsomesan:
I noticed that after successfully applying the templates from master, any subsequent `plan` operation will show that changes are required to update the value of the IOPS attribute in the block device mapping of the ASG launch configuration.
The reason for this is that during a previous apply, for a device type other than IO1 setting IOPS will just ignore the value. When doing a plan later on, the previously set value will not be returned by the AWS API, and so Terraform assumes that it has diverged and attempts to update it.

With this change we only attempt to set the actual IOPS value on volumes of type IO1, while for everything else we pass 0, which is the neutral value for this attribute. This avoids the perceived divergence and will not signal required changes later on.

### DIFF
--- a/modules/aws/etcd/nodes.tf
+++ b/modules/aws/etcd/nodes.tf
@@ -48,6 +48,6 @@ resource "aws_instance" "etcd_node" {
   root_block_device {
     volume_type = "${var.root_volume_type}"
     volume_size = "${var.root_volume_size}"
-    iops        = "${var.root_volume_iops}"
+    iops        = "${var.root_volume_type == "io1" ? var.root_volume_iops : 0}"
   }
 }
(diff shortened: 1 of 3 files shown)


s-urbaniak:
This generally LGTM modulo one open question Doesn't this also need to be set on the etcd node https://github.com/coreos/tectonic-installer/blob/9c27ef3/modules/aws/etcd/nodes.tf#L51 ?

s-urbaniak:
Whop, sorry for closing&reopening, pushed the wrong button ;)

alexsomesan:
You're right!
etcd is needs changes as well.

alexsomesan:
Fixed etcd now too.

bsiegel:
The default for `root_block_device.iops` when using a `gp2` volume type appears to be `100`, so this is still causing an idempotency issue:

```
-/+ module.etcd.aws_instance.etcd_node

    <snip>

    root_block_device.#:                       "1" => "1"
    root_block_device.0.delete_on_termination: "true" => "true"
    root_block_device.0.iops:                  "100" => "0" (forces new resource)
    root_block_device.0.volume_size:           "30" => "30"
    root_block_device.0.volume_type:           "gp2" => "gp2"

    <snip>
```

PR #898, opened 2017-05-26
docs: edit role list to reflect reality

robszumski:


robszumski:
I was asking around and didn't get a really solid answer. This was my best guestimate based on looking at the roles. I was mostly concerned with this table at least matching roles names with what we ship.

PR #906, opened 2017-05-26
frontend: Don't make poweron steps substeps.

ggreer:
Fixes an issue where margins were different between steps.


kans:
screen shot 2017-05-26 at 1 33 00 pm

PR #909, opened 2017-05-26
jenkins: timeouts

kans:
It looks like we don't destroy clusters if we timeout?


kans:
Distribute timeouts so that make destroy gets to run.  Also, snuck in the j4 flag to make.

PR #914, opened 2017-05-26
Test with a max length cluster name.

ggreer:
- Reduce duplicated code by moving common stuff to aws-common.sh
- Truncate or append to cluster name if name is too short or too long
- Set max cluster name length to 19 chars (max length that currently works)

We need this to prevent regressions, because people keep appending characters to the cluster name when creating other aws resources (elbs, s3 buckets, etc).

### DIFF
--- a/test/scripts/aws-common.sh
+++ b/test/scripts/aws-common.sh
@@ -1,0 +1,38 @@
+#!/bin/bash -ex
+set -o pipefail
+shopt -s expand_aliases
+
+# Set the specified vars file
+export TF_VARS_FILE=$1
+TEST_NAME=$(echo "${TF_VARS_FILE}" | cut -d "." -f 1)
+
+# Set required configuration
+export PLATFORM=aws
+CLUSTER="${TEST_NAME}-${BRANCH_NAME}-${BUILD_ID}"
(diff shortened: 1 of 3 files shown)


ggreer:
Related to #875

Quentin-M:
This variable used to be filled by Jenkinsfile. Where did it go?

Quentin-M:
Your change is systematically overridden by https://github.com/coreos/tectonic-installer/blob/master/test/scripts/aws.sh#L14.

ggreer:
If that's the case, then why are tests failing with it?

Quentin-M:
My bad. TF_VAR seem to be lower priority than the tfvars. This means that you are now deploying every single PRs under the same cluster name.

Quentin-M:
Rather than adding a smoke test that will just verify we can create clusters with a name of length > n, I would just truncate the cluster name where we need it to be truncated. So that users do not run into the issue. Adding a test won't prevent users from doing it.

ggreer:
Truncating loses the unique mapping between cluster names and resource names. eg: "tectonic-testing-cluster-100" and "tectonic-testing-cluster-101" would create AWS resources with the same names.

PR #922, opened 2017-05-30
Revert "Remove flock timeout from apiserver."

s-urbaniak:
Reverts coreos/tectonic-installer#915

As per discussion in https://github.com/coreos/tectonic-installer/pull/915#issuecomment-304720940 this was supposed to get in post-1.6.4.

PR #915, opened 2017-05-26
Remove flock timeout from apiserver.

diegs:
It doesn't contribute anything and leads to unnecessary apiserver
restarts during the bootstrap process.

PR #927, opened 2017-05-30
Revert "Revert "Remove flock timeout from apiserver. (#915)" (#922)"

diegs:
Adding back #915 now that the release branch has been cut.


squat:
This is a very good strategy for getting anything approved 😉

PR #928, opened 2017-05-30
Fix checkpointer container name.

diegs:
This reverts commit ee9842145dffaf95dcd926e86105ddcacc729d04.


squat:
ah i see... this is to re-do the change that https://github.com/coreos/tectonic-installer/commit/ee9842145dffaf95dcd926e86105ddcacc729d04 reverted.

PR #934, opened 2017-05-31
DO NOT MERGE - GlusterFS Setup - Do not merge

lpabon:
This is a proof of concept of setting up the nodes for GlusterFS. The setup does the following:

* Creates three disks per node
* Opens ports for Gluster node-node communication
* Sets up dm_thin_pool

### DIFF
--- a/modules/aws/ignition/ignition.tf
+++ b/modules/aws/ignition/ignition.tf
@@ -2,6 +2,7 @@ data "ignition_config" "main" {
   files = [
     "${data.ignition_file.max-user-watches.id}",
     "${data.ignition_file.s3-puller.id}",
+    "${data.ignition_file.dm_thin_pool.id}",
     "${data.ignition_file.init-assets.id}",
     "${data.ignition_file.detect-master.id}",
   ]
@@ -93,6 +94,16 @@ data "ignition_file" "s3-puller" {
 
   content {
     content = "${data.template_file.s3-puller.rendered}"
+  }
+}
+
+data "ignition_file" "dm_thin_pool" {
+  filesystem = "root"
+  path       = "/etc/modules-load.d/dm_thin_pool.conf"
+  mode       = "555"
+
+  content {
+    content = "${file("${path.module}/resources/dm_thin_pool.conf")}"
   }
 }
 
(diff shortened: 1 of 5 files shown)


alexsomesan:
Discussed this offline with @lpabon 
This is a really cool feature to have in Tectonic.

But before we can integrate this in the installer we need to find a way to make those block devices optional in the launch configuration. Reminds of the recently extended `tags` attribute which had similar limitations.

/cc: @s-urbaniak

s-urbaniak:
This sort of limitation was also a pain in the code reuse between openstack nova and neutron modules was limited.

Apparently there is no generic way in Terraform to enable/disable blocks inside a resource.

brianredbeard:
@lpabon A few quick questions to serve as proto documentation.  They're somewhat loaded answers as I'm pretty sure I know, but figured it's easy to document here:

  1)  The EBS devices provisioned in `worker.tf`, those are to be provisioned on each node acting as a gluster replica, correct?
  1)  Is the purpose of three discrete devices is for striping to improve performance?
  1)  What is the minimum size per EBS device?
  1)  If the purpose is for striping, what is the relationship between total space provisioned and available space?

lpabon:
Hi @brianredbeard, answers below
1. *The EBS devices provisioned in worker.tf, those are to be provisioned on each node acting as a gluster replica, correct?* Yes
1. *Is the purpose of three discrete devices is for striping to improve performance?* On baremetal with actual spindles, it would make a difference using devices, but on EBS it shouldn't, afaik.
1. *What is the minimum size per EBS device?* It would be the minimum to hold the requested volume + (XFS + pvcreate + vg metadata)
1. *If the purpose is for striping, what is the relationship between total space provisioned and available space?* There is no striping in GlusterFS replica 3 (which is the default replication style). GlusterFS itself supports striping across bricks (which are the node:directory pairs), but it is not used in Kubernetes because Heketi (The GlusterFS provisioner) does not have the support for that yet.

cpanato:
@lpabon @alexsomesan @s-urbaniak This PR still makes sense? if yes can rebase and we check what is missing? Otherwise, I will close and you can reopen another PR in the future

s-urbaniak:
Let's close this one, as it is pretty stale. We can always revive this PR in the context of a proper sprint planning/priorization, /cc @sym3tri @robszumski.

lpabon:
@s-urbaniak Yep, I agree.

Issue #933, opened 2017-05-30
Odd permissions set on detect-master.sh, s3-puller.sh, & init-assets.sh


ggreer:
When you create a cluster, you get some really weird permissions on some of the shell scripts. Here's the first master in a brand new cluster:

![screen shot 2017-05-30 at 16 34 35](https://cloud.githubusercontent.com/assets/200121/26609441/e904580e-4555-11e7-8ed6-c58a2d915080.png)

I'm pretty sure we don't want these permissions. The only reason stuff comes up is because we specify the shell in the various systemd units. (`ExecStart=/usr/bin/bash /opt/tectonic/init-assets.sh`)

diegs:
Fixed in #936

PR #936, opened 2017-05-31
Fix incorrect file modes.

diegs:
- 555 should be 755.
- 420 translates to 644.
- If octal is being used it should be prefixed by 0.
- Quotes are unnecessary.


ggreer:
Probably fixes #933.

ggreer:
Considering that 13 out of 18 instances of `mode` were incorrect, I'm worried this bug is going to show up again. Can the terraform linting step be modified to catch this error?

Issue #924, opened 2017-05-30
S3 download failure causes cluster bootstrap failure


sym3tri:
If the init-assets unit fails (due to an S3 download error), it causes a cascading failure to the other dependent units: bootkube & tectonic.

#796 documents the issue well, but it's unclear if that is a complete fix.

squat:
#796 is not a complete fix. The configuration in it is actually invalid, but even if it *were* valid downstream services would not be restarted after an initial failure of the init-assets service.

PR #937, opened 2017-05-31
init-assets should run until it works...

kans:
This PR also makes bootkube.sh and tecotnic.sh run sequentially after init-assets completes which was the behavior before [#177](https://github.com/coreos/tectonic-installer/pull/177/files#diff-4d0a4888c7144e1086cc7197d7196df7L53) .  This PR does not delete all dead code or increase test timeouts which we would probably need to bump by 5 minutes.


sym3tri:
please wait for review from @s-urbaniak before merging

s-urbaniak:
To add more solutions to the brainstorm-table:

1. As shown in https://github.com/coreos/coreos-kubernetes/issues/675#issuecomment-248458226 we could `ExecStartPre=/usr/bin/systemctl is-active init-assets.service` in `bootkube.service` and `ExecStartPre=/usr/bin/systemctl is-active bootkube.service` in `tectonic.service`.
2. We could add a `https://www.freedesktop.org/software/systemd/man/systemd.path.html#PathExists=` directive in i.e. `bootkube.service` which will be activated if and only if the downloaded assets exist.

modules/aws/ignition/resources/services/init-assets.service
```diff
@@ -1,19 +1,20 @@
 [Unit]
 Description=Download Tectonic Assets
 ConditionPathExists=!/opt/tectonic/init_assets.done
-Before=bootkube.service kubelet-env.service
+Before=kubelet-env.service
 
 [Service]
-Type=oneshot
-RemainAfterExit=true
+Restart=always
```
kans: What happens if the machine reboots?

modules/aws/ignition/resources/services/init-assets.service
```diff
@@ -1,19 +1,20 @@
...
+RestartSec=10
 WorkingDirectory=/opt/tectonic
-
 User=root
 Group=root
-
-ExecStart=/usr/bin/bash /opt/tectonic/init-assets.sh
-ExecStartPost=/bin/touch /opt/tectonic/init_assets.done
+TimeoutSec=infinity
```
kans: The ExecStartPres timeout otherwise.

modules/aws/ignition/ignition.tf
```diff
@@ -125,14 +123,3 @@ data "ignition_systemd_unit" "init-assets" {
...
 }
-
-data "ignition_systemd_unit" "bootkube" {
-  name    = "bootkube.service"
-  content = "${var.bootkube_service}"
-}
-
-data "ignition_systemd_unit" "tectonic" {
-  name    = "tectonic.service"
-  enable  = "${var.tectonic_service_disabled == 0 ? true : false}"
```
kans: Yes, but why is that a feature we support in the *tectonic* installer?

kans:
> First this PR unfortunately breaks "vanilla" k8s functionality 

Is this feature explicitly supported?  If so, I'd be happy to rework this PR.

> ...and [breaks] the separation of concerns between the three systemd service units init-assets.service, bootkube.service, and tectonic.service.

Init-assets, bootkube, and tectonic are tightly coupled in any case.  They must run sequentially, may take an indefinite amount of time, and they should only run once, *ever*.  Spreading the coupling across different services does not reduce the complexity of the system or make it easier to reason about.

> All of these services have been marked as Type=oneshot conciously to have a clear happens-before-relationship for the startup ordering between.

Sure, but you don't get a cluster if any of the scripts exit for any reason.

> All of this models the serial startup behavior in a pretty simple manner.

Mostly. I believe PostExecs only run if everything exits 0, so the lockfiles won't be around if the units fail.

> We could overcome this too by introducing dedicated target, eventually reached once the preceding service succeeds, but this introduces stage-based complexity bloat.

Agreed.

> ... This PR tries to delegate restart logic towards systemd and the outlined changes navigate around limitations in systemd. While I understand what we are trying to achieve here, I am not convinced that this is the right approach.  [We should]implement the retry logic inside init-assets.sh itself and leave the current (simple) construct

I disagree given the general impossibility of writing *robust*, mission critical bash and the number of existing bugs in the current scripts - init-assets.sh, tectonic.sh and bootkube.sh.  In so many words, if bash were capable of doing this, I suspect systemd wouldn't exist.

modules/aws/ignition/ignition.tf
s-urbaniak: It was requested in https://github.com/coreos/tectonic-installer/issues/133 and this feature is currently used by the Prometheus team in their CI pipeline on AWS to test operator versions.

s-urbaniak:
@kans thanks a lot for the good discussion! I totally agree about the robustness point about bash ;-) I was even thinking about implementing all the `tectonic.sh` and `init-assets` logic in Go (bootkube is already invoked directly) for that reason.

You have a good point about the complexity of orchestrating those three service unit files though which can be avoided. Once I am in the office I'll piggy-pack on this PR to create a follow-up commit using a templated systemd service unit file.

s-urbaniak:
@kans so I iterated on idea 2. described above and tested a commit which builds on top of this PR, PTAL: 

https://github.com/s-urbaniak/tectonic-installer/commit/1f1d8210dcbe1c5fa16bbefc17017ac8209e4a49

Also to have qualified answers to your comments I iterated on the single bullet items:

> > First this PR unfortunately breaks "vanilla" k8s functionality 

> Is this feature explicitly supported?  If so, I'd be happy to rework this PR.

Please see https://github.com/coreos/tectonic-installer/pull/937#discussion_r119776478

> > ...and [breaks] the separation of concerns between the three systemd service units init-assets.service, bootkube.service, and tectonic.service.

> Init-assets, bootkube, and tectonic are tightly coupled in any case.  They must run sequentially, may take an indefinite amount of time, and they should only run once, *ever*.  Spreading the coupling across different services does not reduce the complexity of the system or make it easier to reason about.

You have a point, sure. This reasoning also amplifies my conjecture to program at least `tectonic.sh` and `init-assets.sh` as a Go program which runs atomically in a systemd service unit similar to bootkube.

> > All of these services have been marked as Type=oneshot conciously to have a clear happens-before-relationship for the startup ordering between.

> Sure, but you don't get a cluster if any of the scripts exit for any reason.

Agreed

> > All of this models the serial startup behavior in a pretty simple manner.

> Mostly. I believe PostExecs only run if everything exits 0, so the lockfiles won't be around if the units fail.

That is indeed correct.

> > ... This PR tries to delegate restart logic towards systemd and the outlined changes navigate around limitations in systemd. While I understand what we are trying to achieve here, I am not convinced that this is the right approach.  [We should]implement the retry logic inside init-assets.sh itself and leave the current (simple) construct

> I disagree given the general impossibility of writing *robust*, mission critical bash and the number of existing bugs in the current scripts - init-assets.sh, tectonic.sh and bootkube.sh.  In so many words, if bash were capable of doing this, I suspect systemd wouldn't exist.

Totally agreed as described above.

modules/aws/ignition/ignition.tf
kans: Thanks for the link.

sym3tri:
Note: this is to fix https://github.com/coreos/tectonic-installer/issues/924

alexsomesan:
Approved by mistake. Was trying to read through the issue.
Please disregard.

kans:
@s-urbaniak @alexsomesan 

Thanks for the links and backstory.  

Simple services and `systemctl is-active` seems like the way to go.  I'm going to close this PR and open a new one.

sstarcher:
I looked for other tickets, but it looks like this has not been resolved yet.

another failure example in init-assets

```Sep 15 21:44:14 localhost systemd[1]: Starting Download Assets...
Sep 15 21:44:21 ip-10-0-2-235 bash[758]: pubkey: prefix: "quay.io/coreos/awscli"
Sep 15 21:44:21 ip-10-0-2-235 bash[758]: key: "https://quay.io/aci-signing-key"
Sep 15 21:44:21 ip-10-0-2-235 bash[758]: gpg key fingerprint is: BFF3 13CD AA56 0B16 A898  7B8F 72AB F5F6 799D 33BC
Sep 15 21:44:21 ip-10-0-2-235 bash[758]:         Quay.io ACI Converter (ACI conversion signing key) <support@quay.io>
Sep 15 21:44:21 ip-10-0-2-235 bash[758]: Trusting "https://quay.io/aci-signing-key" for prefix "quay.io/coreos/awscli" without fingerprint review.
Sep 15 21:44:21 ip-10-0-2-235 bash[758]: Added key for prefix "quay.io/coreos/awscli" at "/etc/rkt/trustedkeys/prefix.d/quay.io/coreos/awscli/bff313cdaa560b16a8987b8f72abf5f679
Sep 15 21:44:21 ip-10-0-2-235 bash[758]: Downloading signature:  0 B/473 B
Sep 15 21:44:21 ip-10-0-2-235 bash[758]: Downloading signature:  473 B/473 B
Sep 15 21:44:21 ip-10-0-2-235 bash[758]: Downloading signature:  473 B/473 B
Sep 15 21:44:21 ip-10-0-2-235 bash[758]: Downloading ACI:  0 B/32 MB
Sep 15 21:44:21 ip-10-0-2-235 bash[758]: Downloading ACI:  16.3 KB/32 MB
Sep 15 21:44:22 ip-10-0-2-235 bash[758]: Downloading ACI:  32 MB/32 MB
Sep 15 21:44:22 ip-10-0-2-235 bash[758]: image: signature verified:
Sep 15 21:44:22 ip-10-0-2-235 bash[758]:   Quay.io ACI Converter (ACI conversion signing key) <support@quay.io>
Sep 15 21:44:29 ip-10-0-2-235 bash[758]: Error: Mounting "/etc/rkt-resolv.conf" on "/opt/stage2/awscli/rootfs//etc/resolv.conf" failed: No such file or directory
Sep 15 21:44:29 ip-10-0-2-235 bash[758]: cat: /run/metadata/master: No such file or directory
Sep 15 21:44:30 ip-10-0-2-235 systemd[1]: init-assets.service: Main process exited, code=exited, status=1/FAILURE
Sep 15 21:44:30 ip-10-0-2-235 systemd[1]: Failed to start Download Assets.
Sep 15 21:44:30 ip-10-0-2-235 systemd[1]: init-assets.service: Unit entered failed state.
Sep 15 21:44:30 ip-10-0-2-235 systemd[1]: init-assets.service: Failed with result 'exit-code'.```

PR #940, opened 2017-05-31
Build docker image

### LINKED ISSUE #689: Provide Docker image of current `master` of installer
I am using the Tectonic installer in our CI pipeline of the Tectonic Prometheus operator and (in the future) the Prometheus Operator. In order to do that I have to compile the installer and its front-end first which I am doing with the `quay.io/coreos/tectonic-builder:v1.7` image.

We are already compiling the installer in the CI pipeline of the Tectonic installer. I would propose building a Docker image with terraform and the compiled binaries as a step of the pipeline so anyone can just go ahead and start her cluster without any dependencies by using the `quay.io/coreos/tectonic-installer-dev:master` image,

What do you think? If you aggree I would go ahead and create a PR. (I would replace the current top level `Dockerfile` if it is not needed anymore.)

/CC @s-urbaniak we already dicussed this. I am creating this in case somebody is already working on it or has a better idea.

mxinden:
Add stage to build docker image containing compiled binaries.

> We are already compiling the installer in the CI pipeline of the
Tectonic installer. I would propose building a Docker image with
terraform and the compiled binaries as a step of the pipeline so anyone
can just go ahead and start her cluster without any dependencies by
using the quay.io/coreos/tectonic-installer-dev:master image.

In addition, this specifies a Docker image per stage, and not per entire pipeline.

Fixes #689

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -16,10 +16,7 @@ def creds = [
 
 pipeline {
   agent {
-    docker {
-      image 'quay.io/coreos/tectonic-builder:v1.12'
-      label 'worker'
-    }
+    label 'worker'
   }
 
   options {
@@ -36,14 +33,24 @@ pipeline {
 
   stages {
     stage('TerraForm: Syntax Check') {
+      agent {
+        docker {
+          image 'quay.io/coreos/tectonic-builder:v1.12'
+        }
+      }
       steps {
         sh """#!/bin/bash -ex
-        make structure-check
+          make structure-check
         """
       }
     }
 
     stage('Generate docs') {
+      agent {
+        docker {
+          image 'golang:1.8'
+        }
+      }
       steps {
         sh """#!/bin/bash -ex
 
@@ -59,6 +66,11 @@ pipeline {
     }
 
     stage('Generate examples') {
+      agent {
+        docker {
+          image 'golang:1.8'
+        }
+      }
       steps {
         sh """#!/bin/bash -ex
 
@@ -74,6 +86,11 @@ pipeline {
     }
 
     stage('Installer: Build & Test') {
+      agent {
+        docker {
+          image 'quay.io/coreos/tectonic-builder:v1.12'
+        }
+      }
       steps {
         checkout scm
         sh "mkdir -p \$(dirname $GO_PROJECT) && ln -sf $WORKSPACE $GO_PROJECT"
@@ -100,6 +117,11 @@ pipeline {
     }
 
     stage("Smoke Tests") {
+      agent {
+        docker {
+          image 'quay.io/coreos/tectonic-builder:v1.12'
+        }
+      }
       steps {
         parallel (
           "TerraForm: AWS": {
@@ -142,7 +164,32 @@ pipeline {
         }
       }
     }
+
+    stage('Build docker image')  {
+      when {
+        branch 'master'
+      }
+      steps {
+        unstash 'installer'
+        unstash 'sanity'
+        withCredentials([
+            usernamePassword(
+              credentialsId: 'quay-robot',
+              passwordVariable: 'QUAY_ROBOT_SECRET',
+              usernameVariable: 'QUAY_ROBOT_USERNAME'
+            )
+          ]) {
+          sh """
+            docker build -t quay.io/coreos/tectonic-installer:master -f images/tectonic-installer/Dockerfile .
+            docker login -u="$QUAY_ROBOT_USERNAME" -p="$QUAY_ROBOT_SECRET" quay.io
+            docker push quay.io/coreos/tectonic-installer:master
+            docker logout quay.io
+          """
+        }
+      }
+    }
   }
+
   post {
     always {
       // Cleanup workspace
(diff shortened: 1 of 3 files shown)


test/docker-client-container/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM golang:1.7.5-wheezy
```
sym3tri: Why not use a newer version of go? 1.8.3 just shipped.

Jenkinsfile
```diff
@@ -79,7 +96,44 @@ pipeline {
...
+                dockerfile {
+                    dir 'test/docker-client-container'
+                    args '-v /var/run/docker.sock:/var/run/docker.sock'
+                }
+            }
+            /*
+            when {
+                branch 'master'
+            }
+            */
```
sym3tri: ?

test/docker-client-container/Dockerfile
mxinden: Yes, thanks for the hint.

Jenkinsfile
mxinden: This is still a *work in progress*. I had to create the PR in order to be able to run this on our Jenkins. Sorry for the noise.

This is commented out for now to run it on this `mxinden:build-docker-image` branch as well.

mxinden:
Status: Will rebase to changes by #948

kans:
Do you know how much time does this add to our builds?

mxinden:
The last successful build with the *Build docker image* step was build number [25](https://jenkins-tectonic-installer.prod.coreos.systems/job/coreos%20-%20tectonic-installer/job/tectonic-installer/job/PR-940/25/console). Any future test runs will ignore this step, as it will only run on master (See [condition](https://github.com/coreos/tectonic-installer/pull/940/files#diff-58231b16fdee45a03a4ee3cf94a9f2c3R113)).

The *Build docker image* step took `4min 28s` in total in build number 25. The longest substep was `RUN go get github.com/hashicorp/terraform` which took from `11:58:37` to `12:00:58`.

These four minutes are only added when running on `master`. On any other branch (e.g. PRs) it adds `40ms`.

@kans Is this extra time OK?

mxinden:
@kans @sym3tri Rebased again. Any further changes wanted?

mxinden:
@alexsomesan I have updated the *Tectonic-installer-Dockerfile* as you suggested to download the release of *Terraform* instead of building it itself. @kans This will speed up the build step as well.

@kans I have moved the *Build docker image* step to the very end of the pipeline. Now the Docker image only gets build when all previous steps pass, so the pipeline fails faster.

Jenkinsfile
```diff
@@ -36,14 +33,24 @@ pipeline {
...
-        make structure-check
+          make structure-check
         """
       }
     }
 
     stage('Generate docs') {
+      agent {
+        docker {
+          image 'golang:1.8'
```
sym3tri: why use a different image here? You can just reuse the builder image, it is based off of `golang`

Jenkinsfile
```diff
@@ -100,6 +117,11 @@ pipeline {
     }
 
     stage("Smoke Tests") {
+      agent {
+        docker {
+          image 'quay.io/coreos/tectonic-builder:v1.12'
```
sym3tri: This is repeated in a lot of places. Is there any way to make this a variable that can be reused?

images/tectonic-installer/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM golang:1.8-stretch
+
+ENV TERRAFORM_VERSION="0.9.6"
+
+RUN apt-get update \
+    && apt-get install --no-install-recommends -y -q \
+    unzip
+
+# Install Terraform
+RUN curl https://releases.hashicorp.com/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_linux_amd64.zip | funzip > /usr/local/bin/terraform && chmod +x /usr/local/bin/terraform
```
sym3tri: Would be nice if we could reuse the logic in `make release` https://github.com/coreos/tectonic-installer/blob/master/installer/Makefile#L77 instead of repeating it all here.

Jenkinsfile
Quentin-M: Yeah, I am not sure of this either.

images/tectonic-installer/Dockerfile
Quentin-M: When I commented on the issue initially, I pretty much wanted to use the release scripts as well, as much as possible (so it can be used for nightly builds).

Jenkinsfile
mxinden: @sym3tri @Quentin-M I thought of running each step in its own Docker image with the least amount of dependencies in it as possible, instead of running each step with the same Docker image, where dependencies are getting a little messy. I find this cleaner and easier to reason about. I am fine with changing this back to using the `tectonic-builder` image. Whatever you prefer?

Jenkinsfile
mxinden: We could probably move the `tectonic-builder` image url to a variable?

images/tectonic-installer/Dockerfile
mxinden: Sounds like a good idea. If you don't mind I would just open a new issue for that, so we can fix it in a separate PR. What do you think?

PR #946, opened 2017-05-31
modules/aws removes most timeouts ...

kans:
when installing tectonic components.

### DIFF
--- a/modules/tectonic/resources/tectonic.sh
+++ b/modules/tectonic/resources/tectonic.sh
@@ -18,7 +18,7 @@ KUBECTL="/kubectl --kubeconfig=$KUBECONFIG"
 function kubectl() {
   local i=0
 
-  echo "Executing kubectl $@"
+  echo "Executing kubectl" "$@"
   while true; do
     (( i++ )) && (( i == 100 )) && echo "kubectl failed, giving up" && exit 1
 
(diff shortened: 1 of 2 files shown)


modules/tectonic/resources/tectonic.sh
```diff
@@ -47,9 +47,8 @@ function wait_for_tpr() {
 
   echo "Waiting for TPR $2"
   until $KUBECTL -n "$1" get thirdpartyresources "$2"; do
-    (( i++ )) && (( i == 100 )) && echo "TPR $2 not available, giving up" && exit 1
-
-    echo "TPR $2 not available yet, retrying in 5 seconds"
+    (( i++ ))
```
s-urbaniak: This will now potentially loop forever. I have doubts this is really what we want to achieve. Why not bumping `i` to a higher value?

ggreer: If this script exits early, the cluster won't come up. So in this case, waiting forever seems like a reasonable thing to do. It allows for users on slower connections to bring up clusters.

Quentin-M: We could move the slow / flaky / less critical resources to the bottom. Thus, we'd create most of the cluster regardless. Then we can wait forever or what, it'd matter less.

ggreer: Order of resource creation seems orthogonal to this PR. There is no case in which waiting indefinitely would cause cluster bringup to fail that wouldn't also cause the old code to fail. In short: this change strictly dominates.

kans:
> Why not bumping i to a higher value?

That was my original thought too, but counting goes 1, 2, infinity in CS.  Alternatively, what is the case that we want this script to stop?  As I just ran into this failure mode yet again, I going to merge this PR and If someone has a compelling reason to change it from infinity to a smaller number in the future, please open another PR.

PR #947, opened 2017-05-31
Make detect-master.sh more robust.

ggreer:
- Retry fetching region & instance ID.
- Avoid false positive comparisons that could potentially cause a node to think it's the first master when it isn't.
- Wrap some fields in quotes to prevent potential issues with globbing.

### DIFF
--- a/modules/aws/ignition/resources/detect-master.sh
+++ b/modules/aws/ignition/resources/detect-master.sh
@@ -1,14 +1,18 @@
 #!/bin/bash
 
-REGION=$(wget -q -O - http://169.254.169.254/latest/meta-data/placement/availability-zone | sed s'/[a-zA-Z]$//')
-INSTANCE_ID=$(wget -qO- http://169.254.169.254/latest/meta-data/instance-id)
-ASG_NAME=$(aws autoscaling describe-auto-scaling-instances --region="$REGION" --instance-ids="$INSTANCE_ID" | jq -r ".AutoScalingInstances[0] .AutoScalingGroupName")
-
 # Wait for the ASG to run at the expected scale.
 while true; do
+  REGION=$(wget -q -O - http://169.254.169.254/latest/meta-data/placement/availability-zone | sed s'/[a-zA-Z]$//')
+  INSTANCE_ID=$(wget -q -O - http://169.254.169.254/latest/meta-data/instance-id)
+  ASG_NAME=$(aws autoscaling describe-auto-scaling-instances --region="$REGION" --instance-ids="$INSTANCE_ID" | jq -r ".AutoScalingInstances[0] .AutoScalingGroupName")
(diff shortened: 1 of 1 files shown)


modules/aws/ignition/resources/detect-master.sh
```diff
@@ -1,14 +1,17 @@
...
 while true; do
+  REGION=$(wget -q -O - http://169.254.169.254/latest/meta-data/placement/availability-zone | sed s'/[a-zA-Z]$//')
+  INSTANCE_ID=$(wget -q -O - http://169.254.169.254/latest/meta-data/instance-id)
+  ASG_NAME=$(aws autoscaling describe-auto-scaling-instances --region="$REGION" --instance-ids="$INSTANCE_ID" | jq -r ".AutoScalingInstances[0] .AutoScalingGroupName")
   ASG_DESCRIPTION=$(aws autoscaling describe-auto-scaling-groups --region="$REGION" --auto-scaling-group-names="$ASG_NAME")
-  ASG_DESIRED_CAP=$(echo $ASG_DESCRIPTION | jq ".AutoScalingGroups[0] .DesiredCapacity")
-  ASG_INSTANCE_IDS=$(echo $ASG_DESCRIPTION | jq -r ".AutoScalingGroups[0] .Instances | sort_by(.InstanceId) | .[].InstanceId")
+  # shellcheck disable=SC2181
+  if [ $? -ne 0 ]; then
+    continue
```
s-urbaniak: this could end up in an endless loop. I suggest to include a max-retry break, maybe similar to https://github.com/coreos/tectonic-installer/blob/2bede2c/modules/tectonic/resources/tectonic.sh#L23?

s-urbaniak: Also, does it make sense to sleep before the continue?

alexsomesan: Neat improvement!
But +1 on both the max-retry and sleep.
It is dangerous without a way out of the loop.

ggreer: An endless loop is exactly the point. If detect-master fails, it's never restarted (because init-assets service is a oneshot). No node will think it's the first master and the cluster will never come up.

alexsomesan: But is there a point in hammering the AWS API and getting that account blacklisted with throttling if the failure can't be recovered from within N retries?

kans: This amounts to 1/4 req/s.

ggreer:
Added a sleep before the continue. Nice catch.

modules/aws/ignition/resources/detect-master.sh
ggreer: I think Alex made that comment before I added the sleep to the continue logic.

PR #953, opened 2017-06-01
Calculate root_block_device.iops for gp2 volumes

bsiegel:
This fixes an idempotency issue where even though the `iops` parameter is not used when the volume type is anything other than `io1`, Terraform will report a diff forcing recreation of the instance(s) because the reported `iops` parameter does not match the value specified in the module.

The calculation comes from the AWS docs at https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSVolumeTypes.html, specifically:

>Between a minimum of 100 IOPS (at 33.33 GiB and below) and a maximum of 10,000 IOPS (at 3,334 GiB and above), baseline performance scales linearly at 3 IOPS per GiB of volume size.

Related to / supersedes #896

### DIFF
--- a/modules/aws/etcd/nodes.tf
+++ b/modules/aws/etcd/nodes.tf
@@ -48,7 +48,7 @@ resource "aws_instance" "etcd_node" {
   root_block_device {
     volume_type = "${var.root_volume_type}"
     volume_size = "${var.root_volume_size}"
-    iops        = "${var.root_volume_type == "io1" ? var.root_volume_iops : 100}"
+    iops        = "${var.root_volume_type == "io1" ? var.root_volume_iops : var.root_volume_type == "gp2" ? min(10000, max(100, 3 * var.root_volume_size)) : 0}"
   }
 
   volume_tags = "${merge(map(


bsiegel:
I verified that with an EBS volume of type `standard`, Terraform appears to report `0` for the value of the `iops` parameter, so this should solve the problem for all volume types currently supported by Terraform.

ggreer:
ok to test

bsiegel:
Updated this PR to only apply this change to `aws_instance` resources. `aws_launch_configuration` resources appear to omit the `iops` value when returned from the AWS API unless the volume type is `io1`, so this will continue to appear as `0` to Terraform.

cpanato:
@bsiegel thanks again for this PR, can you please rebase?

squat:
This seems to be getting at the same problem as https://github.com/coreos/tectonic-installer/pull/1826 and will cover more cases.

bsiegel:
@cpanato @squat rebased!

squat:
closing and re-opening to trigger tests.

squat:
Everything looks good. There was an unrelated flake in one of the Azure tests.

squat:
@bsiegel can you please rebase off of master? We just merged a PR that should cover the other half of the volume-based resource churn.

bsiegel:
@squat rebased!

squat:
@bsiegel everything worked except for one unrelated Azure smoke test.

cpanato:
Tests are green. @alexsomesan do you have any comments on this PR? Otherwise, I will merge it

PR #956, opened 2017-06-02
frontend: Move cluster-type.jsx to new Connect/Form pattern.

ggreer:
- Add support for optional errorComponent in forms to customize rendering of errors.
- Change cluster selection page to use form & Connect.

Doesn't change behavior in any way.

### DIFF
--- a/installer/frontend/__tests__/protocols.js
+++ b/installer/frontend/__tests__/protocols.js
@@ -14,6 +14,7 @@ import '../components/etcd';
 import '../components/bm-sshkeys';
 import '../components/bm-nodeforms';
 import '../components/bm-hostname';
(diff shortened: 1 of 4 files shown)


installer/frontend/form.js
```diff
@@ -296,8 +296,9 @@ export class Form extends Node {
     this.errorComponent = connect(
       ({clusterConfig}) => ({
         error: _.get(clusterConfig, toError(id)) || _.get(clusterConfig, toAsyncError(id)),
+        clusterConfig,
```
ggreer: I don't like this, but I want to get at clusterConfig data in error rendering. Other ideas welcome.

Issue #959, opened 2017-06-02
etcd-operator with backup fails to start with permission denied to mkdir /.aws


BdeUtra:
## Versions

* Tectonic version ([release](https://github.com/coreos/tectonic-installer/releases) or commit hash): https://github.com/coreos/tectonic-installer/releases/tag/1.6.4-tectonic.1-rc.3
* Terraform version (`terraform version`): Terraform v0.9.6-dev (fcdf494cffc06aac027f6cf904647051ab6d5015+CHANGES)
* Platform (aws|azure|openstack|metal): AWS

## What happened?
When backup args are passed in to etcd-operator it fails to start with 

`level=fatal msg=controller Run() ended with failure: mkdir /.aws: permission denied`

I suspect this is because it no longer runs as root (see https://github.com/coreos/tectonic-installer/commit/9dd1657a025642cba0b07ac4e495b607817ff03c) and therefore it fails here https://github.com/coreos/etcd-operator/blob/master/pkg/controller/s3.go#L43

## What you expected to happen?
Etcd-operator to start normally.

## How to reproduce it (as minimally and precisely as possible)?

1. Edit `modules/bootkube/resources/experimental/manifests/etcd-operator.yaml` to include backup-aws-secret and backup-aws-config and the operator backup args:
```yaml
apiVersion: v1
data:
  credentials: ....
kind: Secret
metadata:
  name: etcd-backup-aws
  namespace: kube-system
type: Opaque

---

apiVersion: v1
data:
  config: |
    [default]
    region = eu-west-1
kind: ConfigMap
metadata:
  name: aws-region
  namespace: kube-system

---

apiVersion: extensions/v1beta1
kind: Deployment
metadata: 
  labels: 
    k8s-app: etcd-operator
  name: etcd-operator
  namespace: kube-system
spec: 
  replicas: 1
  template: 
    metadata: 
      labels: 
        k8s-app: etcd-operator
    spec: 
      containers: 
        - env: 
            - name: MY_POD_NAMESPACE
              valueFrom: 
                fieldRef: 
                  fieldPath: metadata.namespace
            - name: MY_POD_NAME
              valueFrom: 
                fieldRef: 
                  fieldPath: metadata.name
          image: ${etcd_operator_image}
          name: etcd-operator
          command:
          - /usr/local/bin/etcd-operator
          - --backup-aws-secret=etcd-backup-aws
          - --backup-aws-config=aws-region
[...]

PR #962, opened 2017-06-02
modules/bootkube: fix etcd-operator startup with S3 backups and nobody

Quentin-M:
Fixes https://github.com/coreos/tectonic-installer/issues/959.
See issue for details.

### DIFF
--- a/modules/bootkube/resources/experimental/manifests/etcd-operator.yaml
+++ b/modules/bootkube/resources/experimental/manifests/etcd-operator.yaml
@@ -22,6 +22,8 @@ spec:
               valueFrom: 
                 fieldRef: 
                   fieldPath: metadata.name
+            - name: HOME
+              value: /tmp
           image: ${etcd_operator_image}
           name: etcd-operator
       nodeSelector:


squat:
@Quentin-M Is this ported from upstream?

Quentin-M:
@squat Thanks for checking! The [example](https://github.com/coreos/etcd-operator/blob/master/example/deployment.yaml) manifest upstream runs as root, so this does not apply.

kans:
@diegs should this be backported?

kans:
@Quentin-M any idea if this problem exists elsewhere too?

diegs:
Sorry, I was unaware of this behavior and it appears to not have test coverage.

Looks like this should be backported, or added to known issues with the workaround (add this manifest change between `make plan` and `make apply`).

modules/bootkube/resources/experimental/manifests/etcd-operator.yaml
```diff
@@ -22,6 +22,8 @@ spec:
               valueFrom: 
                 fieldRef: 
                   fieldPath: metadata.name
+            - name: HOME
```
diegs: I don't love that the etcd-operator writes to $HOME/.aws, but if that's an unavoidable side-effect of using AWS then so be it.

sstarcher:
looks like this issue is back in newer versions of etcd operator I'm running 0.5.2 and it's trying to use
```/var/tmp/etcd-operator/aws```
where aws is the credential.

```
time="2017-09-13T21:57:32Z" level=info msg="etcd-operator Version: 0.5.2"
time="2017-09-13T21:57:32Z" level=info msg="Git SHA: 0fa2f7d"
time="2017-09-13T21:57:32Z" level=info msg="Go Version: go1.8.1"
time="2017-09-13T21:57:32Z" level=info msg="Go OS/Arch: linux/amd64"
time="2017-09-13T21:57:32Z" level=info msg="Event(v1.ObjectReference{Kind:"Endpoints", Namespace:"kube-system", Name:"etcd-operator", UID:"8b2da31f-98ce-11e7-bbf2-0aa4c304ed3c", APIVersion:"v1", ResourceVersion:"1545", FieldPath:""}): type: 'Normal' reason: 'LeaderElection' etcd-operator-525807187-xghfx became leader"
2017-09-13 21:57:52.494409 I | warning: ignoring ServerName for user-provided CA for backwards compatibility is deprecated
time="2017-09-13T21:57:52Z" level=error msg="cluster failed to setup: new S3 storage failed: mkdir /var/tmp/etcd-operator/aws: permission denied" cluster-name=kube-etcd pkg=cluster
time="2017-09-13T21:57:52Z" level=warning msg="fail to handle event: ignore failed cluster (kube-etcd). Please delete its CR" pkg=controller
error: unexpected EOF```

Issue #699, opened 2017-05-15
GUI - nav bar updates


meghanschofield:
- Now with a CoreOS topper and a few bits from the nav. 
- Tectonic docs link is included in the nav (this is different from our marketing site nav. I'm including two elements that match but have revealed the specific link for Tectonic Docs that'll take users to the main installer docs page!) 
- The Installer Version is revealed in the nav bar, top right (outlined here: https://github.com/coreos/tectonic-installer/issues/364)
- Data collection note lives at the bottom, on all pages. 
  -- 20px from the bottom of the window
  -- text color #d9d9d9

![installation type_destroy cluster](https://cloud.githubusercontent.com/assets/3373315/26081740/28a24c9a-3981-11e7-87a3-471efc431e22.png)

How does this look @robszumski ?

amrutac:
@meghanschofield: Can you please provide more details about the `Products` and `Open Source` dropdowns?

meghanschofield:
Yup! Those are the same nav items, with the same drop downs as https://coreos.com/

<img width="403" alt="screen shot 2017-05-26 at 5 10 35 pm" src="https://cloud.githubusercontent.com/assets/3373315/26516115/4732611e-4236-11e7-8460-4ea2436ea6f8.png">
<img width="502" alt="screen shot 2017-05-26 at 5 10 40 pm" src="https://cloud.githubusercontent.com/assets/3373315/26516116/483e60d0-4236-11e7-80d7-0aae8f8fe5a7.png">

Does that help?

Issue #364, opened 2017-04-25
GUI: Surface all relevant version information


sym3tri:
It would be nice if the GUI installer surfaced the following info:

- What version you are currently using
- Whether or not a new version is available
- What version of Terraform the installer is compatible with
- All links go to specific versioned docs (i.e. not `docs/latest`)

meghanschofield:
I'm on it. Seems like stuff for the first page. A few questions:

1. Why does this help users? is it really for repeat users or first time?
2. IF there is a new GUI installer version available, do we link to that from older ones? 
3. Why would they ever want to use an old one? How would they get to the old one?
4. For links to the versioned docs. I assume that's just for the options users need, but do we need to give users a way to easily jump to other versioned docs? (maybe for comparison?)

I always worry about optimizing for edge cases, b/c it gives the other 99% of users too much to contemplate when it's not needed. So, is this an interim thing? or a forever thing, b/c of how Terraform works?

Thanks @sym3tri !

robszumski:
I'll take a swing at some of these:

> What version you are currently using
> Why does this help users? is it really for repeat users or first time?
> Whether or not a new version is available

**Supportability:** when encountering issues with the installer, this is the first thing to help everyone get on the same page to triage
**Repeat usage:** if you downloaded a while ago and want to see what has changed since this (installation-wise) you can match this up to release notes
**First time:** someone may have signed up for your company and sent you an out of date installer. Mostly related to below:

> If there is a new GUI installer version available, do we link to that from older ones?

This should always be safe, as we're always fixing bugs, making it more stable, and adding new platforms.

> Why would they ever want to use an old one? How would they get to the old one?

If they want to set up a cluster exactly as they did a while ago.  They would need to swap out the version number in the link: https://releases.tectonic.com/tectonic-1.5.6-tectonic.1.tar.gz aka we don't make this super easy on purpose. We'd rather you run the latest.

> What version of Terraform the installer is compatible with

For this, we should start listing it with the release notes. And then link to them from wherever we display the current version. Something like `Version 1.5.6-tectonic.1 (release notes)` or similar.

> All links go to specific versioned docs (i.e. not docs/latest)
> For links to the versioned docs. I assume that's just for the options users need, but do we need to give users a way to easily jump to other versioned docs? (maybe for comparison?)

This is mostly an intangible UX thing that the engineering team will swap out. At the top of each doc there is a dropdown to swap to an older version. Take a look to see if you think this meets needs.

[...]

PR #963, opened 2017-06-02
frontend: Adds header to the installer

amrutac:
Implements https://github.com/coreos/tectonic-installer/issues/699 and https://github.com/coreos/tectonic-installer/issues/364

Uses a fork(https://github.com/coreos-inc/babel-plugin-git-version) of https://github.com/renke/babel-plugin-git-version to get the nearest and the latest tag.

screen shot 2017-06-02 at 1 24 34 pm

screen shot 2017-06-02 at 1 24 51 pm

screen shot 2017-06-02 at 1 24 59 pm


amrutac:
ok to test

ggreer:
The yarn.lock file should have changes. Looks like our tests don't catch that.

Quentin-M:
We could consider dropping the `version` command of the binary installer if we have that?

amrutac:
@geoff: Good catch, added <code>yarn.lock</code>

installer/frontend/components/header.jsx
```diff
@@ -0,0 +1,62 @@
+import React from 'react';
+import { Dropdown } from './ui';
+
+export class Header extends React.Component {
+  render() {
+    const hasNewVersion = GIT_TAG !== GIT_LATEST_TAG;
```
ggreer: Won't this only be true if one builds a release of an old tag? eg: If I build v1.6.4 today, `GIT_TAG` and `GIT_LATEST_TAG` will both be v1.6.4 in the generated js.

Is one of the requirements that it checks for new versions when run?

amrutac: Ah, yes. You are right, I tested this locally by rebasing off of older tag, but this will not work for generated js. Probably need an external API to provide the latest version then?

ggreer: Yeah. I don't think that should block this PR.

amrutac: How do you feel about scraping the latest version from the http://coreos.com/tectonic/releases/?

ggreer: There's probably some programmatic way to get it from github. eg, a machine-readable version of https://github.com/coreos/tectonic-installer/releases

Issue #893, opened 2017-05-26
kube-apiserver constantly logs TLS handshake errors


ggreer:
Even when my tectonic console comes up successfully, I see constant TLS handshake errors in kube-apiserver. Here's the output of `kubectl --kubeconfig=generated/auth/kubeconfig logs kube-apiserver-xmqwb --namespace=kube-system -f`: 

https://gist.github.com/ggreer/42a7cc780bf52a2ec7dcb0cfa1569338

And here's the output of `kubectl --kubeconfig=generated/auth/kubeconfig get pods --all-namespaces`:
https://gist.github.com/ggreer/2a4444ff187bd79acdb061a064c2092d

I've since deleted the cluster those logs came from, but I can reproduce this 100% of the time. I have no idea if it's "normal" or bad.

philips:
this is "normal" and due to ELB health checking :( :( :(

Quentin-M:
Addressed by https://github.com/coreos/tectonic-installer/pull/964.

PR #964, opened 2017-06-02
modules/aws/master-asg: switch ELBs’ health checks to using SSL targets

Quentin-M:
This avoids having the API servers’ logs being hammered with:

```
http: TLS handshake error from 10.0.44.6:28942: EOF
```

Fixes https://github.com/coreos/tectonic-installer/issues/893.


kans:
thanks for tracking this down!

s-urbaniak:
Post-merge note: good catch! :+1:

PR #967, opened 2017-06-02
git-version: always use commit as part of the version

Quentin-M:
This is a very small improvement to what we have. Instead of putting the build artifacts with the tag if available, we will always include the commit as well, thus hinting if someone weird happened..

### DIFF
--- a/git-version
+++ b/git-version
@@ -3,16 +3,13 @@
 set -e
 
 # pull the current git commit hash
-COMMIT=$(git rev-parse HEAD)
+VERSION=$(git rev-parse HEAD)
 
-# check if the current commit has a matching tag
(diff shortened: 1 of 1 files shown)


bison:
This seems reasonable to me. This script is used outside of the upload process though. Not sure if this should just be specific to the use in: `upload_installer_bins.sh`

Quentin-M:
We don't actually really use that script but to embed the version in the binary, and upload the artifacts. I want both.

ggreer:
I think this will cause some silliness in the GUI once #963 is merged.

Quentin-M:
Oh, interesting! I don't think they are related tho, as it is not using the `git-version` script, which output is injected by the linker into the main package of the binary.

ggreer:
Yeah, you're right. There shouldn't be an issue.

PR #968, opened 2017-06-02
tests: add a script to implement kubernetes conformance tests.

rithujohn191:
This script will be implemented by a Jenkins job periodically every night.

### DIFF
--- a/tests/conformance/conformance.sh
+++ b/tests/conformance/conformance.sh
@@ -1,0 +1,119 @@
+#!/bin/bash
+set -exo pipefail
+
+PROJECT=/go/src/github.com/coreos/tectonic-installer
+
+export PLATFORM=aws
+export CLUSTER="tf-${PLATFORM}-${BUILD_ID}"
+export TF_VAR_tectonic_pull_secret_path=${TF_VAR_tectonic_pull_secret_path}
+export TF_VAR_tectonic_license_path=${TF_VAR_tectonic_license_path}
+export TECTONIC_BUILDER=quay.io/coreos/tectonic-builder:v1.18
+export KUBE_CONFORMANCE=quay.io/coreos/kube-conformance:v1.6.2_coreos.0
+
+# Create an env var file
+cat <<EOF >> env.list
+PLATFORM=aws
+CLUSTER="tf-${PLATFORM}-${BUILD_ID}"
+TF_VAR_tectonic_cluster_name=$(echo "${CLUSTER}" | awk '{print tolower($0)}')
+AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID}
+AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY}
+TF_VAR_tectonic_aws_region="us-west-2"
+TF_VAR_tectonic_pull_secret_path=${TF_VAR_tectonic_pull_secret_path}
+TF_VAR_tectonic_license_path=${TF_VAR_tectonic_license_path}
+TF_VAR_tectonic_aws_ssh_key="jenkins"
+EOF
+
+# Mount secret command for docker run
+MNT_SECRETS="-v ${TF_VAR_tectonic_license_path}:${TF_VAR_tectonic_license_path} -v ${TF_VAR_tectonic_pull_secret_path}:${TF_VAR_tectonic_pull_secret_path}"
+
+function cleanup() {
+    docker run --env-file ./env.list -i -v "${WORKSPACE}":"${PROJECT}" "${MNT_SECRETS}" "${TECTONIC_BUILDER}" /bin/bash <<EOF
+cd ${PROJECT}
+make destroy
+EOF
+}
+# Trap to ensure cluster clean up before exit
+trap cleanup EXIT
+
+export KUBECTL=${WORKSPACE}/kubectl
+curl -o "${KUBECTL}" https://storage.googleapis.com/kubernetes-release/release/v1.6.3/bin/linux/amd64/kubectl && chmod +x "${KUBECTL}"
+
+function kubectl() {
+    local i=0
+
+    echo "Executing kubectl" "$@"
+    while true; do
+        (( i++ )) && (( i == 100 )) && echo "kubectl failed, giving up" && exit 1
+
+        set +e
+        out=$($KUBECTL "$@" 2>&1)
+        status=$?
+        set -e
+  
+        if [[ "$out" == *"AlreadyExists"* ]]; then
+            echo "$out, skipping"
+            return
+        fi
+  
+        echo "$out"
+        if [ "$status" -eq 0 ]; then
+            return
+        fi
+  
+        echo "kubectl failed, retrying in 5 seconds"
+        sleep 5
+    done
+}
+ 
+function wait_for_pods() {
+    echo "Waiting for pods in namespace $1"
+    while true; do
+  
+        set +e
+        out=$($KUBECTL -n "$1" get po -o custom-columns=STATUS:.status.phase,NAME:.metadata.name)
+        set -e
+        status=$?
+        echo "$out"
+  
+        if [ "$status" -ne "0" ]; then
+           echo "kubectl command failed, retrying in 5 seconds"
+           sleep 5
+           continue
+        fi
+  
+        # make sure kubectl does not return "no resources found"
+        if [ "$(echo "$out" | tail -n +2 | grep -c '^')" -eq 0 ]; then
+           echo "no nodes were found, retrying in 5 seconds"
+           sleep 5
+           continue
+        fi
+  
+        stat=$( echo "$out"| tail -n +2 | grep -v '^Running')
+        if [[ "$stat" == "" ]]; then
+            return
+        fi
+  
+        echo "Pods not available yet, waiting for 5 seconds"
+        sleep 5
+    done
+}
+
+docker run --env-file ./env.list -i -v $"{WORKSPACE}":"${PROJECT}" "${MNT_SECRETS}" -w "${PROJECT}" "${TECTONIC_BUILDER}" /bin/bash <<EOF
+  
+mkdir -p ${PROJECT}/build/${CLUSTER}/
+ln -sf ${PROJECT}/tests/smoke/aws/vars/aws.tfvars ${PROJECT}/build/${CLUSTER}/terraform.tfvars
+  
+make plan
+make apply
+EOF
+  
+export KUBECONFIG=${WORKSPACE}/build/${CLUSTER}/generated/auth/kubeconfig
+  
+wait_for_pods kube-system
+echo "API Server UP!"
+  
+kubectl get pods --all-namespaces
+kubectl get nodes
+  
+docker pull "${KUBE_CONFORMANCE}"
+docker run -v "${KUBECONFIG}":/kubeconfig "${KUBE_CONFORMANCE}"


tests/conformance/conformance.sh
```diff
@@ -0,0 +1,100 @@
...
+  export AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY}
+  export TF_VAR_tectonic_pull_secret_path=${TF_VAR_tectonic_pull_secret_path}
+  export TF_VAR_tectonic_license_path=${TF_VAR_tectonic_license_path}
+  export TF_VAR_tectonic_cluster_name=${TF_VAR_tectonic_cluster_name}
+  
+  make localconfig
+  ln -sf ${PROJECT}/test/aws.tfvars ${PROJECT}/build/${CLUSTER}/terraform.tfvars
+  ls -l ${PROJECT}/build/${CLUSTER}
+  
+  make plan
```
squat: We recently went through a good bit of iteration to get  a lot of this working well in the smoke.sh script. It would make sense instead or recreating all this logic here, to simply call `smoke.sh create ...` and `smoke.sh destroy ...` to create and destroy your cluster respectively.

tests/conformance/conformance.sh
```diff
@@ -0,0 +1,100 @@
+#!/bin/bash
```
rithujohn191: Done

tests/conformance/conformance.sh
rithujohn191: Added a function to handle checking cluster name and randomizing AWS region. Exporting the function from smoke.sh requires a lot more restructuring.

tests/conformance/conformance.sh
```diff
@@ -0,0 +1,133 @@
+#!/bin/bash -ex
```
squat: please also set `-o pipefail` to catch errors that would be hidden by a pipe.

tests/conformance/conformance.sh
```diff
@@ -0,0 +1,133 @@
...
+PROJECT=/go/src/github.com/coreos/tectonic-installer
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+source "$DIR/ ../smoke/aws/smoke.sh"
```
squat: are you using smoke.sh? I thought you re-implemented the region and cluster name generation.

tests/conformance/conformance.sh
```diff
@@ -0,0 +1,133 @@
...
+
+export PLATFORM=aws
+export CLUSTER="tf-${PLATFORM}-${BUILD_ID}"
+export TF_VAR_tectonic_cluster_name=$(echo ${CLUSTER} | awk '{print tolower($0)}')
+
+MNT_SECRETS="-v ${TF_VAR_tectonic_license_path}:${TF_VAR_tectonic_license_path} -v ${TF_VAR_tectonic_pull_secret_path}:${TF_VAR_tectonic_pull_secret_path}"
+
+function cluster_check() {
+	CLUSTER=$1
```
squat: these statements should all be indented equally. It looks like they're indented 1 space character.

tests/conformance/conformance.sh
```diff
@@ -0,0 +1,133 @@
...
+	MAX_LENGTH=28
+
+	LENGTH=${#CLUSTER}
+    	if [ "$LENGTH" -gt "$MAX_LENGTH" ]
```
squat: the indentation should match this `if` statement.

tests/conformance/conformance.sh
```diff
@@ -0,0 +1,133 @@
...
+        	echo "Cluster name too long. Truncated to $CLUSTER"
+    	elif [ "$LENGTH" -lt "$MAX_LENGTH" ]
+    	then
+        	APPEND=$(( MAX_LENGTH - LENGTH ))
+        	APPEND_STR="012345678901234567890123456789"
+        	CLUSTER="$CLUSTER${APPEND_STR:0:APPEND}"
+        	echo "Cluster name too short. Appended to $CLUSTER"
+    	fi
+
+	CLUSTER=$(echo "${CLUSTER}" | awk '{print tolower($0)}')
```
squat: also indent.

tests/conformance/conformance.sh
```diff
@@ -0,0 +1,133 @@
...
+    	export CHANGE_ID=${CHANGE_ID:-${BUILD_ID}}
+    	i=$(( CHANGE_ID % ${#REGIONS[@]} ))
+    	export TF_VAR_tectonic_aws_region="${REGIONS[$i]}"
+    	export AWS_REGION="${REGIONS[$i]}"
+    	echo "selected region: $TF_VAR_tectonic_aws_region"
+    	echo "cluster name: $CLUSTER"
+}
+
+function cleanup() {
+	docker run -i -v ${WORKSPACE}:${PROJECT} ${MNT_SECRETS} quay.io/coreos/tectonic-builder:v1.9 /bin/bash <<EOF
```
squat: this statement should be indented and then the heredoc should be unindented since whitespace is interpreted literally in a heredoc.

tests/conformance/conformance.sh
```diff
@@ -0,0 +1,133 @@
...
+
+
+export KUBECTL=${WORKSPACE}/kubectl
+curl -o ${KUBECTL} https://storage.googleapis.com/kubernetes-release/release/v1.6.3/bin/linux/amd64/kubectl && chmod +x ${KUBECTL}
+function kubectl() {
+  local i=0
+
+  echo "Executing kubectl $@"
+  while true; do
+    (( i++ )) && (( i == 100 )) && echo "kubectl failed, giving up" && exit 1
```
squat: this will work but it is a little fragile. If you happened to move the increment to a different line from the value test, then this would fail on the first try since `(( 0++ ))` returns error code `1`.

tests/conformance/conformance.sh
```diff
@@ -0,0 +1,133 @@
...
+    sleep 5
+  done
+  set -e
+}
+
+kubectl help
+
+docker pull quay.io/coreos/tectonic-builder:v1.9
+docker run -i -v ${WORKSPACE}:${PROJECT} ${MNT_SECRETS} quay.io/coreos/tectonic-builder:v1.9 /bin/bash <<EOF
+  cd ${PROJECT}
```
squat: same thing with this heredoc.

squat: this in-line bash script needs to have `set -exo pipefail` as well since it is executing a bunch of bash and make file invocations.

tests/conformance/conformance.sh
```diff
@@ -0,0 +1,133 @@
...
+  
+  export PLATFORM=${PLATFORM}
+  cluster_check "${CLUSTER}"
+  export AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID}
```
squat: you will need to ensure that these credentials do not leak.

tests/conformance/conformance.sh
```diff
@@ -0,0 +1,133 @@
...
+    echo "kubectl failed, retrying in 5 seconds"
+    sleep 5
+  done
+}
+
+function wait_for_pods() {
+  set +e
+  local i=0
+  echo "Waiting for pods in namespace $1"
+  while $KUBECTL -n "$1" get po -o custom-columns=STATUS:.status.phase,NAME:.metadata.name | tail -n +2 | grep -v '^Running'; do
```
squat: once you add `-o pipefail` this while check will not do what we want. if kubectl returns an error, then the error will not be swallowed by the pipe, even if you do `+e`. As a result, if kubectl returns any error, then the while condition will evaluate to false and the function will exit.

tests/conformance/conformance.sh
```diff
@@ -0,0 +1,133 @@
...
+      return
+    fi
+
+    echo "kubectl failed, retrying in 5 seconds"
+    sleep 5
+  done
+}
+
+function wait_for_pods() {
+  set +e
```
squat: if possible, we should avoid having +e here

tests/conformance/conformance.sh
rithujohn191: any specifics on how?

squat: Setting `+x` before line 105 and then `-x` on line 123 may be sufficient

rithujohn191:
addressed all the feedback :)

squat:
ok to test

Quentin-M:
ok to test

tests/conformance/conformance.sh
```diff
@@ -0,0 +1,136 @@
...
+        fi
+
+        echo "Pods not available yet, waiting for 5 seconds"
+        sleep 5
+    done
+}
+
+kubectl help
+
+docker pull quay.io/coreos/tectonic-builder:v1.9
```
Quentin-M: Can we have the image as a variable at the top so it's easier to update? This is already super old. Also do you need to pull before running?

tests/conformance/conformance.sh
```diff
@@ -0,0 +1,136 @@
...
+docker run -i -v ${WORKSPACE}:${PROJECT} ${MNT_SECRETS} quay.io/coreos/tectonic-builder:v1.9 /bin/bash <<EOF
+set  -eo
+cd ${PROJECT}
+ 
+export PLATFORM=${PLATFORM}
+cluster_check "${CLUSTER}"
+export AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID}
```
Quentin-M: This is showing publicly in the logs..

tests/conformance/conformance.sh
```diff
@@ -0,0 +1,136 @@
...
+export AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY}
```
Quentin-M: these are shown in the logs because you didn't `set -x/+x` around.

tests/conformance/conformance.sh
```diff
@@ -0,0 +1,136 @@
...
+        echo "Pods not available yet, waiting for 5 seconds"
+        sleep 5
+    done
+}
+
+kubectl help
+
+docker pull quay.io/coreos/tectonic-builder:v1.9
+docker run -i -v ${WORKSPACE}:${PROJECT} ${MNT_SECRETS} quay.io/coreos/tectonic-builder:v1.9 /bin/bash <<EOF
+set  -eo
```
Quentin-M: why do we need to disable errors? looks like it will also affect `make apply`?

tests/conformance/conformance.sh
```diff
@@ -0,0 +1,136 @@
...
+
+export KUBECONFIG=${WORKSPACE}/build/${CLUSTER}/generated/auth/kubeconfig
+kubectl version
+
+echo "API Server UP!"
+wait_for_pods kube-system
+kubectl get pods --all-namespaces
+kubectl get nodes
+
+docker pull quay.io/coreos/kube-conformance:v1.6.2_coreos.0
```
Quentin-M: do you need to pull? can you move this as a variable?

tests/conformance/conformance.sh
rithujohn191: I see all the other scripts pulling before running it, so thought it would be a safe move. Changed it to a var.

tests/conformance/conformance.sh
rithujohn191: done

rithujohn191:
@Quentin-M @squat made quite a few changes since the Makefile changed.

Quentin-M:
I'd be more than happy to get this in, especially considering how much work you put in it and considering the fact that it works on Jenkins! So we can go with the pipeline and Slack notifications.

Will approve as soon as shellcheck passes!

rithujohn191:
ok to test

rithujohn191:
ok to test

PR #969, opened 2017-06-02
installer/scripts: Use HTTPS in awsutil.sh

bison:
Not sure why this is using HTTP.


Quentin-M:
Thanks for re-submitting this. It was first submitted on March 24 https://github.com/coreos-inc/tectonic/pull/1620 but got closed for a reason that was entirely unrelated. Both the fact that it got closed and the DMs I have gotten on Slack for it are ridiculous and quite outrageous to me.

PR #974, opened 2017-06-03
Jenkinsfile,tests/smoke: assume minimum role

squat:
We want to ensure that our smoke tests run with limited privileges so we
can catch when a change adds a dependency on a new privilege.

@ggreer @Quentin-M

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -29,6 +29,7 @@ pipeline {
   }
 
   environment {
+    TECTONIC_INSTALLER_ROLE= 'tectonic-installer'
     GO_PROJECT = '/go/src/github.com/coreos/tectonic-installer'
     MAKEFLAGS = '-j4'
   }
@@ -100,11 +101,13 @@ pipeline {
             unstash 'sanity'
             withCredentials(creds) {
               timeout(30) {
+                sh 'set +x -e && eval "$(${WORKSPACE}/tests/smoke/aws/smoke.sh assume-role "$TECTONIC_INSTALLER_ROLE")"'
                 sh '${WORKSPACE}/tests/smoke/aws/smoke.sh plan vars/aws.tfvars'
                 sh '${WORKSPACE}/tests/smoke/aws/smoke.sh create vars/aws.tfvars'
                 sh '${WORKSPACE}/tests/smoke/aws/smoke.sh test vars/aws.tfvars'
               }
               timeout(10) {
+                sh 'set +x -e && eval "$(${WORKSPACE}/tests/smoke/aws/smoke.sh assume-role "$TECTONIC_INSTALLER_ROLE")"'
                 sh '${WORKSPACE}/tests/smoke/aws/smoke.sh destroy vars/aws.tfvars'
               }
             }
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -100,11 +101,13 @@ pipeline {
             unstash 'sanity'
             withCredentials(creds) {
               timeout(30) {
+                sh 'source ${WORKSPACE}/tests/smoke/aws/smoke.sh assume-role "$TECTONIC_INSTALLER_ROLE"'
```
squat: we need to source this script so that exported AWS credentials are set on the shell and not a subshell.

tests/smoke/aws/smoke.sh
```diff
@@ -53,7 +81,7 @@ common() {
 
 create() {
     common "$1"
-    make plan | filter
+    assume_role
```
squat: we don't need to plan here. if you want to plan and create then execute two calls: `smoke.sh plan...` followed by `smoke.sh create...`

tests/smoke/aws/smoke.sh
```diff
@@ -80,33 +108,41 @@ test_cluster() {
...
+    printf "\t test <tfvars>           \ttest a Tectonic cluster parameterized by <tfvars>\n"
     printf "\n"
 }
 
-COMMAND=$1
-if [ $# -eq 0 ]; then
-    usage
-    exit 1
-fi
+main () {
```
squat: make this a function so we can make the variables local. this is useful so that sourcing the script does not set a bunch of variables like `COMMAND` on the calling shell.

tests/smoke/aws/smoke.sh
```diff
@@ -2,15 +2,43 @@
...
+    # shellcheck disable=SC2155
+    export AWS_ACCESS_KEY_ID="$(echo "$credentials" | jq -r '.AccessKeyId')"
+    # shellcheck disable=SC2155
+    export AWS_SECRET_ACCESS_KEY="$(echo "$credentials" | jq -r '.SecretAccessKey')"
+    # shellcheck disable=SC2155
+    export AWS_SESSION_TOKEN="$(echo "$credentials" | jq -r '.SessionToken')"
+}
 
 common() {
+    DIR="$( cd "$( dirname "$0" )" && pwd )"
```
squat: these are common dependencies so it makes sense for them to go in `common`.

Jenkinsfile
```diff
@@ -92,6 +93,14 @@ pipeline {
       }
     }
 
+    stage("Assume Role") {
+        steps {
+            withCredentials(creds) {
+                sh 'set +x -e && eval "$(${WORKSPACE}/tests/smoke/aws/smoke.sh assume-role "$TECTONIC_INSTALLER_ROLE")"'
```
squat: set +x so that the eval'd credentials are not printed out to stdout

squat: we use `eval` rather than `source` as the method for passing variables because if we source the script, we try to execute bash-specific commands like `set -o pipefail` in `sh` which causes an error.

Quentin-M:
Is there a simple way to do that? Stages are not meant to run individual steps like assuming a role.

squat:
@Quentin-M the other option is to run "assume-role" in every parallel stage before the make apply/plan/destroy. Do you prefer that style?

The Jenkins step must call +x to avoid printing the eval'd statements and the smoke.sh must call set +x to avoid printing the echo'd statements. We need both because smoke.sh does set -x at the very top when it is executed.

Quentin-M:
The reason we can't do that is that, stages/steps are not all going to run on the same machines (see bare-metal vs. AWS) - thus the environment won't be everywhere.

PR #976, opened 2017-06-05
Jenkinsfile: Fail tests if someone forgets to commit yarn.lock

ggreer:


ggreer:
This correctly fails if package.json is updated but yarn.lock isn't. See https://jenkins-tectonic-installer.prod.coreos.systems/blue/organizations/jenkins/coreos%20-%20tectonic-installer%2Ftectonic-installer/detail/PR-976/4/

ggreer:
Tests passed: https://jenkins-tectonic-installer.prod.coreos.systems/blue/organizations/jenkins/coreos%20-%20tectonic-installer%2Ftectonic-installer/detail/PR-976/5/pipeline

 I think GitHub outage caused the PR status not to update.

Quentin-M:
I disagree with this change:

```
1. Stages should simply invoke a make target or a self-contained script. Do not write testing logic in this Jenkinsfile.
```

alexsomesan:
@Quentin-M has a good point!
We discussed this convention early on. We all agreed no logic should go into the Jenkins file.
Everything should be either a make target or a script. Jenkins should just invoke those.

This allows for those operations to be reproducible outside of Jenkins, by running them manually if necessary. Logic that is in the Jenkins file can't easily be dealt with outside of Jenkins.

PR #980, opened 2017-06-05
images/tectonic-builder: add awscli to image

squat:
We need awscli for assuming the limited role for #974.
cc @ggreer @kans


kans:
Whats the motivation?

Quentin-M:
Giphy

PR #984, opened 2017-06-05
modules/aws/ignition: template awscli

squat:
We should be using a variable for the awscli container image tag.

cc @kans @Quentin-M


Quentin-M:
Looks like you forgot to add container_images 's default in config.tf

squat:
@Quentin-M that image already exists in config.tf https://github.com/coreos/tectonic-installer/blob/master/config.tf#L46

squat:
@Quentin-M fixed the tempate. PTAL

PR #987, opened 2017-06-06
Jenkinsfile,examples,tests/smoke/aws: add custom ca test

squat:
We need to expand our smoke test matrix; this commit adds a smoke test
that uses a custom CA for all certificate generation. Eventually, we
should make these tests do a full `smoke.sh create && smoke.sh test`.

@ggreer @kans @sym3tri


kans:
+1

PR #988, opened 2017-06-06
modules/*: Update kubelet.service to save kubelet pod uuid in less volatile location

yifan-gu:
Otherwise the uuid file won't survive the reboot which makes `rkt rm --uuid=` useless.

Xref: https://github.com/kubernetes-incubator/bootkube/pull/562


yifan-gu:
kubernetes-incubator/bootkube#562 is merged.

Issue #337, opened 2017-04-25
azure: tighten security groups according to the generic platform spec


s-urbaniak:
https://github.com/coreos/tectonic-installer/blob/master/Documentation/generic-platform.md specifies settings which are not reflected in the current setup of security groups on azure: https://github.com/coreos/tectonic-installer/blob/master/modules/azure/vnet/security-groups.tf.

s-urbaniak:
xref #102, as it was originally requested

PR #991, opened 2017-06-06
azure/network: Conform Azure Network Security Groups to Tectonic generic platform standards

justaugustus:
Initial work to conform Azure NSGs (Network Security Groups) to [Tectonic Generic Platform guidelines](https://github.com/coreos/tectonic-installer/blob/master/Documentation/generic-platform.md)

This PR introduces NSGs (and supporting vars) for:
* etcd Availability Set
* master nodes
* worker nodes
* load balancers

In addition, the ability to conditionally create rules based on Tectonic node roles.

Fixes: #337
Supercedes: #865


justaugustus:
@Quentin-M / @alexsomesan / @s-urbaniak --
Let me know your thoughts on this!

examples/terraform.tfvars.azure
```diff
@@ -150,6 +195,20 @@ tectonic_service_cidr = "10.3.0.0/16"
 // 
 tectonic_ssh_key = ""
 
+// (optional) Network (external) to allow SSH access from. Maps to `source_address_prefix` in Azure.
+// Defaults to `*`. Can be external to Azure environment.
+// Allowed values: [network CIDR (i.e., 10.0.0.0/16) | `VirtualNetwork` | `Internet` | `*` ]
+// tectonic_ssh_network_ext = "VirtualNetwork"
```
s-urbaniak: For the sake of consistency I would also call this `tectonic_ssh_network_external`

Documentation/variables/azure.md
```diff
@@ -22,5 +31,9 @@ This document gives an overview of variables used in the Azure platform of the T
 | tectonic_azure_use_custom_fqdn | (optional) If set to true, assemble the FQDN from the configuration. Otherwise, use the FQDN set up by Azure. | string | `false` |
 | tectonic_azure_vnet_cidr_block | Block of IP addresses used by the Resource Group. This should not overlap with any other networks, such as a private datacenter connected via ExpressRoute. | string | `10.0.0.0/16` |
 | tectonic_azure_worker_vm_size | Instance size for the worker node(s). Example: Standard_DS2_v2. | string | `Standard_DS2_v2` |
+| tectonic_enable_ssh_external | Specifies if SSH access should be allowed from external networks | string | `false` |
```
s-urbaniak: Please name these new variables concerning azure only with a `tectonic_azure_...` prefix, i.e. `tectonic_enable_ssh_external` would become `tectonic_azure_enable_ssh_external`

examples/terraform.tfvars.azure
s-urbaniak: The default indiciates `VirtualNetwork` rather than `*`

examples/terraform.tfvars.azure
```diff
@@ -150,6 +195,20 @@ tectonic_service_cidr = "10.3.0.0/16"
...
+
+// Network (internal) to allow SSH access from. Maps to `source_address_prefix` in Azure.
+// Defaults to `VirtualNetwork`. Should be internal to Azure environment.
+// Allowed values: [network CIDR (i.e., 10.0.0.0/16) | `VirtualNetwork` | `Internet` | `*` ]
+tectonic_ssh_network_int = "VirtualNetwork"
+
+// (optional) Specifies whether a jumpbox should be created to manage cluster nodes.
+// Experimental - DO NOT USE
+// tectonic_use_jumpbox = "false"
```
s-urbaniak: That is not a bad idea, indeed. Especially in the context of testing this might be useful.

modules/azure/vnet/nsg-etcd.tf
```diff
@@ -0,0 +1,154 @@
+resource "azurerm_network_security_group" "etcd" {
+  count               = "${var.external_etcd_nsg_name == "" ? 1 : 0}"
+  name                = "${var.tectonic_cluster_name}-etcd-nsg"
```
s-urbaniak: This object is already of type "network security group", hence I would not redundantly encode it in the name. I suggest to call it simply `${var.tectonic_cluster_name}-etcd`.

modules/azure/vnet/nsg-master.tf
```diff
@@ -0,0 +1,342 @@
...
+  destination_port_range      = "*"
+  source_address_prefix       = "${var.vnet_cidr_block}"
+  destination_address_prefix  = "*"
+  resource_group_name         = "${var.external_nsg_rsg_name == "" ? var.resource_group_name : var.external_nsg_rsg_name}"
+  network_security_group_name = "${var.external_master_nsg_name == "" ? join("",azurerm_network_security_group.master.*.name) : var.external_master_nsg_name }"
+}
+
+resource "azurerm_network_security_rule" "master_ingress_ssh" {
+  count                       = "${var.create_master_nsg_rules ? 1 : 0}"
+  name                        = "${var.tectonic_cluster_name}-master_ingress_ssh"
```
s-urbaniak: We should be consistent with underscores `_` and dashes `-` across the names. We tend to use underscores `_` only.

platforms/azure/variables.tf
```diff
@@ -7,6 +7,49 @@ EOF
...
+(optional) Network (external) to allow SSH access from. Maps to `source_address_prefix` in Azure.
+Defaults to `*`. Can be external to Azure environment.
+Allowed values: [network CIDR (i.e., 10.0.0.0/16) | `VirtualNetwork` | `Internet` | `*` ]
+EOF
+
+  default = "VirtualNetwork"
+}
+
+# TODO: Check if similar var exists for AWS
+variable "tectonic_use_jumpbox" {
```
s-urbaniak: We don't have jumpbox creation for AWS, but I believe this is a great idea, also for the sake of testing :+1:

justaugustus: @s-urbaniak thoughts on moving this to `config.tf`?
Right now, it's inactive, but the intention would be to introduce it for each of the cloud platforms.

modules/azure/vnet/nsg-master.tf
justaugustus: @s-urbaniak Ideally, I'd like to keep this as is for now.
There are several instances where hyphens are used over underscores e.g., https://github.com/coreos/tectonic-installer/blob/master/modules/azure/master/master.tf#L23-L28

If we make the decision to definitively go one way or another, it should probably be its' own PR with a wholesale cleanup.

Let me know what you think!

s-urbaniak: Agreed, I am fine this not being part of the PR.

s-urbaniak: Created #993 as a follow-up issue.

platforms/azure/variables.tf
s-urbaniak: I am fine with moving this to `config.tf` with a disclaimer in the description, that this works only for Azure right now.

But to be sure, let's gather also the opinion of @alexsomesan about the possibility to add an option to instanciate a bastion/jumpbox vm for encapsulated networks.

platforms/azure/main.tf
```diff
@@ -9,13 +9,29 @@ module "resource_group" {
...
+  ssh_network_external      = "${var.tectonic_azure_ssh_network_external}"
+  external_nsg_rsg_name     = "${var.tectonic_azure_external_nsg_rsg_name}"
+  external_etcd_nsg_name    = "${var.tectonic_azure_external_etcd_nsg_name}"
+  external_api_nsg_name     = "${var.tectonic_azure_external_api_nsg_name}"
+  external_master_nsg_name  = "${var.tectonic_azure_external_master_nsg_name}"
+  external_worker_nsg_name  = "${var.tectonic_azure_external_worker_nsg_name}"
+  create_etcd_nsg_rules     = "${var.tectonic_azure_create_etcd_nsg_rules}"
+  create_api_nsg_rules      = "${var.tectonic_azure_create_api_nsg_rules}"
+  create_master_nsg_rules   = "${var.tectonic_azure_create_master_nsg_rules}"
+  create_worker_nsg_rules   = "${var.tectonic_azure_create_worker_nsg_rules}"
```
s-urbaniak: This seems like an extremely fine-grained possibility to tweak security group creation vs. external reference. Do we actually this level of granularity?

I.e. in AWS we simply create those security groups, always. I wonder why in the case of Azure we need to optionally reference them?

justaugustus: @s-urbaniak For context, this was a requirement coming from a customer, due to a fairly restrictive IAM policy.

It would be useful in other similar scenarios.

s-urbaniak:
(modulo review comments from @alexsomesan regarding the jumpbox setting)

platforms/azure/variables.tf
alexsomesan: I agree it's a great idea. We should backport this to the other platforms.

platforms/azure/main.tf
alexsomesan: I second what @s-urbaniak suggested. This is a level of granularity we can't realistically sustain in Tectonic from a product perspective. If we set such a precedent for one corner-case, it could backfire into an unsustainable set of corner-case uses that ask to be supported. We want to keep the product maintainable and generic enough to appeal to the maximum number of users.

The way we currently tackle this particular aspect usage of security groups in AWS is better suited for generic consumption. I think we should use that as a guideline. So in AWS, we allow users to opt out of tectonic creating networking resources and instead bring in their own, pre-configured network. In this mode, we stay COMPLETELY hands off. Meaning we don't try to create anything, but also expect that the user takes full control of configuring those resources.

How that would translate in here is: just ask the user to supply us with correctly configured NSGs, which have all the rules. When users supply NSG, we completely refrain from creating them or their rules.

I suggest we move in the direction of that approach and drop these `create_*_nsg_rules` flags altogether. The user input would then become just a set of NSG ids.

platforms/azure/variables.tf
justaugustus: @s-urbaniak @alexsomesan Awesome. Moved to `config.tf` with a note to work on backporting it.

metral:
If the rest of this PR can get in and the only bits left out are the `create_*_nsg_rules` I can live with that being in the fork for the customer.

@s-urbaniak @alexsomesan thoughts? If so, then @justaugustus remove the `create_*_nsg_rules` bits from this PR.

justaugustus:
@s-urbaniak @alexsomesan @metral --> https://github.com/coreos/tectonic-installer/pull/991/commits/13db235fe64e96b52fbcc1e6da0b08b87ca8c247

I think that's sort of a happy medium, where you can choose a few paths:
1. TF creates NSGs and rules for you
2. User supplies NSG(s) and TF creates ALL rules
3. User supplies NSG(s) and TF creates NO rules; user must following Generic Platform guidelines to ensure correct rules

`tectonic_azure_create_{api,etcd,master,worker}_nsg_rules` vars have been folded into `tectonic_azure_create_nsg_rules` and some var names have been cleaned up.

metral:
@justaugustus the changes in https://github.com/coreos/tectonic-installer/commit/13db235fe64e96b52fbcc1e6da0b08b87ca8c247 LGTM

justaugustus:
@alexsomesan / @alexsomesan --
https://github.com/coreos/tectonic-installer/pull/991/commits/3b86e23914e60859a065941552e261169dde65d8 reintroduces that "all or nothing" logic for creating the NSGs and accompanying rules.

modules/azure/vnet/variables.tf
```diff
@@ -88,18 +88,6 @@ variable "external_nsg_worker" {
...
-variable "create_etcd_nsg_rules" {
-  default = false
-}
-
-variable "create_master_nsg_rules" {
-  default = false
-}
-
-variable "create_worker_nsg_rules" {
+variable "create_nsg_rules" {
```
metral: This can be removed since its no longer used

alexsomesan: What Mike said :)

alexsomesan:
Merging since PR builder doesn't check Azure.

PR #994, opened 2017-06-06
vmware: Fix incorrect file modes

alekssaul:
Change permissions of `/etc/sysctl.d/max-user-watches.conf` and `/etc/kubernetes/kubelet.env` similar to https://github.com/coreos/tectonic-installer/pull/936

### DIFF
--- a/modules/vmware/node/ignition.tf
+++ b/modules/vmware/node/ignition.tf
@@ -82,7 +82,7 @@ data "ignition_systemd_unit" "kubelet-env" {
 data "ignition_file" "max-user-watches" {
   filesystem = "root"
   path       = "/etc/sysctl.d/max-user-watches.conf"
-  mode       = "420"
+  mode       = "0644"
 
   content {
     content = "fs.inotify.max_user_watches=16184"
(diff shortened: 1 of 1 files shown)


modules/vmware/node/ignition.tf
```diff
@@ -82,7 +82,7 @@ data "ignition_systemd_unit" "kubelet-env" {
 data "ignition_file" "max-user-watches" {
   filesystem = "root"
   path       = "/etc/sysctl.d/max-user-watches.conf"
-  mode       = "420"
+  mode       = "0644"
```
ggreer: I have no idea if the quotes change the behavior, but https://github.com/coreos/tectonic-installer/pull/936 didn't use quotes.

ggreer:
As [I said before](https://github.com/coreos/tectonic-installer/pull/936#issuecomment-305266731), I think we're going to continue to have this issue until some linting step detects it. Terraform makes it incredibly easy to make this mistake.

modules/vmware/node/ignition.tf
diegs: Don't use quotes. Agree that a linter would help with this.

alekssaul:
Thanks @ggreer & @diegs 

Removed the quotes. Did a quick search to make sure `mode       = "` is not used in modules folder.

PR #1006, opened 2017-06-06
frontend: Validate lengths of AWS key/secret.

ggreer:
This prevents us from hitting the AWS API every time these fields are changed.


kans:
Interesting... although I'd like to debug the actual problem.

kans:
did you look up those values in docs?

ggreer:
I couldn't find the actual constraints in any AWS docs. Googling gave me those lengths, and I couldn't find any counterexamples.

PR #931, opened 2017-05-30
pod-checkpointer: 4e7a7dab10bc4d895b66c21656291c6e0b017248

diegs:
Includes fixes for atomic writes and crashlooping.

Issue #926, opened 2017-05-30
API server fails to start when bootstrapping cluster


sym3tri:
There have been various reports of this during installation. It seems to occur semi-frequently but we've thus far been unable to consistently reproduce.

/cc @ggreer @kans please add any additional details.

kans:
The init-assets service is the only known cause atm.  I hope to have a PR by eod.

diegs:
Assigning to @kans because this doesn't appear to be a bootkube issue ATM.

Please feel free to assign back to me if you encounter any bootkube-related failure modes.

sym3tri:
sounds like this may just be a dupe of #924

kans:
partially addressed by #937 #947 #946

sym3tri:
closing b/c dupe of #924

PR #1007, opened 2017-06-07
pod-checkpointer: 4e7a7dab10bc4d895b66c21656291c6e0b017248

diegs:


diegs:
Cherry-pick of https://github.com/coreos/tectonic-installer/pull/931

Open to discussion, but this an issue that at least some people have seen, and could be a possible cause of apiserver unavailability (though only really if a node reboots, so not exactly https://github.com/coreos/tectonic-installer/issues/926)

aaronlevy:
@brianredbeard just ran into this, and it's something we should get fixed. We had found this issue a couple weeks ago - but when we were unsure about all of the update instability (terminating pods and such) it was backed out as a possible cause.

Now that we know this is unrelated, this has been running as part of the bootkube test suite for 2weeks - and seems stable.

Ideally we would get this also into an update payload - but because of the late timing, we could fix it in the installer, and follow up with the fix in the update payload in the next release).

PR #1024, opened 2017-06-07
Integration gui installer tests to run on jenkins

gopilal:


installer/Makefile
```diff
@@ -105,19 +105,20 @@ bin/sanity:
 	go test tests/sanity/k8s_test.go -c -o bin/sanity
 
 .PHONY: launch-installer
-launch-installer: build
+launch-installer:
```
ggreer: I'm pretty sure the installer needs to be built before it can be run.

installer/Makefile
```diff
@@ -105,19 +105,20 @@ bin/sanity:
...
 	mkdir -p frontend/ui-tests/reports/
 	./bin/$(OS)/installer $(INSTALLER_OPTS) & echo $$! > frontend/ui-tests/reports/installer.pid
 
 .PHONY: launch-installer-guitests
 launch-installer-guitests: launch-installer
-	cd frontend && yarn run installer-tests
+	Xvfb -ac :99 -screen 0 1280x1024x16 & echo $$! > frontend/ui-tests/reports/xvfb.pid
+	cd frontend && export DISPLAY=:99; yarn run installer-tests
```
ggreer: The second part of this line (`yarn run installer-tests`) is run in the current directory, not frontend. You probably want to remove the semicolon and the export. eg: `DISPLAY=:99 yarn run installer-tests`

installer/Makefile
gopilal: since its already build as part of previous steps on jenkins file which gets stashed. I'm testing these changes though

ggreer: That stash command only stashes the installer binary. It doesn't stash node_modules.

installer/Makefile
gopilal: cannot run ```yarn run installer-tests``` as package.json is in forntend.

Jenkinsfile
```diff
@@ -131,9 +131,24 @@ pipeline {
                 }
               }
             }
-          }
-        )
-      }
+          },
+          "IntegrationTest Installer Gui": {
```
ggreer: I think this step needs the environment stuff from up on line 37-40.

images/tectonic-builder/Dockerfile
```diff
@@ -53,6 +53,16 @@ RUN cd /tmp && \
     mv /tmp/dist /usr/local/yarn && \
     ln -s /usr/local/yarn/bin/yarn /usr/local/bin/yarn
 
+
+### install chrome & xvfb to run the installer-gui tests.
+
+RUN wget --quiet -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | sudo apt-key add - && \
+    sudo sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google-chrome.list'
```
ggreer: Not a blocker, but I think it's possible to avoid adding another apt source by using Chromium instead of Chrome.

Quentin-M:
Congrats!

Quentin-M:
Thanks for the hard work :)

PR #1036, opened 2017-06-08
Add existing certificate support.

coresolve:
The premise here is to seperate the certs into a certs used to terminate
tls and certificates used for auth.

- Refactor existing certs to a map called tectonic-existing-certs.
  This may be a breaking change. Need to figure out how the console does
this. I went to a model where we provide a path similar to how we
provide paths for the ssh public key.

- Support existing certs for securing apiserver and ingress controller
and the ca key..

- If the user provides a ca.key path we will use that to sign the certs we generate
using the provided ca key and cert.

- If the user provides only a ca.crt we will generate a ca that will be
used to sign those certs that we generate for client auth. We will also
bundle the ca.crt we create and the ca.crt provided so that all the
things trust certs created by both.


coresolve:
@alexsomesan this is the pr for existing certs.

s-urbaniak:
@coresolve thanks a lot for the PR! The "map as variable container" idiom is something we avoided until now. Honestly, I am not aware any more of the gotchas involved, if any.

The only nit I have is, if we go the map-as-vars route, we should also factor the existing `tectonic_etcd_ca_cert_path`, `tectonic_etcd_client_cert_path`, and `tectonic_etcd_client_key_path` variables into this cert map which I use heavily in introducing TLS communication to etcd in #751.

From my side this generally LGTM modulo any concerns from @alexsomesan.

coresolve:
@s-urbaniak I am happy to roll the etcd certs in. I am still a little concerned about the changes to the ui though.

modules/bootkube/resources/manifests/kube-apiserver.yaml
```diff
@@ -44,10 +44,10 @@ spec:
         - --tls-ca-file=/etc/kubernetes/secrets/ca.crt
         - --tls-cert-file=/etc/kubernetes/secrets/apiserver.crt
         - --tls-private-key-file=/etc/kubernetes/secrets/apiserver.key
-        - --kubelet-client-certificate=/etc/kubernetes/secrets/apiserver.crt
-        - --kubelet-client-key=/etc/kubernetes/secrets/apiserver.key
+        - --kubelet-client-certificate=/etc/kubernetes/secrets/kubelet.crt
```
aaronlevy: Name changes here require an upgrade path to all existing clusters as well - otherwise we have some clusters with one set of flags and another set with different flags. This could cause problems in the future.

Are the file name changes a requirement, and if so - this needs to have an upgrade path prior to merging.

cc @diegs as he's been working on flag merging logic.

metral: I'll let @coresolve clarify, but IIUC, this switch happened because client auth with Kubelet's using the apiserver cert & key required access to the CA key: https://github.com/coreos/tectonic-installer/pull/1036/files#diff-a6092b44e14d3f0ebd9d657ab9ba5998L85.

Providing the CA Key is understandably something not all users may not want to do.

The kubelet's instead generate their own self-signed cert to get around the CA key requirement, while still being able to encrypt internal comms.

However, I agree that an upgrade path is in order to manage the different sets of flags clusters can have.

aaronlevy: None of these should necessitate the use of the CA key. The CA key doesn't/shouldn't even need to exist at all in the cluster right now (until we actually use the CSR endpoint). This is simply saying that the apiserver key/cert should be used as the client cert -- which is pretty valid: the client is the apiserver (as a client) is reaching out to the kubelet api.

Could be misunderstanding something (or maybe tectonic does this differently than bootkube).

metral: @coresolve I probably misunderstood the change here. Care to clear things up?

aaronlevy: From an offline discussion:

It sounds like the need is to allow the apiserver to use separate client and server certificates (right now it uses the same certificate for both). We shouldn't be using the kubelet cert for this, however, so as follow up this should change to generating 2 certs for the apiserver, one with server usage, one with client.

@ericchiang any particular rbac group that the client usage cert should get?

ericchiang:
Some notes on a potential upgrade path.

Currently the API server cert secret looks something like:

```yaml
kind: Secret
apiVersion: v1
data:
  apiserver.crt: (data)
  apiserver.key: (data)
```

This PR would add two additional fields to the secret:

```yaml
kind: Secret
apiVersion: v1
data:
  apiserver.crt: (data)
  apiserver.key: (data)
  # API server -> kubelet client cert
  kubelet.crt: (data)
  kubelet.key: (data)
```

Upgrade is as follows:

1. Copy `apiserver.crt` and `apiserver.key` to `kubelet.crt` and `kubelet.key` respectively. Since the current certs has both serving and client usages it can act as either.
2. Modify API server flags `--kubelet-client-certificate` and `--kubelet-client-key` to point to the new secret fields.

Not that this unblocks this PR, but it gives us an upgrade path while letting new installs use different serving and client certs.

@diegs does that sound sane?

aaronlevy:
As long as we're okay with the inserted certificate still having both usages (which I have no objections to). This should be doable.

I do think we need to come up with different names for the files - but we can wait until this PR settles before working on the migration code

modules/bootkube/resources/manifests/kube-apiserver.yaml
coresolve: Thanks @aaronlevy I will address this change.

metral:
Do we have any updates on where this PR stands?

sym3tri:
@metral @coresolve This is a very useful feature. I'd like to get it merged too, but the remaining tasks are:

- [ ] Rebase & resolve merge conflicts
- [ ] Ensure smoke tests pass again for all platforms
- [ ] Make requested flag changes
- [ ] Get final technical sign-off from @alexsomesan (esp WRT variable map)
- [ ] Coordinate upgrade path with Core Services team (@aaronlevy)
- [ ] Coordinate any UI changes with me or @squat 
- [ ] Get product sign-off from @robszumski 
- [ ] File issue for docs team to update the docs.

s-urbaniak:
Closing as bring-your-own ingress certificates landed in master. Let's create separate issues for the remaining TLS assets.

PR #1041, opened 2017-06-09
docs: updating VPC peering note.

zbwright:


colhom:
thanks @zbwright. non-generated docs change, no retest needed.

PR #1043, opened 2017-06-09
Use Availability Set for masters and workers

### LINKED ISSUE #93: azure: put masters and workers into AvailabilitySets and ScaleSets
ScaleSets currently cannot attach different disks to different virtual machines. This means that Kubernetes persistent volumes will not work with our setup for multi-machine.

To fix this we need to offer the ability to spin-up via AvailabilitySets instead. Should we create a separate module if someone wants to use AvailabilitySets vs ScaleSets similar to how you do with AWS?

cc @alexsomesan

discordianfish:
Instances in a Scale Set are not supported by the kubernetes
cloud-provider integration.

This closes #93


alexsomesan:
@discordianfish 
Thanks for the contribution.
Coincidentally, we have a similar change in the works. However, we planned it to be a bit different.

We're aiming to have alternative implementations for ScaleSets and AvailabilitySets. Because of the current limitations with ScaleSets, the plan is to default to using ASes for as long as necessary and then flip back to ScaleSets once they get the necessary features to work well with K8S.

Would you mind reworking your change to make these a new pair of modules, called `modules/azure/master-as` and `modules/azure/worker-as` and rename the existing ones to be prefixed with `-ss`?

discordianfish:
@alexsomesan Like this?
Feel free to pick up my changes and adjust them as you see fit, I won't be able to get back as this in the next days.

alexsomesan:
Yep, that's what I meant.
Thanks a lot! 👍 

I will bring this in and adjust it myself when necessary.

PR #1059, opened 2017-06-12
contrib,tests: add private vpc smoke test

squat:
Enable a smoke test that creates Tectonic cluster in a private VPC. In
order for this to work, we need to first create a VPN connection into
the private VPC. This commit adds two commands to smoke.sh for managing
the state of the private VPC `create-vpc` and `destroy-vpc`.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -102,7 +102,7 @@ pipeline {
               }
             }
           },
-          "SmokeTest TerraForm: AWS (Experimental)": {
+          "SmokeTest TerraForm: AWS (experimental)": {
             node('worker && ec2') {
               withCredentials(creds) {
                 withDockerContainer(builder_image) {
@@ -118,7 +118,7 @@ pipeline {
               }
             }
           },
-          "SmokeTest TerraForm: AWS (custom-ca)": {
+          "SmokeTest TerraForm: AWS (custom ca)": {
             node('worker && ec2') {
               withCredentials(creds) {
                 withDockerContainer(builder_image) {
@@ -128,6 +128,27 @@ pipeline {
                     sh """#!/bin/bash -ex
                     . ${WORKSPACE}/tests/smoke/aws/smoke.sh assume-role "$TECTONIC_INSTALLER_ROLE"
                     ${WORKSPACE}/tests/smoke/aws/smoke.sh plan vars/aws-ca.tfvars
+                    """
+                  }
+                }
+              }
+            }
+          },
+          "SmokeTest TerraForm: AWS (private vpc)": {
+            node('worker && ec2') {
+              withCredentials(creds) {
+                withDockerContainer(image: builder_image, args: '--device=/dev/net/tun --cap-add=NET_ADMIN') {
+                  checkout scm
+                  unstash 'installer'
+                  unstash 'sanity'
+                  timeout(40) {
+                    sh """#!/bin/bash -ex
+                    . ${WORKSPACE}/tests/smoke/aws/smoke.sh create-vpc
+                    ${WORKSPACE}/tests/smoke/aws/smoke.sh plan vars/aws-vpc.tfvars
+                    ${WORKSPACE}/tests/smoke/aws/smoke.sh create vars/aws-vpc.tfvars
+                    ${WORKSPACE}/tests/smoke/aws/smoke.sh test vars/aws-vpc.tfvars
+                    ${WORKSPACE}/tests/smoke/aws/smoke.sh destroy vars/aws-vpc.tfvars
+                    ${WORKSPACE}/tests/smoke/aws/smoke.sh destroy-vpc
                     """
                   }
                 }
@@ -160,12 +181,12 @@ pipeline {
                   make launch-installer-guitests
                   make gui-tests-cleanup
                   """
-               }
-             }
-           }
-         }
-       )
-     }
+                }
+              }
+            }
+          }
+        )
+      }
       post {
         failure {
           node('worker && ec2') {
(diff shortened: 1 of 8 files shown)


kans:
```vars/aws-vpc.tfvars': No such file or directory```

**Comment by squat** [comment]
thanks, @kans already pushed a fix. Everything runs locally; I just need to codify for Jenkins now. I'll ping when it's ready.

**Comment by ggreer** [comment]
Just FYI, you'll want to rebase master & push a new builder image to quay, now that #1024 is merged.

**Comment by squat** [comment]
ok to test

**Comment by squat** [comment]
@Quentin-M @kans PTAL!

File: Jenkinsfile [hunk 1]
```diff
@@ -176,7 +197,8 @@ pipeline {
                 retry(3) {
                   timeout(15) {
                     sh """#!/bin/bash -ex
-                    . ${WORKSPACE}/tests/smoke/aws/smoke.sh assume-role "$TECTONIC_INSTALLER_ROLE"
```
squat: There is no reason to mess around with assuming roles when we are recovering from a failure. If the assumed role is not privileged enough to destroy then that failure will be caught in the respective stage.

Quentin-M:
ok to test

tests/smoke/aws/smoke.sh
```diff
@@ -85,6 +93,77 @@ create() {
...
 }
 
+common_vpc() {
+    random_region
+    export TF_VAR_vpc_aws_region="$TF_VAR_tectonic_aws_region"
+    TF_VAR_vpc_name="$(echo "vpc-$BRANCH_NAME-$BUILD_ID" | awk '{print tolower($0)}')"
+    export TF_VAR_vpc_name
+}
+
+create_vpc() {
```
Quentin-M: should it be called something mentioning internal cluster / vpc?

tests/smoke/aws/smoke.sh
```diff
@@ -85,6 +93,77 @@ create() {
...
+    TF_VAR_tectonic_aws_external_private_zone="$(terraform output -json | jq -r '.private_zone_id.value')"
+    export TF_VAR_tectonic_aws_external_private_zone
+    TF_VAR_tectonic_aws_external_vpc_id="$(terraform output -json | jq -r '.vpc_id.value')"
+    export TF_VAR_tectonic_aws_external_vpc_id
+    # shellcheck disable=SC2183,SC2046
+    TF_VAR_tectonic_aws_external_master_subnet_ids="$(printf "[%s, %s]" $(terraform output -json | jq '.subnets.value[0,1]'))"
+    export TF_VAR_tectonic_aws_external_master_subnet_ids
+    # shellcheck disable=SC2183,SC2046
+    TF_VAR_tectonic_aws_external_worker_subnet_ids="$(printf "[%s, %s]" $(terraform output -json | jq '.subnets.value[2,3]'))"
+    export TF_VAR_tectonic_aws_external_worker_subnet_ids
```
Quentin-M: could you squash the var definition / export statements?
and jump some lines maybe between individual logic blocks?

tests/smoke/aws/smoke.sh
squat: Remember, this command does not create a cluster. It only creates a vpc into which we can launch a cluster. Maybe "create-private-vpc" but now we start getting into long command names. I can live with that if you feel strongly about it.

Jenkinsfile
```diff
@@ -176,7 +197,8 @@ pipeline {
...
+                    ${WORKSPACE}/tests/smoke/aws/smoke.sh destroy vars/aws-vpc.tfvars
+                    ${WORKSPACE}/tests/smoke/aws/smoke.sh destroy-vpc
```
Quentin-M: It is okay to do that all the time? Does not look bad, but figured I'd ask.

squat: You can run destroy-vpc multiple times in a row with no error. Unless terraform destroy fails, this command should exit 0.

Jenkinsfile
```diff
@@ -176,7 +197,8 @@ pipeline {
                 retry(3) {
                   timeout(15) {
                     sh """#!/bin/bash -ex
-                    . ${WORKSPACE}/tests/smoke/aws/smoke.sh assume-role "$TECTONIC_INSTALLER_ROLE"
+                    ${WORKSPACE}/tests/smoke/aws/smoke.sh destroy vars/aws-vpc.tfvars
```
kans: I think we need an independent retry block per apply.

Quentin-M: yeap

kans: Its maybe OK for two clusters.  What happens if the first destroy throws, or if it doesn't halt, or even if it just takes a really long time?

squat:
@Quentin-M I consolidated the var declaration and exporting. Tests are passing. Let's get this merged soon.

Jenkinsfile
squat: @kans I changed it so that there are two retries. one for every cluster that is actually created.

Quentin-M:
Yass!

tests/smoke/aws/smoke.sh
```diff
@@ -85,6 +93,72 @@ create() {
...
+    echo "nameserver 8.8.8.8"  >> /etc/resolv.conf
+    # Export all of the VPC details.
+    # shellcheck disable=SC2155,SC2183,SC2046
+    {
+        export TF_VAR_tectonic_aws_external_private_zone="$(terraform output -json | jq -r '.private_zone_id.value')"
+        export TF_VAR_tectonic_aws_external_vpc_id="$(terraform output -json | jq -r '.vpc_id.value')"
+        export TF_VAR_tectonic_aws_external_master_subnet_ids="$(printf "[%s, %s]" $(terraform output -json | jq '.subnets.value[0,1]'))"
+        export TF_VAR_tectonic_aws_external_worker_subnet_ids="$(printf "[%s, %s]" $(terraform output -json | jq '.subnets.value[2,3]'))"
+        export TF_VAR_tectonic_base_domain="$(terraform output -json | jq -r '.base_domain.value')"
+    }
```
ggreer: Should we export these in common or something? Otherwise vpc-destroy won't work, since it will lack these env vars. See https://jenkins-tectonic-installer.prod.coreos.systems/job/tectonic-installer/job/PR-1075/19/execution/node/198/log/ for an example of this failing.

PR #1063, opened 2017-06-12
GUI Tests: show browser console logs

ggreer:


kans:
aws_s3_bucket.tectonic: Error creating S3 bucket: TooManyBuckets: You have attempted to create more buckets than allowed

PR #1064, opened 2017-06-12
frontend: Fetches the latest release from github to display in the header

amrutac:
Implements https://github.com/coreos/tectonic-installer/issues/364

### DIFF
--- a/installer/frontend/components/header.jsx
+++ b/installer/frontend/components/header.jsx
@@ -1,9 +1,78 @@
 import React from 'react';
-
+import semver from 'semver';
 import { Dropdown } from './ui';
 
+
+const fetchLatestRelease = () => {
+  return fetch('/releases/latest',
+    { method: 'GET' }).then((response) => {
+      if (response.ok) {
+        return response.text();
+      }
+      return response.text().then(err => Promise.reject(err));
+    });
+};
+
+const parseLatestVersion = (html) => {
+  const parser = new DOMParser();
+  const htmlDoc = parser.parseFromString(html, "text/html");
+  return htmlDoc.getElementsByClassName('latestReleaseTag')[0].textContent.trim();
+};
+
+const hasNewVersion = (latestRelease) => {
+  if (semver.major(latestRelease) > semver.major(GIT_TAG)) {
+    return true;
+  }
+
+  if (semver.minor(latestRelease) > semver.minor(GIT_TAG)) {
+    return true;
+  }
+
+  if (semver.patch(latestRelease) > semver.patch(GIT_TAG)) {
+    return true;
+  }
+
+
+  const latestReleasePr = semver.prerelease(latestRelease);
+  const gitTagPr = semver.prerelease(GIT_TAG);
+
+  //No rc string in the latest release
+  if (latestReleasePr.length < gitTagPr.length) {
+    return true;
+  }
+
+  //compares rc-x string
+  if (latestReleasePr[1] > gitTagPr[1]) {
+    return true;
+  }
+
+  //rc version
+  if (latestReleasePr[2] > gitTagPr[2]) {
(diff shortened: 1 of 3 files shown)


installer/frontend/components/header.jsx
```diff
@@ -3,14 +3,40 @@ import semver from 'semver';
...
+};
+
 export class Header extends React.Component {
+  constructor (props) {
+    super(props);
+    this.state = { latestRelease: null };
+  }
+
+  componentDidMount() {
+    if (!this.state.latestRelease) {
```
kans: You don't need this `if`.

installer/frontend/components/header.jsx
```diff
@@ -3,14 +3,40 @@ import semver from 'semver';
...
+      getLatestRelease().then((release) => {
+        this.setState({ latestRelease: release.tag_name });
+      }).catch(() => this.setState({ latestRelease: null }));
```
kans: I have no idea what to do with this error?  Maybe just log it.

installer/frontend/components/header.jsx
```diff
@@ -3,14 +3,40 @@ import semver from 'semver';
 
 import { Dropdown } from './ui';
 
+
+const getLatestRelease = () => {
+  return fetch('https://api.github.com/repos/coreos/tectonic-installer/releases/latest',
```
kans: I see`	"API rate limit exceeded ...)"` when I curl this URL.  I'm guessing this will be common at the office :-/

amrutac: Yeah, just saw that too. Looking into it.

diegs: Sorry for the additional chime in here, but shouldn't we use a URL we control completely (e.g. somewhere under https://coreos.com) rather than relying on github?

amrutac: Yeah, I thought about scraping it from our docs site originally, looking into that again now that we are hitting api rate limit.

Quentin-M:
How is this doing? Would you feel comfortable removing the `-version` flag as a follow-up PR right after this (https://github.com/coreos/tectonic-installer/issues/1100)?

amrutac:
@Quentin-M: Almost done with this. Resolving one minor issue to finish this.

installer/frontend/components/header.jsx
```diff
@@ -1,9 +1,75 @@
...
+    return true;
+  }
+
+  if (semver.minor(latestRelease) > semver.minor(GIT_TAG)) {
+    return true;
+  }
+
+  if (semver.patch(latestRelease) > semver.patch(GIT_TAG)) {
+    return true;
+  }
```
ggreer: I think you can replace all three of these conditionals with `semver.gt(latestRelease, GIT_TAG);`

amrutac: It doesn't recognize <code>rc</code> as the prerelease string and returns <code>false</code> for <code>semver.gt('1.6.4-tectonic.1', '1.6.4-tectonic.1-rc.3');</code>

ggreer: Ah, nevermind then.

installer/frontend/components/header.jsx
```diff
@@ -1,9 +1,75 @@
...
 export class Header extends React.Component {
+  constructor (props) {
+    super(props);
+    this.state = { latestRelease: null };
+  }
+
+  componentDidMount() {
+    fetchLatestRelease().then((release) => {
+      this.setState({ latestRelease: parseLatestVersion(release) });
+    }).catch(() => this.setState({ latestRelease: null }));
```
ggreer: Probably makes sense to log the caught error.

amrutac:
Done!

PR #1065, opened 2017-06-12
tf/all: use regex for repo/tag parsing

### LINKED ISSUE #1003: incorrect docker image string parsing
Throughout the codebase, container image strings are split on colon to separate the image repo URI and the tag. 

Example:
```
kubelet_image_url = "${element(split(":", var.tectonic_container_images["hyperkube"]), 0)}"
kubelet_image_tag = "${element(split(":", var.tectonic_container_images["hyperkube"]), 1)}"
```

This becomes problematic because colon can also used as an optional separator for `hostname:port` in a container image string:

Example:

`registry.local:5000/coreos/hyperkube:v1.6.2_coreos.0`

Possible fixes include:
* use terraform `replace` function with a regexp and capture groups
* write a terraform function to only split on the last occurrence of `:` in the string

colhom:
fixes #1003


modules/aws/ignition/variables.tf
```diff
@@ -46,3 +46,12 @@ variable "tectonic_service_disabled" {
...
   type = "string"
 }
+
+variable "image_re" {
+  description = <<EOF
+(internal) Regular expression used to extract repo and tag components from image strings
+EOF
+
+  type    = "string"
+  default = "/^([^/]+/[^/]+/[^/]+):(.*)$/"
```
s-urbaniak: instead of duplicating the regex, can't we just pass `image_re` from `config.tf`?

alexsomesan:
LGTM, but we should get the tests green.
The current test failure is unrelated to this change, but is intermittent.
Restarting tests one more time.

@colhom can you make sure the PR is rebased on the latest master. We recently made some changes to mitigate that particular problem.

alexsomesan:
retest this please

colhom:
It seems ones of the alert manager containers is failing to come up after the rebase.

colhom:
@s-urbaniak @alexsomesan another rebase and tests are green.

PR #1075, opened 2017-06-13
GUI tests: Use headless Chrome

ggreer:
This makes it so we don't need to start X, avoiding any issues with multiple xvfbs running on a single worker.

### DIFF
--- a/images/tectonic-builder/Dockerfile
+++ b/images/tectonic-builder/Dockerfile
@@ -54,14 +54,14 @@ RUN cd /tmp && \
     ln -s /usr/local/yarn/bin/yarn /usr/local/bin/yarn
 
 
-### install chrome & xvfb to run the installer-gui tests.
-
+# install chrome for installer gui tests
+# using chrome beta because v60 or higher is needed for headless mode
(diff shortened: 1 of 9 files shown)


images/tectonic-builder/Dockerfile
```diff
@@ -54,14 +54,14 @@ RUN cd /tmp && \
     ln -s /usr/local/yarn/bin/yarn /usr/local/bin/yarn
 
 
-### install chrome & xvfb to run the installer-gui tests.
-
+# install chrome for installer gui tests
+# using chome beta because v60 or higher is needed for headless mode
```
amrutac: nit: <code>chrome<code>

ggreer: me spel gud

ggreer:
Even in the latest version of Chrome, we still need xvfb to simulate keypresses: https://bugs.chromium.org/p/chromedriver/issues/detail?id=1772

I'll update this PR with some `xvfb-run` stuff, so at least we won't get conflicting displays on the same builder.

ggreer:
Alright. Now this is ready.

Issue #903, opened 2017-05-26
Embed flag for GA into the automated release process


robszumski:
In dev, do not even give the option to turn on GA. Instead, turn it on as part of our release builds only. This will stop polluting our funnel data when devs aren't using dev mode.

PR #1076, opened 2017-06-13
frontend: Ensures that GA events are sent only for release tags

amrutac:
Implements https://github.com/coreos/tectonic-installer/issues/903

### DIFF
--- a/installer/frontend/.eslintrc
+++ b/installer/frontend/.eslintrc
@@ -60,6 +60,6 @@
   "globals": {
     "GIT_COMMIT": false,
     "GIT_TAG": false,
-    "GIT_LATEST_TAG": false,
+    "GIT_RELEASE_TAG": false,
   }
 }
--- a/installer/frontend/components/header.jsx
+++ b/installer/frontend/components/header.jsx
@@ -1,16 +1,9 @@
 import React from 'react';
-import semver from 'semver';
 
 import { Dropdown } from './ui';
 
 export class Header extends React.Component {
   render() {
-    let hasNewVersion = false;
-    try {
-      hasNewVersion = semver.lt(GIT_TAG, GIT_LATEST_TAG);
-    } catch (unused) {
-      // unused
-    }
     const productDdItems = {
       'Tectonic - Kubernetes': 'https://coreos.com/tectonic/',
       'Quay - Registry': 'https://coreos.com/quay-enterprise',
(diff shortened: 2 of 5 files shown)


installer/frontend/components/header.jsx
```diff
@@ -1,16 +1,9 @@
...
 import { Dropdown } from './ui';
 
 export class Header extends React.Component {
   render() {
-    let hasNewVersion = false;
-    try {
-      hasNewVersion = semver.lt(GIT_TAG, GIT_LATEST_TAG);
-    } catch (unused) {
-      // unused
-    }
```
amrutac: Removing this since it's not working as expected and I am refactoring this in https://github.com/coreos/tectonic-installer/pull/1064

installer/frontend/.eslintrc
```diff
@@ -60,6 +60,6 @@
   "globals": {
     "GIT_COMMIT": false,
     "GIT_TAG": false,
-    "GIT_LATEST_TAG": false,
+    "GIT_RELEASE_TAG": false,
```
amrutac: This gets the tag(if any) associated to the commit. If not then it returns <code>unknown</code> (https://github.com/coreos/babel-plugin-git-version-build/blob/master/lib/index.js#L107)

installer/frontend/utils.js
```diff
@@ -61,3 +62,7 @@ export const toInFly = toPath('inFly');
 export const toExtraData = toPath('extra');
 export const toExtraDataError = toPath('extraError');
 export const toExtraDataInFly = toPath('extraInFly');
+
+export const isReleaseVersion = () => {
+  return GIT_RELEASE_TAG !== 'unknown' ? semver.valid(GIT_RELEASE_TAG) && !GIT_RELEASE_TAG.includes('-rc.') : false;
```
ggreer: Could the ternary be switched and !== be changed to ===? eg:

    return GIT_RELEASE_TAG === 'unknown' ? false : semver.valid(GIT_RELEASE_TAG) && !GIT_RELEASE_TAG.includes('-rc.');

installer/frontend/tectonic-ga.js
```diff
@@ -37,7 +39,7 @@ export const TectonicGA = {
     ga('create', 'UA-42684979-10', 'none', 'TectonicInstaller');
     ga('create', 'UA-42684979-6', 'none', 'CoreOS');
 
-    if (window.config.devMode) {
+    if (!isReleaseVersion()) {
       ga('TectonicInstaller.set', 'sendHitTask', null);
```
ggreer: Do you remember what this is for? I totally forget. I don't know why we'd want to send any ga events in dev mode or non-release versions.

amrutac: This was used being used for debugging. Deleted it.

installer/frontend/utils.js
amrutac: Right you are! Done!

PR #1081, opened 2017-06-14
Azure: configure corresponding AS for VMs

alexsomesan:
This change assigns the newly introduced Availability Sets to the corresponding Virtual Machines.
Without this, the node VMs would not actually be placed in the respective AvSets.

/cc: @discordianfish


discordianfish:
Oh oops, good catch! LGTM

PR #1082, opened 2017-06-14
vmware: Add VMWare LUN support

hhoover:
There is an open issue for Terraform not supporting LUNs:
https://github.com/terraform-providers/terraform-provider-vsphere/issues/2
https://groups.google.com/d/msg/terraform-tool/rSjjZi4ojf8/5F5y0XB-CAAJ

This fix enables users to specify a LUN for each node type.


platforms/vmware/main.tf
```diff
@@ -81,6 +83,7 @@ module "workers" {
   vm_memory               = "${var.tectonic_vmware_worker_memory}"
   vm_network_label        = "${var.tectonic_vmware_network}"
   vm_disk_datastore       = "${var.tectonic_vmware_datastore}"
+  vm_lun                  = "${var.tectonic_vmware_worker_lun}"
```
alekssaul: not super opinionated but perhaps `vm_disk_lun` may flow better as a variable name?

platforms/vmware/variables.tf
```diff
@@ -120,6 +120,11 @@ variable "tectonic_vmware_etcd_gateway" {
   description = "Default Gateway IP address for etcd nodes(s)"
 }
 
+variable "tectonic_vmware_etcd_lun" {
+  type        = "string"
+  description = "The storage LUN used by etcd"
+}
+
```
alekssaul: can we set the defaults to "" for the `tectonic_vmware_etcd_lun` , `tectonic_vmware_master_lun` and `tectonic_vmware_worker_lun` ? I believe having those as defaults should allow the users to use datastore directly (if they are not part of datastore cluster.

alekssaul:
hey @hhoover , thanks for submitting this!!

Not sure why the build is failing, but as a general measure please make sure to run `make examples docs` and `make structure-check` before submitting PRs to tectonic-installer repo. `terraform fmt .` on root folder should handle the formatting issues. This has been why most of my builds fail :)

Made two comments in code that I am hoping makes sense. Also opened an issue in the repo. I am hoping if terraform gets storage cluster support, we won't need the extra variable. 

Would you mind also either adding additional documentation into the vmware-install document (https://github.com/coreos/tectonic-installer/blob/master/Documentation/install/vmware/vmware-terraform.md) or perhaps in the variable descriptions. A good example of notes in variables is; https://github.com/coreos/tectonic-installer/blob/master/platforms/metal/variables.tf#L1-L8 .

modules/vmware/etcd/nodes.tf
```diff
@@ -13,7 +13,7 @@ resource "vsphere_virtual_machine" "etcd_node" {
   }
 
   disk {
-    datastore = "${var.vm_disk_datastore}"
+    datastore = "${var.vm_disk_datastore}/${var.vm_disk_lun[count.index]}"
```
alekssaul: var.vm_disk_lun seems to be a string. [count.index] requires it to be a map. Should remove [count.index]

modules/vmware/node/nodes.tf
```diff
@@ -13,7 +13,7 @@ resource "vsphere_virtual_machine" "node" {
   }
 
   disk {
-    datastore = "${var.vm_disk_datastore}"
+    datastore = "${var.vm_disk_datastore}/${var.vm_disk_lun[count.index]}"
```
alekssaul: var.vm_disk_lun seems to be a string. [count.index] requires it to be a map. Should remove [count.index]

alekssaul:
Hey @hhoover,

Made two comment on the code.

We already have "tectonic_vmware_datastore" variable that can potentially be set to something in lines of "StorageCluster/Datastore" as a value rather than introduce new variables. 
ex: for variables
```
tectonic_vmware_datastore = storagecluster01
tectonic_vmware_worker_lun = datastore01
```
```
tectonic_vmware_datastore/tectonic_vmware_worker_lun = storagecluster01/datastore01
```
whereas setting tectonic_vmware_datastore directly 
```
tectonic_vmware_datastore = storagecluster01/datastore01
```
to my best understanding can produce the same results.

Your PR introduces ability to deploy etcd, master, worker nodes to individual LUNs, which is valuable. I've toyed with this idea and sent a PR to your branch to demonstrate what I think can be a good approach, please let me know your thoughts.

hhoover:
@alekssaul Thanks for the additions! I like your way better. Merged to my branch, waiting on checks here.

alekssaul:
Tested my PR with deploying to a $datastorecluster/$datastore in my lab. Once provisioned, I've migrated machines from one datastore to another withing the datastorecluster ex: deployed to `dscluster/datastore01` and migrated the machine to `dscluster/datastore02`. Hit `terraform plan` after and terraform said nothing needed to be changed (good), and also I was able to `terraform destroy` as expected.

So it seems like the only thing that's missing is the ability to directly target storage cluster with initial provisioning, but lifecycle functionality (VMs moving within the LUNs in the datastore cluster) shouldn't have a negative impact. Eventually, we may want to fix it upstream in terraform-provider-vsphere cod, so we can directly target it without specifying the individual datastore within the cluster.

s-urbaniak:
@hhoover Also from my side thanks a lot for the contribution! do you mind to rebase on master and address the final comments from @alekssaul?

metral:
Any updates on this PR?

PR #920, opened 2017-05-27
Test with long cluster names (28 chars)

ggreer:
Commit on top of #914.

Once this test passes, https://github.com/coreos/tectonic-installer/issues/413 will be fixed.

PR #1083, opened 2017-06-14
ELB Name is CLUSTER-api-int

sstarcher:
The ELB name was changed in #920 from `CLUSTER-api-internal` to `CLUSTER-int`
In addition the tag Name is set for the AWS ELB, but is not the same as the actual ELB name.  In the AWS console `CLUSTER-int` shows up on the right side which is the actual ELB name, but in the tags for the ELB `Name=CLUSTER-api-internal` is listed which may cause confusion.

### DIFF
--- a/modules/aws/ignition/resources/detect-master.sh
+++ b/modules/aws/ignition/resources/detect-master.sh
@@ -23,7 +23,7 @@ while true; do
   sleep 15
 done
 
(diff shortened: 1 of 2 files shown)


Quentin-M:
ok to test

Quentin-M:
ok to test

PR #1091, opened 2017-06-14
Move sanity tests

squat:
This PR move the "sanity" tests to the smoke test directory, since that is what they really are. This is just a first step as we refactor the project's testing infrastructure, add integration tests, and automate the QA checklist.

cc @kans @ggreer


kans:
maybe you broke the vendoring?

```k8s_test.go:11:2: cannot find package "k8s.io/client-go/kubernetes" ```

squat:
figured it out.. it's an unfortunate combination of issues between go not letting you use vendor directories outside of a $GOPATH and Make resolving the symlink of the current directory so that it is no longer technically in a $GOPATH. the fix is to not explicitly use $TOP_DIR, just like before.

rithujohn191:
would stuff on Jenkins need to be updated too? :thinking:

squat:
This does update all the jenkins stuff. PTAL at the Jenkinsfile changes

Quentin-M:
Oops it got merged while I was reading. This opens the way to re-implementing our bash scripts in Go, looks good, thanks! 👍

PR #1098, opened 2017-06-15
tests/smoke/aws/vars: reduce test node count

squat:
This commit reduces the node count for the private vpc test in order to
accomplish two goals:
1. smoke test bringing up a single master cluster; and
2. reduce total resource utilization.

This test should give us greater confidence in the installer since it
tests a new configuration.

cc @Quentin-M @kans


Quentin-M:
Thanks for the cc. Will update our matrix.

PR #999, opened 2017-06-06
Add support for cloud provider config on azure

discordianfish:
This is WIP, I just wanted to share this early for feedback.

This closes #84 and should be fairly self-explanatory. One change in behaviour is that now you have to specify the provider config as tf variables instead of env variables so we can render the cloud provider config which include these.

I'm also not sure if the cloud config should be a config map or secret. It contains secrets but it's called cloud config, so dunno.. Probably better to use a secret.

Will update PR once I've tested everything in Azure.

PR #1101, opened 2017-06-16
Enforce Real HA of critical K8's components

Keidrych:
Ensure that critical control-plane scheduling always occurs in multiple nodes & AZ's where possible for cluster disaster recovery.

Old scheduling rules would allow all pods to run on the same host bringing down the cluster with extended AZ unavailability or Host Corruption

### DIFF
--- a/modules/bootkube/resources/manifests/kube-controller-manager.yaml
+++ b/modules/bootkube/resources/manifests/kube-controller-manager.yaml
@@ -21,6 +21,9 @@ spec:
           preferredDuringSchedulingIgnoredDuringExecution:
           - weight: 100
             podAffinityTerm:
+              topologyKey: failure-domain.beta.kubernetes.io/zone
+          - weight: 50
+            podAffinityTerm:
               labelSelector:
                 matchExpressions:
                 - key: tier
@@ -31,7 +34,6 @@ spec:
                   operator: In
                   values:
                   - kube-contoller-manager
-              topologyKey: kubernetes.io/hostname
       containers:
       - name: kube-controller-manager
         image: ${hyperkube_image}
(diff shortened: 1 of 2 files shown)


robszumski:
Thanks for the PR! I'd like to verify that this label is set on all of our platforms, or confirm the behavior if it isn't before merging.

alekssaul:
re:platforms/vmware testing - Ran against Keidrych:master with
```
tectonic_master_count = 3
tectonic_worker_count = 2
tectonic_experimental = true
```
`kubectl get pods --all-namespaces -o wide | grep -E 'kube-scheduler|kube-controller'`
```
kube-system       kube-controller-manager-1453561898-9tsr8          1/1       Running            0          9m        10.2.0.6       master0
kube-system       kube-controller-manager-1453561898-db5kl          1/1       Running            0          9m        10.2.0.5       master0
kube-system       kube-scheduler-551508370-8c10c                    1/1       Running            0          9m        10.2.0.4       master0
kube-system       kube-scheduler-551508370-xw68f                    1/1       Running            0          9m        10.2.0.3       master0
```
cluster builds as expected, however scheduler and controllers may still get scheduled on the same node. This PR may be valuable for AWS/GCE platforms, without impact to others.

Keidrych:
That's correct, it was designed for cloud based deployments, not bare metal as the topologykey for Region would need to be specified in a Bare Metal Install for HA to be implemented.

Quentin-M:
We should be extremely careful when doing changes to these manifests without having a corresponding change to https://github.com/kubernetes-incubator/bootkube (cc @aaronlevy).

cehoffman:
This field does exist in Azure when using availability sets. It does require the cloud provider configuration integration from #999

cpanato:
@Keidrych Thanks for your PR, can you please rebase, then we can continue the reviewing process?

cpanato:
closing due to inactivity. Please open another PR if you want to get this in.

PR #1102, opened 2017-06-16
modules/tectonic/resources: move/create tectonic-monitoring-config.yaml

Quentin-M:
This manifest is not created, nor is it in the right folder apparently. This PR moves it and creates it. Should we drop it instead?

@brancz


brancz:
It is actually optional, but yes I intended it to be created.

PR #1093, opened 2017-06-15
monitoring: update to latest releases

brancz:
The base64 blob is a small change in the default Alertmanager config.

Other than that this just bumps Prometheus/Alertmanager/Prometheus Operator to their latest releases.

@s-urbaniak @alexsomesan

PR #1105, opened 2017-06-16
Monitoring fixes

brancz:
This PR

- fixes the Alertmanger config as reported in https://github.com/coreos/tectonic-installer/pull/1093#issuecomment-308902332
- moves the tectonic-monitoring config back to `updater/`, as the configuration is _only_ used by the Tectonic Prometheus Operator. If a user is not installing self-driving it is never used (I didn't pay enough attention to https://github.com/coreos/tectonic-installer/pull/1102 and missed that the file was moved)
- fixes a problem of Tectonic Monitoring 1.3, where the Prometheus Operator was not properly added as a scrape target

I just brought up a cluster with this, and it was working. The Tectonic Prometheus Operator 1.3.0 release was adapted appropriately with these fixes.

Sorry for the trouble.

@alexsomesan @s-urbaniak @Quentin-M


Quentin-M:
Thanks a lot for the follow-up!

PR #1106, opened 2017-06-16
tectonic: bump ingress replicas 1 -> 2

alekssaul:
Increase the size of replicas for ingress controller (nodeport) and ingress default-backend from 1 to 2

### DIFF
--- a/modules/tectonic/resources/manifests/console/deployment.yaml
+++ b/modules/tectonic/resources/manifests/console/deployment.yaml
@@ -25,6 +25,22 @@ spec:
         component: ui
       name: tectonic-console
     spec:
+      affinity:
+        podAntiAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+          - weight: 100
+            podAffinityTerm:
+              labelSelector:
+                matchExpressions:
+                - key: component
+                  operator: In
+                  values:
+                  - ui
+                - key: k8s-app
+                  operator: In
+                  values:
+                  - tectonic-console
+              topologyKey: kubernetes.io/hostname
       containers:
       - command:
         - /opt/bridge/bin/bridge
--- a/modules/tectonic/resources/manifests/console/deployment.yaml
+++ b/modules/tectonic/resources/manifests/console/deployment.yaml
@@ -27,7 +27,7 @@ spec:
     spec:
       affinity:
         podAntiAffinity:
-          preferredDuringSchedulingIgnoredDuringExecution:
+          requiredDuringSchedulingIgnoredDuringExecution:
           - weight: 100
             podAffinityTerm:
               labelSelector:
--- a/modules/tectonic/resources/manifests/identity/deployment.yaml
+++ b/modules/tectonic/resources/manifests/identity/deployment.yaml
@@ -25,7 +25,7 @@ spec:
     spec:
       affinity:
         podAntiAffinity:
-          preferredDuringSchedulingIgnoredDuringExecution:
+          requiredDuringSchedulingIgnoredDuringExecution:
           - weight: 100
             podAffinityTerm:
               labelSelector:
--- a/modules/tectonic/resources/manifests/identity/deployment.yaml
+++ b/modules/tectonic/resources/manifests/identity/deployment.yaml
@@ -23,6 +23,22 @@ spec:
         k8s-app: tectonic-identity
         component: identity
     spec:
+      affinity:
+        podAntiAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+          - weight: 100
+            podAffinityTerm:
+              labelSelector:
+                matchExpressions:
+                - key: component
+                  operator: In
+                  values:
+                  - identity
+                - key: k8s-app
+                  operator: In
+                  values:
+                  - tectonic-identity
+              topologyKey: kubernetes.io/hostname
       volumes:
       - name: config
         configMap:
--- a/modules/tectonic/resources/manifests/ingress/default-backend/deployment.yaml
+++ b/modules/tectonic/resources/manifests/ingress/default-backend/deployment.yaml
@@ -4,7 +4,7 @@ metadata:
   name: default-http-backend
   namespace: tectonic-system
 spec:
-  replicas: 1
+  replicas: 2
   template:
     metadata:
       labels:
--- a/modules/tectonic/resources/manifests/ingress/default-backend/deployment.yaml
+++ b/modules/tectonic/resources/manifests/ingress/default-backend/deployment.yaml
@@ -10,6 +10,18 @@ spec:
       labels:
         k8s-app: default-http-backend
     spec:
+      affinity:
+        podAntiAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+          - weight: 100
+            podAffinityTerm:
+              labelSelector:
+                matchExpressions:
+                - key: k8s-app
+                  operator: In
+                  values:
+                  - default-http-backend
+              topologyKey: kubernetes.io/hostname
       terminationGracePeriodSeconds: 60
       containers:
       - name: default-http-backend
--- a/modules/tectonic/resources/manifests/ingress/default-backend/deployment.yaml
+++ b/modules/tectonic/resources/manifests/ingress/default-backend/deployment.yaml
@@ -12,7 +12,7 @@ spec:
     spec:
       affinity:
         podAntiAffinity:
-          preferredDuringSchedulingIgnoredDuringExecution:
+          requiredDuringSchedulingIgnoredDuringExecution:
           - weight: 100
             podAffinityTerm:
               labelSelector:
--- a/modules/tectonic/resources/manifests/ingress/nodeport/deployment.yaml
+++ b/modules/tectonic/resources/manifests/ingress/nodeport/deployment.yaml
@@ -8,7 +8,7 @@ metadata:
     component: ingress-controller
     type: nginx
 spec:
-  replicas: 1
+  replicas: 2
   template:
     metadata:
       labels:
--- a/modules/tectonic/resources/manifests/ingress/nodeport/deployment.yaml
+++ b/modules/tectonic/resources/manifests/ingress/nodeport/deployment.yaml
@@ -18,7 +18,7 @@ spec:
     spec:
       affinity:
         podAntiAffinity:
-          preferredDuringSchedulingIgnoredDuringExecution:
+          requiredDuringSchedulingIgnoredDuringExecution:
           - weight: 100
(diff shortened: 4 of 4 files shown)


kans:
Would you provide some context on this change?

alekssaul:
Hi @kans ,

Apologies for the lack of detail. Currently, we have single replica for ingress controller and backend. If the node running the ingress controller dies (or during updates when kube-proxy is replaced), Tectonic-console and identity connectivity would be impacted eventhough we have 2 copies of tectonic-console and identity running.

Quentin-M:
Sounds legit. Any thoughts @alexsomesan @s-urbaniak?

Quentin-M:
Without hard requirement with anti-affinity tho, it is extremely likely that these two replicas will get scheduled on the same node during bootstrapping, when a single node is there.

alekssaul:
This is also the case for tectonic-console and tectonic-identity deployments. I can take a stab adding affinity rules in?

alekssaul:
Added Affinity spec similar to
```
      affinity: 
        podAntiAffinity: 
          preferredDuringSchedulingIgnoredDuringExecution: 
          - weight: 100 
            podAffinityTerm: 
              labelSelector: 
                matchExpressions: 
                - key: component 
                  operator: In 
                  values: 
                  - ui 
                - key: k8s-app 
                  operator: In 
                  values: 
                  - tectonic-console 
              topologyKey: kubernetes.io/hostname 
```
for console, identity, http-backend, tectonic-ingress for testing. Seem to work out ok, would it be appropriate the PR with it?
```
kubectl get pods --all-namespaces -o wide | grep console
tectonic-system   tectonic-console-3000036105-8s2xf                                  1/1       Running            1          21m       10.2.2.2      ip-10-0-73-121.us-east-2.compute.internal
tectonic-system   tectonic-console-3000036105-chgp2                                  1/1       Running            1          21m       10.2.1.4      ip-10-0-92-151.us-east-2.compute.internal
alekssaul@Alekss-MacBook-Pro ~/d/g/a/t/b/aws> kubectl get pods --all-namespaces -o wide | grep ingress
tectonic-system   tectonic-ingress-controller-1922552424-3234p                       1/1       Running            0          21m       10.2.1.6      ip-10-0-92-151.us-east-2.compute.internal
tectonic-system   tectonic-ingress-controller-1922552424-cw486                       1/1       Running            0          21m       10.2.2.6      ip-10-0-73-121.us-east-2.compute.internal
alekssaul@Alekss-MacBook-Pro ~/d/g/a/t/b/aws> kubectl get pods --all-namespaces -o wide | grep identity
tectonic-system   tectonic-identity-2587106793-29930                                 1/1       Running            0          22m       10.2.2.3      ip-10-0-73-121.us-east-2.compute.internal
tectonic-system   tectonic-identity-2587106793-tj8hd                                 1/1       Running            0          22m       10.2.1.3      ip-10-0-92-151.us-east-2.compute.internal
alekssaul@Alekss-MacBook-Pro ~/d/g/a/t/b/aws> kubectl get pods --all-namespaces -o wide | grep back
tectonic-system   default-http-backend-1320719044-xdmdb                              1/1       Running            0          21m       10.2.2.7      ip-10-0-73-121.us-east-2.compute.internal
tectonic-system   default-http-backend-1320719044-zvmtf                              1/1       Running            0          21m       10.2.1.7      ip-10-0-92-151.us-east-2.compute.internal

```

alekssaul:
Also, scheduler seem to spread console/identity even without pod-affinity spec. I am guessing scheduler by default will try to spread the same workload to different nodes within the cluster.

Quentin-M:
If it is simply up to me, I'd say go for it. `preferredDuringSchedulingIgnoredDuringExecution` won't hurt!

alekssaul:
went for it

alekssaul:
@coresolve

colhom:
This functionality is sorely needed. I suggest: 
* Change the anti-affinity to `requiredDuringSchedulingIgnoredDuringExecution`. There's not really a good reason to double up on ingress controller pods per host- the goal here is availability.
* Setting default number of replicas to `min( count(worker nodes), 3)`

Users are having ingress ELB alarms go off when the CL nodes update- as ingress service regularly experiences complete service outages as a result of nodes rebooting. 

\cc @Quentin-M @alekssaul

robszumski:
Bump on this

alekssaul:
Colin's recommendation makes sense to me. I can attempt to revise the change(s) if the team agrees. There has been a discussion by @joshrosso and @aaronlevy in https://github.com/coreos/tectonic-installer/pull/1616 that seem to be similar in nature.

aaronlevy:
If this merges it either needs an upgrade path, or it needs to be tracked as skew that has been introduced and what version it was introduced in. @robszumski @sym3tri 

I am unsure if this will affect how ingress is handled because it is sometimes a daemonset and sometimes a deployment. /cc @squat

aaronlevy:
Further, if you're setting anti-affinity on a node-basis .. you should probably just use daemonsets.

colhom:
I could also seen daemonsets being the right answer here as well- my main reservation is locking the number of ingress pods to the number of worker nodes. In clusters w/ large numbers of workers this could be seen as undesirable, as there are diminishing returns w.r.t availability as the replica count grows

alekssaul:
>  In clusters w/ large numbers of workers this could be seen as undesirable
I agree with this.
Committed change in Pod Affinity to `requiredDuringSchedulingIgnoredDuringExecution`. 

@colhom wrt `min( count(worker nodes), 3)` any strong objection to keeping replicate count to 2 as a default?

aaronlevy:
Ingress controllers are a total mess right now and has essentially become unmanageable.

- If you're on bare-metal it's a daemonset + host-port service
- If you're on AWS it's a deployment + node-port service (no idea about azure)
- The monitoring team needs to make some changes (not aware of details, cc @pst).

A brief look at the last time we updated the manifest we did a wholesale replacement of the manifest - which means the entire thing is going to get reset with whatever ships in the update anyway (so if you're using a deployment your replica count doesn't matter). This won't always be the case, but only other option is a bunch of custom code right now for likely next release or two.

Basically - this component can't be safely updated or changed unless someone has time to unpack all the various deployment methods that have been used, and come up with a plan on how we can move forward. From my end we're not going to have generic ways of handling this for a few releases.

So your options are essentially:

- Make your replica count + affinity changes in the installer only, and those changes might get overridden in an update (so I'd prob stick with static value for now for replicas). If you go this route - @robszumski please track the additional skew that is being introduced and the version it is introduced in.

- Someone spends the time to unpack all this and come up with a plan to get everything back into a single deployment method that can be reasonably managed - and that is implemented in a future release.

- Wait until we can bring all of this back under sane management and the changes can go out in a normal update.

cc:

@dghubble - as I believe you know most about why we added different ingress deployment methods for baremetal.

@squat I believe you dealt with this in an update last time

@pst - as this brings up an earlier discussed issue about updating ingress for monitoring - but that's going to be a big pain, because it's not consistently deployed, so whoever that lands on might have to spend some time on this.

@diegs to keep aware of some more craziness we need to support

@mike-saparov to keep you up to date on the debt

dghubble:
I don't see creating a single deployment/service for Ingress controllers across all clouds as a viable option. Currently AWS and Azure use the "NodePort" setup. Bare-Metal, Openstack, VMWare, and Sandbox use "HostPort". You'll find when you get to Google Cloud and Digital ocean clusters, you'll find a similar distinction. Its not a one-solution-fits-all kinda situation. See https://gist.github.com/dghubble/24b3def990791a27ee286a54f5b99e39

If you absolutely must have the same architecture regardless of platform, HostPort is the lowest common denominator you could standardize on. There are drawbacks to this and I wouldn't recommend it.

Over time, I think the need for different clusters to have different manifests (and different update payloads) will grow. Unifying the setup here kicks the can down the road, but this broader issue will resurface.

aaronlevy:
If it really is provider specific, this makes things like multi-provider clusters (some bare-metal nodes, some AWS) another knob to keep in mind. 

Mid-term target would be install-time decision tracking for update handling (e.g. "this cluster uses ingress flavor-A"). I don't think we can reasonably tie something like "provider=aws" to "use deployment for ingress" as that could reasonably change over time. Instead I guess, we just ship multiple flavors of this component. Unfortunately, we don't have support for that right now - so we're left to not-so-great runtime heuristics to try and figure this out.

dghubble:
> I don't think we can reasonably tie something like "provider=aws" to "use deployment for ingress" as that could reasonably change over time

Agreed. Even now, for bare-metal one of the first things I do is switch from a HostPort style to a ClusterIP style, but it requires control over the routing layer. So I definitely see a platform offering multiple options one day.

alexsomesan:
I just ran into this as a blocker while working private clusters for Azure.

Essentially, we can't use Azure LBs in the private setting because of an upstream limitation of Azure (lack of SNAT on private LBs).
Because of that, we need to reference API and Console by multi-entry A records in DNS.

However, for console, that means the ingress needs to run on all masters, otherwise users will experience intermittent "Connection refused" as the console FQDN gets round-robin resolved to one master or the other.

My vote goes to daemon-sets, which seem to work well in all scenarios and seem like the simplest to manage, form the manifest perspective (no templating of any node count, etc).

I'd love to get consensus on this soonish to unblock the Azure private work.

Just my two cents.

dghubble:
@alexsomesan interesting, I haven't used Azure. Isn't it DNAT that's the issue? Also, when you say you favor the damonset, are you arguing for exposing it via hostport? I suspect not bc you said you're running them on controllers which wouldn't work with hostport due to the port collision.

I originally called them "nodeport" and "hostport" styles to focus on the way they're exposed, bc the Daemonset vs Deployment distinction can depend just on whether the LB offering has health checks or whether you want it on every worker.

alexsomesan:
@dghubble it's SNAT that is actually the problem, as described in the "Limitations" section of [this doc](https://github.com/MicrosoftDocs/azure-docs/blob/master/articles/load-balancer/load-balancer-internal-overview.md) and few other places.

I'm arguing for running the ingress with a daemonset on a port other than 443 (obviously taken by the API). Currently the NodePort deployment uses 32000 & 32001 for this (part of the service range, so should change).
This would in all cases but metal and Azure private be fronted with an LB that would expose the console further under standard HTTPS and it's own VIP. For Azure private we don't have too many options other than keeping it on a different port (or complicating things further with some sort of proxy than we'd have to manage).

sym3tri:
It seems the consensus is on DaemonSet and there a number of other things that need to be done in order to complete this. I've documented everything in a new issue:

https://coreosdev.atlassian.net/browse/INST-146

dghubble:
As long as you understand the tradeofffs

PR #1108, opened 2017-06-16
There is a mistake in the kubectl delete line.

dorongrinstein:
I am correcting the instruction to delete to newly created service. It was kubectl delete deploy/simple-service svc/simple-service but it needs to be kubectl delete deploy/simple-deployment svc/simple-service


robszumski:
ok to test

kans:
Tests failed because we hit a resource limit on AWS.

Issue #904, opened 2017-05-26
installer: submit GA events per page


robszumski:
Feature request:

Right now platforms share some URLs and not others, which makes it hard to construct a funnel between pages. Let's add an event on each "next" click with:

category: metal, aws, etc
label: next on `pagetitle`
action: click

robszumski:
Going to re-open this, I had added `next on <pagetitle>` but Github stripped out the page title part because it looked like an HTML tag. Updated above.

Can we update to send the page title?

Quentin-M:
Addressed by #1110!

PR #1110, opened 2017-06-16
frontend: Updates GA events to have send value as 'platformType version'

amrutac:


amrutac:
Based on a discussion with @robszumski and addresses https://github.com/coreos/tectonic-installer/issues/904

installer/frontend/components/base.jsx
```diff
@@ -16,11 +16,12 @@ import { Header } from './header';
 import { Footer } from './footer';
 
 const downloadState = (state) => {
+  console.log('state', state);
   const toSave = savable(state);
   const saved = JSON.stringify(toSave, null, 2);
   const stateBlob = new Blob([saved], {type: 'application/json'});
   saveAs(stateBlob, 'tectonic.progress');
-  TectonicGA.sendEvent('Installer Link', 'click', 'User downloads progress file');
+  TectonicGA.sendEvent('Installer Link', 'click', 'User downloads progress file', state.clusterConfig.platformType);
```
kans: This should use PLATFORM_TYPE.

installer/frontend/components/success.jsx
```diff
@@ -14,11 +14,11 @@ const stateToProps = ({cluster, clusterConfig}) => {
     const domain = _.get(clusterConfig, ['extra', AWS_HOSTED_ZONE_ID, 'zoneToName', hostedZoneID]);
     tectonicConsole = clusterConfig[CLUSTER_SUBDOMAIN] + (clusterConfig[CLUSTER_SUBDOMAIN].endsWith('.') ? '' : '.') + domain;
   }
-  return {tectonicConsole};
+  return {tectonicConsole, platformType: clusterConfig.platformType};
```
kans: This guy should also use the `const`.

installer/frontend/components/base.jsx
```diff
@@ -16,11 +16,12 @@ import { Header } from './header';
 import { Footer } from './footer';
 
 const downloadState = (state) => {
+  console.log('state', state);
```
kans: Some debugging code?

kans:
This should pass tests if you rebase on master.

Quentin-M:
Are we good here? 👍

installer/frontend/tectonic-ga.js
```diff
@@ -44,8 +44,13 @@ export const TectonicGA = {
...
   },
 
-  sendEvent: (category, action, label, value) => {
-    send({ type: 'event', category, action, label, value});
+  sendEvent: (category, action, label, value = "") => {
+    send({
+      type: 'event',
+      category, action,
+      label,
+      value: `${value} ${GIT_TAG}`,
```
ggreer: Is this what we actually want? It seems like it'd make it harder to carve-up things in GA. I'm no expert though.

robszumski: @ggreer the tool where we consume this can do fancy string manipulation and regex searches so we can break this out. The problem is that we only have a few things that we report back, so we have to stuff both of these in here.

Quentin-M:
Thanks!

PR #1123, opened 2017-06-19
platforms: add vagrant platform

AlexNPavel:
Container Linux version 1451 and above will support ignition
for vagrant+virtualbox. This will add initial support for it via a
Vagrantfile that can bring up tectonic given the right ignition
configs.

This obviously still requires a lot of work, but vagrant+virtualbox support would be a very nice feature for tectonic, especially for demos. This Vagrantfile is what we will be using for this new platform. The ignition configs need to be written to config-controller.ign and config-worker.ign before the machine is started. For sshing in, the vagrant insecure key will be used for now (and there are multiple ways to do that; for instance, vagrant supports ssh based provisioning by adding a few lines to the Vagrantfile).

This requires Container Linux version 1451, which is the next alpha and has not been released as of the time of writing this PR. It also requires the vagrant-ignition plugin. Both will be released at the end of the week. If any Coreos want to start working or testing this, you can stop by the OS office and I can provide you with a testing box with the new ignition support and the required plugin.

### DIFF
--- a/platforms/vagrant/Vagrantfile
+++ b/platforms/vagrant/Vagrantfile
@@ -1,0 +1,116 @@
+# -*- mode: ruby -*-
+# # vi: set ft=ruby :
+
+require 'fileutils'
+require 'open-uri'
+require 'tempfile'
+require 'yaml'
+
+Vagrant.require_version ">= 1.6.0"
+
+# Make sure the vagrant-ignition plugin is installed
+required_plugins = %w(vagrant-ignition)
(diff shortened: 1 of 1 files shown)


alexsomesan:
This looks like an early version of what eventually became the Sandbox project.
However, integrating the sandbox into the the installer is still a topic of interest and should be further explored when capacity allows.

I'm going to close this now as it's not relevant anymore in this form.

Issue #1113, opened 2017-06-17
TLS support for self hosted etcd cluster


xiang90:
etcd operator already supports TLS for self hosted cluster. And we are finishing up the TLS support for bootkube: https://github.com/kubernetes-incubator/bootkube/pull/592.

Since tectonic installer does not use bootkube render directly, we still need to make changes in tectonic installer to support TLS.

/cc @hongchaodeng @robszumski @s-urbaniak

hongchaodeng:
ref https://github.com/kubernetes-incubator/bootkube/issues/582
The render code are in step 1, 2

Quentin-M:
Assigning this to you @s-urbaniak as you've been working on it for the statically provisioned etcd. Tell me if you'd like someone else to do it :)

Quentin-M:
Tentatively adding release blocker. Will see how it goes.

PR #1066, opened 2017-06-13
installer/tests/sanity: add a test that ensure all manifests are created

Quentin-M:
This PR makes the sanity tests verify that all the resources defined by the generated manifests are created and present on the provisioned clusters. This aims at avoiding situations where we thought we were deploying Prometheus, while we were not due to a bug.

The PR is huge because we *had* to update the vendors. They were from mid-2016 and because we are now using a `kubectl` command, conflicts arise as the different packages in Kubernetes were not using the same client. Now, everything uses `client-go`. **You want to look at the third commit.** If you prefer, I can make a PR that is only doing the vendor bump, wait for it to be merged, and rebase.

Fixes https://github.com/coreos/tectonic-installer/issues/1020
Needs to be updated after https://github.com/coreos/tectonic-installer/issues/1037 (set of experimental manifests)

**How to test?**
```
export MANIFEST_EXPERIMENTAL=false
export MANIFEST_PATHS=build/<CLUSTER>/generated/
export TEST_KUBECONFIG=build/<CLUSTER>/generated/auth/kubeconfig

go test -v github.com/coreos/tectonic-installer/tests/smoke -cluster
```

You can toggle the `MANIFEST_EXPERIMENTAL` flag to generate a failure depending on whether you deployed a clluster with experimental features or not. You may also create extra or wrong manifests in the generated manifest folder and see the test complaining because the manifest is not there, or cannot be parsed, respectively.

PR #1124, opened 2017-06-19
*: add manifests for tectonic-etcd-operator and tectonic-etcd appversion

hasbro17:
This PR adds the manifests for tectonic-etcd-operator(TEO) `v0.0.1` and the appversion `tectonic-etcd` to the installer.

The TEO will reconcile the versions for its components, etcd-operator and the etcd-cluster, according to the version map maintained by the TEO. 
For the starting TEO version `v0.0.1` these will be etcd-operator(`v0.3.2`) and etcd-cluster(`v3.1.7`). Both these versions have been updated in config.tf.

@yifan-gu @sym3tri 

/cc @xiang90 @hongchaodeng

### DIFF
--- a/config.tf
+++ b/config.tf
@@ -44,10 +44,11 @@ variable "tectonic_container_images" {
     kubedns_sidecar                 = "gcr.io/google_containers/k8s-dns-sidecar-amd64:1.14.1"
     flannel                         = "quay.io/coreos/flannel:v0.7.1-amd64"
     etcd                            = "quay.io/coreos/etcd:v3.1.6"
-    etcd_operator                   = "quay.io/coreos/etcd-operator:v0.3.0"
+    etcd_operator                   = "quay.io/coreos/etcd-operator:v0.3.2"
     kenc                            = "quay.io/coreos/kenc:48b6feceeee56c657ea9263f47b6ea091e8d3035"
     awscli                          = "quay.io/coreos/awscli:025a357f05242fdad6a81e8a6b520098aa65a600"
     kube_version                    = "quay.io/coreos/kube-version:0.1.0"
+    tectonic_etcd_operator          = "quay.io/coreos/tectonic-etcd-operator:v0.0.1"
   }
 }
 
@@ -56,12 +57,13 @@ variable "tectonic_versions" {
   type        = "map"
 
   default = {
-    etcd         = "3.1.6"
-    prometheus   = "v1.7.1"
-    alertmanager = "v0.7.1"
-    monitoring   = "1.3.0"
-    kubernetes   = "1.6.4+tectonic.1"
-    tectonic     = "1.6.4-tectonic.1"
+    etcd                   = "3.1.7"
+    prometheus             = "v1.7.1"
+    alertmanager           = "v0.7.1"
+    monitoring             = "1.3.0"
+    kubernetes             = "1.6.4+tectonic.1"
+    tectonic               = "1.6.4-tectonic.1"
+    tectonic_etcd_operator = "0.0.1"
   }
 }
 
--- a/modules/tectonic/resources/manifests/updater/app-version-tectonic-etcd.yaml
+++ b/modules/tectonic/resources/manifests/updater/app-version-tectonic-etcd.yaml
@@ -1,0 +1,13 @@
+apiVersion: coreos.com/v1
+kind: AppVersion
+metadata:
+  name: tectonic-etcd
+  namespace: tectonic-system
+  labels:
+    managed-by-channel-operator: "true"
+spec:
+  desiredVersion: ${tectonic_etcd_operator_version}
+  paused: false
+status:
+  currentVersion: ${tectonic_etcd_operator_version}
+  paused: false
--- a/modules/tectonic/resources/manifests/updater/tectonic-etcd-operator.yaml
+++ b/modules/tectonic/resources/manifests/updater/tectonic-etcd-operator.yaml
@@ -1,0 +1,36 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: tectonic-etcd-operator
+  namespace: tectonic-system
+  labels:
+    k8s-app: tectonic-etcd-operator
+    managed-by-channel-operator: "true"
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      k8s-app: tectonic-etcd-operator
+  template:
+    metadata:
+      labels:
+        k8s-app: tectonic-etcd-operator
+        tectonic-app-version-name: tectonic-etcd
+    spec:
+      containers:
+      - image: ${tectonic_etcd_operator_image}
+        name: tectonic-etcd-operator
+        command: ["/usr/local/bin/tectonic-etcd-operator"]
+        resources:
+          limits:
+            cpu: 100m
+            memory: 128Mi
+          requests:
+            cpu: 20m
+            memory: 50Mi
+      restartPolicy: Always
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 65534
+      imagePullSecrets:
+      - name: coreos-pull-secret
--- a/modules/tectonic/resources/tectonic.sh
+++ b/modules/tectonic/resources/tectonic.sh
@@ -186,12 +186,14 @@ kubectl create -f updater/kube-version-operator.yaml
 kubectl create -f updater/tectonic-channel-operator.yaml
 kubectl create -f updater/tectonic-monitoring-config.yaml
 kubectl create -f updater/tectonic-prometheus-operator.yaml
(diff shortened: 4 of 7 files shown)


config.tf
```diff
@@ -56,12 +57,12 @@ variable "tectonic_versions" {
   type        = "map"
 
   default = {
-    etcd         = "3.1.6"
-    prometheus   = "v1.7.1"
-    alertmanager = "v0.7.1"
```
yifan-gu: why this gets removed?

yifan-gu:
LGTM except one question.
@hasbro17 Have you tried `make payload`, does it generate the correct payload?

config.tf
hasbro17: Yeah I think I messed up the rebase from master. Let me fix this.

hasbro17:
@yifan-gu The payload seems to be generated fine once the changes from #1125 are made.
https://gist.github.com/hasbro17/c223c38559f1a63fb109745333b653c5

config.tf
```diff
@@ -44,10 +44,11 @@ variable "tectonic_container_images" {
     kubedns_sidecar                 = "gcr.io/google_containers/k8s-dns-sidecar-amd64:1.14.1"
     flannel                         = "quay.io/coreos/flannel:v0.7.1-amd64"
     etcd                            = "quay.io/coreos/etcd:v3.1.6"
-    etcd_operator                   = "quay.io/coreos/etcd-operator:v0.3.0"
+    etcd_operator                   = "quay.io/coreos/etcd-operator:v0.3.2"
```
xiang90: @hasbro17 ideally this should be matched the version teo (line51) currently manages. or we have to keep them in-sync manually every time.

but it should not block this pr. just something we should keep in mind.

Quentin-M:
Will this be compatible out of the box with https://github.com/coreos/tectonic-installer/issues/1113?

config.tf
hasbro17: Yeah currently this would have to be done manually. I'll try to think of a way to match the etcd-operator tag with the one that the TEO expects for its current mapping.

hasbro17:
@Quentin-M enabling TLS for the cluster should have no effect on the TEO. The TEO updates the etcd-operator directly via the deployment spec, and the etcd-cluster via a client that only modifies the TPR. The TEO does not talk to the operator or cluster directly for it to require TLS.

So yes, this should be compatible with #1113 although I can test that out once it's in place.

Quentin-M:
Thank you for your prompt answer!

Could you please fix the weird merge and run `terraform fmt` on this?
Other than that, I will let @yifan-gu and @xiang90 review this, then let's merge!

config.tf
hasbro17: fixed

modules/tectonic/resources/manifests/updater/app-version-tectonic-etcd.yaml
```diff
@@ -0,0 +1,13 @@
...
+kind: AppVersion
+metadata:
+  name: tectonic-etcd
+  namespace: tectonic-system
+  labels:
+    managed-by-channel-operator: "true"
+spec:
+  desiredVersion: ${tectonic_etcd_operator_version}
+  paused: false
+status:
```
xiang90: probably we do not need the status section.

modules/tectonic/resources/manifests/updater/tectonic-etcd-operator.yaml
```diff
@@ -0,0 +1,36 @@
...
+        tectonic-app-version-name: tectonic-etcd
+    spec:
+      containers:
+      - image: ${tectonic_etcd_operator_image}
+        name: tectonic-etcd-operator
+        command: ["/usr/local/bin/tectonic-etcd-operator"]
+        resources:
+          limits:
+            cpu: 20m
+            memory: 50Mi
```
xiang90: 50mi memory seems too small...
100m cpu + 128mb memory makes more sense to me.

xiang90:
defer to @yifan-gu

modules/tectonic/resources/manifests/updater/app-version-tectonic-etcd.yaml
hasbro17: I think the status needs to be set since the TEO will compare the `spec.desiredVersion` with the `status.currentVersion`. If `status.currentVersion` is not present it will try to reconcile.

xiang90: /cc @yifan-gu what other operator does?

Quentin-M:
@hasbro17 Needs rebase. Sorry!
@yifan-gu thinks it looks good offline.

modules/tectonic/resources/manifests/updater/app-version-tectonic-etcd.yaml
yifan-gu: Same as @hasbro17 said above.

yifan-gu:
Yeah, LGTM

hasbro17:
Rebased. Waiting for tests to pass.

modules/tectonic/resources/tectonic.sh
```diff
@@ -186,12 +186,14 @@ kubectl create -f updater/kube-version-operator.yaml
 kubectl create -f updater/tectonic-channel-operator.yaml
 kubectl create -f updater/tectonic-monitoring-config.yaml
 kubectl create -f updater/tectonic-prometheus-operator.yaml
+kubectl create -f updater/tectonic-etcd-operator.yaml
```
s-urbaniak: This should be in the `if [ "$EXPERIMENTAL" = "true" ]` section, as the Tectonic etcd operator is for self-hosted etcd only, correct?

modules/tectonic/resources/tectonic.sh
```diff
@@ -186,12 +186,14 @@ kubectl create -f updater/kube-version-operator.yaml
...
 wait_for_tpr tectonic-system channel-operator-config.coreos.com
 kubectl create -f updater/tectonic-channel-operator-config.yaml
 wait_for_tpr tectonic-system app-version.coreos.com
 kubectl create -f updater/app-version-tectonic-cluster.yaml
 kubectl create -f updater/app-version-kubernetes.yaml
 kubectl create -f updater/app-version-tectonic-monitoring.yaml
+kubectl create -f updater/app-version-tectonic-etcd.yaml
```
s-urbaniak: same comment applies as above.

Quentin-M:
Oh good catch!
I can follow-up!

alekssaul:
Came across this PR as I noticed 
```
tectonic-system   tectonic-etcd-operator-1532894501-zx40f           0/1       ImagePullBackOff   0          11m       10.2.0.16      worker0
```
against master. Sounds like Quentin will follow up. Anything we can improve from testing end?

Quentin-M:
I have a PR that pretty much covers this kind of misbehaviors: https://github.com/coreos/tectonic-installer/pull/1066

Quentin-M:
Interestingly having the following on my test cluster too:
```
  2h		3m		37	kubelet, ip-10-0-65-23.eu-west-1.compute.internal	spec.containers{tectonic-etcd-operator}	Warning		Failed		Failed to pull image "quay.io/coreos/tectonic-etcd-operator:v0.0.1": rpc error: code = 2 desc = Error: Status 403 trying to pull repository coreos/tectonic-etcd-operator: "{\"error\": \"Permission Denied (user: coreos+tec2_ifbugljtha3egrcggy3c2mjvg)\"}"
  2h		3m		37	kubelet, ip-10-0-65-23.eu-west-1.compute.internal						Warning		FailedSync	Error syncing pod, skipping: failed to "StartContainer" for "tectonic-etcd-operator" with ErrImagePull: "rpc error: code = 2 desc = Error: Status 403 trying to pull repository coreos/tectonic-etcd-operator: \"{\\\"error\\\": \\\"Permission Denied (user: coreos+tec2_ifbugljtha3egrcggy3c2mjvg)\\\"}\""
```

Looks like the Quay repository misses the required groups.. Adding them now.

hasbro17:
@Quentin-M do you know which user accounts the `coreos-pull-secret` credentials are for? I'm looking at the robot accounts on quay to grant read permissions for the [tectonic-etcd-operator](https://quay.io/repository/coreos/tectonic-etcd-operator) repo  but there are a ton of accounts with the name `coreos+tec2_ifbug******`.

Quentin-M:
I have already added the appropriate groups to your repository :) Check it out

hasbro17:
Ah, just saw it. Had to refresh the tab. Thanks!!

Quentin-M:
Sure, thanks for all the hard work! We are all looking forward to seeing the first etcd update on Tectonic!

PR #1126, opened 2017-06-19
Shellcheck everything

ggreer:
- `make shellcheck` now checks shell scripts in `tests/` and  `modules/`
- Update all shell scripts in those dirs to pass, either by fixing the warning/error or disabling the rule for that line.

Shellcheck is part of the Jenkins tests, so new lint errors should fail the build. We can clean up the existing stuff in time. Basically, this stops us from making new mistakes.

### DIFF
--- a/modules/aws/ignition/resources/s3-puller.sh
+++ b/modules/aws/ignition/resources/s3-puller.sh
@@ -5,6 +5,7 @@ if [ "$#" -ne "2" ]; then
     exit 1
 fi
 
+# shellcheck disable=SC2086,SC2154
 /usr/bin/sudo /usr/bin/rkt run \
     --net=host --dns=host \
     --trust-keys-from-https ${awscli_image} \
@@ -18,4 +19,5 @@ fi
         done
     '
 
+# shellcheck disable=SC1001,SC1083,SC2086
 /usr/bin/sudo mv /tmp/$${1//\//+} $2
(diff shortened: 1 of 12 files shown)


ggreer:
Tests failed because of AWS limits. Shellcheck stuff is fine.

modules/aws/ignition/resources/s3-puller.sh
```diff
@@ -18,4 +19,5 @@ fi
         done
     '
 
+# shellcheck disable=SC1001,SC1083,SC2086
 /usr/bin/sudo mv /tmp/$${1//\//+} $2
```
kans: Does anyone know what this does?

modules/update-payload/make-update-payload.sh
```diff
@@ -45,17 +45,23 @@ echo "Using deployments: [${deployments[*]}]" >&2
...
 for f in ${deployments[*]}; do
   tmpfile=$(mktemp /tmp/deployment.XXXXXX)
+  # shellcheck disable=SC2086
   yaml2json < ${ASSETS_DIR}/${f} > ${tmpfile}
   tmpfiles+=(${tmpfile})
 done
 
+# TODO: (ggreer) I'm pretty sure ">" is *not* what we want here
+# shellcheck disable=SC2071
 if [[ ${#tmpfiles[*]} > 0 ]]; then
```
kans: I don't think this is even close.   Lets fix it now?

modules/aws/ignition/resources/s3-puller.sh
ggreer: It is a mystery.

modules/update-payload/make-update-payload.sh
ggreer: I'm worried by "fixing" it, we'll break something. For all we know, this script depends on broken behavior. :(

Issue #347, opened 2017-04-25
Control plane pods are all scheduled to same node when installing w/ multiple masters


chancez:
Currently the scheduler/controller-manager pods are setup with `preferredDuringSchedulingIgnoredDuringExecution` to attempt to have the pod's spread across different nodes for increased redundancy. 

Unfortunately, in practice, everytime I have installed tectonic, these pods are all landing on the same node. 2 copies of scheduler, and 2 copies of controller-manager will be running on the master which successfully bootstraps first. This makes sense, because once the first master is ready, it will be the only node marked as ready for scheduling these pods, and it will schedule all of them to itself.

One solution is to change the affinity we use from`preferredDuringSchedulingIgnoredDuringExecution` to `requiredDuringSchedulingIgnoredDuringExecution` when the user chooses to use more than 1 master. As mentioned by @aaronlevy, we cannot use `requiredDuringSchedulingIgnoredDuringExecution` with single master setups, as this will break self-hosted upgrades of master components.

Going this route means that scaling up from 1 master to multiple masters will have a similar problem, and the affinity won't be updated with addition of another master node in the ASG, and all pods will still be running on the original master when the other master comes up. 

This is not any worse than today's situation, and the change proposed above solves the issue for multi-master. Longer-term, we can consider building out tooling/features to help people go from single master to multi-master, since this won't go away soon, but is likely to be a common use-case. As much as it would be nice to keep manifests/config for multi-master and single-master unified, I think that having a different configuration based on the setup is unavoidable if we want to optimize for true HA.

chancez:
cc @aaronlevy since we discussed this a bit in slack.

metral:
Outside of kicking the tires & P.O.C's, do we have any insight into how often folks are wanting to provision a single master cluster vs. a HA setup?

It seems that defaulting to multi-master, and enforcing `requiredDuringSchedulingIgnoredDuringExecution` would be preferable for most use-cases.

aaronlevy:
fwiw it wouldn't just be a provision time, if they ever scaled down to 1 master - upgrades would stop working if we enforce 'require'.

Also, the control-plane components will be naturally re-balanced during the first update (because the pods would be re-created).

My initial preference would be that we don't make assumptions based on initial number of nodes -- because those assumptions could later be wrong (e.g. scale down).

[...]

PR #1129, opened 2017-06-19
bootkube: Add required pod anti-affinity field with proper label selectors.

yifan-gu:
Per discussion at https://github.com/coreos/tectonic-installer/issues/347#issuecomment-307177869,
the core components will now have a required pod anti-affinity field
with a label selector that selects labels of the pods of the same version,
so that multiple pods of the same version don't get scheduled into one node while
the pods of different versions can still be scheduled together

/cc @aaronlevy @alexsomesan @s-urbaniak @Quentin-M

### DIFF
--- a/modules/bootkube/assets.tf
+++ b/modules/bootkube/assets.tf
@@ -96,6 +96,8 @@ resource "template_dir" "bootkube" {
     etcd_client_key = "${base64encode(file(
       var.etcd_client_key != "" ? var.etcd_client_key : "/dev/null"
     ))}"
+
+    tectonic_version = "${var.versions["tectonic"]}"
   }
 }
 
--- a/modules/bootkube/assets.tf
+++ b/modules/bootkube/assets.tf
@@ -98,6 +98,8 @@ resource "template_dir" "bootkube" {
     ))}"
 
     tectonic_version = "${var.versions["tectonic"]}"
+
+    master_count = "${var.master_count}"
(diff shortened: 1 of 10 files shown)


Quentin-M:
Is this change happening as well on bootkube? What would be a good way to test for the expected behavior and for any regressions?

yifan-gu:
@Quentin-M 
First question: Yes, it's happening in kvo. 
Second question: I am adding a test in kvo for this but not sure what to test here.

Another thing I found is that the replica number of the controller manager and scheduler is now 2. Can we change it to the number of nodes?

Quentin-M:
@yifan-gu I am not exactly sure what this was set to 2 only. I agree that it should probably be the number of master nodes? Note however that for bare-metal, the `tectonic_master_count` variable is unused and `${length(var.tectonic_metal_controller_names)}` should be used instead. Ping me if you need help!

yifan-gu:
@Quentin-M OK, sounds good, I will tackle this after writing my kvo tests. But so far the `make apply` works as expected.

Quentin-M:
@yifan-gu Thanks lots!

Quentin-M:
@yifan-gu Needs rebase.

Quentin-M:
Tests passed. @aaronlevy you feel okay with this?

aaronlevy:
lgtm

Issue #1100, opened 2017-06-15
1.6.4-tectonic.1 installer --version not working right.


coresolve:
## Versions
tectonic-1.6.4-tectonic.1

* Tectonic version ([release](https://github.com/coreos/tectonic-installer/releases) or commit hash):
1.6.4.1
* Terraform version (`terraform version`):
All
* Platform (aws|azure|openstack|metal):
All
## What happened?
When you run ./installer --version on darwin the output is: 
```
~/tectonic/tectonic-1.6.4-tectonic.1/tectonic-installer/darwin
$ ./installer --version
was not built properly
```

## What you expected to happen?
In the 1.6.2 release:
```
~/tectonic/tectonic-1.6.2-tectonic.1/tectonic/tectonic-installer/darwin
$ ./installer --version
1.6.2-tectonic.1
```

## How to reproduce it (as minimally and precisely as possible)?
See above.

Quentin-M:
We'd like to remove this command and move towards using a banner in the installer.

Quentin-M:
@amrutac Could you please remove this command from the installer once the version banner is merged (or at the same time)?

amrutac:
Sure, will work on that.

Quentin-M:
@amrutac Thanks a lot! Ask me directly if you need help! It should be fairly straight forward (remove it from [cmd/installer/main.go](https://github.com/coreos/tectonic-installer/blob/master/installer/cmd/installer/main.go#L54) and from the [Makefile](https://github.com/coreos/tectonic-installer/blob/master/installer/Makefile#L5)).

PR #1131, opened 2017-06-19
installer: Removes --version flag the installer

amrutac:
Fixes https://github.com/coreos/tectonic-installer/issues/1100


git-version
```diff
@@ -1,20 +0,0 @@
-#!/usr/bin/env bash
```
Quentin-M: This guy is still used by other apps (like release). We should keep it 👍

Quentin-M:
Thanks a lot!

amrutac:
My bad, re-added it.

Quentin-M:
Thanksssssss

PR #1137, opened 2017-06-20
platforms/azure: Remove etcd-lb and connect to etcd directly

### LINKED ISSUE #464: azure: etcd machines are on the Internet and the VMs have a static password

discordianfish:
This closes #464.

There is no reason for using an lb for that, so it should be removed to reduce attack surface and simplify configuration.

### DIFF
--- a/platforms/azure/tectonic.tf
+++ b/platforms/azure/tectonic.tf
@@ -23,7 +23,7 @@ module "bootkube" {
   oidc_groups_claim   = "groups"
   oidc_client_id      = "tectonic-kubectl"
 
-  etcd_endpoints   = ["${module.vnet.etcd_public_ip}"]
+  etcd_endpoints   = "${module.vnet.etcd_private_ips}"
   etcd_ca_cert     = "${var.tectonic_etcd_ca_cert_path}"
   etcd_client_cert = "${var.tectonic_etcd_client_cert_path}"
   etcd_client_key  = "${var.tectonic_etcd_client_key_path}"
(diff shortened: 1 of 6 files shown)


platforms/azure/tectonic.tf
```diff
@@ -23,7 +23,7 @@ module "bootkube" {
   oidc_groups_claim   = "groups"
   oidc_client_id      = "tectonic-kubectl"
 
-  etcd_endpoints   = ["${module.vnet.etcd_public_ip}"]
```
discordianfish: Should we use the DNS name generated by `modules/azure/dns/etcd.tf`? That way we don't need to update the masters if the etcd ips change. Not sure what the reasoning was to use IPs here.

justaugustus:
With the move of LB code into `modules/vnet`, the etcd-lb was removed.
I believe this can be closed out.

alexsomesan:
@justaugustus Right, this has been solved as you mentioned.

Issue #1037, opened 2017-06-08
remove operators from experimental flag


sym3tri:
- [ ] Move self-hosted etcd to be its own option on the etcd screen of the installer
- [ ] Deploy all other operators by default
- [ ] Remove all monitoring manifests that are currently conditionally deployed if the experimental flag is _off_ (b/c they will get deployed by the Prometheus operators)
- [ ] Remove experimental checkbox from the UI

Quentin-M:
This involves both Terraform and frontend changes. They can be executed in parallel as long as we agree on a variable name, such as "tectonic_etcd_self_hosted" (bool).

I started hacking on the Terraform side of things but anyone should feel free to tackle this if they feel comfortable. @ggreer Do you feel like you'll be able to pull this off or maybe @kans can give it a shot? This is high priority for this sprint.

Quentin-M:
Edited my comment above regarding to varible name change.

Quentin-M:
The decision for now is to remove everything from the experimental flag BUT self-hosted etcd. After the code freeze, we will add a feature flag for self-hosted etcd specifically.

PR #1143, opened 2017-06-20
Self-hosted etcd UI

ggreer:
- Hide experimental option (which is really just self-hosted etcd option)
- Unify booleans (self-hosted, external) to a single UI state var: etcd_option.
- Update toAWS & toBareMetal to correctly change UI state to tfvars.
- Update test cases & progress files to pass with new etcd UI.
- UI tests: Give better error message if missing env vars.

Here's what the new radio buttons look like:

screenshot from 2017-06-20 14-28-09

And when provision nodes is selected:

screenshot from 2017-06-20 14-38-02

For #1037.


ggreer:
I assume some copy needs to be updated, but besides that it should be good.

Also important: this changes the default to self-hosted etcd.

Quentin-M:
UI looks good to me! Thanks a lot for the change!

robszumski:
To make it clear that we're creating vs connecting to existing, how about "Create self-hosted etcd cluster (alpha)"

My thinking is using a verb will help you grok this.

ggreer:
So we need the default to be self-hosted on bare metal, but provisioned on AWS? With the current UI framework, our options are:

- Leave the default as self-hosted for both baremetal & AWS.
- Change the default to provisioned. Bare metal users will see no option selected.

Anything requiring "dynamic defaults" (that change based on other form values changing) means writing a new feature in the UI framework. That'd probably take Matt & I a day.

robszumski:
@ggreer where did the bare metal requirement come from? Self-hosted etcd has not graduated to stable and should not be the default anywhere.

ggreer:
We can't boot AWS instances for bare metal, so the only choices are external etcd & self-hosted.

Quentin-M:
Unlike what's stated in Slack (no support for external) or in PR (no support for provisioned etcd), the bare-metal platform supports both: on controller provisioned etcd, external etcd, and self-hosted etcd.

ggreer:
OK, all three options enabled and default is provision or use controllers, not self hosted.

Here's what it looks like now:

screenshot from 2017-06-20 17-54-41

screenshot from 2017-06-20 17-55-49

PR #1146, opened 2017-06-20
m/tectonic/r/m/etcd/cluster-config.yaml: add resource requests

Quentin-M:
As suggested by @xiang90, this PR introduces resource requests to self-hosted etcd in order to guarantee that they will get enough resources to support the cluster.

Since we document that `t2.medium` instances are the bare-minimum, and since we default to that, this sets the recommended resources for a 10 nodes Kubernetes cluster (500m CPU / 300MB RAM), which should comply with what we currently have.

### DIFF
--- a/modules/tectonic/resources/manifests/etcd/cluster-config.yaml
+++ b/modules/tectonic/resources/manifests/etcd/cluster-config.yaml
@@ -6,3 +6,8 @@ metadata:
 spec:
   size: ${etcd_cluster_size}
   version: "${etcd_version}"
+  pod:
+    resources:
+      requests:
+        cpu: 500m
+        memory: 300Mi


xiang90:
lgtm

xiang90:
we should make this more dynamic in the future (probably based on node counts).

/cc @coresolve

coresolve:
according to https://github.com/kubernetes/community/blob/master/contributors/design-proposals/resource-qos.md#qos-classes

[4:39] 
 ```If limits and optionally requests (not equal to 0) are set for all resources across all containers and they are equal, then the pod is classified as Guaranteed.
```

[4:40] 
maybe we should do limits and requests to get the Gauranteed thing

PR #1151, opened 2017-06-21
vmware: 1409.2.0 updates/remove vmtoolsd from ignition

alekssaul:
Ignition v0.14.0 (introduced to stable in Container Linux 1409.2.0) [enables VMware Tools services](https://github.com/coreos/ignition/pull/339) by default. Enabling the service via ignition is no longer required & causes issues with booting up CL 1409.2.0 via Terraform. Removing the dependency from Ignition resources.

Also updates the documentation as `DHCP=no` is the default and makes recommendations to use thin provisioning in template and making the VM template name reflect CL version number.

This change assumes user provisioning to VMware infrastructure is using the latest stable CL image.

### DIFF
--- a/Documentation/install/vmware/vmware-terraform.md
+++ b/Documentation/install/vmware/vmware-terraform.md
@@ -7,7 +7,7 @@ Generally, the VMware platform templates adhere to the standards defined by the 
 ## Prerequsities
 
 1. Download the latest Container Linux Stable OVA from; https://coreos.com/os/docs/latest/booting-on-vmware.html.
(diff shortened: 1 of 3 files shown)


robszumski:
> This change assumes user provisioning to VMware infrastructure is using the latest stable CL image.

Would anything break on beta or stable?

alekssaul:
beta and alpha match ignition version with stable branch. It should be ok, however I only tested the PR against stable image.  

Removal of comment `however in "Customize template", "DHCP Support for Interface 0" *must* be changed to **"no"**.`  impacts versions < 1409.0.0 . 

To my best knowledge, the risk is for users building Tectonic on VMware in master branch against an image that's not current with alpha/beta/stable channels. We did push out a release on all channels today, which may increases the probability.

alekssaul:
tested against `1353.8.0` seems like the change works in both previous stable and current, whereas master seem to have issues against `1409.2.0`

PR #1154, opened 2017-06-21
Azure cleanup

alexsomesan:
This change attempts to clean-up the Azure code base in preparation for promoting it to Beta state.
No major features have been added, but the behaviour of some has changed slightly.

Notable changes:
* all the networking resources, including NICs for each VM, have been moved to the vnet module
* external networking resources, like existing VNet, existing NSGs, etc are now configured by passing in FULL RESROUCE IDs rather than separate names and resource groups. This is to simplify the use API and decrease the change of them not being in sync with each other.
* public DNS is now optional without an explicit flag
  * when not setting "tectonic_base_domain" we assume to use Azure default DNS (*.<location>.cloudapp.azure.com)
  * when "tectonic_base_domain" is set to a domain name, we expect `tectonic_azure_external_dns_zone_id` to pass to a pre-configured Azure DNS zone, that is resolvable via the internet. We don't manage this resource and terraform no longer tries to delete it.
  * the default Azure DNS module can be swapped out for a generic DNS module based on the Terraform [DNS](https://www.terraform.io/docs/providers/dns/index.html) provider. This can publish record to any RFC-2136 compliant DNS server (supporting DNS UPDATE).
* etcd module correctly detects self-hosted mode and avoids creating any resources.
* etcd is also fully TLS enabled

Generally, there should be less input variables to configure now and they should have slightly better descriptions

Integration test configurations have been added, that can also serve as examples of different cluster configurations. They have all been confirmed to produce functional clusters.

Various smaller changes might have happened that are not called out here. 
See the PR contents for the entire truth.

/cc: @s-urbaniak @mxinden @coresolve @metral @justaugustus


justaugustus:
@alexsomesan @s-urbaniak `etcd_group` NSG and ruleset is no longer required... the etcd rules get handled in `modules/vnet/nsg-etcd.tf` now.

alexsomesan:
@justaugustus Got it! I'm dropping it.

justaugustus:
@alexsomesan @s-urbaniak I'd like to see the master / worker AS / SS configurations melded into one `.tf` with a var like `tectonic_azure_use_scale_sets` to control behavior.

I can work on that in a separate, smaller PR, if you'd like.

justaugustus:
@alexsomesan @s-urbaniak we should also use the opportunity to cleanup var descriptions and some of the TODOs floating around the module. 
A [few|lot] of the TODOs were my doing, so feel free to call them out if you need me to clarify anything.

alexsomesan:
@justaugustus 
If that's achievable without too much complexity we could do it.
However, there are too many moving slates right now. Let's look into that once we've stabilized the overall design of this. We're trying to align to the reference implementation first so we can declare Azure as Beta state.

justaugustus:
@alexsomesan / @s-urbaniak #1169 shortens the node / resource names for Azure to get under some of the Azure resource character limits.
It should go in before this cleanup.

config.tf
```diff
@@ -68,13 +68,14 @@ variable "tectonic_versions" {
...
 
   default = {
-    etcd          = "3.1.8"
-    prometheus    = "v1.7.1"
-    alertmanager  = "v0.7.1"
-    monitoring    = "1.3.0"
-    kubernetes    = "1.6.6+tectonic.1"
-    tectonic      = "1.6.6-tectonic.1"
-    tectonic-etcd = "0.0.1"
+    container_linux = "1353.8.0"
```
s-urbaniak: this seems to be in conflict with `tectonic_cl_channel` and is also not very portable across platforms. i.e. AWS uses the channel semantics, OpenStack expects an explicit VM image. I am afraid we are introducing redundancies and inconsistencies here.

examples/terraform.tfvars.azure
```diff
@@ -10,97 +10,113 @@ tectonic_admin_email = ""
...
-tectonic_azure_dns_resource_group = "tectonic-dns-group"
-
-// Storage account type for the etcd node(s). Example: Premium_LRS.
-tectonic_azure_etcd_storage_account_type = "Premium_LRS"
+// (optional) Instance size for the etcd node(s). Example: Standard_DS2_v2.
+// tectonic_azure_etcd_vm_size = "Standard_DS2_v2"
 
-// Instance size for the etcd node(s). Example: Standard_DS2_v2.
-tectonic_azure_etcd_vm_size = "Standard_DS2_v2"
+// (optional) The ID of the external Azure DNS zone used for nodes & endpoints FQDNs.
```
s-urbaniak: We should describe the behavior if this field is left empty.

examples/terraform.tfvars.azure
```diff
@@ -10,97 +10,113 @@ tectonic_admin_email = ""
...
+// Note this is the complete ID as returned in the "id" field by the Azure client, not just the name of the resource.
+// 
+// Example: "/subscriptions/b520eabf-d6c0-4757-8a5e-0fa7d2fe12b5/resourceGroups/my-k8s-cluster/providers/Microsoft.Network/virtualNetworks/test-pre-vnet/subnets/my-k8s-subnet
 // tectonic_azure_external_master_subnet_id = ""
 
-// (optional) The name of the external Network Security Group used for the
-// Tectonic Console and Kubernetes API Server. This depends on
-// `tectonic_azure_external_resource_group` to also be specified.
-// tectonic_azure_external_nsg_api = ""
+// (optional) The ID of the external Network Security Group used for the Tectonic Console and Kubernetes API Server.
```
s-urbaniak: Same here, we should describe the behavior if this field is left empty and all the `*_id` fields below. I suppose this is for bring-your-own use cases?

examples/terraform.tfvars.azure
```diff
@@ -10,97 +10,113 @@ tectonic_admin_email = ""
...
+// Note this is the complete ID as returned in the "id" field by the Azure client, not just the name of the resource.
+// 
+// Example: "/subscriptions/b520eabf-d6c0-4757-8a5e-0fa7d2fe12b5/resourceGroups/my-k8s-cluster/providers/Microsoft.Network/networkSecurityGroups/my-worker-nsg
+// tectonic_azure_external_nsg_worker_id = ""
 
-// (optional) The name of the resource group of the external Network Security
-// Group used. This is required if specifying `tectonic_external_nsg_etcd`,
-// `tectonic_external_nsg_master`, and/or `tectonic_external_nsg_worker`,
+// (optional) Name of an external resource group to place created resources in.
+// Leave blank to create a new one.
```
s-urbaniak: 👍

config.tf
alexsomesan: First, great catch with `tectonic_cl_channel`. 
It seems hardcored for Azure. I'll make it draw from that var.

As to the version number, I used it to pin down to a lower version recently when the latest CL had networking regressions on Azure. I suspect this would come in handy in future similar situations. 

Should we rather try to port it to the other platforms the same way? What do you think?

examples/terraform.tfvars.azure
alexsomesan: Yeah, all the `*_id` are for bring your own stuff.
You mean make it explicit that setting them to empty creates new resources of that type?

config.tf
s-urbaniak: @alexsomesan We could try that, but I am afraid not all platforms will be able to support it. As a data-point:

- AWS: queries `tectonic_cl_channel` only
- OpenStack: refers to a hard-coded image referenced by `tectonic_openstack_image_id`.
- VMWare: expects a converted VM image template referenced by `tectonic_vmware_vm_template`
- Bare Metal: needs both a CL version as here referenced by `tectonic_metal_cl_version` which is in sync with `tectonic_cl_channel`

examples/terraform.tfvars.azure
s-urbaniak: @alexsomesan yes, i.e. `if left empty/blank, a new network security group will be created.`

modules/azure/master-ss/api-lb.tf
```diff
@@ -1,9 +1,9 @@
 resource "azurerm_public_ip" "tectonic_api_ip" {
-  name                         = "tectonic_api_ip"
+  name                         = "${var.cluster_name}_tectonic_api_ip"
   location                     = "${var.location}"
   resource_group_name          = "${var.resource_group_name}"
   public_ip_address_allocation = "static"
-  domain_name_label            = "${var.cluster_name}-k8s"
+  domain_name_label            = "${var.cluster_name}-api"
 
   tags {
```
s-urbaniak: I guess we should add custom tags like `tectonic_aws_extra_tags` to all the things.

s-urbaniak: And probably also add the same cluster ID tag per default as we do in AWS:
```
...
"tectonicClusterID", "${var.cluster_id}"
...
```

examples/terraform.tfvars.azure
alexsomesan: 👍 makes sense. let me do that.

modules/azure/master-ss/api-lb.tf
s-urbaniak: having said that, this is probably less critical in Azure as we have resource groups.

modules/azure/dns/variables.tf
```diff
@@ -22,14 +26,34 @@ variable "master_ip_addresses" {
   type = "list"
 }
 
-variable "console_ip_address" {
-  type = "string"
+variable "worker_ip_addresses" {
```
s-urbaniak: just a naming nit: `worker_ips`/`master_ips`/.... would be way shorter.

justaugustus:
@alexsomesan / @s-urbaniak / @mxinden Not much to say except awesome work on the cleanup, guys!
Looking forward to this massive improvement on the Azure side getting merged in. :)

@metral / @coresolve PTAL when you have a chance.

config.tf
```diff
@@ -68,13 +68,14 @@ variable "tectonic_versions" {
...
 
   default = {
-    etcd          = "3.1.8"
-    prometheus    = "v1.7.1"
-    alertmanager  = "v0.7.1"
-    monitoring    = "1.3.0"
-    kubernetes    = "1.6.7+tectonic.1"
-    tectonic      = "1.6.7-tectonic.1"
-    tectonic-etcd = "0.0.1"
+    container_linux = "1353.8.0"
```
metral: Pinning to 1353.8.0 gets around https://github.com/coreos/tectonic-installer/issues/1171, but with locksmithd and update engine still enabled, the CLUO can still bump this up to 1409.x.0 creating dropped network packet errors on Azure.

modules/azure/master-ss/api-lb.tf
metral: Along the same token as cluster ID tagging, resources like Azure storage account, containers etc. do not utilize the cluster name in their respective names. This results in some resources to be easily identified when listed by the cluster name set since its used as a prefix for the resource name, where as others don't have it. Thoughts?

platforms/azure/main.tf
```diff
@@ -1,60 +1,68 @@
...
 
   location             = "${var.tectonic_azure_location}"
   resource_group_name  = "${module.resource_group.name}"
-  image_reference      = "${var.tectonic_azure_image_reference}"
   vm_size              = "${var.tectonic_azure_etcd_vm_size}"
   storage_account_type = "${var.tectonic_azure_etcd_storage_account_type}"
+  container_image      = "${var.tectonic_container_images["etcd"]}"
 
-  etcd_count            = "${var.tectonic_etcd_count}"
+  etcd_count            = "${var.tectonic_experimental ? 0 : min(var.tectonic_etcd_count, 1)}"
```
metral: If experimental is not enabled, this will always default the count of etcd to be 1 instance per the `min` interpolation syntax. How come we are not using the tectonic_etcd_count?

alexsomesan: You're right. Good catch!
The logic is wrong here. It should have been a max() not a min().

The aim was to make sure that when experimental is disabled, we create at least 1 etcd node.
This comes after debugging an issue for a user who had set both etcd_count to 0 and experimental to false. In that case everything would still start up but API would be unresponsive due to no etcd available.

Will change it to max().

metral: 👍

alexsomesan:
modules/generic-dns is a rough POC I did earlier. It's not well tested. 

I then discussed the idea with Sergiusz and he implemented modules/dns/ddns. He tested that and it works as expected, so that's the recommended one to use. 

I will delete modules/generic-dns. Thanks for spotting the duality.

config.tf
alexsomesan: Makes total sense. 
The measure was meant to unblock me from making progress with these changes, but apparently other users have been asking for such an option.
As you rightfully point out, it's pretty useless without the corresponding support in CLUO.

@robszumski Are there any plans for CLUO to support version pinning, similar to this?

Jenkinsfile
```diff
@@ -183,7 +202,227 @@ pipeline {
...
+                withDockerContainer(builder_image) {
+                  sshagent(['azure-smoke-ssh-key-kind-ssh']) {
+                    checkout scm
+                    unstash 'installer'
+                    unstash 'smoke'
+                    script {
+                      try {
+                        timeout(45) {
+                          sh """#!/bin/bash -ex
+                          ${WORKSPACE}/tests/smoke/azure/smoke.sh plan vars/azure-example.tfvars
```
mxinden: How about using `examples/terraform.tfvars.azure` here instead?

s-urbaniak: Sounds good 👍 Let's iterate on this, once CI passes green. I also have a better commit for the `etcd_count` fix. The only diff to the native example I see is the worker count (1 vs. 3) and azure location ("" vs. "eastus").

mxinden: As terraform favours file defined variables over env defined variables, this is quite a messy change?!

mxinden: I talked to @s-urbaniak. We decided to use `tfvars` file shadowing in the long run, but to delay this task to a different PR.

mxinden:
@s-urbaniak I have rebased the branch so it contains the Jenkins fixes. In addition I triggered the tests twice in a row to test out the cpu limits. Let's see tomorrow how well that works.

s-urbaniak:
CI failure:

```
3 error(s) occurred:

* module.workers.azurerm_virtual_machine.tectonic_worker[0]: 1 error(s) occurred:

* azurerm_virtual_machine.tectonic_worker.0: compute.VirtualMachinesClient#CreateOrUpdate: Failure responding to request: StatusCode=409 -- Original Error: autorest/azure: Service returned an error. Status=409 Code="PropertyChangeNotAllowed" Message="Changing property 'osDisk.vhd.uri' is not allowed."
* module.workers.azurerm_virtual_machine.tectonic_worker[1]: 1 error(s) occurred:

* azurerm_virtual_machine.tectonic_worker.1: compute.VirtualMachinesClient#CreateOrUpdate: Failure responding to request: StatusCode=409 -- Original Error: autorest/azure: Service returned an error. Status=409 Code="PropertyChangeNotAllowed" Message="Changing property 'osDisk.vhd.uri' is not allowed."
* module.masters.azurerm_virtual_machine.tectonic_master: 1 error(s) occurred:

* azurerm_virtual_machine.tectonic_master: compute.VirtualMachinesClient#CreateOrUpdate: Failure responding to request: StatusCode=409 -- Original Error: autorest/azure: Service returned an error. Status=409 Code="PropertyChangeNotAllowed" Message="Changing property 'osDisk.vhd.uri' is not allowed."
```

I am not sure what that means, as this test passed before the CI hickup, so it might be a flake.

alexsomesan:
I've seen this issue occurs when leftover resources from previously failed
builds collide with newly created ones, or when you try to apply again over
a failed build.

Try cleaning up any leftover resources. You can easily do it by deleting
any existing resource groups that are named like the ones created by CI.

Use 'az group list' to describe them and the 'as group delete
<name-of-group>' to clean each one up.

config.tf
robszumski: Yes, I believe there is a way to stop a roll out. @euank or @dghubble, is this true?

justaugustus:
Wooooooo!

Issue #1134, opened 2017-06-20
aws: api-external elb listening on ssh port


alekssaul:
## Versions

* Tectonic version ([release](https://github.com/coreos/tectonic-installer/releases) or commit hash): Tectonic 1.6.X
* Platform (aws|azure|openstack|metal): AWS

## What happened?
API-External ELB is setup with [listening on SSH ports](https://github.com/coreos/tectonic-installer/blob/master/modules/aws/master-asg/elb.tf#L56-L61) however security groups [do not allow SSH traffic](https://github.com/coreos/tectonic-installer/blob/master/modules/aws/vpc/sg-elb.tf)

I believe this is an old configuration from older versions where SSHing into master was done via ELB. 

## Anything else we need to know?
Opening an issue to determine weather we can remove the configuration block in; https://github.com/coreos/tectonic-installer/blob/master/modules/aws/master-asg/elb.tf#L56-L61

Quentin-M:
I do not personally see it being used.

Quentin-M:
Assigning as there is already a PR.

PR #1159, opened 2017-06-21
aws: remove SSH listen from API-external ELB

alekssaul:
Cleanup

Fixes https://github.com/coreos/tectonic-installer/issues/1134

API-External ELB is setup with listening on SSH ports however security groups do not allow SSH traffic.

I believe this is an old configuration from older versions where SSHing into master was done via ELB.


Quentin-M:
ok to test

PR #1166, opened 2017-06-21
modules/aws/etcd/nodes.tf: change hardcoded value of IOPS to 100

### LINKED ISSUE #1035: aws: Amazon assigns 100 IOPS to gp2 storage device causing terraform to destroy etcd on apply after create

Choose one: BUG REPORT or FEATURE REQUEST

## Versions

* Tectonic version: master commit e99319a
* Terraform version (`terraform version`): v0.9.4
* Platform (aws|azure|openstack|metal): aws

## What happened?
After creating a new cluster via terraform apply, performing a terraform plan shows that the ectd instances will be destroyed due to the root_block_device iops being set to 0.  

The defaults for the root_block_device for ectd nodes are:
etcd_root_volume_type: gp2
etcd_root_volume_size: 30
etcd_root_volume_iops: 100

## What you expected to happen?
No changes to the etcd nodes if no variables have changed in terraform.tfvars from the user.

1. Create a new cluster via terraform apply with defaults for root_volumes for etcd
2. After create, perform a terraform plan

It seems that the etcd_node resource in `modules/aws/etcd/nodes.tf` develops a bug once a cluster exists:

```
resource "aws_instance" "etcd_node" {
  count = "${length(var.external_endpoints) == 0 ? var.instance_count : 0}"
  ami   = "${data.aws_ami.coreos_ami.image_id}"

  instance_type          = "${var.ec2_type}"
  subnet_id              = "${element(var.subnets, count.index)}"
  key_name               = "${var.ssh_key}"
  user_data              = "${data.ignition_config.etcd.*.rendered[count.index]}"
  vpc_security_group_ids = ["${var.sg_ids}"]

  lifecycle {
    # Ignore changes in the AMI which force recreation of the resource. This
    # avoids accidental deletion of nodes whenever a new CoreOS Release comes
    # out.
    ignore_changes = ["ami"]
  }

  tags = "${merge(map(
      "Name", "${var.cluster_name}-etcd-${count.index}",
      "kubernetes.io/cluster/${var.cluster_name}", "owned",
      "tectonicClusterID", "${var.cluster_id}"
    ), var.extra_tags)}"

  root_block_device {
    volume_type = "${var.root_volume_type}"
    volume_size = "${var.root_volume_size}"
    iops        = "${var.root_volume_type == "io1" ? var.root_volume_iops : 0}"
  }
}
```

Amazon assigns 100 iops for general purpose root devices (at least on t2.mediums).  The config here set's it to zero if the user chooses to use gp2.  This causes terraform to destroy the instance and recreate it.  The user can't even set the variable to 100 for the match to override it due to the conditional setting it to 0.  The user would need to modify this file and replace 0 to 100 so that terraform sees it as not changed.

Even though this same logic is used for the master and worker nodes, in those situations the installer is creating a launch configuration which seems to hold the value at 0; therefore no changes are made.  Due to actually launching etcd instances without an auto-scaling-group, terraform sees that the root_block_device is 100 and therefore needs to destroy the instance to set it back to zero.

Is there a reason why the etcd nodes are not in a ASG (not an ectd expert, perhaps a bigger discussion)?  Quick solution would be to change 0 to 100 if the user is using the general purpose ssd.

Quentin-M:
This is to prevent Terraform from destroying/re-creating the etcd nodes right away
after a deployment where root_volume_type is not io1.

I am tempted to add ``lifecycle { ignore_changes = [ "*" ] }``.

Fixes #1035
Also created https://github.com/coreos/tectonic-installer/issues/1167 for future ref

### DIFF
--- a/modules/aws/etcd/nodes.tf
+++ b/modules/aws/etcd/nodes.tf
@@ -48,6 +48,6 @@ resource "aws_instance" "etcd_node" {
   root_block_device {
     volume_type = "${var.root_volume_type}"
     volume_size = "${var.root_volume_size}"
-    iops        = "${var.root_volume_type == "io1" ? var.root_volume_iops : 0}"
+    iops        = "${var.root_volume_type == "io1" ? var.root_volume_iops : 100}"
   }
 }


Quentin-M:
ok to test

Quentin-M:
Several AWS tests passed. Also tested manually.

ggreer:
Do https://github.com/coreos/tectonic-installer/blob/master/modules/aws/worker-asg/worker.tf#L46 and https://github.com/coreos/tectonic-installer/blob/master/modules/aws/master-asg/master.tf#L81 need to be changed as well?

Quentin-M:
@ggreer They don't, because they are part of launch configurations, which default to 0.

Quentin-M:
ok to test

Quentin-M:
Tested again manually.

```

This means that Terraform did not detect any differences between your
configuration and real physical resources that exist. As a result, Terraform
doesn't need to do anything.
```

jescarri:
This is still not solved, and from this change:
https://github.com/coreos/tectonic-installer/pull/1397/files

````
  root_block_device {
    volume_type = "${var.root_volume_type}"
    volume_size = "${var.root_volume_size}"
    iops        = "${var.root_volume_type == "io1" ? var.root_volume_iops : 100}"
  }
}
````

````
The conditional syntax is the well-known ternary operation:

CONDITION ? TRUEVAL : FALSEVAL
````

It will read as follows:
if var.root_volume_type == "io1" then use var.root_volume_iops otherwise if is gp2/standard etc use 100

But terraform documentation says:

https://www.terraform.io/docs/providers/aws/r/instance.html#block-devices

````
iops - (Optional) The amount of provisioned IOPS. This is only valid for volume_type of "io1", and must be specified if using that type
````

I think the proper fix will be:

````
iops        = "${var.root_volume_type == "io1" ? var.root_volume_iops : ""}"
````

Issue #1115, opened 2017-06-19
Make sure resources are all tagged


Quentin-M:
Tracking the issue mentioned in https://github.com/coreos/tectonic-installer/pull/218#issuecomment-308591039.

The fix should be in the next release.

Quentin-M:
Would there be a way to add an automated test for that one?

robszumski:
Automated tests would be great. Manually booting 1.6.4 release shows that these items are tagged with custom tags:

 - security groups
 - VPC
 - Subnets

Not tagged with custom tags:

 - EC2 vms
 - ELBs
 - ASGs

Not tagged at all:

 - EBS volumes
 - Internet Gateways
 - Security groups

Quentin-M:
Assigning to @estroz avec discussed offline! Thanks!

estroz:
Tags by Resource Types as of 1.6.4:

*: Only some were tagged, not all

Resource Type | ClusterID | Extra tags
:--- | :---: | :---:
Instance | x | x*
ELB | x | x
Security Group | x | x 
Route53 HostedZone | x | x
Route Table | x | 
Subnet | x |
VPC | x | x
Internet Gateway |  | 
Volume |  |
S3 Bucket |  | 
Autoscaling Group | x | 

Note:
- ASG-controlled instances do not get extra tags

rithujohn191:
Needs more investigation for desired user experience.

robszumski:
In the planning meeting we discussed:

 - upstream Terraform doesn't support tagging for all resource types
 - we could ship a binary that uses the raw APIs to accomplish this

I have reached out to our AWS contacts to see if they can lend any resources to improving the terraform support.

Shipping a separate binary or requiring users to interact with the raw providers API does not meet our UX requirements.

@estroz what was the limitation around tagging the auto-scaled instances?

estroz:
All resource types that support tagging in AWS can be tagged by Terraform
if explicitly created. However, default resources, ex. network interfaces
created when creating an EC2 instance, or a resource created when creating
a parent, cannot be tagged in certain cases.

Regarding volumes of auto-scaled instances, `root_block_device` volumes
specified by launch configurations cannot be tagged by Terraform, according
to https://github.com/hashicorp/terraform/issues/14558.

sym3tri:
This is complete. Just need to document the caveats for which i created a separate issue.

ggreer:
[...]

PR #1169, opened 2017-06-22
azure: Fix node naming

justaugustus:
Tweaks some of the resources names to adhere to Azure limits:

* Add cluster prefixes for AS, fix computer names
* Add `role` var to further parameterize node names
* Shorten `random_id` byte length to 2
* Set etcd to use tectonic module def for cluster name
* Fix node naming => `${cluster_name}{e,m,w}###`
* Use `cluster_name` as prefix for storage objects

### DIFF
--- a/modules/azure/master-as/master.tf
+++ b/modules/azure/master-as/master.tf
@@ -6,11 +6,11 @@
 
 # Generate unique storage name
 resource "random_id" "tectonic_master_storage_name" {
-  byte_length = 4
+  byte_length = 2
 }
 
 resource "azurerm_storage_account" "tectonic_master" {
-  name                = "${random_id.tectonic_master_storage_name.hex}"
+  name                = "${var.cluster_name}${random_id.tectonic_master_storage_name.hex}m"
   resource_group_name = "${var.resource_group_name}"
   location            = "${var.location}"
   account_type        = "${var.storage_account_type}"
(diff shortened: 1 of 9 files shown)


modules/azure/worker-as/variables.tf
```diff
@@ -75,3 +75,7 @@ variable "cloud_provider" {
 variable "kubelet_node_label" {
   type = "string"
 }
+
+variable "role" {
```
alexsomesan: Do we really need this as a variable? Is it reasonable to just make it part of the interpolated string?
The resources where it's being used wouldn't be reused across modules.

justaugustus: @alexsomesan the thought was that this could be used for an Azure companion PR of #1115.
I can pull it out though, just let me know.

modules/azure/master-as/master.tf
```diff
@@ -72,7 +72,7 @@ resource "azurerm_virtual_machine" "tectonic_master" {
   }
 
   os_profile {
-    computer_name  = "${var.cluster_name}-master${count.index}"
+    computer_name  = "${format("%s%s%03d", var.cluster_name, "m", count.index + 1)}"
```
cehoffman: This name needs to exactly match the resource name in order for the azure cloud provider to work when searching for the machines definition in the azure api.

modules/azure/worker-as/workers.tf
```diff
@@ -73,7 +73,7 @@ resource "azurerm_virtual_machine" "tectonic_worker" {
   }
 
   os_profile {
-    computer_name  = "${var.cluster_name}-worker${count.index}"
+    computer_name  = "${format("%s%s%03d", var.cluster_name, "w", count.index + 1)}"
```
cehoffman: Same comment as the master name.

cpanato:
@justaugustus thanks! can you please rebase and address the comments above? thanks again!

justaugustus:
@cpanato will do when I have a chance next week!

cpanato:
closing. Please reopen another PR if you want this change.

PR #1170, opened 2017-06-22
monitoring: add monitoring ingress and auth proxies

brancz:
This PR adds an Ingress object to expose Prometheus and Alertmanager. Between the Ingress and Prometheus/Alertmanager is a modified version of the bitly oauth2 proxy supporting oidc and some other patches. To move quickly product decided to roll our own version for now, which is referred to as `tectonic-monitoring-auth`.

Note this functionality is not yet built into the TPO, so upgrading from the previous Tectonic version will not yet create these manifests, however, once merged here I will take care of that, I just want to agree on these manifests first. Regardless everything should work experimental on or off.

This also means that from now on Tectonic Monitoring components are accessible with authentication via the console address `/prometheus` and `/alertmanager`. As a check that this works when spinning up a cluster head to `/alertmanager` (and login) and you should see 1 firing alert called "DeadMansSwitch", indicating that the alerting pipeline does indeed work. Once you click on "Source" for that alert you should land on the Prometheus UI with the graph for the alert that caused this alert. For the "DeadMansSwitch" this should be a constant line.

@alexsomesan @s-urbaniak @Quentin-M @sym3tri 

/cc @fabxc @mxinden @rezaloo @robszumski


fabxc:
Wuhoo, working after all!

Quentin-M:
Awesome!

Quentin-M:
ok to test

brancz:
ok to test

s-urbaniak:
And of course also from my side: great PR! 👍

brancz:
For some guidance @aaronlevy : 

All of these manifests will be updated by the Tectonic Prometheus Operator except for the changes in the Tectonic Identity configmap. I'm not sure that is actually self-driving yet. The scheme for the callback that have been added there is

- `https://<cluster-dns>/alertmanager/auth/callback`
- `https://<cluster-dns>/prometheus/auth/callback`

Also adding @rithujohn191 to.

robszumski:
As a follow up, can we figure out some resource limits for these and add them to the manifest?

brancz:
@robszumski totally. I'll monitor them to find out appropriate values.

brancz:
If anyone already goes ahead and adds resource requests and limits, then please tag us, so we can take over the same change into the TPO.

Quentin-M:
Testing it ! Yass !

brancz:
The resource usage I've been observing we're around 8Mb memory so a request of 10 and limit of 20 should be fine.

Quentin-M:
The only thing that surprised me a little when I tried it is that I needed to auth again when I was on the console and wanted to switch to Prometheus/AlertManager, I had to login again. Otherwise, that's great.

Quentin-M:
@robszumski @amrutac Are we adding links in the console to these new dashboards?

brancz:
@robszumski said we'll add a rudimentary page in the admin section of console. Would be great if we can just get a simple page with the links in for this release.

Regarding the authentication. Yes that is that way for now, I could come up with a way to generate the cookies in the same way. I'd say it's an improvement for the future.

aaronlevy:
@brancz Will the missing values in the configMap cause problems for existing clusters (when they update)?

brancz:
@aaronlevy problems as in authentication will not work for Prometheus and Alertmanager until the config is adapted and der restarted/reloaded. However, existing clusters will not suffer in any way except for the above, as far as I've been able to tell (I ran the update from the current 1.6.4 and everything seemed to have kept working as before).

aaronlevy:
@brancz Is there a plan for bringing existing clusters into sync? Ideally we're not leaving deltas in place (but understand sometimes we will need to).

robszumski:
@brancz shouldn't be a problem if we know where we want to send folks too, and what the plan going forward it when we have stuff like Grafana.

PR #1172, opened 2017-06-22
modules/tectonic,tests/smoke: make stats URL configurable

squat:
This commit makes the stats reporting URL configurable so that all test
clusters will report statistics to the staging stats database. Although
we already filter out all test data from the production database, it is
useful to submit statistics to the staging database so that we can
happily automate tests that fetch data from that database without
interfering with or causing problems in prod.

Updated: turns out to be consistent with how we setup vars, for the
Tectonic module on any platform, we need to be more verbose.

cc @Quentin-M @kans


kans:
Error creating EIP: AddressLimitExceeded: The maximum number of addresses has been reached.

squat:
ok to test

squat:
ok to test

squat:
ok to test

squat:
ok to test

squat:
passed. testing again.

squat:
ok to test

PR #1174, opened 2017-06-22
Jenkinsfile: change the AWS credential being used & assume-role in private VPC smoke test

Quentin-M:


squat:
@Quentin-M thanks for updating this. I am not sure you should be assuming a role before creating the VPC. I intentionally did not do this because the Tectonic Installer policy should _not_ have privileges to create a VPN gateway or connection, neither of which are required for the installer, but both of which are required for getting the private VPC up and working.

Quentin-M:
Let me add the rights to run `create-vpc` to the smoke user then, and we'll assume the role right afterwards.

Quentin-M:
ok to test

Quentin-M:
ok to test

Quentin-M:
ok to test

Quentin-M:
ok to test

PR #1177, opened 2017-06-23
Jenkins: print systemd logs

ggreer:
- Add forensic script to ssh to machine & print debugging info
- Add cluster-foreach script to run forensics script on each machine in the ASG.
- Get jenkins & smoke tests to run these scripts automatically (even on failure).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -79,7 +79,7 @@ pipeline {
           "SmokeTest TerraForm: AWS": {
             node('worker && ec2') {
               withCredentials(creds) {
-                withDockerContainer(builder_image) {
+                withDockerContainer(args: '-v /etc/passwd:/etc/passwd:ro', image: builder_image) {
                   checkout scm
                   unstash 'installer'
                   unstash 'smoke'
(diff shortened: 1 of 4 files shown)


ggreer:
Closed #1168 & opened a new PR so that it will run in another region. I'm tired of waiting for cleanup scripts to run.

tests/smoke/aws/cluster-foreach.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+set -x
+set -e
+
```
squat: Could you `set -o pipefail`?

kans: I don't think we want x or pipefail here - this is just for debugging.

tests/smoke/forensics.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+set -x
+
+echo -n "hostname: "
```
squat: These would all be more succinct with `printf "hostname %s\n" "$(hostname)"`

tests/smoke/aws/cluster-foreach.sh
squat: Ah ok fair enuff. No pipefail, e; yes x.

Jenkinsfile
```diff
@@ -79,7 +80,7 @@ pipeline {
           "SmokeTest TerraForm: AWS": {
             node('worker && ec2') {
               withCredentials(creds) {
-                withDockerContainer(builder_image) {
+                withDockerContainer(args: '-v /etc/passwd:/etc/passwd', image: builder_image) {
```
Quentin-M: Is there something else we can do..? Like useradd on the executor/in the image, changing the UID/GID mapping..? Terraform could be made to do anything with that.

ggreer:
`/etc/passwd` doesn't contain password hashes. It's just a list of usernames, user ids, group ids, and default shells. Also, I changed jenkins to mount it read-only.

ggreer:
Anyone want to chime in here? I'm pretty sure it's fine to mount `/etc/passwd` as read-only.

kans:
It would probably be a good idea to add `kubelet` to the list - [see also](https://github.com/coreos/tectonic-installer/blob/master/Documentation/troubleshooting/master-nodes.md)

ggreer:
Probably a good idea, but this PR has been open long enough that I just want to get it merged. We can improve on the reporting after that.

Issue #783, opened 2017-05-19
Document minimum requirements for a bare metal install


brianredbeard:
Currently when using the [documentation](https://coreos.com/tectonic/docs/latest/install/bare-metal/) there are no specified requirements for hosts.  At the same time there exists tribal knowledge about _actual_ requirements that are not enumerated in user facing documentation (e.g "The Prometheus process *alone* currently consumes 2GB of RAM, thus you will never be able to run a controller with only 2GB of RAM", "Well all of our _testing_ uses t2.Medium machines and it works" implying a 4GB of RAM minimum).

I personally wasted a number of hours on these bugs within our documentation and similar concerns have been called out in other issues (#563 for example).

dghubble:
https://coreos.com/tectonic/docs/latest/install/bare-metal/requirements.html#machines

sym3tri:
@brianredbeard Going to close this since the docs are already there. Please reopen if we're lacking something specific or you think there's a better way to surface them

dghubble:
I think we could address a few docs issues here. I can see how a reader of the page Redbeard linked might see the Overview section and think that was all the background requirement detail they get. How about we remove that section and simply link to the more extensive bare-metal requirements doc I linked above? cc @zbwright 

@brianredbeard Can you check the existing requirements document to see if it contains the information and caveats you were concerned about?

Finally, to address the larger than expected minimum node sizes, I asked prometheus operator to support configuration of the resource requests of the pods it deploys so then Tectonic could set a more reasonable default. https://github.com/coreos-inc/tectonic-prometheus-operator

zbwright:
let me take a look, and I'll work with @brianredbeard and @dghubble to figure this out. we'll sort this out.

zbwright:
here's the plan:
1. Will will QA the Requirements doc, and flesh it out / correct it as necessary.
2. I'll point out to the Requirements doc from the Installing doc more prominently, and right at the top of the page.
3. I will simultaneously flesh out that doc to match Will and my install test from last week.

(don't know why, but I can neither @mention nor assign him in this issue.)

dghubble:
@zbwright you should be able to @mention CoreOS employees now, the full time team needs to be added as read.

zbwright:
QA'd install guide, with pointer to the requirements doc: https://github.com/coreos/tectonic-installer/pull/1032

zbwright:
[...]

PR #1182, opened 2017-06-23
docs: edits as a result of qa'ing requirements.md.

zbwright:


zbwright:
finishes out https://github.com/coreos/tectonic-installer/issues/783

PR #1195, opened 2017-06-26
docs: adding rhel installing-workers.md.

zbwright:


Documentation/install/rhel/installing-workers.md
```diff
@@ -0,0 +1,156 @@
+# Installing Tectonic workers on Red Hat Enterprise Linux
+
+The flexibility and ease of adding drivers to Red Hat Enterprise Linux allows Kubernetes worker components to be deployed with specialized workloads and hardware. This guide describes using Tectonic Installer to launch these components on a RHEL environment.
```
robszumski: a note about using this for workers but running masters on Container Linux would be great here. Just a quick sentence and link to the general platform section on the index should suffice.

Documentation/install/rhel/installing-workers.md
```diff
@@ -0,0 +1,163 @@
+# Installing Tectonic workers on Red Hat Enterprise Linux
+
+Tectonic Installer creates two types of nodes when deploying a Kubernetes cluster: worker nodes (which execute containers), and master nodes (which handle cluster management tasks, and include the API server).
+
+In large scale clusters, worker nodes may hold containers running several different types of applications. The flexibility and ease of adding drivers to Red Hat Enterprise Linux allows Kubernetes worker components to be deployed with specialized workloads and hardware.
```
robszumski: what do you think of something like this at the end of this paragraph:

```
You can join these Red Hat Enterprise Linux workers to the same cluster as your Container Linux workers.
```

Documentation/install/rhel/installing-workers.md
```diff
@@ -0,0 +1,163 @@
...
+
+If you have not yet created a cluster, use [CoreOS Container Linux][container-linux] as the operating system for the deployment. Then, use Tectonic Installer to deploy a cluster on [bare metal][bare-install] (or [AWS][aws-install]). When defining nodes for the cluster, set the worker node count to 0.
```
robszumski: Slight restructuring, what do you think?

> Before continuing on this guide, use Tectonic Installer to deploy a cluster on [bare metal][bare-install] (or [AWS][aws-install]). The Installer will use Container Linux for the master nodes, and you may additionally configure additional Container Linux worker nodes. 

> Once the cluster is deployed, follow this guide to configure and join additional worker nodes running Red Hat Enterprise Linux.

robszumski: Some random notes:
- using `configure and join` vs `launch` because technically we aren't launching these, the users brings them to us
- didn't want to encourage setting workers to 0, but they can

Documentation/install/rhel/installing-workers.md
```diff
@@ -0,0 +1,163 @@
...
+
+```
+$ systemctl enable kubelet.service
+```
+
+### SELinux
+
+To allow the Tectonic worker, run SELinux in Permissive mode. Running in Enforcing mode will block permissions for worker nodes.
+
+
```
robszumski: I feel like we need an ending here, just to confirm that you're done. Something like:

> You're done! The new worker nodes should be visible in the Tectonic Console and are ready to start running your containers.

robszumski:
These are just docs for a platform not even covered by tests as it's pre-alpha, merging.

PR #1198, opened 2017-06-27
tests/smoke/aws/: grafiti cleanup hook if smoke test fails

### LINKED ISSUE #1116: Hook grafiti in every PRs
In addition to running Grafiti daily, we should replace our Jenkin's failure hook by Grafiti, therefore we will be able to garbage collect any resource regardless of Terraform's state.

estroz:
Jenkinsfile: added step after second destroy attempt that cleans all
created resources

tests/smoke/aws/: function that runs the grafiti Docker container and
deletes resources based on `tectonicClusterID`

In the case of `terraform destroy` failure, `grafiti delete` will clean up resources unsuccessfully destroyed using their `tectonicClusterID` tag.

Fixes #1116

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -207,24 +207,27 @@ pipeline {
       post {
         failure {
           node('worker && ec2') {
-            withCredentials(creds) {
+            withCredentials([creds, quay_creds]) {
               withDockerContainer(builder_image) {
                 checkout scm
                 unstash 'installer'
-                retry(3) {
-                  timeout(15) {
-                    sh """#!/bin/bash -ex
-                    ${WORKSPACE}/tests/smoke/aws/smoke.sh destroy vars/aws.tfvars
-                    """
-                  }
+                timeout(15) {
+                  sh """#!/bin/bash -x
+                  ${WORKSPACE}/tests/smoke/aws/smoke.sh destroy vars/aws.tfvars
+                  """
                 }
-                retry(3) {
-                  timeout(20) {
-                    sh """#!/bin/bash -ex
-                    ${WORKSPACE}/tests/smoke/aws/smoke.sh destroy vars/aws-vpc.tfvars
-                    ${WORKSPACE}/tests/smoke/aws/smoke.sh destroy-vpc
-                    """
-                  }
+                timeout(20) {
+                  sh """#!/bin/bash -x
+                  ${WORKSPACE}/tests/smoke/aws/smoke.sh destroy vars/aws-vpc.tfvars
+                  ${WORKSPACE}/tests/smoke/aws/smoke.sh destroy-vpc
+                  """
+                }
+                timeout(10) {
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -226,6 +226,15 @@ pipeline {
                     """
                   }
                 }
+                retry(2) {
```
kans: Why two retries?

Jenkinsfile
```diff
@@ -226,6 +226,15 @@ pipeline {
...
+                  timeout(10) {
+                    sh """#!/bin/bash -ex
+                    docker login -u="$QUAY_ROBOT_USERNAME" -p="$QUAY_ROBOT_SECRET" quay.io
+                    docker pull quay.io/coreos/grafiti:latest
+                    ${WORKSPACE}/tests/smoke/aws/smoke.sh grafiti-clean vars/aws.tfvars
```
kans: We create two clusters on aws - `aws.tfvars` and `aws-vpc.tfvars`.

kans:
Maybe we should drop the [retries around destroy](https://github.com/estroz/tectonic-installer/blob/22e3fa44fea44bbd7af11a85e7873e701e97b7c2/Jenkinsfile#L214)?

estroz:
@kans retries on both destroys? or just the one you linked

kans:
@estroz Sorry, the old post/failure retry loops.  @ggreer @Quentin-M

Quentin-M:
I think we could drop the `terraform destroy` altogether in the failure hooks now and go straight up with grafiti. We just want to clean up.

kans:
> I think we could drop the terraform destroys altogether in the failure hook now and go straight up with grafiti.

Keeping both but no retries gives the best chance of cleanup I think.

Jenkinsfile
```diff
@@ -207,24 +207,28 @@ pipeline {
...
-                    ${WORKSPACE}/tests/smoke/aws/smoke.sh destroy-vpc
-                    """
-                  }
+                timeout(20) {
+                  sh """#!/bin/bash -ex
+                  ${WORKSPACE}/tests/smoke/aws/smoke.sh destroy vars/aws-vpc.tfvars
+                  ${WORKSPACE}/tests/smoke/aws/smoke.sh destroy-vpc
+                  """
+                }
+                timeout(10) {
```
kans: Is this long enough for two grafiti runs?

estroz: Definitely. `grafiti delete` takes less than 10 minutes to delete an entire AWS region, so 1 cluster shouldn't be more than that.

kans: One last thing: if `destroy` is failing, we this code run?

estroz: Now that I've removed `-e` from destroy step shebangs, they should fall through to grafiti

PR #1200, opened 2017-06-27
Merge images

alexsomesan:
This allows users to pass individual container images as overrides instead of having to provide the whole map in tfvars.

/cc: @Quentin-M @mxinden


mxinden:
@alexsomesan Updated documentation and examples.

alexsomesan:
@mxinden good catch! thanks for that.

alexsomesan:
@mxinden 
We forgot the other platforms.
Can you please apply the same changes to the remaining platforms?
I've already done it for AWS here https://github.com/coreos/tectonic-installer/pull/1200/commits/108d6e5e5a56270c4bb17bf9ae8c21caa6cf1cca
Use that as an example.

platforms/azure/main.tf
```diff
@@ -92,9 +93,9 @@ module "workers" {
   cluster_name                 = "${var.tectonic_cluster_name}"
   public_ssh_key               = "${var.tectonic_azure_ssh_key}"
   virtual_network              = "${module.vnet.vnet_id}"
-  subnet                       = "${module.vnet.worker_subnet}"
```
mxinden: @alexsomesan Just to make sure: This is a wanted change, right?

Quentin-M:
This adds quite a bit of complexity.. Can't a jq command achieve the same? Or sed even. It is extremely easy to change an image's version programmatically with a sed regexp.

PR #1206, opened 2017-06-27
Use shellcheck v0.4.6

ggreer:
- Update builder image to use shellcheck v0.4.6.
- Fix new shellcheck errors.
- Update Jenkinsfile to use new builder image.

Ubuntu & Debian are on v0.4.4. People using homebrew or other installation methods are likely on newer versions that cause `make lint` to report shellcheck errors that Jenkins doesn't catch.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -22,7 +22,7 @@ def quay_creds = [
   )
 ]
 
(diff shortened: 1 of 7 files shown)


Quentin-M:
Thanks!

ggreer:
OK, now it should be good.

PR #1216, opened 2017-06-28
frontend/tests: adds a unit test for synchronous invalidation.

kans:


kans:
smoke tests have failed twice in 3 different ways...

PR #1224, opened 2017-06-28
platforms/aws: set controller manager’s timers to work around ELB’s TTL

Quentin-M:
The default behavior of Kubernetes's controller manager is to mark a node
as Unhealthy after 40s without an update from the node's kubelet. However,
AWS ELB's Route53 records have a fixed TTL of 60s. Therefore, when an ELB's
node disappears (e.g. scaled down or crashed), kubelet might fail to report
for a period of time that exceed the default grace period of 40s and the
node might become Unhealthy. While the eviction process won't start until
the pod_eviction_timeout is reached, 5min by default, certain operators
might already have taken action. This is the case for the etcd operator as
of v0.3.3, which removes the likely-healthy etcd pods from the the
cluster, potentially leading to a loss-of-quorum as generally all kubelets
are affected simultaneously.

To cope with this issue, we increase the grace period, and reduce the
pod eviction time-out accordingly so pods still get evicted after an total
time of 340s after the first post-status failure.

[Here](https://github.com/kubernetes-incubator/kubespray/blob/master/docs/kubernetes-reliability.md) is a good read explaining how the various grace periods and update
periods work together and how their values should be calculated. Note that 
we do not let the users configure kubelet's `--node-status-update-frequency`
and controller-manager's `–-node-monitor-period` due to the fact that these flags are
set in different Terraform modules. Therefore, this PR does not introduce the ability to 
set fast-paced failure detection and pod eviction. The grace period must still be 
set > 10-20sec, with 40sec being the minimum recommended. This is a non-goal as
this PR mainly tries to address the issues we are facing with self-hosted etcd on AWS. 

Fixes https://github.com/coreos-inc/tectonic-issues/issues/268

Ref: https://github.com/kubernetes/kubernetes/issues/41916
Ref: https://github.com/kubernetes-incubator/kube-aws/issues/598

/cc @xiang90 @sym3tri @robszumski @bison

### DIFF
--- a/modules/bootkube/assets.tf
+++ b/modules/bootkube/assets.tf
@@ -99,7 +99,7 @@ resource "template_dir" "bootkube" {
 
     tectonic_version = "${var.versions["tectonic"]}"
 
-    master_count = "${var.master_count}"
+    master_count              = "${var.master_count}"
     node_monitor_grace_period = "${var.node_monitor_grace_period}"
     pod_eviction_timeout      = "${var.pod_eviction_timeout}"
   }
--- a/modules/bootkube/assets.tf
+++ b/modules/bootkube/assets.tf
@@ -100,6 +100,8 @@ resource "template_dir" "bootkube" {
     tectonic_version = "${var.versions["tectonic"]}"
 
     master_count = "${var.master_count}"
+    node_monitor_grace_period = "${var.node_monitor_grace_period}"
+    pod_eviction_timeout      = "${var.pod_eviction_timeout}"
   }
 }
 
--- a/modules/bootkube/resources/manifests/kube-controller-manager.yaml
+++ b/modules/bootkube/resources/manifests/kube-controller-manager.yaml
@@ -38,6 +38,8 @@ spec:
         - --root-ca-file=/etc/kubernetes/secrets/ca.crt
         - --service-account-private-key-file=/etc/kubernetes/secrets/service-account.key
         - --leader-elect=true
+        - --node-monitor-grace-period=${node_monitor_grace_period}
+        - --pod-eviction-timeout=${pod_eviction_timeout}
         - --cloud-provider=${cloud_provider}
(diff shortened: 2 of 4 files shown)


xiang90:
@Quentin-M I think this is an awesome improvement. But I also want us to do some QA on this to ensure it does fix our problem.

Quentin-M:
Agreed~

hongchaodeng:
Thanks @Quentin-M . This change looks great!

I just want to add that, besides `node-monitor-period` which fixes the node becoming unready problem, we also have two other problems that needs to fix:
- **Kubelet** already failed to talk to one endpoint of APIServer, why it didn't cut its connection and retry? 
From https://github.com/kubernetes-incubator/kubespray/blob/master/docs/kubernetes-reliability.md:
> Kubelet uses http.Client() Golang method, but has no specified timeout.

- **ELB** knew that one endpoint is gone/unhealthy, why it didn't cut kubelet's connection? To be verified that ELB won't cut the long connection after node disappears.

- `node monitor grace period` is 40s, buth `pod-eviction-timeout` by default is 5min. That means kubelet & ELB combined didn't cut the connection after 5*60 + 40 = 340s. So I'm worried about there is some other problematic timeout values lying around that would cause trouble to long connections.

Quentin-M:
@hongchaodeng Agreed with you 100% on this one. This is another point we discussed this morning where we were wondering how was this possible that after 5m40s, the communication was not re-established. My guess is that the Idle Timeout configured (3600s) on the ELB defines the TCP Keep-Alive, and that due to that and AWS never terminating the connection forcefully on node destruction, the connection is never interrupted and Kubelet thinks the TCP stream is still valid.

We need to verify the behavior manually on a simple case. This might take a while though because we have no control on the ELB's internal ASG.

robszumski:
Great work tracking this down everyone!

PR #1192, opened 2017-06-26
installer ux: add K8s client progress into installer status screen

LyonesGamer:
Fixes #561.

 * Separates the Tectonic and Terraform status APIs so they don't rely on each other
 * Splits the "Starting Tectonic Console" status check into multiple Kubernetes and Tectonic components

Still in progress.

PR #1229, opened 2017-06-28
	installer: update vendors for k8s.io/client-go/kubernetes to compile

Quentin-M:
Otherwise, the gogo/protobuf and k8s.io/client-go versions are not in
sync and it won’t compile as soon as it is imported.

The same issue was tackled for smoke vendors in:
coreos#1066.

```
vendor/k8s.io/client-go/pkg/api/unversioned/generated.pb.go:74:
undefined: proto.GoGoProtoPackageIsVersion1
```

```
ProtoPackageIsVersion1 is referenced from generated protocol buffer
files to assert that that code is compatible with this version of the
proto package.
```

**How to test?**
Add the following in any Go file:
```
import _ "k8s.io/client-go/kubernetes"
```

And try to compile:
```
$ make backend
```

This unblocks @LyonesGamer's work.


robszumski:
Can we get this reviewed and merged so we can merge https://github.com/coreos/tectonic-installer/pull/1192?

Quentin-M:
It even has a `How to test` section. Shouldn't take more than 5min of your time.

PR #1233, opened 2017-06-28
modules/tectonic/resources/tectonic.sh: update wait for pods logic

rithujohn191:
Account for the case when `kubectl` command returns no resources

### DIFF
--- a/modules/tectonic/resources/tectonic.sh
+++ b/modules/tectonic/resources/tectonic.sh
@@ -57,11 +57,32 @@ function wait_for_tpr() {
 
 function wait_for_pods() {
   set +e
-  local i=0
   echo "Waiting for pods in namespace $1"
-  while $KUBECTL -n "$1" get po -o custom-columns=STATUS:.status.phase,NAME:.metadata.name | tail -n +2 | grep -v '^Running'; do
-    (( i++ ))
-    echo "Pods not available yet, waiting for 5 seconds ($i)"
+  while true; do
+  
+    out=$($KUBECTL -n "$1" get po -o custom-columns=STATUS:.status.phase,NAME:.metadata.name)
+    status=$?
+    echo "$out"
+  
+    if [ "$status" -ne "0" ]; then
+      echo "kubectl command failed, retrying in 5 seconds"
+      sleep 5
+      continue
+    fi
+  
+    # make sure kubectl does not return "no resources found"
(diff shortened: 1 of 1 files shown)


Quentin-M:
ok to test

rithujohn191:
ok to test

s-urbaniak:
ok to test

modules/tectonic/resources/tectonic.sh
```diff
@@ -56,15 +56,36 @@ function wait_for_tpr() {
...
+  
+    if [ "$status" -ne "0" ]; then
+      echo "kubectl command failed, retrying in 5 seconds"
+      sleep 5
+      continue
+    fi
+  
+    # make sure kubectl does not return "no resources found"
+    if [ "$(echo "$out" | tail -n +2 | grep -c '^')" -eq 0 ]; then
+      echo "no nodes were found, retrying in 5 seconds"
```
s-urbaniak: nit: `no resources were found, retrying in 5 seconds`

modules/tectonic/resources/tectonic.sh
```diff
@@ -56,15 +56,36 @@ function wait_for_tpr() {
...
+      sleep 5
+      continue
+    fi
+  
+    stat=$( echo "$out"| tail -n +2 | grep -v '^Running')
```
s-urbaniak: - nit (remove the extraneous whitespace) in `$( echo...`
- additionally we have to wrap this call with `set +e`/`set -e`. `grep` will return with exit code 1, causing the whole script to fail. I think this is also the cause for the CI failure.

modules/tectonic/resources/tectonic.sh
rithujohn191: Done

modules/tectonic/resources/tectonic.sh
rithujohn191: Done. Thought it would be more clean to `set +e/set -e` for the whole function.

rithujohn191:
ok to test

s-urbaniak:
ok to test

modules/tectonic/resources/tectonic.sh
s-urbaniak: @rithujohn191 agreed 👍

PR #1235, opened 2017-06-28
tectonic: use regex for image string parsing

colhom:


colhom:
ok to test

s-urbaniak:
ok to test

modules/tectonic/resources/manifests/monitoring/alertmanager.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
     alertmanager: main
 spec:
   replicas: 2
-  baseImage: ${element(split(":",alertmanager_image),0)}
+  baseImage: ${replace(alertmanager_image,image_re,"$1")}
```
alexsomesan: The established practice throughout the rest of the code is to just have variables interpolated into templates. Any calculations necessary to get to the values of those template variables should ideally be done higher up the stack, in the template_* resource inputs or the module inputs.

Not a technical issue, just a code consistency one.

platforms/vmware/tectonic.tf
```diff
@@ -77,6 +77,8 @@ module "tectonic" {
   experimental      = "${var.tectonic_experimental}"
   master_count      = "${var.tectonic_master_count}"
   stats_url         = "${var.tectonic_stats_url}"
+
+  image_re = "${var.tectonic_image_re}"
```
alexsomesan: We generally want to keep the number of user-facing input variables to the lowest amount possible and try hard to keep it that way. These inputs are our API to the users. The less inputs we expose the simpler the tool is to understand and use. Also, the easier it is to manage / evolve the API of the installer as it inevitably grows.

Now, in my opinion there's very little reason for this to be a user input. Under the vast majority of use-cases this value will not change and I'd really appreciate if users had one less item to worry about when creating their cluster configs.

If you agree with my point above, let's make this an internal variable to the module and not expose it.

As a side note, I'm using the very same approach for parsing Azure resource IDs and have also added a similar constant variable to the azure modules.

alexsomesan: Now I noticed that the input variable is already there, introduced by your earlier PR, which already got merged. My assessment still holds and I think the input var should be removed.
Since it's already in, no reason to block on it this time. I should have payed better attention to the original PR.

Can you please then follow-up with a PR to remove the input var?

s-urbaniak:
@colhom CI failure reveals an unresolved variable:
```
* template_dir.tectonic: failed to render /home/.../build/aws-tls-pr-1235-.../.terraform/modules/.../resources/manifests/monitoring/alertmanager.yaml: 11:43: unknown variable accessed: image_re
```

Do you mind to have a look?

platforms/vmware/tectonic.tf
colhom: The only reason I have it as a user input is to avoid re-defining the regex in all the modules, instead allowing the value of `tectonic_image_re` to propagate down to all the modules. Is there a way to accomplish this consolidation without defining the variables as we have here?

colhom:
@s-urbaniak missed a var definition :tongue:

platforms/vmware/tectonic.tf
colhom: ping @alexsomesan

alexsomesan: Gave more thought to this issue. We're seeing more often the need for similar "global constant" type of vars, that only make sense internally. We need to find a way to all of them out of the sight of the docs & examples generators, which requires a bit of refactoring.
Since we need to do this anyway for all other situations, let's keep things as they are for now here too, also since this change is not the one that introduced that top-level var.

dghubble:
@colhom how is your image replacement plan going to work with lots of operators spawning deployments on their own, with their own images. This seems unmaintainable.

dghubble:
Also, why was this merged after code freeze? @alexsomesan

squat:
@dghubble as long as it's not cherry-picked into the release branch, it still observes the freeze.

dghubble:
Ah, thanks @squat

yifan-gu:
@colhom Do we need to cherry-pick this to the 1.7 release? If so, we also need to run `make payload` to include the change on payload as well.
/cc @Quentin-M

Quentin-M:
@yifan-gu I don't feel like this is a bug fix, we should be fine.

modules/tectonic/resources/manifests/updater/operators/container-linux-update-operator.yaml
```diff
@@ -23,6 +23,8 @@ spec:
         command:
         - "/bin/update-operator"
         - "--manage-agent=true"
+        - "-agent-image-repo"
```
Quentin-M: May I ask why is this flag modification here? /cc @dghubble

colhom:
@Quentin-M this is definitely a bug. If you specify an image repo like `quay.local:4430/coreos/hyperkube:v1.6.7` the installer is broken

Quentin-M:
@colhom Sorry, I didn't express myself properly, it is indeed. Could you confirm whether this is a hard product requirement for 1.7.1? Thank you! /cc @robszumski.

robszumski:
This is not a hard requirement from my end

modules/tectonic/resources/manifests/updater/operators/container-linux-update-operator.yaml
Quentin-M: Bump ^

dghubble: I did not approve of this

dghubble: After talking with Colin, its for his plan to try to make other image caches work. However, we're dropping all operator/management flags from CLUO (the application) in the release after this one so it won't be here for long.

dghubble: If this change isn't going in v1.7.1 there's nothing to worry about :), the `-agent-image` flag will be a NoOp or removed in the release after this one.

Quentin-M: My understanding is that It'd apparently break some migration work, and it affects the payload. It should probably have been discussed a bit more before merging. But as this is not going to be in the 1.7.1 release, we are probably fine. In the next CLUO release, this flag will be NoOp, so it should not affect it. Thanks for the explanations!

dghubble: It'd actually make the migration easier if this were reverted to remove this before the next Tectonic release. CLUO will not be responsible for "managing" or "operating" other Kubernetes objects then, so no reason to have them at all. We could delete the code. Thoughts? @colhom

dghubble:
The more I consider this the more I worry it may break all operators. This allows manifests to contain custom images and adds flags without notice. This may not be known to TCO and KVO and will be a surprise when updating those components, payloads have no knowledge of these per-cluster customizations. Ideally there are no conflicts in the patch merge, but I had to handle this specially in the prototype operator for CLUO because other flags were being changed. I suspect other operators will have to account for this too. Are they informed?

cc @yifan-gu @derekparker

modules/tectonic/resources/manifests/updater/operators/container-linux-update-operator.yaml
colhom: @dghubble sgtm. sorry for the difficulty

derekparker:
@dghubble not sure I fully understand, but generally for "custom images" and per-cluster configuration, that certainly won't work for any component KVO manages. The "per-cluster" configuration could work eventually once we get our 3-way merge logic fully implemented for flags and such, but as of right now KVO will overwrite the image field or replace the manifest during the upgrade, losing any custom image or per-cluster configuration set.

Quentin-M:
No further actions are needed for 1.7.1. This won't get cherry picked. As far as I'm concerned, this can go in master.

dghubble:
IMO, Tectonic should revert this until there is an idea of how it will work with the upgrades.

dghubble:
@Quentin-M Seems there is a mutual understanding that this won't go in 1.7.1. Now what do we do about master though? I don't want this to be forgotten and accidentally ship

Quentin-M:
@dghubble This is unfortunately not my call to make. I'd say this is the Installer's team one, Core Services's (KVO) and yours (CLUO flag). My understanding regarding to updates is that the images are going to be overwritten by an update - thus breaking the cluster if images have been changed because there is no internet connectivity.

yifan-gu:
@Quentin-M How can a cluster without internet connection then get update payload??

dghubble:
@alexsomesan @robszumski I think we wanna revert this before it makes it in a Tectonic 1.7.2 release

PR #1236, opened 2017-06-28
tests/conformance: remove double quotes for docker run command

rithujohn191:
${MNT_SECRETS} is a multi-valued variable that cannot have quotes around it.


squat:
@rithujohn191 there should still be quotes around all the other variables on that line apart from the mnt_secrets

PR #1237, opened 2017-06-29
frontend: Change AWS docs link to open GUI installer docs

kyoto:
Shouldn't this link open the GUI installer docs?


robszumski:
Yeah, that seems correct to be.

PR #1250, opened 2017-06-29
modules/update-payload: Retain cluo manage-agent behavior

dghubble:
* CLUO will default to manage-agent=false in a future release. Retain the original behavior in Tectonic until the new version operator for this component is ready.
* Corresponds to this change which will go into a future CLUO release https://github.com/coreos/container-linux-update-operator/pull/104

### DIFF
--- a/modules/tectonic/resources/manifests/updater/operators/container-linux-update-operator.yaml
+++ b/modules/tectonic/resources/manifests/updater/operators/container-linux-update-operator.yaml
@@ -22,6 +22,7 @@ spec:
         image: ${container_linux_update_operator_image}
         command:
         - "/bin/update-operator"
+        - "--manage-agent=true"
         env:
         - name: POD_NAMESPACE
           valueFrom:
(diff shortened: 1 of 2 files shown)


euank:
Will tectonic be migrating existing deployments, not just new installs?

aaronlevy:
It will need to be done for both new/existing.

dghubble:
The proposed operator will manage both the update-operator deployment and the update-agent daemonset. Details are in https://github.com/coreos/container-linux-update-operator/issues/98.

This change just ensures the operator continues to behave exactly as it does now, until we're ready to make that change. It does not lock us into that migration either. This just allows the CLUO default behavior to be changed, which helps those using it directly.

dghubble:
Yes, lemme make the change explicit for new installs as well.

s-urbaniak:
@dghubble thanks for the contribution! Is this ready to merge?

yifan-gu:
lgtm

PR #1263, opened 2017-06-30
Run 'grafiti' as cleanup step after AWS smoke test failure

estroz:
Jenkinsfile: removed grafiti docker container dependency
tests/smoke/aws: removed grafiti docker container dependency

Environment variables required by `grafiti` to delete orphaned cluster resources are present within `tectonic-builder` containers and not accessible by a `grafiti` container. Cluster cleanup should be done by the same container that created the cluster.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -99,6 +99,7 @@ pipeline {
     stage("Tests") {
       environment {
         TECTONIC_INSTALLER_ROLE = 'tectonic-installer'
+        GRAFITI_DELETER_ROLE = 'grafiti-deleter'
       }
       steps {
         parallel (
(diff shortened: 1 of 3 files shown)


Quentin-M:
cool

kans:
| Cluster cleanup should be done by the same container that created the cluster.
👍

`go get github.com/coreos/grafiti/cmd/grafiti`

This makes me a bit nervous because if grafiti ever breaks, we probably won't notice until we hit limits.

estroz:
@kans any suggestions on how to lock the version (without using Docker image tags)?

estroz:
ok to test

estroz:
ok to test

estroz:
ok to test

estroz:
ok to test

estroz:
ok to test

mxinden:
In regards to extending the Tectonic builder Dockerfile I would like to start the discussion whether we should break this container up into smaller containers. 

Right now the image:
* is 600 MB big, 
* takes quite a while to  build, 
* and needs to be updated in many places each time we update one small build step.

If I am not mistaking we could run the Grafiti steps inside a small vanilla golang image with just Grafiti installed inside.

Let me know what you think. I do see the advantages of a single image that we can use everywhere and I am not sure what the best practice should be here.

Quentin-M:
The release automation work that I am starting soon will definitely include a revamp of the building process in order to make it streamlined and producing a release tarball everytime.

estroz:
@mxinden I agree. The current images' size and build time are too large. For now this is the best solution though, until @Quentin-M revamps the release process. There is too much state necessary to run grafiti contained in the smoke scripts to run grafiti in its own container at this point.

mxinden:
@estroz @Quentin-M This plan sounds good to me. Let's split the Tectonic builder image up, once we rework the release process. Thanks.

Jenkinsfile
```diff
@@ -217,37 +238,44 @@ pipeline {
...
+                      """
                     }
                   }
                 }
               }
             }
-          },
-          "SmokeTest: Azure": {
+          }
+/*          "SmokeTest: Azure": {
```
estroz: Changed the title back. All this is commented out to test failure recovery, and will be un-commented before merge.

estroz:
Update: `grafiti delete` is not picking up `AWS_REGION` or the `region` config field when run via `smoke.sh`. I'm in the process of debugging why it isn't.

estroz:
ok to test

estroz:
ok to test

estroz:
ok to test

estroz:
ok to test

estroz:
@mxinden @kans @Quentin-M good to review

mxinden:
@estroz Could you rename the PR. If I understand correctly #1442 adds *Grafiti* to the Tectonic builder Docker image. This PR adds *Grafiti* to our Jenkins test pipeline.

PR #1268, opened 2017-06-30
Makefile: support Terraform v0.10.x

alexsomesan:
This change adds support for the new `init` command in Terraform 0.10+ which will download the provider modules on demand.

It will only perform the init command if the version of Terraform used is greater then 0.9.x

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -127,4 +127,4 @@ vendor-smoke: $(TOP_DIR)/tests/smoke/glide.yaml
 	@cd $(TOP_DIR)/tests/smoke && glide up -v
 	@cd $(TOP_DIR)/tests/smoke && glide-vc --use-lock-file --no-tests --only-code
 
-.PHONY: make clean terraform terraform-dev structure-check docs examples terraform-get
+.PHONY: make clean terraform terraform-dev structure-check docs examples terraform-get terraform-init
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -22,7 +22,12 @@ localconfig:
 	mkdir -p $(BUILD_DIR)
 	cp examples/*$(subst /,-,$(PLATFORM)) $(BUILD_DIR)/terraform.tfvars
 
-terraform-get:
+terraform-init:
```
s-urbaniak: nit: this should be added to `.PHONY`

alexsomesan: Very true! Let me fix that.

Makefile
```diff
@@ -22,7 +22,12 @@ localconfig:
...
+ifneq ($(shell $(TF_CMD) version | grep -E "Terraform v0\.1[0-9]\.[0-9]+"), )
```
s-urbaniak: small nit though: this probably won't work with v1.0/0.2x/0.3x....

alexsomesan: That's right, it wouldn't.
I'm hoping we can get rid of the check though, once we're a few versions over 0.1x.x

Mattias-:
You could also update the required version in config.tf: https://github.com/coreos/tectonic-installer/blob/master/config.tf#L20

alexsomesan:
retest this please

PR #1275, opened 2017-07-01
Omit console AWS infra for vanilla clusters

Mattias-:
When creating a cluster on AWS an ELB, two Route 53 routes and a security group entry is created for the purpose of accessing the Tectonic Console service. This service isn't running on vanilla clusters so the infrastructure doesn't need to exist.

### DIFF
--- a/modules/aws/vpc/variables.tf
+++ b/modules/aws/vpc/variables.tf
@@ -56,3 +56,8 @@ variable "master_azs" {
 variable "worker_azs" {
   type = "list"
 }
+
+variable "vanilla_k8s" {
+  description = "If true, no infrastructure will be created to access tectonic services on any nodes."
+  default     = false
+}
--- a/platforms/aws/main.tf
+++ b/platforms/aws/main.tf
@@ -16,6 +16,7 @@ module "vpc" {
   cluster_id              = "${module.tectonic.cluster_id}"
   extra_tags              = "${var.tectonic_aws_extra_tags}"
   enable_etcd_sg          = "${!var.tectonic_experimental && length(compact(var.tectonic_etcd_servers)) == 0 ? 1 : 0}"
+  vanilla_k8s             = "${var.tectonic_vanilla_k8s}"
 
   # VPC layout settings.
   #
@@ -127,6 +128,7 @@ module "masters" {
   extra_tags                   = "${var.tectonic_aws_extra_tags}"
   autoscaling_group_extra_tags = "${var.tectonic_autoscaling_group_extra_tags}"
   custom_dns_name              = "${var.tectonic_dns_name}"
+  vanilla_k8s                  = "${var.tectonic_vanilla_k8s}"
 
   root_volume_type = "${var.tectonic_aws_master_root_volume_type}"
   root_volume_size = "${var.tectonic_aws_master_root_volume_size}"
--- a/modules/aws/master-asg/elb.tf
+++ b/modules/aws/master-asg/elb.tf
@@ -89,6 +89,7 @@ resource "aws_route53_record" "api-external" {
 }
 
 resource "aws_elb" "console" {
+  count           = "${var.vanilla_k8s ? 0 : 1}"
   name            = "${var.custom_dns_name == "" ? var.cluster_name : var.custom_dns_name}-con"
   subnets         = ["${var.subnet_ids}"]
   internal        = "${var.public_vpc ? false : true}"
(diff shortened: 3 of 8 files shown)


modules/aws/vpc/variables.tf
```diff
@@ -56,3 +56,8 @@ variable "master_azs" {
 variable "worker_azs" {
   type = "list"
 }
+
+variable "tectonic_vanilla_k8s" {
```
squat: Our convention is to omit the lead in `tectonic_` on variable names in the actual modules.

platforms/aws/main.tf
```diff
@@ -16,6 +16,7 @@ module "vpc" {
   cluster_id              = "${module.tectonic.cluster_id}"
   extra_tags              = "${var.tectonic_aws_extra_tags}"
   enable_etcd_sg          = "${!var.tectonic_experimental && length(compact(var.tectonic_etcd_servers)) == 0 ? 1 : 0}"
+  tectonic_vanilla_k8s    = "${var.tectonic_vanilla_k8s}"
```
squat: So this would just be `vanilla_k8s =...`

modules/aws/vpc/variables.tf
Mattias-: Fixed in f74f0a5, thank you @squat!

squat:
Looks good! I will test manually early next week since our smoke tests do not currently cover vanilla k8s clusters, though they really should if we are offering this feature :/

squat:
ok to test

s-urbaniak:
ok to test

s-urbaniak:
@squat pinging about the manual testing, this LGTM from my side.

squat:
@Mattias- Sorry for the delay in providing more feedback. I tested this PR and found that it failed to build because there is a dangling referenced to console infrastructure that is not created. Specifically, I found the following by grepping for `console.id`:
```
modules/aws/master-asg/master.tf
33:  load_balancers = ["${aws_elb.api-internal.id}", "${join("",aws_elb.api-external.*.id)}", "${aws_elb.console.id}"]
```
The master ASG still references the console elb but since that elb is not createdTerraform fails. We need to make this reference conditional on vanilla_k8s.

Also, we are still provisioning a console AWS security group; we should not do if we are eliminating console infrastructure. I found the following references to that resource:
```
modules/aws/vpc/sg-elb.tf
26:resource "aws_security_group" "console" {

modules/aws/vpc/outputs.tf
32:  value = "${aws_security_group.console.id}"

modules/aws/vpc/sg-worker.tf
174:  source_security_group_id = "${aws_security_group.console.id}"
```

Finally, you correctly modified the master_ingress_services_from_console security group rule for the master ASG and we must similarly add a conditional count to the worker ASG's console security group rule:
```
modules/aws/vpc/sg-worker.tf
171:resource "aws_security_group_rule" "worker_ingress_services_from_console" {
```

Thank you very much for this work, @Mattias-!

cc @s-urbaniak

Mattias-:
Thank's for reviewing @squat. Fixes in 65a756e

squat:
Thanks. Testing again now...

squat:
After much testing on Saturday, I am still seeing errors with this PR with the current changes. I am unable to `terraform apply`, however errors look like red herrings: Terraform is failing to resolve the first variables in the first templates it renders in both the Tectonic and Bootkube modules. Have you had any luck launching a cluster with these changes? If so, can you provide a tfvars so I can binary search what it is in my configuration the breaks? Otherwise, we'll need to search through the diff to find the breakage.

squat:
Hi @Mattias- after more digging I found why this does not work. Currently, `platforms/aws/tectonic.tf` uses
```
oidc_issuer_url    = "https://${module.masters.ingress_internal_fqdn}/identity"
```
to configure the oidc provider for the k8s apiserver in bootkube and
```
base_address       = "${module.masters.ingress_internal_fqdn}"
```
`module.masters.ingress_internal_fqdn` in turn comes from the Route53 record for the private-ingress in `modules/aws/master-asg/elb.tf`:
```
value = "${aws_route53_record.ingress-private.name}"
```
which you are no longer creating. This causes the bootkube module to fail to render templates and causes the tectonic module to fail to provision TLS certs for the private ingress. This will require some reworking in order to elimnate the ELB and Route53 records for private ingress.

cpanato:
closing due to inactivity. Please open another PR if you want this change. Thanks!

Issue #1285, opened 2017-07-05
Add AWS and BM to "advanced installer" opt-group


robszumski:
Right now the UI doesn't show the AWS or BM terraform guides under the "advanced installer" section. We should add them.

robszumski:
Other related changes to this screen are contained in https://github.com/coreos/tectonic-installer/issues/462 which is about destroying clusters in an easier way.

PR #1292, opened 2017-07-06
frontend: Add AWS and Bare Metal options under "Advanced Installer"

kyoto:
Also clean up code a bit
- Remove `isTerraform()` since everything is Terraform now.
- Remove `SELECTED_PLATFORMS` sorting since it is not used.

Fixes: #1285

screenshot-1
screenshot-2
screenshot-3
screenshot-1

### DIFF
--- a/installer/frontend/platforms.js
+++ b/installer/frontend/platforms.js
@@ -1,33 +1,33 @@
+export const AWS = 'aws';
 export const AWS_TF = 'aws-tf';
 export const AZURE = 'azure';
+export const BARE_METAL = 'bare-metal';
 export const BARE_METAL_TF = 'bare-metal-tf';
 export const OPENSTACK = 'openstack';
 
 export const isSupported = platform => [AWS_TF, BARE_METAL_TF].includes(platform);
 
 export const PLATFORM_NAMES = {
+  [AWS]: 'Amazon Web Services',
   [AWS_TF]: 'Amazon Web Services',
   [AZURE]: 'Microsoft Azure',
+  [BARE_METAL]: 'Bare Metal',
   [BARE_METAL_TF]: 'Bare Metal',
   [OPENSTACK]: 'Openstack',
 };
 
 export const optGroups = [
   ['Graphical Installer (default)', AWS_TF, BARE_METAL_TF],
-  ['Advanced Installer', AZURE, OPENSTACK],
+  ['Advanced Installer', AWS, BARE_METAL, AZURE, OPENSTACK],
 ];
 
-const platOrder = [
-  AWS_TF, BARE_METAL_TF, AZURE, OPENSTACK,
-];
-
-export const SELECTED_PLATFORMS = (window.config ? window.config.platforms : []).sort((a, b) =>
-  platOrder.indexOf(a) - platOrder.indexOf(b)
-);
+export const SELECTED_PLATFORMS = window.config ? window.config.platforms : [];
 
 export const DOCS = {
-  [BARE_METAL_TF]: 'https://coreos.com/tectonic/docs/latest/install/bare-metal/index.html',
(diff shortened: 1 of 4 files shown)


installer/frontend/platforms.js
```diff
@@ -1,33 +1,31 @@
...
   ['Graphical Installer (default)', AWS_TF, BARE_METAL_TF],
-  ['Advanced Installer', AZURE, OPENSTACK],
+  ['Advanced Installer', AWS, BARE_METAL, AZURE, OPENSTACK],
 ];
 
-const platOrder = [
-  AWS_TF, BARE_METAL_TF, AZURE, OPENSTACK,
-];
-
-export const SELECTED_PLATFORMS = (window.config ? window.config.platforms : []).sort((a, b) =>
```
kans: `window.config.platforms` comes from a command line flag.  I believe the old behavior was that we would remove items from the platform dropdown in the case that `platforms` were specified.

installer/frontend/platforms.js
```diff
@@ -1,35 +1,31 @@
...
 
 export const isSupported = platform => [AWS_TF, BARE_METAL_TF].includes(platform);
 
 export const PLATFORM_NAMES = {
+  [AWS]: 'Amazon Web Services',
   [AWS_TF]: 'Amazon Web Services',
   [AZURE]: 'Microsoft Azure',
+  [BARE_METAL]: 'Bare Metal',
   [BARE_METAL_TF]: 'Bare Metal',
   [OPENSTACK]: 'Openstack',
```
robszumski: I think this is `OpenStack` with a capital S

installer/frontend/platforms.js
kyoto: Thanks @kans. I brought back `SELECTED_PLATFORMS` and added "aws" and "bare-metal" as additional options accepted by `-platforms` to preserve the current behavior.

installer/frontend/platforms.js
kyoto: Updated

kyoto:
@kans RFR

installer/frontend/platforms.js
```diff
@@ -1,35 +1,33 @@
...
-);
+export const SELECTED_PLATFORMS = window.config ? window.config.platforms : [];
 
 export const DOCS = {
-  [BARE_METAL_TF]: 'https://coreos.com/tectonic/docs/latest/install/bare-metal/index.html',
+  [AWS]: 'https://coreos.com/tectonic/docs/latest/install/aws/aws-terraform.html',
   [AWS_TF]: 'https://coreos.com/tectonic/docs/latest/install/aws/index.html',
   [AZURE]: 'https://coreos.com/tectonic/docs/latest/install/azure/azure-terraform.html',
+  [BARE_METAL]: 'https://coreos.com/tectonic/docs/latest/install/bare-metal/metal-terraform.html',
+  [BARE_METAL_TF]: 'https://coreos.com/tectonic/docs/latest/install/bare-metal/index.html',
```
cpanato: why there are two bare metals docs and the `BARE_METAL` points to the terraform doc and the `BARE_METAL_TF` points to another doc, should not be the other way round?

sym3tri:
@squat can you pls review?

installer/frontend/platforms.js
kyoto: `BARE_METAL` / `metal-terraform.html` is for installing using the command line (using Terraform).
`BARE_METAL_TF` / `index.html` is for installing using the GUI (`_TF` suffix because this also uses Terraform).

It is confusing and should probably be renamed, but I don't think that should be part of this change because the naming pattern existed before this PR.

cpanato: thanks for the explanation

PR #1299, opened 2017-07-06
Bump hyperkube to v1.6.7_coreos.0

aaronlevy:


Quentin-M:
@robszumski @amrutac @sym3tri ^ This would fix the etcd upgrade issue, but means we'd ship 1.6.7 rather than 1.6.6.

PR #1314, opened 2017-07-07
Apply 'extra tags' to autoscaling groups

estroz:
installer/frontend/: 'tectonic_autoscaling_group_extra_tags' will be populated by data from user-provided tags.

User-defined 'extra' tags are not being applied to autoscaling groups because tag data doesn't populate `tectonic_autoscaling_group_extra_tags`.

Applies to #1115

### DIFF
--- a/installer/frontend/__tests__/examples/tectonic-aws-vpc.progress
+++ b/installer/frontend/__tests__/examples/tectonic-aws-vpc.progress
@@ -1,7 +1,7 @@
 {
   "clusterConfig": {
     "platformType": "aws-tf",
-    "aws_ssh": "some-ssh-key", 
+    "aws_ssh": "some-ssh-key",
     "aws_etcds-numberOfInstances": 3,
     "aws_etcds-instanceType": "t2.large",
     "aws_etcds-storageSizeInGiB": 300,
@@ -28,7 +28,7 @@
     "extra": {
         "awsHostedZoneId": {
             "zoneToName": {
-                "asfd": "example.com"    
+                "asfd": "example.com"
             }
         }
     },
--- a/installer/frontend/cluster-config.js
+++ b/installer/frontend/cluster-config.js
@@ -217,9 +217,11 @@ export const toAWS_TF = (cc, FORMS, opts={}) => {
   }
 
   const extraTags = {};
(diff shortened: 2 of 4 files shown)


installer/frontend/cluster-config.js
```diff
@@ -223,6 +223,15 @@ export const toAWS_TF = (cc, FORMS, opts={}) => {
     }
   });
 
+  const extraAutoScalingTags = {};
+  _.each(cc[AWS_TAGS], ({key, value}) => {
+    if(key && value) {
+      extraAutoScalingTags.key = key;
```
kans: I think you meant `extraAutoScalingTags[key] = value`.  It would also be OK to dump these three lines into the [loop above](https://github.com/estroz/tectonic-installer/blob/1b16a5a5a638e0fc2ee03352e4bcf8f629572bab/installer/frontend/cluster-config.js#L222).

estroz: The structure of autoscaling group tags is like so:
```json
{
    "key": "key-name",
    "value": "value-data",
    "propogate_at_launch": bool
}
```
I kept the autoscaling tags in a separate loop to keep the code clean.

installer/frontend/__tests__/examples/tectonic-aws-vpc.progress
```diff
@@ -54,6 +54,13 @@
         "value": "testing"
       }
     ],
+    "awsASGTags": [
```
kans: You should remove "awsASGTags" from both progress files (which will also get rid of the console.errors) - you'd only want these here if we were storing the tags in memory at this location, but we only really want to output a new tfvar.

installer/frontend/cluster-config.js
```diff
@@ -20,6 +20,7 @@ export const AWS_SECRET_ACCESS_KEY = 'awsSecretAccessKey';
 export const AWS_SESSION_TOKEN = 'awsSessionToken';
 export const AWS_SSH = 'aws_ssh';
 export const AWS_TAGS = 'awsTags';
+export const AWS_ASG_TAGS = 'awsASGTags';
```
kans: I don't think this is being used.

PR #1323, opened 2017-07-07
docs: print platform lines explictly

robszumski:
During user testing, two things tripped people up that are very easy to fix:

1. The `$PLATFORM` variable is strange if you are not used to the command line, plus we don't actually do the variable interpolation.

2. Some users were unfamiliar with `darwin` = macOS, as this is not a way it is branded.

This change simply gives macOS and Linux users the command to run.

cc: @meghanschofield

### DIFF
--- a/Documentation/install/bare-metal/index.md
+++ b/Documentation/install/bare-metal/index.md
@@ -169,10 +169,14 @@ tar xzvf tectonic-1.6.4-tectonic.1.tar.gz
 cd tectonic/tectonic-installer
 ```
 
-Run the Tectonic Installer that matches your platform (`linux` or `darwin`):
+Run the Tectonic Installer for your platform:
 
-```sh
-./$PLATFORM/installer
+```bash
+# macOS users
+$ ./tectonic-installer/darwin/installer
+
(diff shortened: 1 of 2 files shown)


Documentation/install/bare-metal/index.md
```diff
@@ -169,10 +169,14 @@ tar xzvf tectonic-1.6.4-tectonic.1.tar.gz
...
 
-Run the Tectonic Installer that matches your platform (`linux` or `darwin`):
+Run the Tectonic Installer for your platform:
 
-```sh
-./$PLATFORM/installer
+```bash
+# macOS users
+$ ./tectonic-installer/darwin/installer
+
```
zbwright: even though it starts getting super wordy, I think if you want to do this, you should break the two commands apart (on both pages).

Run the Tectonic Installer for your platform. 

For macOS users:
```bash
$  ./tectonic-installer/darwin/installer
```

For Linux users:
```bash
$ ./tectonic-installer/linux/installer
```

robszumski:
@zbwright updated, PTAL

PR #1338, opened 2017-07-10
tests: Make tectonic builder-image configurable

mxinden:
Currently we are using a custom terraform version. Among others it increases time outs in terraform itself. In order to stay in sync with upstream terraform we want to do nightly tests with vanilla terraform. In order to be able to accomplish this, we have to make the tectonic builder-image configurable.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -22,7 +22,7 @@ def quay_creds = [
   )
 ]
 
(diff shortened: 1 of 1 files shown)


mxinden:
I would merge tomorrow if there are no objections from San Francisco.

PR #1341, opened 2017-07-10
Revert "Merge pull request #1314 from estroz/autoscaling-group-tags"

estroz:
This reverts commit b018c80dc6c4c62fcc38ff9836d1f56e9917e8ce, reversing changes made to d684cca43d80fe92ce94186ccaee9ed7d590c5f6.

When running `terraform plan` to deploy a cluster, the following error message is received:
```
1 error(s) occurred:

* variable tectonic_autoscaling_group_extra_tags should be type list, got map
```

Comments like https://github.com/hashicorp/terraform/issues/7705#issuecomment-300564571 indicate that terraform cannot handle complicated data structures, such as lists of maps whose structures are not explicitly declared, i.e. user-created variables. Therefore the changes created by b018c80dc6c4c62fcc38ff9836d1f56e9917e8ce should be reverted until lists of arbitrary maps are supported. In the meantime a workaround will be created and merged.

### DIFF
--- a/installer/frontend/__tests__/examples/aws-vpc.json
+++ b/installer/frontend/__tests__/examples/aws-vpc.json
@@ -31,13 +31,6 @@
     "tectonic_aws_extra_tags": {
       "test_tag": "testing"
     },
-    "tectonic_autoscaling_group_extra_tags": [
-      {
-        "key": "test_tag",
-        "value": "testing",
-        "propagate_at_launch": true
-      }
-    ],
     "tectonic_aws_master_ec2_type": "t2.large",
     "tectonic_aws_master_root_volume_size": 33,
     "tectonic_aws_master_root_volume_type": "gp2",
--- a/installer/frontend/__tests__/examples/aws.json
+++ b/installer/frontend/__tests__/examples/aws.json
@@ -14,13 +14,6 @@
     "tectonic_aws_extra_tags": {
       "test_tag": "testing"
     },
-    "tectonic_autoscaling_group_extra_tags": [
-      {
-        "key": "test_tag",
-        "value": "testing",
-        "propagate_at_launch": true
-      }
-    ],
     "tectonic_aws_master_custom_subnets": {
       "us-west-1a": "10.0.0.0/19",
       "us-west-1c": "10.0.32.0/19"
--- a/installer/frontend/cluster-config.js
+++ b/installer/frontend/cluster-config.js
@@ -217,11 +217,9 @@ export const toAWS_TF = (cc, FORMS, opts={}) => {
   }
 
   const extraTags = {};
-  const extraAutoScalingTags = [];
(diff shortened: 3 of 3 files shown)


estroz:
ok to test

sym3tri:
@estroz what was the workaround?

estroz:
@sym3tri users must manually tag their autoscaling groups. I can make a PR documenting this if you'd like.

s-urbaniak:
This is very strange, but I believe I have the explanation. We are generating a JSON based `terraform.tfvars` file.

Using a list of maps is fine in HCL syntax (I am using the following for all my clusters):
```
tectonic_autoscaling_group_extra_tags = [
  {
    key = "expirationDate",
    value = "yyyy-mm-dd",
    propagate_at_launch = true,
  },
]
```

When I convert the above to JSON syntax:
```
{
  "tectonic_autoscaling_group_extra_tags": [
    {
      "key": "expirationDate",
      "propagate_at_launch": true,
      "value": "yyyy-mm-dd"
    },
  ],
}
```

Then I also get the same reported error:
```
1 error(s) occurred:

* variable tectonic_autoscaling_group_extra_tags should be type list, got map
```

This error message doesn't make any sense, until we convert the above using https://github.com/kvz/json2hcl (which simply uses the Terraform JSON parser) back to HCL:
```
"tectonic_autoscaling_group_extra_tags" = {
  "key" = "expirationDate"

  "propagate_at_launch" = true

  "value" = "2017-09-20"
}
```

One can see, that the JSON parser simply omitted the list declaration and generated a simple map. Now the error message makes total sense.

s-urbaniak:
Other interesting failure cases. When I declare two list entries like this:
```
  "tectonic_autoscaling_group_extra_tags": [
    {
      "key": "expirationDate",
      "propagate_at_launch": true,
      "value": "2017-09-20"
    },
    {
      "key": "foo",
      "propagate_at_launch": true,
      "value": "bar"
    }
  ],
```

Terraform generates two HCL variables of the same name:
```
"tectonic_autoscaling_group_extra_tags" = {
  "key" = "expirationDate"

  "propagate_at_launch" = true

  "value" = "2017-09-20"
}

"tectonic_autoscaling_group_extra_tags" = {
  "key" = "foo"

  "propagate_at_launch" = true

  "value" = "bar"
}
```

The `terraform plan` error message here obviously also makes sense:
```
invalid value "terraform.tfvars" for flag -var-file-default: multiple map declarations not supported for variables
```

lucab:
For reference https://github.com/hashicorp/hcl/issues/209.

s-urbaniak:
Another reference https://github.com/hashicorp/terraform/issues/15549

s-urbaniak:
two ways out of this:
- we generate `terraform.tfvars` in HCL syntax (which supports lists of maps)
- we fix the JSON parser (that's obviously the cleaner way)

PR #1347, opened 2017-07-10
frontend: removes split-dns off option...

kans:
because it does not work

### DIFF
--- a/installer/frontend/components/aws-vpc.jsx
+++ b/installer/frontend/components/aws-vpc.jsx
@@ -49,7 +49,7 @@ import {
   SERVICE_CIDR,
   toVPCSubnetID,
   SPLIT_DNS_ON,
-  SPLIT_DNS_OPTIONS,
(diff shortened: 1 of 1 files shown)


installer/frontend/components/aws-vpc.jsx
```diff
@@ -49,7 +49,7 @@ import {
   SERVICE_CIDR,
   toVPCSubnetID,
   SPLIT_DNS_ON,
-  SPLIT_DNS_OPTIONS,
+  // SPLIT_DNS_OPTIONS,
```
amrutac: Can we just delete this code?

installer/frontend/components/aws-vpc.jsx
```diff
@@ -355,7 +355,7 @@ class AWS_VPCComponent extends React.Component {
           </div>
         </div>
       </div>
-      { !privateZone &&
+      {/* privateZone &&
```
amrutac: This too

PR #1348, opened 2017-07-10
Cherry picks ui change for the installer and docs change

amrutac:


alekssaul:
I believe the current stable is; 1409.6.0

amrutac:
@alekssaul: This specifies the min version required for upgrades to work.

PR #1350, opened 2017-07-11
tests: Update tectonic-builder to allow writes to coreos go folder 

mxinden:
With this PR comes a new version of the tectonic-builder image (v1.29).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -22,7 +22,7 @@ def quay_creds = [
   )
 ]
 
-def default_builder_image = 'quay.io/coreos/tectonic-builder:v1.25'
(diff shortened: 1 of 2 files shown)


mxinden:
After quite some debugging I still can't tell you what changed on our Jenkins setup. Previously we were able to change `resolv.conf` and the `/go` path, owned by root, with the `uid 1000` user. Somehow this changed (like it should be). Now this can only be done by root.

Solution: Make `/go` path writable to non-root users, like the underlying vanilla `golang` image does as well and run the `private vpn` smoke test as root inside the docker image so that it can alter the `resolv.conf` file.

Feel free to merge.

mxinden:
@squat @s-urbaniak Thanks for the quick review!

PR #1373, opened 2017-07-12
frontend: ignores the value of AWS_SPLIT_DNS in progress files

kans:


Quentin-M:
This prevents having the feature enabled when a progress file that has it is used.

PR #1376, opened 2017-07-12
Full permissions for $WORKSPACE

estroz:
tests/smoke/aws: full permissions for `$WORKSPACE`

Jenkins smoke tests flake randomly due to permissions errors, likely due to the Docker user not having full access to the `$WORKSPACE` directory. Opening permissions to that dir should solve this issue.

Tests: none

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -69,17 +69,18 @@ pipeline {
           withDockerContainer(params.builder_image) {
             checkout scm
             sh """#!/bin/bash -ex
-            mkdir -p \$(dirname $GO_PROJECT) && ln -sf $WORKSPACE $GO_PROJECT
+            chmod -R 777 "$WORKSPACE"
+            mkdir -p \$(dirname "$GO_PROJECT") && ln -sf "$WORKSPACE" "$GO_PROJECT"
 
             # TODO: Remove me.
             go get github.com/segmentio/terraform-docs
             go get github.com/s-urbaniak/terraform-examples
 
-            cd $GO_PROJECT/
+            cd "$GO_PROJECT"/
             make structure-check
             make bin/smoke
 
-            cd $GO_PROJECT/installer
+            cd "$GO_PROJECT"/installer
             make clean
             make tools
             make build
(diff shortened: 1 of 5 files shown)


estroz:
ok to test

estroz:
ok to test

Makefile
```diff
@@ -116,8 +116,8 @@ structure-check:
 	$(eval FMT_ERR := $(shell terraform fmt -list -write=false .))
 	@if [ "$(FMT_ERR)" != "" ]; then echo "misformatted files (run 'terraform fmt .' to fix):" $(FMT_ERR); exit 1; fi
 
-	@if make docs && ! git diff --exit-code; then echo "outdated docs (run 'make docs' to fix)"; exit 1; fi
-	@if make examples && ! git diff --exit-code; then echo "outdated examples (run 'make examples' to fix)"; exit 1; fi
+	@if make docs && ! git diff -G. --exit-code; then echo "outdated docs (run 'make docs' to fix)"; exit 1; fi
```
mxinden: Just as a hint for other reviewers: "Makefile,installer/Makefile: turn off file permission checks"

Jenkinsfile
```diff
@@ -69,17 +69,18 @@ pipeline {
           withDockerContainer(params.builder_image) {
             checkout scm
             sh """#!/bin/bash -ex
-            mkdir -p \$(dirname $GO_PROJECT) && ln -sf $WORKSPACE $GO_PROJECT
+            chmod -R 777 "$WORKSPACE"
```
mxinden: `withDockerContainer` executes code inside the container as uid1000 by default. As far as I can tell the *file-permission* errors are due to the fact that we are running as uid1000 but some folders and files are owned by root. If this is the case, `chmod -R 777` executed by uid1000 will fail, right? Am I missing something?

estroz: Yes you're correct. Weirdly enough, permissions failures only occurred on one or two stages per test run, and those seem to have disappeared from Jenkins as of now.

ggreer:
This PR might not be the fix, but the permission errors are definitely a problem with tests right now.

estroz:
Unfortunately we saw permission errors when calling `chmod`, so we need to find another way to fix this.

PR #1378, opened 2017-07-13
frontend: Remove modal outline when focussed

kyoto:
### Before
screenshot-1

### After
screenshot-2


kyoto:
@robszumski Separate PR for adding that text: #1395

PR #1398, opened 2017-07-14
vmware: calico network policy support fixes

alekssaul:
Follow up on https://github.com/coreos/tectonic-installer/pull/780

fixes:
- syntax error for archive_file resource for assets (copy pasted from AWS; https://github.com/coreos/tectonic-installer/blob/master/platforms/aws/tectonic.tf#L139
- add `kubelet_cni_bin_dir` missing variable to workers https://github.com/coreos/tectonic-installer/blob/master/platforms/aws/main.tf#L138

/cc : @abhinavdahiya

### DIFF
--- a/platforms/vmware/main.tf
+++ b/platforms/vmware/main.tf
@@ -79,6 +79,7 @@ module "workers" {
 
   kubelet_node_label  = "node-role.kubernetes.io/node"
   kubelet_node_taints = ""
+  kubelet_cni_bin_dir = "${var.tectonic_calico_network_policy ? "/var/lib/cni/bin" : "" }"
   kube_dns_service_ip = "${module.bootkube.kube_dns_service_ip}"
   container_images    = "${var.tectonic_container_images}"
   bootkube_service    = ""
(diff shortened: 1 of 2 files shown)


alekssaul:
Prior to the change; 
image

Tested without; `tectonic_calico_network_policy = true` flag. Builds successfully;
image

image

alekssaul:
also builds with `tectonic_calico_network_policy = true` and calico containers deployed; 
image

alekssaul:
Azure smoketests are failing. PR only changes platforms/vmware, I believe this is a test flake.

platforms/vmware/main.tf
```diff
@@ -98,4 +98,5 @@ module "workers" {
   kubeconfig              = "${module.bootkube.kubeconfig}"
   private_key             = "${var.tectonic_vmware_ssh_private_key_path}"
   image_re                = "${var.tectonic_image_re}"
+  kubelet_cni_bin_dir     = "${var.tectonic_calico_network_policy ? "/var/lib/cni/bin" : "" }"
```
abhinavdahiya: nit: maybe move this option to just after `kubelet_node_taints` option.

alekssaul: makes sense.

alekssaul:
Thanks @abhinavdahiya , don't have access to merge. Can you squash/merge when you get a moment please.

PR #1400, opened 2017-07-14
tests: Prevent *permission denied* errors

mxinden:
* Wrap clean up steps of tests, executed by root, inside *try catch
finally* blocks. This way whenever *plan* or *apply* fails, the cleanup
steps are still executed.

* Remove `-e` (# exit immediately when a command fails) from shebang in
clean up steps, so if one clean up step fails others are still executed.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -211,18 +211,37 @@ pipeline {
                   checkout scm
                   unstash 'installer'
                   unstash 'smoke'
-                  timeout(45) {
-                    sh """#!/bin/bash -ex
-                    . ${WORKSPACE}/tests/smoke/aws/smoke.sh create-vpc
-                    ${WORKSPACE}/tests/smoke/aws/smoke.sh plan vars/aws-vpc.tfvars
-                    ${WORKSPACE}/tests/smoke/aws/smoke.sh create vars/aws-vpc.tfvars
-                    ${WORKSPACE}/tests/smoke/aws/smoke.sh test vars/aws-vpc.tfvars
-                    ${WORKSPACE}/tests/smoke/aws/smoke.sh destroy vars/aws-vpc.tfvars
-                    ${WORKSPACE}/tests/smoke/aws/smoke.sh destroy-vpc
-                    # As /build is created by root, make sure to remove it again
-                    # so non-root can create it later.
-                    rm -r ${WORKSPACE}/build
-                    """
+                  script {
+                    try {
+                      timeout(45) {
+                        sh """#!/bin/bash -ex
+                        . ${WORKSPACE}/tests/smoke/aws/smoke.sh create-vpc
+                        ${WORKSPACE}/tests/smoke/aws/smoke.sh plan vars/aws-vpc.tfvars
+                        ${WORKSPACE}/tests/smoke/aws/smoke.sh create vars/aws-vpc.tfvars
+                        ${WORKSPACE}/tests/smoke/aws/smoke.sh test vars/aws-vpc.tfvars
+
+                        ${WORKSPACE}/tests/smoke/aws/smoke.sh destroy vars/aws-vpc.tfvars
+                        ${WORKSPACE}/tests/smoke/aws/smoke.sh destroy-vpc
+                        """
+                      }
+                    }
+                    catch (error) {
+                        throw error
+                    }
+                    finally {
+                      retry(3) {
+                        timeout(15) {
+                          sh """#!/bin/bash -x
+                            ${WORKSPACE}/tests/smoke/aws/smoke.sh destroy vars/aws-vpc.tfvars
+                            ${WORKSPACE}/tests/smoke/aws/smoke.sh destroy-vpc
+
+                            # As /build is created by root, make sure to remove it again
+                            # so non-root can create it later.
+                            rm -r ${WORKSPACE}/build
+                          """
+                        }
+                      }
+                    }
                   }
                 }
               }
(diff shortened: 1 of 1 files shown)


mxinden:
This is a follow up for https://github.com/coreos/tectonic-installer/pull/1376.

kans:
@mxinden 

Would you explain the root cause of the permission problems or how this fixes them?

mxinden:
@kans We have to run some tests as root inside the docker container for example to be able to edit `/etc/resolv.conf`. The tests run as root create files and folders which are only editable by root itself. If they leave these files after their execution other tests fail, as those non-root tests are not able to edit/delete the folders and files.

But why do these files still exist after our root-tests run even though we have proper clean up commands in place? The issue is, that if previous steps inside a test failed, like e.g. `make apply`, the clean up steps would never be executed, either because Jenkins stops after one command failed, or because we had the `-e` parameter in our clean up bash scripts.

This PR makes sure the clean up steps are run both when previous steps succeed or fail. This is done in two ways:
1. Via Groovy with `try catch finally`
2. Removing the `-e` parameter from the clean up step bash scripts. (Not all scripts)

Let me know if this answers your question. This will fix some of our tests flakes, sadly not all of them.

Jenkinsfile
```diff
@@ -288,7 +303,7 @@ pipeline {
                       finally {
                         retry(3) {
                           timeout(15) {
-                              sh """#!/bin/bash -ex
+                              sh """#!/bin/bash -x
```
ggreer: We want the tests to fail if destroy fails. This changes that behavior.

mxinden: Ah, I see. I don't think this is an issue in this line, as bash without '-e' just returns the last exit code. There is only one line here, whose exit code will be returned. But this is definitely a concern in other changes I made.

Thanks for the catch. I have updated the PR. Will follow up with a comment.

Jenkinsfile
```diff
@@ -211,18 +211,40 @@ pipeline {
...
+                            # fails, still return non-zero if one command
+                            # returns non-zero
+                            EXIT_STATUS=0
+
+                            ${WORKSPACE}/tests/smoke/aws/smoke.sh destroy vars/aws-vpc.tfvars || EXIT_STATUS=$?
+                            ${WORKSPACE}/tests/smoke/aws/smoke.sh destroy-vpc || EXIT_STATUS=$?
+
+                            # As /build is created by root, make sure to remove it again
+                            # so non-root can create it later.
+                            rm -r ${WORKSPACE}/build || EXIT_STATUS=$?
```
estroz: If either `destroy` step fails but this succeeds, `EXIT_STATUS` will be overwritten and exit 0.

mxinden: @kans The right side of the `||` will only be executed when the left side fails. See: http://www.unix.com/shell-programming-and-scripting/42417-what-does-mean-double-pipe.html

Hope I understood your comment correctly.

estroz: Good point. I would still refactor this as suggested to make it cleaner.

mxinden:
As @ggreer @kans  mentioned [here](https://github.com/coreos/tectonic-installer/pull/1400#discussion_r127758670) I was missing the fact, that my changes ignore if e.g. a `make destroy` fails.

I have now added the following construct around our cleanup steps:

``` bash
# No "-e" so all commands are executed even if one
# fails, still return non-zero if one command
# returns non-zero
EXIT_STATUS=0

command1 || EXIT_STATUS=$?
command2 || EXIT_STATUS=$?
command3 || EXIT_STATUS=$?

exit EXIT_STATUS
```

If one of the commands fails, others are still executed, but the script fails with the respective exit code. I have wrapped this even around single line scripts, so whenever we extend the script we don't forget this.

I am up for better suggestions. I do see that this is very verbose. Let me know what you think. Thanks for the feedback so far.

Jenkinsfile
mxinden: @estroz Are other `timeout` blocks still executed even if one of them fails?

estroz: No they won't be if you're using `-e`, but outside of the `catch` block you want to check for failure

mxinden: But this would defeat the purpose of this PR. If `make destroy` fails I still want to execute the following clean up steps. 

I could move the `make destroy` step into the `try` block. But if `make apply` fails in the middle, ressources will not be cleaned up properly, as `-e` makes the script fail entirely after the `make apply`. What do you think?

estroz: To confirm, you're talking about the timeouts in the [case I've linked](https://github.com/estroz/tectonic-installer/blob/9f075b498a5283ae08c7deedfb470160183ca7d9/Jenkinsfile#L113-L152) right?

In the above case, `destroy` has a chance to run first in the `try` block. If either `apply` or `destroy` fail, both of which we want to test for failure, then we should try `destroy` again in the `catch` block. This will give the behavior that you want (run clean-up steps no matter what). Does that make sense?

mxinden: @estroz Ah, thanks for the explanation. Understood. I will adjust the PR accordingly, remove my verbose logic and duplicate the make destroy. Looks a lot cleaner.

estroz: No problem! Apologies if I was being unclear earlier. Thanks!

mxinden:
Third iteration: As @kans pointed out [here](https://github.com/coreos/tectonic-installer/pull/1400#discussion_r127795460) I can duplicate the `make destroy` command, one in `try` one in `finally`. This way the tests still fail if `make destroy` fails. In addition we keep the `-e` flag removed so all clean up steps, including a second `make destroy` are executed.

@kans @ggreer Let me know what you think.

s-urbaniak:
Impressive distributed consensus! :-)

PR #1403, opened 2017-07-14
config.tf: bump image deps (incl. hyperkube 1.7.1)

Quentin-M:
Let's see how that baby goes.

### DIFF
--- a/config.tf
+++ b/config.tf
@@ -25,8 +25,8 @@ variable "tectonic_container_images" {
   type        = "map"
 
   default = {
-    hyperkube                       = "quay.io/coreos/hyperkube:v1.6.7_coreos.0"
-    pod_checkpointer                = "quay.io/coreos/pod-checkpointer:4e7a7dab10bc4d895b66c21656291c6e0b017248"
+    hyperkube                       = "quay.io/coreos/hyperkube:v1.7.1_coreos.0"
+    pod_checkpointer                = "quay.io/coreos/pod-checkpointer:980d1b4b4b8374240c240fb0f85e3a8d9c51663c"
     bootkube                        = "quay.io/coreos/bootkube:v0.5.0"
     console                         = "quay.io/coreos/tectonic-console:v1.7.4"
     identity                        = "quay.io/coreos/dex:v2.4.1"
@@ -42,9 +42,9 @@ variable "tectonic_container_images" {
     node_exporter                   = "quay.io/prometheus/node-exporter:v0.14.0"
     kube_state_metrics              = "quay.io/coreos/kube-state-metrics:v0.5.0"
     config_reload                   = "quay.io/coreos/configmap-reload:v0.0.1"
-    prometheus_config_reload        = "quay.io/coreos/prometheus-config-reloader:v0.0.1"
-    heapster                        = "gcr.io/google_containers/heapster:v1.3.0"
-    addon_resizer                   = "gcr.io/google_containers/addon-resizer:1.7"
+    prometheus_config_reload        = "quay.io/coreos/prometheus-config-reloader:v0.0.2"
+    heapster                        = "gcr.io/google_containers/heapster:v1.4.0"
+    addon_resizer                   = "gcr.io/google_containers/addon-resizer:2.0"
     stats_emitter                   = "quay.io/coreos/tectonic-stats:6e882361357fe4b773adbf279cddf48cb50164c1"
     stats_extender                  = "quay.io/coreos/tectonic-stats-extender:487b3da4e175da96dabfb44fba65cdb8b823db2e"
     error_server                    = "quay.io/coreos/tectonic-error-server:1.0"
@@ -52,16 +52,16 @@ variable "tectonic_container_images" {
     kubedns                         = "gcr.io/google_containers/k8s-dns-kube-dns-amd64:1.14.4"
     kubednsmasq                     = "gcr.io/google_containers/k8s-dns-dnsmasq-nanny-amd64:1.14.4"
     kubedns_sidecar                 = "gcr.io/google_containers/k8s-dns-sidecar-amd64:1.14.4"
-    flannel                         = "quay.io/coreos/flannel:v0.7.1-amd64"
+    flannel                         = "quay.io/coreos/flannel:v0.8.0-amd64"
     flannel_cni                     = "quay.io/coreos/flannel-cni:0.1.0"
     etcd                            = "quay.io/coreos/etcd:v3.1.8"
-    etcd_operator                   = "quay.io/coreos/etcd-operator:v0.4.0"
+    etcd_operator                   = "quay.io/coreos/etcd-operator:v0.4.2"
     kenc                            = "quay.io/coreos/kenc:8f6e2e885f790030fbbb0496ea2a2d8830e58b8f"
-    calico                          = "quay.io/calico/node:v1.3.0"
-    calico_cni                      = "quay.io/calico/cni:v1.9.1-4-g23fcd5f"
     awscli                          = "quay.io/coreos/awscli:025a357f05242fdad6a81e8a6b520098aa65a600"
     kube_version                    = "quay.io/coreos/kube-version:0.1.0"
     tectonic_etcd_operator          = "quay.io/coreos/tectonic-etcd-operator:v0.0.1"
+    calico                          = "quay.io/calico/node:v1.3.0"
+    calico_cni                      = "quay.io/calico/cni:v1.9.1-4-g23fcd5f"
   }
 }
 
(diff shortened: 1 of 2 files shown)


Quentin-M:
Looks like all AWS/Bare-metal tests passed successfully.
I'd like to deploy that manually and try it extensively on Monday.

/cc @sym3tri @aaronlevy

s-urbaniak:
AWS went green, restarted due to Azure flake, opened #1420 to track this flake.

s-urbaniak:
Oh, and btw ... this is great! Small change, big impact ;-) 👍

s-urbaniak:
ok to test

aaronlevy:
I'd also like to see the conformance tests updated/passed for this. Our PR tests do pretty minimal validation.

Quentin-M:
ok to test

Quentin-M:
I'm about to run the conformance tests.

Quentin-M:
There is a cluster up with that PR. Please ask me in private to get access to it and test it.

Quentin-M:
Conformance tests report:

```
Summarizing 3 Failures:

[Fail] [k8s.io] SchedulerPredicates [Serial] [BeforeEach] validates that NodeSelector is respected if not matching [Conformance]
/go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/test/e2e/scheduling/predicates.go:84

[Fail] [k8s.io] SchedulerPredicates [Serial] [BeforeEach] validates resource limits of pods that are allowed to run [Conformance]
/go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/test/e2e/scheduling/predicates.go:84

[Fail] [k8s.io] SchedulerPredicates [Serial] [BeforeEach] validates that NodeSelector is respected if matching [Conformance]
/go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/test/e2e/scheduling/predicates.go:84

Ran 149 of 651 Specs in 3579.389 seconds
```

Quentin-M:
Reproduced these three failures twice ^

Quentin-M:
We've seen the following one consistently on 1.6.7:
```
[Fail] [k8s.io] SchedulerPredicates [Serial] [It] validates that NodeSelector is respected if matching [Conformance] 
/go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/test/e2e/scheduling/predicates.go:286
```

s-urbaniak:
Another observation from within the tectonic service:
```
$ journalctl -u tectonic
Jul 19 12:23:40 tectonic-master-1 bash[3798]: [  397.829608] hyperkube[5]: 2017-07-19 12:23:40.513641 I | proto: duplicate proto type registered: google.protobuf.Any
Jul 19 12:23:40 tectonic-master-1 bash[3798]: [  397.830366] hyperkube[5]: 2017-07-19 12:23:40.514446 I | proto: duplicate proto type registered: google.protobuf.Duration
Jul 19 12:23:40 tectonic-master-1 bash[3798]: [  397.832052] hyperkube[5]: 2017-07-19 12:23:40.516128 I | proto: duplicate proto type registered: google.protobuf.Timestamp
```

s-urbaniak:
I have one failed cluster deployment.

My initial observation is that `tectonic-console` (being of type `NodePort`) got a port `32001` assigned, which conflicts with `tectonic-lb`:
```
$ kubectl get svc --all-namespaces 
NAMESPACE         NAME                                           CLUSTER-IP     EXTERNAL-IP   PORT(S)             AGE
...
tectonic-system   tectonic-console                               10.3.131.8     <nodes>       80:32001/TCP        33m
```

I am wondering though, that we did not hit this earlier.

s-urbaniak:
Since we don't set `--service-node-port-range` on the API server and the default service node port range is `30000-32768` I just hit a 3/2768 chance of getting the ports 32000/32001/32002 for the console, which is created before tectonic-lb ... what a lucky incident!

Quentin-M:
Just verified with the etcd team that etcd v3.1.8 and etcd-operator v0.4.2 are good for this release.

Quentin-M:
Conformance tests on 1.7.1:
```
Summarizing 1 Failure:

[Fail] [k8s.io] SchedulerPredicates [Serial] [It] validates that NodeSelector is respected if matching [Conformance]
/go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/test/e2e/scheduling/predicates.go:292

Ran 149 of 652 Specs in 3473.418 seconds
FAIL! -- 148 Passed | 1 Failed | 0 Pending | 503 Skipped --- FAIL: TestE2E (3473.46s)
FAIL
```

squat:
@rithujohn191 can you PTAL? Is that one of our known failures?

Quentin-M:
@squat This is the one failure we have every single time (since 1.6.x). The etcd request times out for the particular condition that waits for the node to be labeled properly.

rithujohn191:
Yes this was 1 among the 5 known failures. After including a fix that @ethernetdan got in to 1.7, 4 of them got fixed.

Quentin-M:
This is what is reported by Kubernetes about our AWS nodes:
```
Addresses:
  InternalIP:   10.0.35.57
  InternalDNS:  ip-10-0-35-57.ca-central-1.compute.internal
  Hostname:     ip-10-0-35-57
```
The hostname matches the hostname that is set on CoreOS Linux. Therefore the following node label is defined: `kubernetes.io/hostname=ip-10-0-35-57`. The node names, however are of the form: `ip-10-0-35-57.ca-central-1.compute.internal`.

The test runs a pod on a random node that can schedule the pod and [gets](https://github.com/kubernetes/kubernetes/blob/1dc5c66/test/e2e/scheduling/predicates.go#L800) its `NodeName` to schedule a pod that has an extra label. The NodeSelector is `kubernetes.io/hostname=ip-10-0-35-57.ca-central-1.compute.internal` + the extra label.

This is why the pod cannot be scheduled and the test fails.

Quentin-M:
Might be related to https://github.com/coreos/bugs/issues/1983.

euank:
Probably unrelated to that bug.

Those two values are determined via the following on aws:

* Node name &mdash; Use AWS's api for private ip  (see [kubelet](https://github.com/kubernetes/kubernetes/blob/v1.7.1/pkg/kubelet/kubelet.go#L318-L325) which depends on [aws](https://github.com/kubernetes/kubernetes/blob/v1.7.1/pkg/cloudprovider/providers/aws/aws.go#L3343-L3346)).
* Hostname &mdash; Call the go `os.Hostname` (see [here](https://github.com/kubernetes/kubernetes/blob/v1.7.1/pkg/util/node/node.go#L43-L54))

Go doesn't depend on the os libs or hostname binary, so that bug is unrelated (not to mention that it's running in a chroot with a diff hostname binary anyways). Furthermore, there's not really much on the os we can do to change it.

Similarly, we can't control AWS's output.

Assuming those two things are the same is obviously wrong though. The real issue is just that the upstream test seems to be making a broken assumption.

Quentin-M:
```
$ /home/core/hostname --fqdn
ip-10-0-16-137.ca-central-1.compute.internal

core@ip-10-0-16-137 ~ $ hostname --fqdn
ip-10-0-16-137
```

Quentin-M:
We could fix with a `sudo /usr/bin/hostnamectl set-hostname $(curl -s http://169.254.169.254/latest/meta-data/local-hostname)` before kubelet on AWS.

euank:
@Quentin-M 

The less hacky way to do that would be setting the kubelet's `--node-name=$EC2_HOSTNAME` + using coreos-metadata's generated environmentfile for getting that value.

However, the proper fix is to fix the upstream test. The test is not trying to determine that your hostname matches your nodename. It's simply a poorly written test which should be updated to not accidentally make that assumption.

Quentin-M:
@euank Yeah. Using coreos-metadata templating is not practical for us. I am looking at the upstream behavior now and trying to understand whether [what they currently do](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/scheduling/predicates.go#L285) is what they actually wanted to do. In other words, if changing that will solve our case but break their cases.

aaronlevy:
Hmmm looks like a unique label is already being added to the node (https://github.com/kubernetes/kubernetes/blob/master/test/e2e/scheduling/predicates.go#L274)

I don't immediately see why they actually need the hostname label as part of the node selector (to test that a node selector is respected): https://github.com/kubernetes/kubernetes/blob/master/test/e2e/scheduling/predicates.go#L285

Might be able to just drop that line altogether

Quentin-M:
@aaronlevy Interestingly, `validates that required NodeAffinity setting is respected if matching` also have the assumption that the `hostname == nodename`, but that test does not have the `[Conformance]` tag .. so even tho we use that feature, it's not part of our E2E test suite.

Quentin-M:
Submitted https://github.com/kubernetes/kubernetes/pull/49253.

Issue #1054, opened 2017-06-12
CI failure umbrella issue


s-urbaniak:
The following failure modes have been seen lately:

https://github.com/coreos/tectonic-installer/pull/1052#issuecomment-307841897
https://github.com/coreos/tectonic-installer/pull/1053#issuecomment-307825313
https://github.com/coreos/tectonic-installer/pull/1053#issuecomment-307819614

---

EDIT: added by Ed

Things we know we need to fix:
- [ ] Jenkins dies
- [x] AWS limits
- [ ] TF destroy issues with timeouts or dependencies
- [ ] Some smoke tests are flaky
- [ ] Smoke tests timeout waiting for Tectonic components to come up
- [ ] Creating EIPs for NAT Gateways (associating the EIPs to the NAT Gateways)

Quentin-M:
Just got:

```
* module.etcd.aws_route53_record.etc_a_nodes[2]: index 2 out of range for list aws_instance.etcd_node.*.private_ip (max 2) in:

${aws_instance.etcd_node.*.private_ip[count.index]}
```

Quentin-M:
https://github.com/coreos/tectonic-installer/issues/1017 is still an issue.

Quentin-M:
From https://jenkins-tectonic-installer.prod.coreos.systems/blue/organizations/jenkins/tectonic-installer/detail/PR-968/4/pipeline:

```
[31mError applying plan:

3 error(s) occurred:

* module.vpc.aws_security_group.console (destroy): 1 error(s) occurred:

* aws_security_group.console: DependencyViolation: resource sg-f25bb183 has a dependent object
	status code: 400, request id: 52d71468-4208-4725-818a-f9e7beb361ba
* module.vpc.aws_security_group.api (destroy): 1 error(s) occurred:

* aws_security_group.api: DependencyViolation: resource sg-155cb664 has a dependent object
	status code: 400, request id: 5799d0da-f069-44b3-bf71-9b535548eb90
* module.vpc.aws_internet_gateway.igw (destroy): 1 error(s) occurred:

* aws_internet_gateway.igw: Error waiting for internet gateway (igw-72198414) to detach: couldn't find resource (31 retries)
```

The IGW issue also appeared in https://jenkins-tectonic-installer.prod.coreos.systems/blue/organizations/jenkins/tectonic-installer/detail/PR-1074/2/pipeline/ again. So https://github.com/coreos/tectonic-installer/issues/1017 is still happening.

[...]

PR #1406, opened 2017-07-14
Smoke tests: Print logs if matchbox or dnsmasq fail.

ggreer:
Should help troubleshoot #1054.


kans:
see https://github.com/coreos/tectonic-installer/issues/1054#issuecomment-314254670

Issue #1402, opened 2017-07-14
Add Calico network policy smoke test using `apply`, not just `plan`


squat:
Calico network policies are a big change. We need to add a smoke test to verify that the calico network policy works using `Terraform apply`. We can modify one of the existing tests to do this.

mxinden:
Closing here as #1410 is merged. @squat Please re-open if something is missing.

PR #1410, opened 2017-07-14
tests/smoke/: networkpolicy smoke tests

abhinavdahiya:
#1402 
Add smoke tests for network policy if calico was deployed

/cc: @squat @aaronlevy


abhinavdahiya:
ok-to-test

abhinavdahiya:
@aaronlevy

Jenkinsfile
```diff
@@ -172,18 +172,40 @@ pipeline {
...
                 withDockerContainer(params.builder_image) {
                   checkout scm
                   unstash 'installer'
-                  timeout(5) {
+                  unstash 'smoke'
+                  timeout(45) {
                     sh """#!/bin/bash -ex
                     . ${WORKSPACE}/tests/smoke/aws/smoke.sh assume-role "$TECTONIC_INSTALLER_ROLE"
                     ${WORKSPACE}/tests/smoke/aws/smoke.sh plan vars/aws-exp-np.tfvars
+                    ${WORKSPACE}/tests/smoke/aws/smoke.sh create vars/aws-exp-np.tfvars
```
cpanato: if it is not more experimental we maybe can remove the `exp` from the tfvars file. eg: `vars/aws-exp-np.tfvars` to be `vars/aws-np.tfvars` or to be `vars/aws-net-policy.tfvars`

Jenkinsfile
```diff
@@ -172,18 +172,40 @@ pipeline {
...
+                        . ${WORKSPACE}/tests/smoke/aws/smoke.sh common vars/aws-exp-np.tfvars
+                        ${WORKSPACE}/tests/smoke/aws/cluster-foreach.sh ${WORKSPACE}/tests/smoke/forensics.sh
+                        """
+                      }
+                    }
+                  }
+                  retry(3) {
+                    timeout(15) {
+                      sh """#!/bin/bash -ex
+                      . ${WORKSPACE}/tests/smoke/aws/smoke.sh assume-role "$TECTONIC_INSTALLER_ROLE"
```
mxinden: If `smoke.sh assume-role` fails three times, `smoke destroy` will never be executed, because with `-e` bash exits as soon as one command fails.

I think we don't really care whether `assume-roles` is working correctly, but we do want the tests to fail whenever `smoke.sh destroy` fails. So what I would suggest is to remove the `-e` so that the exit code of the last command (`smoke.sh destroy`) is returned and all commands are always executed.

What do you think?

Jenkinsfile
abhinavdahiya: Sure, done

Jenkinsfile
abhinavdahiya: @mxinden  if `smoke.sh assume-role` fails, I don't think `smoke destroy` will work as assume-role setups the AWS keys.... ?

Jenkinsfile
cpanato: thanks do much 👍

Jenkinsfile
mxinden: @abhinavdahiya True. I was not aware of the full functionality. Sorry for the noise.

mxinden: @abhinavdahiya Actually I think the AWS credentials should still be set due to the credential injection in the Jenkinsfile. So that should be enough to just destroy the resources. I am sorry if I am misunderstanding this here. So my question would be, why are we still running `assume-role` right before destroying the cluster?
https://github.com/coreos/tectonic-installer/blob/master/Jenkinsfile#L23

abhinavdahiya: @mxinden looks like assume-role creates temporary secrets here for use https://github.com/coreos/tectonic-installer/blob/master/tests/smoke/aws/smoke.sh#L23. I tried to use this as template for running the test (https://github.com/coreos/tectonic-installer/blob/master/Jenkinsfile#L105-L142).

mxinden: @abhinavdahiya Ok, nevermind then. Sorry for the noise. Thanks for the comments.

abhinavdahiya:
@s-urbaniak

PR #1417, opened 2017-07-15
vmware: fix etcd vs master race condition

### LINKED ISSUE #1405: vmware: race condition with non self-hosted etcd
## Versions

* Tectonic version ([release](https://github.com/coreos/tectonic-installer/releases) or commit hash):
master
* Terraform version (`terraform version`):
0.9.6+
* Platform (aws|azure|openstack|metal):
vmware

## What happened?
Commit; https://github.com/coreos/tectonic-installer/commit/710cebc04fc9665053e36238eaaaa86629c270fe introduced a race condition by having the bootkube module depend on etcd module output. In some instances this may cause failure to provision Tectonic successfully. 

In VMware provisioning all instances start getting generated in parallel. Using etcd module output as a variable in bootkube; https://github.com/coreos/tectonicinstaller/blob/master/platforms/vmware/tectonic.tf#L29 causes bootkube templates to be generated only once etcd cluster is up. This, in turn may result in `./generated` folder transferred to master node without the bootkube contents in it.

## What you expected to happen?
bootkube module should not depend on etcd. Or master/worker nodes should depend successful completion of etcd/bootkube/tectonic modules

Provision tectonic on vmware with non self hosted etcd several times.

Wanted to create the issue for tracking/documentation purposes. Will be working on a fix in the next few days.

alekssaul:
Intend to fix #1405 

**Background:**

Bootkube module requires a list of `etcd_endpoints`. In current implementation, list was generated thru modules.etcd output in platform/vmware. This creates implicit dependency for bootkube module on completion of module.etcd. On platform/vmware, all instances are (attempted to be) generated at the same time, constrained by the infrastructure  (vmware) provider's configuration limits. 

Consider a tectonic cluster with 3 etcd nodes, 2 masters and 2 works. All instance/VM generations are executed in parallel. If master nodes "win the race", bootkube module is not 100% completed (missing bootstrap assets as they depend on etcd nodes output) and `./generated` folder is SCP'ed without bootstrap manifests. If etcd nodes "win the race" of being provisioned first, module bootkube completes and all manifests are copied over to master to bootstrap the cluster succesfully.

**Change:**

Instead of relying on module.etcd output to generate bootkube assets, change transforms `tectonic_vmware_etcd_hostnames` map into fully qualified domain name list of etcd nodes. This removes the dependancy of module.bootkube-> module.etcd and ensures that bootkube assets get generated prior to etcd/master/worker nodes coming up online.

**Testing/Confirmation:**

To my best understanding, we've seen a prospect hit the issue. Working with @snsumner to confirm the issue and resolution.

### DIFF
--- a/platforms/vmware/tectonic.tf
+++ b/platforms/vmware/tectonic.tf
@@ -26,7 +26,7 @@ module "bootkube" {
   oidc_groups_claim   = "groups"
   oidc_client_id      = "tectonic-kubectl"
 
-  etcd_endpoints   = ["${module.etcd.endpoints}"]
+  etcd_endpoints   = "${formatlist("%s.%s", values(var.tectonic_vmware_etcd_hostnames), var.tectonic_base_domain)}"
   etcd_ca_cert     = "${var.tectonic_etcd_ca_cert_path}"
   etcd_client_cert = "${var.tectonic_etcd_client_cert_path}"
   etcd_client_key  = "${var.tectonic_etcd_client_key_path}"


snsumner:
I've been assisting Terradata within Tectonic install on VMware and they hit this race condition with etcd (non-experimental mode).  After provide this fix to them they got past this issue confirming at least that it solve the problem for one prospective customer.

alekssaul:
@s-urbaniak don't have access to self merge. would be great if you can assist whenever you get a chance.

s-urbaniak:
@snsumner @alekssaul thanks a lot for the verification, merging!

PR #1419, opened 2017-07-17
vmware - cloud-provider support

alekssaul:
PR to support VMware Kubernetes integration, building on the Azure cloud-provider work.

**Background:**

Upstream VMware vSphere cloud-provider documentation is located in; https://kubernetes.io/docs/getting-started-guides/vsphere/#configuring-vsphere-cloud-provider. 

At a high level:
- cloud-provider integration allows Storage Class/Dynamic storage provisioning in VMware environment
- Requires ini formatted cloud-provider configuration file to be available to; kubelet, api-server, controller-manager components.
- ini file contains variables mostly collected already with exceptions of; `tectonic_vmware_username` , `tectonic_vmware_password`. 
- [Upstream documentation](https://kubernetes.io/docs/getting-started-guides/vsphere/#configuring-vsphere-cloud-provider) notes the use of `port` parameter. Per [upstream code](https://github.com/kubernetes/kubernetes/blob/master/pkg/cloudprovider/providers/vsphere/vsphere.go#L335-L337) ; "port is a deprecated field in vsphere.conf and will be removed in future release.", thus not included in the configuration

**Changes**

- Introduces additional variables required for setting up cloud-provider, namely username/password. Currently, at runtime, user is prompted for username/password pairs require to provision the infrastructure. This set of credentials is different than the cloud-provider username/password. At provision time, end user will need additional rights in VMware infrastructure then [cloud-provider requirements](https://kubernetes.io/docs/getting-started-guides/vsphere/#configuring-vsphere-cloud-provider) for cloud-provider credentials, which only need to provision disks and attach them.
- Uses a template_file under platforms/vmware in order to generate `/etc/kubernetes/cloud/config` file contents. 
- kubelet service is configured to use `cloud-provider` and `cloud-config` flags on all (non-etcd) nodes 
- `disk.enableUUID` is added to Virtual Machine provisioning per; https://kubernetes.io/docs/getting-started-guides/vsphere/#configuring-vsphere-cloud-provider

**Need Feedback**

VMware cloud provider requires variable `vm-uuid = <VM Instance UUID of virtual machine which can be retrieved from instanceUuid property in VmConfigInfo, or also set as vc.uuid in VMX file. If empty, will be retrieved from sysfs (requires root)>`

This requirement puts constrains to implementation such as;
- cannot use ignition to provision cloud-config file with `vm-uuid` statically, as this variable is generated once the machine is provisioned
- cannot use kubernetes secrets/configmaps to set `vm-uuid` as this would restrict control-manager pod's mobility (vm-uuid is bound per Virtual Machine/instance) to a single node.
- control-manager runs as user "nobody" introducing access denied issues to volume mount for `/sys/class/dmi/id/product_serial`

The implementation attempts to walk around these constrains via;
- using host mounts for api-server, controller-manager specifically to vmware environment, so `vm-uuid` can be dynamically retrieved from `/sys/class/dmi/id/product_serial`
- sets `/sys/class/dmi/id/product_serial` to readable by un-privileged user via systemd unit drop-in to VMware only systemd unit.

Would love to hear feedback from the team for cleaner implementation.

### DIFF
--- a/modules/bootkube/assets.tf
+++ b/modules/bootkube/assets.tf
@@ -77,8 +77,8 @@ resource "template_dir" "bootkube" {
     cloud_provider             = "${var.cloud_provider}"
     cloud_provider_config      = "${var.cloud_provider_config}"
     cloud_provider_config_flag = "${var.cloud_provider_config != "" ? "- --cloud-config=/etc/kubernetes/cloud/config" : "# no cloud provider config given"}"
-    cloud_provider_volume      = "${var.cloud_provider == "vsphere" ? "- name: vspheredmi \n        hostPath: \n          path: /sys/class/dmi/id/product_serial" : ""}"
-    cloud_provider_volumemount = "${var.cloud_provider == "vsphere" ? "- mountPath: /sys/class/dmi/id/product_serial \n          name: vspheredmi" : ""}"
+    cloud_provider_volume      = "${var.volume_name == "" ? "" : "- name: ${var.volume_name} \n        hostPath: \n          path: ${var.volume_host_path}" }"
+    cloud_provider_volumemount = "${var.volume_name == "" ? "" : "- mountPath: ${var.volume_mount_path} \n          name: ${var.volume_name}" }"
 
     cluster_cidr        = "${var.cluster_cidr}"
     service_cidr        = "${var.service_cidr}"
@@ -137,8 +137,8 @@ resource "template_dir" "bootkube_bootstrap" {
     cloud_provider             = "${var.cloud_provider}"
     cloud_provider_config      = "${var.cloud_provider_config}"
     cloud_provider_config_flag = "${var.cloud_provider_config != "" ? "- --cloud-config=/etc/kubernetes/cloud/config" : "# no cloud provider config given"}"
-    cloud_provider_volume      = "${var.cloud_provider == "vsphere" ? "- name: vspheredmi \n    hostPath: \n      path: /sys/class/dmi/id/product_serial" : ""}"
-    cloud_provider_volumemount = "${var.cloud_provider == "vsphere" ? "- mountPath: /sys/class/dmi/id/product_serial \n      name: vspheredmi" : ""}"
+    cloud_provider_volume      = "${var.volume_name == "" ? "" : "- name: ${var.volume_name} \n    hostPath: \n      path: ${var.volume_host_path}" }"
+    cloud_provider_volumemount = "${var.volume_name == "" ? "" : "- mountPath: ${var.volume_mount_path} \n      name: ${var.volume_name}" }"
 
     advertise_address = "${var.advertise_address}"
     cluster_cidr      = "${var.cluster_cidr}"
--- a/modules/bootkube/assets.tf
+++ b/modules/bootkube/assets.tf
@@ -77,6 +77,8 @@ resource "template_dir" "bootkube" {
     cloud_provider             = "${var.cloud_provider}"
     cloud_provider_config      = "${var.cloud_provider_config}"
     cloud_provider_config_flag = "${var.cloud_provider_config != "" ? "- --cloud-config=/etc/kubernetes/cloud/config" : "# no cloud provider config given"}"
+    cloud_provider_volume      = "${var.cloud_provider == "vsphere" ? "- name: vspheredmi \n        hostPath: \n          path: /sys/class/dmi/id/product_serial" : ""}"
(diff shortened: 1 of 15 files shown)


platforms/vmware/main.tf
```diff
@@ -32,6 +32,20 @@ module "etcd" {
   vmware_folder           = "${vsphere_folder.tectonic_vsphere_folder.path}"
 }
 
+data "template_file" "cloud-provider" {
+  template = "${file("${path.module}/resources/cloud-config.ini")}"
+
+  vars {
+    cloud-config-username     = "${var.tectonic_vmware_username}"
```
s-urbaniak: let's use underscores, as we do for other variables, i.e. `cloud_config_username`.

modules/bootkube/assets.tf
```diff
@@ -77,6 +77,8 @@ resource "template_dir" "bootkube" {
     cloud_provider             = "${var.cloud_provider}"
     cloud_provider_config      = "${var.cloud_provider_config}"
     cloud_provider_config_flag = "${var.cloud_provider_config != "" ? "- --cloud-config=/etc/kubernetes/cloud/config" : "# no cloud provider config given"}"
+    cloud_provider_volume      = "${var.cloud_provider == "vsphere" ? "- name: vspheredmi \n        hostPath: \n          path: /sys/class/dmi/id/product_serial" : ""}"
+    cloud_provider_volumemount = "${var.cloud_provider == "vsphere" ? "- mountPath: /sys/class/dmi/id/product_serial \n          name: vspheredmi" : ""}"
```
s-urbaniak: This ternary `var.cloud_provider == "vsphere" ? .. : ..` introduces programmatic logic inside bootkube depending on only vmware. If we (for some reason) need other provider mounts, this will become hard to maintain and introduce ternary spaghetti.

Let's simply inject `volume_name`, `volume_host_path` and `volume_mount_path` from the outside, leaving default values as empty string. The var declarations would look like:
```
volume = ${var.volume_name == "" ? "" : <<EOF
      - name: ${var.volume_name}
        hostPath:
          path: ${var.volume_host_path}
EOF
```
and
```
volume_mount = ${var.volume_name == "" ? "" : <<EOF
      - mountPath: ${volume_mount_path}
        name: ${var.volume_name}
EOF
```

alekssaul:
Thanks @s-urbaniak for comments and feedback. 

Proposal of reading and redirecting output from `/sys/class/dmi/id/product_serial` into a location user nobody can read makes sense as well. I can make the necessary changes, however your feedback sounded like the current implementation looked-ok, hence I've added commits to address your review only.

I've addressed the first 2 reviews. 3rd review I could only partially fix. For your comment on changes requested on `modules/bootkube/assets.tf` file. I moved the platform logic out of module/bootkube, however terraform conditionals  doesn't seem to like HEREDOC style multi strings. When attempted to setup something like;
```
volume = ${var.volume_name == "" ? "" : <<EOF
      - name: ${var.volume_name}
        hostPath:
          path: ${var.volume_host_path}
EOF
```
I was getting error with `Error loading modules: module bootkube: Error loading .terraform/modules/0391b25f8df845582a6b6a555d42296b/assets.tf: Error reading config for template_dir[bootkube]: parse error at 1:32: expected expression but found "<"`

I created a test case to validate this with;
```
$ cat terraform.tf 
data "template_file" "heredoc-test" {
  template = "${file("./testfile.txt")}"
  test = "true"
  vars {
    testvar = "${var.test == "false" ? "" : <<EOF
Multi line
string
goes
here
EOF
}"
  }
}
$ terraform plan
Failed to load root config module: Error loading /private/tmp/heredoctest/terraform.tf: Error reading config for template_file[heredoc-test]: parse error at 1:30: expected expression but found "<"
```
Which seems to be a limitation of Terraform conditionals. I don't know the Terraform core code layout to confirm this as limitation and couldn't find much details about it based on few searches. 

Please let me know if you see something that I may be missing

alekssaul:
Baremetal GUI and Smoketest failed. Smoketest failed with setting up DNS for the cluster, not sure about why UI tests fail.

s-urbaniak:
ok to test

alekssaul:
Please do not merge, need to update

cpanato:
@alekssaul can you please update this PR as you mentioned? thanks!

s-urbaniak:
@alekssaul this PR changes manifests by adding volume mounts to the api server and controller-manager/. We cannot get this in that easily any more without ensuring a proper upgrade path in KVO. The way to move forward here is to ensure we plan and implement a corresponding bump in KVO itself.

/cc @aaronlevy @robszumski @sym3tri

alekssaul:
@s-urbaniak : Thanks for the review Sergiusz.

Your comment makes sense, I'll close the PR for now.

Issue #864, opened 2017-05-25
Add an easy way to specify OS image ID


sym3tri:
Today users can only specify "alpha", "beta", or "stable" for the Container Linux channel. There is no way to pin to a specific AMI.

squat:
This also applies to Azure.

s-urbaniak:
just one thought from my side: As soon as we pin to a specific AMI version, we should also disable CLUO deployment such that the pinned AMI version is not updated immediately. The deployment manifest would still be present in `assets.zip` such that the user still can redeploy CLUO to kick off the update process.

squat:
@s-urbaniak very good call; otherwise this would all be in vain.

sym3tri:
should be covered by: https://coreosdev.atlassian.net/browse/INST-120 now

PR #1426, opened 2017-07-17
modules/aws: add ec2_ami_override to etcd, master, worker nodes

arithx:
Adds three terraform variables which can be used to override the
AMI rather than filtering out the latest Container Linux version
for the given channel.


s-urbaniak:
@arithx thanks a lot for the contribution. CI complains about the following though:
```
misformatted files (run 'terraform fmt .' to fix): modules/aws/master-asg/master.tf
Makefile:130: recipe for target 'structure-check' failed
```

Do you mind to fix the offending file? Thanks!

s-urbaniak:
xref #864, since this PR already partially fixes it.

I have just one concern with this PR, because with this we only solve half the problem. As soon as CLUO [1] kicks in, the pinned AMI version will be updated.

To make this a complete solution, I suggest to also disable deploying CLUO [2] when an AMI is being specified.

/cc @squat @sym3tri for verification

[1] https://github.com/coreos/container-linux-update-operator
[2] https://github.com/coreos/tectonic-installer/blob/3288bad/modules/tectonic/resources/manifests/updater/operators/container-linux-update-operator.yaml

arithx:
@s-urbaniak Is there a way to specify for the CLUO update operator to not target certain nodes/node types? If not then I think that disabling it should only be done when all three node types are using the override (and maybe I should rework this into using a single variable to set all three node types to the same AMI?). Currently it would be possible to deploy every node of a type to a specific AMI while both other node types are using the default filtering method to acquire the latest AMI for the given channel, and I'm not sure that disabling CLUO for all nodes just because you're running certain nodes on an older image is the ideal functionality.

Also, is there a preferred way to actually disable it or is unsetting `tectonic_container_images["container_linux_update_operator"]` the correct way?

s-urbaniak:
@arithx CLUO iterates only over those nodes who are running the update-agent. Since we set `--manage-agent=true` in Tectonic, CLUO also creates the update-agent daemonset according to the following template: https://github.com/coreos/container-linux-update-operator/blob/489de05/pkg/operator/agent_manager.go#L130.

The podspec in the above template does not seem to be configurable, hence it simply deploys the update-agent everywhere. To make this more flexible we'd need to set `--manage-agent=false`, include our own update-agent daemonset template and make this more configurable using node selectors/taint tolerations.

I think this introduces way too much complexity. My suggestion is to collapse `tectonic_aws_[master|worker|etcd]_ec2_ami_override` into `tectonic_aws_ec2_ami_override` which instruments all node AMIs and disables CLUO at all in this case.

Since the CLUO manifest would be created anyways, the user can still `kubectl create -f cluo.yaml` after cluster creation to opt into CLUO which would update the master/worker nodes. In the case of etcd, the user would still need to ssh into the etcd node machine and unmask the locksmith service (given it will be masked if the user specifies a custom AMI, which would also needed to be implemented in this PR).

dghubble:
@arithx @s-urbaniak there have been many meetings and designs to get Tectonic away from using `-manage-agent=true`. On CLUO master, `manage-agent` is now marked deprecated. Work is underway to migrate Tectonic away from the flag and use a new internal operator (ping me). I'd ask that we not try to redesign that here, its more complex than it appears.

* https://github.com/coreos/container-linux-update-operator/issues/98
* https://github.com/coreos/container-linux-update-operator/issues/76
* https://github.com/coreos/container-linux-update-operator/pull/117

Moreover, CLUO is just the reboot coordinator. To test clusters with a nightly or one-off AMI, you want to mask update-engine for that test run. CLUO components will loop doing nothing.

Beyond that, from discussion, the nightly image will not see the latest stable/beta/alpha as an update so it shouldn't be a problem even if you leave update-engine on.

I don't think anything is needed with regard to updates or cluo.

cpanato:
Tests failed in the structure check.
it says to run `outdated docs (run 'make docs' to fix)`

Can you please run and push again?

thanks

arithx:
@cpanato ran & pushed

cpanato:
I was reviewing some PR and saw this have a conflict. @arithx can you please resolve the conflict and rebase?
I will try to get this done.

sym3tri:
I would consider booting with a particular image a separate feature/flag from disabling CLUO.

arithx:
I rebased and updated the variable to be optional.

I'm not really sure why but on my system I have `make structure-check` fail every time telling me to run `make docs`, even after I just ran `make docs`. Here's a gist of the output.

https://gist.github.com/arithx/3a2cb15dcac8f3fdac7814288df8d744

enxebre:
This looks like a necessary first and non intrusive first step.
CLUO coordination can be approached later if required. At the moment:
- track-1:
explicitly deploys node-agent.yaml
- master:
relies on kube_core_renderer

Overall this lgtm, @arithx can you:
Consider to change the variable names to include *_id - e.g ec2_ami_id
Rebase this PR with latest branch changes

enxebre:
retest this please

enxebre:
retest this please

enxebre:
retest this please

enxebre:
retest this please

PR #1427, opened 2017-07-17
Jekinsfile: Don't run baremetal GUI tests as root.

ggreer:


ggreer:
We need root to edit `/etc/hosts`. Nevermind this PR.

PR #1452, opened 2017-07-20
fix condition where both flannel and calico drop same binaries

abhinavdahiya:
Both calico install-cni & flannel install-cni drop cni binaries. [this](https://github.com/projectcalico/cni-plugin/commit/23fcd5ff1990b1c9a33c7d1f6f4c437d0e6100e5) allows us to prevent calico drop binaries that flannel already does.

PR on [projectcalico/cni-plugin](https://github.com/projectcalico/cni-plugin/pull/341)


abhinavdahiya:
@s-urbaniak fixed the link.

yifan-gu:
LGTM

PR #1462, opened 2017-07-21
monitoring: restore cAdvisor metrics

brancz:
Respective to https://github.com/coreos-inc/tectonic-prometheus-operator/pull/87 here are the changes to restore the cAdvisor container metrics, that we're accidentally removed from the kubelet's insecure port.

The TPO v1.4.0 image might not be the final release, I just already pushed that tag for this to be fully testable.

@mxinden @Gouthamve @s-urbaniak 

/cc @robszumski @sym3tri @squat @metral @Quentin-M


brancz:
Thanks, but I just connected the pieces together, thank our awesome community for contributing it before we hit the issue ourselves! :slightly_smiling_face:

Issue #1439, opened 2017-07-19
Make kubelet node labels a key=value pair


jescarri:
## FEATURE REQUEST

  - As part of https://github.com/kubernetes/kubernetes/pull/39112 the tectonic installer is setting the node labels without value `node-role.kubernetes.io/master=""` this leads to an un-usable prometheus service discovery.

Why is un-usable:
There isn't an easy way to filter controller nodes vs worker nodes, we need to scrape etcd that lives in the same nodes as the control plane.

On tectonic 1.5.x there was a master / worker labels set to true, and made k8s sd super easy.

## Versions

* Tectonic version ([release](https://github.com/coreos/tectonic-installer/releases) or commit hash): 1.6.4 / 1.6.7
* Platform (aws|azure|openstack|metal): metal/aws

## What you expected to happen?

Being able to easily filter the worker nodes and controllers.

## How to reproduce it (as minimally and precisely as possible)?

Deploy a prometheus, using the k8s node SD try to filter only controllers or only workers.

jescarri:
@joshrosso

joshrosso:
xref: https://github.com/coreos/tectonic-installer/issues/667

Somewhat related, could give us more flexibility in node labeling via the installer and, by side-effect, provide a workaround until we solve this 'the right way'.

brancz:
Yes as I was troubleshooting this yesterday, I looked at Prometheus code and realized that empty label value is basically equivalent to no label. So this is indeed the kind of change we need.

robszumski:
@brancz playing devils advocate for a sec, why not change Prometheus, as a value-less label query is a valid thing in Kubernetes.

brancz:
To be honest, I'm not sure how much effort it would be upstream. In my opinion we should support it, just a question of how breaking this is and how much effort.

luxas:
This would be quite painful to change at this point in k8s... and I guess there are other uses for valueless labels in Prom as well

dghubble:
@brancz what was the conclusion on the Prometheus side? In my testing, Prometheus 2.1 and 2.2 don't distinguish between non-existent key-value pairs and pairs with an "" value.

I'm thinking of adding another label (node-role.kubernetes.io/controller="true") to masters (i.e. controllers) in the Typhoon distro to unblock some etcd scraping work. Label not final. https://github.com/poseidon/typhoon/pull/160

brancz:
[...]

Issue #667, opened 2017-05-12
Support configuring additive nodeLabels for clusters


joshrosso:
## Story

It's desirable to be able to add custom node labels on to any nodes in a Tectonic cluster.  For example, in an AWS context, you may wish for nodes that are launched from one ASG to have a specific label representing their purpose, such as `machine/highMem`.

The installer should support this ability and inject the resulting configuration in the kubelet wrapper or systemd unit in order to ensure the label is added during node registration.

PR #1468, opened 2017-07-21
Support additive node labels

joshrosso:

Solves #667 and partially impacts #1439

### DIFF
--- a/config.tf
+++ b/config.tf
@@ -482,3 +482,27 @@ variable "tectonic_bootstrap_upgrade_cl" {
   default     = "true"
   description = "(internal) Whether to trigger a ContainerLinux upgrade on node bootstrap."
 }
+
+variable "tectonic_additional_master_node_labels" {
+  type    = "list"
+  default = [""]
+
+  description = <<EOF
+Specifies additional node labels to be added to master nodes. This correlates to the --node-labels 
+kubelet flag. All labels required by Tectonic will still be present.
+
+Example: `["tier=app", "purpose=data"]`
+EOF
+}
+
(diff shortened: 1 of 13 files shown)


config.tf
```diff
@@ -391,3 +391,27 @@ WARNING: Enabling an alpha feature means that future updates may become unsuppor
 This should only be enabled on clusters that are meant to be short-lived to begin validating the alpha feature.
 EOF
 }
+
+variable "tectonic_additional_master_node_labels" {
+  type    = "list"
+  default = [""]
+
+  description = <<EOF
+Specifies additional node labels to be added to master nodes. This translates to the --node-labels
```
squat: maybe s/translates/correlates/

config.tf
```diff
@@ -391,3 +391,27 @@ WARNING: Enabling an alpha feature means that future updates may become unsuppor
...
 EOF
 }
+
+variable "tectonic_additional_master_node_labels" {
+  type    = "list"
+  default = [""]
+
+  description = <<EOF
+Specifies additional node labels to be added to master nodes. This translates to the --node-labels 
+kubelet flag. All label(s) required by Tectonic will still be present.
```
squat: all labels can be a single label. I don't think you need `(s)`

joshrosso:
Thanks @squat. Both nits have been addressed.

I believe before merging this, we'll want to update baremetal too. I haven't had much bandwidth to look into how it's rendered. As mentioned in my original comment, seems to be based on go templates?

cpanato:
closing due to inactivity. thanks for your PR. if you think this is still useful, please re submit.

PR #1471, opened 2017-07-21
modules/tectonic/resources moves ingress installation earlier...

kans:
so you can console and stuff while other operators are coming up.  This probably saves a few minutes.


s-urbaniak:
ok to test

PR #1473, opened 2017-07-21
modules/update-payload: Add an option to enforce payload.

yifan-gu:
Be default, when upgrading from an old version (several releases behind),
the intermediate payloads will be skipped unless they are "enforced".

The "enforced" field tells tco NOT to skip this payload during
the upgrade.

### DIFF
--- a/modules/update-payload/make-update-payload.sh
+++ b/modules/update-payload/make-update-payload.sh
@@ -87,13 +87,26 @@ if [[ ${#tmpfiles[*]} > 0 ]]; then
     rm ${tmpfiles[*]}
 fi
 
+echo "Is this payload enforced? (y/N)" >&2
+
+read -r payload_enforced
+
+PAYLOAD_ENFORCED=false
+if [[ ${payload_enforced} == "y" ]]; then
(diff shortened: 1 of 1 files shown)


rithujohn191:
There seems to be a shellcheck error:
In ../modules/update-payload/make-update-payload.sh line 92:
read payload_enforced
^-- SC2162: read without -r will mangle backslashes.

yifan-gu:
@rithujohn191 Fixed.

s-urbaniak:
ok to test

s-urbaniak:
no objections from my side, but leaving the other reviews enough air to react.

aaronlevy:
@yifan-gu will this new payload field be safe for older TCO's that do not understand/expect this field?

aaronlevy:
One question, but otherwise lgtm

yifan-gu:
@aaronlevy Yes, old tco won't see the the field, and will run its old logic which updates itself first anyway. When the new tco comes up, it will know it's during the update (`targetVersion` in non-empty), and continue to upgrade.

PR #1482, opened 2017-07-24
tests: Replace Bash scripts with Ruby RSpec

### LINKED ISSUE #1445: tests: Replace Bash scripts with Ruby RSpec
Replace the current bash scripts used for testing with a proper testing framework. The decision landed on RSpec, a Ruby based testing framework.

I am setting up a simple test scenario in RSpec and will introduce it peace by peace into our Jenkins pipeline instead of a big bang roll out. This simple scenario and the helper code around it can then be used to transition all tests into this framework.

In my opinion this should include the conformance tests as well in the long run. (@rithujohn191 correct me if I am wrong.)

mxinden:
Fixes #1445


Jenkinsfile
```diff
@@ -103,6 +108,25 @@ pipeline {
       }
       steps {
         parallel (
+          "RSpec": {
+            node('worker && ec2') {
+              withCredentials(creds) {
+                checkout scm
+                unstash 'installer'
+                unstash 'smoke'
+                sh('docker build -t kubectl-ruby -f images/kubectl-ruby/Dockerfile .')
```
cpanato: do we need to build every time this image?

mxinden: @cpanato Just out of simplicity I do right now as the image is changing quite a bit. In the long run I will push it to quay.

tests/e2e/lib/cluster.rb
```diff
@@ -0,0 +1,124 @@
...
+  end
+
+  private
+
+  def env_variables
+    { 'CLUSTER' => @name, 'TF_VAR_tectonic_cluster_name' => @name }
+  end
+
+  def prepare_assets
+    puts Dir.pwd
```
cpanato: need this? or is just for debug?

tests/e2e/lib/cluster.rb
```diff
@@ -0,0 +1,124 @@
...
+    FileUtils.cp(
+      @tfvars_file.path,
+      Dir.pwd + "/../../build/#{@name}/terraform.tfvars"
+    )
+  end
+
+  def localconfig
+    puts env_variables
```
cpanato: same here

mxinden:
@alexsomesan Can you give this a review? I will bring the other test stages (azure, bare metal, ...) back in, once we are ready to merge.

tests/rspec/lib/shared_examples/k8s.rb
```diff
@@ -0,0 +1,18 @@
+require 'smoke_test'
+
+RSpec.shared_examples 'withCluster' do |tf_vars_path|
+  before(:all) do
+    AWS.check_prerequisites
+    # TODO: Infer cluster name from tfvars file name
+    @cluster = Cluster.new('aws', tf_vars_path)
```
mxinden: `aws` should be inferred from the tfvars file.

tests/rspec/Gemfile
```diff
@@ -0,0 +1,3 @@
+source 'https://rubygems.org'
+
+gem 'rspec', '~>3.0'
```
alexsomesan: Let's pin this down to the latest available and update explicitly when new versions come out.

mxinden: For documentation purposes, as discussed: Gem file versions are enforced via 'Gemfile.lock` which is checked in as well. I added a commit that executes RSpec via Bundler to make sure these versions are enforced. Thanks for the hint.

mxinden:
@alexsomesan Comment is addressed. Any other objections?

images/kubectl-terraform-ruby/Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM ruby:2.4-slim
+
+ENV TERRAFORM_VERSION="v0.9.11-coreos.2"
+
+RUN apt-get update && \
+    apt-get install --no-install-recommends -y -q \
+    curl make unzip && \
+    apt-get clean
+
+RUN gem install bundler rake rubocop rspec
```
alexsomesan: This should actually not be necessary. The ruby base image already provides bundler and rake, while rubocop and rspec should be installed via bundler from the project Gemfile (so we know which versions we run with).

tests/rspec/Gemfile
alexsomesan: Sounds good. That's a sane plan. Thanks.

tests/rspec/lib/smoke_test.rb
```diff
@@ -0,0 +1,38 @@
...
+    'SMOKE_NODE_COUNT' => cluster.tfvars_file.node_count.to_s,
+    'SMOKE_MANIFEST_PATHS' => cluster.manifest_path,
+    'SMOKE_MANIFEST_EXPERIMENTAL' => bool_to_string(
+      cluster.tfvars_file.experimental?
+    ),
+    'SMOKE_CALICO_NETWORK_POLICY' => bool_to_string(cluster.tfvars_file.calico?)
+  }
+end
+
+def bool_to_string(bool)
```
alexsomesan: There is a simpler way to achieve this in ruby - just call the `to_s` method on the boolean var:
```ruby
flag = true
flag.to_s
``

File: tests/rspec/lib/tfvars_file.rb [hunk 8]
```diff
@@ -0,0 +1,38 @@
+require 'json'
+
+# TFVarsFile represents a Terraform configuration file describing a Tectonic
+# cluster configuration
+class TFVarsFile
+  attr_reader :path, :json
+  def initialize(file_path)
+    @path = file_path
+    raise "file #{file_path} does not exist" unless file_exists?
+    @json = JSON.parse(File.read(path))
```
alexsomesan: Let's call this something neutral like `data` or `values` or something, just in case we end up also supporting parsing the key-value format of tfvars.

mxinden:
I would like to delay restructuring the `/tests` folder. See https://github.com/coreos/tectonic-installer/issues/1531

tests/rspec/spec/tfvars_file_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+TFVARS_FILE_PATH = '../smoke/aws/vars/aws.tfvars.json'.freeze
+
+RSpec.describe 'tfvars_file' do
+  before(:all) do
+    @tfvars_file = TFVarsFile.new(TFVARS_FILE_PATH)
+  end
+
+  it '.new raises exception with wrong file path' do
+    expect { TFVarsFile.new('wrong-file-path') }.to raise_error(RuntimeError)
+  end
```
GerritSe:
```ruby
context 'with a wrong file path' do
  subject { describe_class.new('wrong-file-path') }

  it { is_expected.to raise_error(RuntimeError) }
end
```

This is the one-liner syntax previously referred to. `subject` is, just like `let` lazy evaluated.

tests/rspec/spec/aws_spec.rb
```diff
@@ -0,0 +1,7 @@
+require 'cluster'
+require 'aws'
+require 'shared_examples/k8s'
+
+RSpec.describe 'aws-standard' do
+  include_examples('withCluster', '../smoke/aws/vars/aws.tfvars.json')
```
GerritSe: ???

tests/rspec/lib/shared_examples/k8s.rb
```diff
@@ -0,0 +1,18 @@
+require 'smoke_test'
+
+RSpec.shared_examples 'withCluster' do |tf_vars_path|
```
GerritSe: Shared examples are used to reuse code across various different specs. You use it only once and do not refer to the object under test at all. This should be a stand alone spec imho.

tests/rspec/lib/shared_examples/k8s.rb
```diff
@@ -0,0 +1,18 @@
...
+    @cluster = Cluster.new('aws', tf_vars_path)
+    @cluster.start
+  end
+
+  after(:all) do
+    @cluster.stop
+  end
+
+  it 'succeeds with the golang test suit' do
+    SmokeTest.run(@cluster)
```
GerritSe: Expect this to not raise an error.

tests/rspec/lib/aws.rb
```diff
@@ -0,0 +1,13 @@
+require 'env_var'
+
+# AWS contains helper functions for the AWS cloud provider
+module AWS
+  def self.check_prerequisites
+    credential_names = %w[AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY]
+    raise 'AWS credentials are not defined.' unless
+      EnvVar.set?(credential_names)
```
mxinden: Went with the first solution for now. Once the function gets out of hand I will split it up further. Good to know about the bang convention! Thanks.

tests/rspec/lib/env_var.rb
```diff
@@ -0,0 +1,6 @@
+# EnvVars contains helper functions for system environment variables
+module EnvVar
+  def self.set?(vars)
+    vars.all? { |cred| !ENV[cred].nil? }
+  end
```
mxinden: Uuh, this is fancy. Adjusted accordingly.

tests/rspec/lib/cluster.rb
```diff
@@ -0,0 +1,125 @@
...
+# Cluster represents a k8s cluster
+class Cluster
+  attr_reader :tfvars_file, :kubeconfig, :manifest_path
+
+  def initialize(prefix, tfvars_file_path)
+    # Enable local testers to specify a static cluster name
+    @name = ENV['CLUSTER']
+    @name = generate_name(prefix) if @name.nil?
+    # S3 buckets can only handle lower case names
+    @name = @name.downcase
```
mxinden: This looks a lot better. Thanks

tests/rspec/lib/cluster.rb
```diff
@@ -0,0 +1,125 @@
...
+    localconfig
+    prepare_assets
+    plan
+    apply
+    wait_til_ready
+  end
+
+  def stop
+    destroy
+  end
```
mxinden: I will probably extend this method in the near future. Should I still make it an alias? 

Didn't know about alias feature. Thanks.

tests/rspec/lib/cluster.rb
```diff
@@ -0,0 +1,125 @@
...
+
+  def amount_nodes
```
mxinden: I have removed it, as we will probably not use it in the near future. Is there any tooling around finding unused code even though it is a dynamic language?

This still looks interesting. I didn't know ruby's syntax allows the `|| =`, cool, thanks!

tests/rspec/lib/cluster.rb
```diff
@@ -0,0 +1,125 @@
...
+    out = KubeCTL.run_and_parse(@kubeconfig, 'get nodes')
+    out['items'].length
+  end
+
+  def check_prerequisites
+    license_path = 'TF_VAR_tectonic_license_path'
+    pull_secret_path = 'TF_VAR_tectonic_pull_secret_path'
+
+    return if EnvVar.set?([license_path, pull_secret_path])
```
mxinden: Moved to separate private method. Thanks for the hint.

tests/rspec/lib/cluster.rb
```diff
@@ -0,0 +1,125 @@
...
+    {
+      'CLUSTER' => @name,
+      'TF_VAR_tectonic_cluster_name' => @name
+    }
+  end
+
+  def prepare_assets
+    FileUtils.cp(
+      @tfvars_file.path,
+      Dir.pwd + "/../../build/#{@name}/terraform.tfvars"
```
mxinden: That's true, but only because they don't throw proper exceptions. `Fileutils.cp` does. Is it a bad coding style to let this exception just propagate down to RSpec?

tests/rspec/lib/cluster.rb
```diff
@@ -0,0 +1,125 @@
...
+  def destroy
+    retries = 0
+    succeeded = false
+    while retries < 3 && !succeeded
+      succeeded = system(env_variables, 'make -C ../.. destroy')
+      retries += 1
+    end
+
+    raise 'Destroying cluster failed' unless succeeded
+  end
```
mxinden: Now this is just marvellous! Changed. Thanks.

tests/rspec/lib/cluster.rb
```diff
@@ -0,0 +1,125 @@
...
+    begin
+      KubeCTL.run(@kubeconfig, 'cluster-info')
+    rescue KubectlCmdFailed
+      if retries < 100
+        retries += 1
+        sleep 10
+        retry
+      end
+    end
+  end
```
mxinden: This is a lot better to read. Thanks. Function now raises an exception as well. Thanks for raising the awareness here.

tests/rspec/lib/cluster.rb
```diff
@@ -0,0 +1,125 @@
...
+
+  MAX_NAME_LENGTH = 28
+  RANDOM_HASH_LENGTH = 5
```
mxinden: 👍 Thanks

tests/rspec/lib/cluster.rb
```diff
@@ -0,0 +1,125 @@
...
+
+    if Jenkins.environment?
+      build_id = ENV['BUILD_ID']
+      branch_name = ENV['BRANCH_NAME']
+      name = "#{prefix}-#{branch_name}-#{build_id}"
+    end
+
+    name = name[0..(MAX_NAME_LENGTH - RANDOM_HASH_LENGTH)]
+    name += SecureRandom.hex[0..RANDOM_HASH_LENGTH - 1]
+    name
```
mxinden: Changed to the triple dots. `tab` syntax seems too complicated for me as a ruby beginner, but good to know for my ruby future.

tests/rspec/lib/kubectl_helpers.rb
```diff
@@ -0,0 +1,24 @@
...
+# kubectl
+module KubeCTL
+  def self.run(kubeconfig, args)
+    out = `KUBECONFIG=#{kubeconfig} kubectl #{args}`
+    raise KubectlCmdFailed if $CHILD_STATUS.exitstatus != 0
+    out
+  end
+
+  def self.run_and_parse(kubeconfig, args)
+    out = kubectl(kubeconfig, args + ' -ojson')
```
mxinden: Good catch. Fixed. Thanks.

If `run` throws an error, `run_and_parse` will throw the error on to the initial caller, right?

tests/rspec/lib/kubectl_helpers.rb
```diff
@@ -0,0 +1,24 @@
...
+    JSON.parse(out)
+  end
+
+  # KubectlCmdFailed is raised whenever the shell command 'kubectl' fails
+  class KubectlCmdFailed < StandardError
```
mxinden: Sounds better, thanks.

tests/rspec/lib/shared_examples/k8s.rb
mxinden: So far there is only one spec, but there will be many more once this PR is merged.

The shared example will actually create the object under test (`@cluster`). The `withCluster` shared example defines a minimum set of things, every test should have, namely:
1. Spin up cluster
2. Stop cluster
3. Run the Smoketest suit

From here we can define specific test steps inside the tests themselves with additional `it` blocks, which access the `@cluster` created by the shared example.

Do you think this is a bad practice?

tests/rspec/lib/shared_examples/k8s.rb
mxinden: 👍 thanks

tests/rspec/lib/smoke_test.rb
```diff
@@ -0,0 +1,32 @@
...
+    )
+    raise 'SmokeTests failed' unless succeeded
+  end
+
+  def self.compiled?
+    File.file?('../../bin/smoke')
+  end
+end
+
+def env_variables(cluster)
```
mxinden: I don't want to export this function to the outside of this file. So I thought not to include it in the module. What would be a better way to make this private?

tests/rspec/lib/smoke_test.rb
```diff
@@ -0,0 +1,32 @@
+# SmokeTest contains helper functions to operate the smoke tests written in
+# golang
+module SmokeTest
```
mxinden: But then I need to keep a reference to the class around as well. I think I prefer the module for now. Or is there anything that speaks against modules in general? I am mostly using it to structure the code, and to emphasize, what should be used outside the file, and what not.

tests/rspec/spec/aws_spec.rb
mxinden: The plan is to add the `include_examples` to every test that needs a running k8s cluster. Do you think this is a bad practice? `withCluster` will inject a set of basic logic into the test. Specific test logic can then be added to the test e.g. with adding `it` blocks here.

tests/rspec/spec/tfvars_file_spec.rb
```diff
@@ -0,0 +1,29 @@
+require 'tfvars_file'
+
+TFVARS_FILE_PATH = '../smoke/aws/vars/aws.tfvars.json'.freeze
+
+RSpec.describe 'tfvars_file' do
```
mxinden: 👍 Cool, thanks

tests/rspec/spec/tfvars_file_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  before(:all) do
+    @tfvars_file = TFVarsFile.new(TFVARS_FILE_PATH)
+  end
```
mxinden: Ah, cool. Changed that. Wow, RSpec has a lot of features.

tests/rspec/spec/tfvars_file_spec.rb
mxinden: This is fancy, thanks! Always wondered how to catch exceptions that are raised in the prepare steps.

tests/rspec/spec/tfvars_file_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  it '.new raises exception with wrong file path' do
+    expect { TFVarsFile.new('wrong-file-path') }.to raise_error(RuntimeError)
+  end
+
+  it '.path returns the file path' do
+    expect(@tfvars_file.path).to be(TFVARS_FILE_PATH)
+  end
```
mxinden: Do I need to wrap the `describe` around each `it`, even though there is only one `it`?

I don't want to test the attribute readers in particular, but the parsing logic of the file underneath. Let me know if you can think of a better way of doing that?

tests/rspec/lib/jenkins.rb
```diff
@@ -0,0 +1,8 @@
+require 'env_var'
+
+# Jenkins contains helper functions to interact with Jenkins
```
mxinden: Good hint. Will do that once this gets out of hand.

mxinden:
@GerritSe Thanks a lot for the review! Appreciate the hidden ruby insights.

tests/rspec/lib/cluster.rb
GerritSe: Well no you don't have to. I think readability-wise there's not much of a difference.

tests/rspec/lib/cluster.rb
GerritSe: I'm not sure about that. What I do know is that there are gems to measure your test coverage. If you test conscientious and don't dynamically define methods then you should be on the right track.

tests/rspec/lib/cluster.rb
GerritSe: That's up to you. If you catch the exception and raise your own, then you lose the information from the initial exception and won't see it in the stack trace. You could also define your own custom error class like you did with `KubectlCmdError` and pass it the message of the initially raised exception.

tests/rspec/lib/smoke_test.rb
GerritSe: Not if you use a class and save the state statically. If you insist on using the module then you could just remove the `build unless compiled?` line and the `compiled?` method. It's an internal API and if it's properly communicated and documented that `build` has to be called before `run` then there shouldn't be a problem. This way you can spare yourself the file system access on EVERY call of `run`.

tests/rspec/lib/smoke_test.rb
GerritSe: Well, it is exported anyway and exposed to the global namespace as soon as you require it from another file. You cannot really cherry pick afaik. You can make the method private though:

```ruby
module SmokeTest
  ...
  def self.env_variables(cluster)
    ...
  end

  private_class_method :env_variables
end
```

tests/rspec/spec/tfvars_file_spec.rb
GerritSe: Well it's just a suggestion for structuring your tests:
```ruby
describe '#method_name' do
  it 'does this and that' do
  end

  it 'does something else with different parameters' do
  end
end
```

GerritSe: For the second part, I don't understand what you mean. What you do with this specific test mentioned is to test if the attribute reader returns the path that you assigned to it in the initializer. There is no logic under test in this method at all.

tests/rspec/lib/shared_examples/k8s.rb
GerritSe: I think I'd rather appear.in about this than to discuss this in text form :D

tests/rspec/spec/tfvars_file_spec.rb
```diff
@@ -2,28 +2,29 @@
...
-  it '.experimental? returns false if not set' do
-    expect(@tfvars_file.experimental?).to eq(false)
+  it '#experimental? returns false if not set' do
+    expect(subject.experimental?).to eq(false)
   end
 
-  it '.calico? returns false if not set' do
-    expect(@tfvars_file.experimental?).to eq(false)
+  it '#calico? returns false if not set' do
+    expect(subject.experimental?).to eq(false)
```
GerritSe: You used the wrong method name here.

tests/rspec/spec/tfvars_file_spec.rb
mxinden: Ah, yes, you are right.

tests/rspec/spec/tfvars_file_spec.rb
mxinden: Good catch, thanks.

alexsomesan:
This PR has clocked 90 test runs on Jenkins none of which went fully green for reasons not always pertaining to this change.
I've merged this since it's just the beginning of a string of changes meant to improve the testing experience, so we're going to inevitably work on the flakiness that has prevented a green run now as part of that initiative.

PR #1487, opened 2017-07-24
kvo: disable image pre-warming.

diegs:
This is broken for people upgrading from previous versions, so turn it
off until it is fixed.


diegs:
Actually I'm only going to merge this into the 1.7 branch, leaving the master correct for the next version.

squat:
mm not sure about that @diegs. if you go that route, please make sure we file an issue for a blocker for 1.7.2 to track this issue until it is fixed

diegs:
@squat this will be fixed in 1.7.2+. It relies on a `node-agent` manifest fix in the 1.7.1 upgrade: https://github.com/coreos-inc/kube-version-operator/pull/200

PR #1488, opened 2017-07-24
Fix undefined variables in upload-payload/assets.tf.

yifan-gu:
Also plans to add a check in jenkins job for each PR so 'make payload'
won't get broken.

### DIFF
--- a/modules/update-payload/assets.tf
+++ b/modules/update-payload/assets.tf
@@ -9,6 +9,7 @@ resource "template_dir" "payload_operators" {
     tectonic_channel_operator_image       = "${var.tectonic_container_images["tectonic_channel_operator"]}"
     tectonic_prometheus_operator_image    = "${var.tectonic_container_images["tectonic_prometheus_operator"]}"
     tectonic_etcd_operator_image          = "${var.tectonic_container_images["tectonic_etcd_operator"]}"
+    image_re                              = "${var.tectonic_image_re}"
   }
 }
 


yifan-gu:
Here https://github.com/yifan-gu/tectonic-installer/blob/23fc5aeb7ad1474268af80708507c92d7356e5e9/modules/update-payload/assets.tf#L3

It's used by the [cluo manifest](https://github.com/coreos/tectonic-installer/blob/master/modules/tectonic/resources/manifests/updater/operators/container-linux-update-operator.yaml#L27)

dghubble:
What is that line doing in the CLUO manifest? cc @colhom

yifan-gu:
Working on making the verification of `make payload` as part of the CI, so it will be more clear. I can either create a new PR or add commits to this one.

yifan-gu:
@dghubble 
```
@@ -31,7 +31,9 @@
               {
                 "command": [
                   "/bin/update-operator",
-                  "--manage-agent=true"
+                  "--manage-agent=true",
+                  "-agent-image-repo",
+                  "quay.io/coreos/container-linux-update-operator"
                 ],
```

dghubble:
Hm, I suppose its needed for now for the image replacement Colin is doing. It'll be removed next release when that flag is removed. A higher level operator (take to internal chat) will take on creating these manifests and if you need to edit the images it deploys, better ask for that feature soon @colhom

PR #1490, opened 2017-07-24
bootkube/tls: fix bootstrap secrets for external etcd cluster

### LINKED ISSUE #1489: Support external etcd clusters when tls is provided.
## Versions

* Tectonic version ([release](https://github.com/coreos/tectonic-installer/releases) or commit hash):
* Terraform version (`terraform version`): 1.6.7
* Platform (aws|azure|openstack|metal): metal

## What happened?
Configured terraform to point at an existing etcd cluster with tls enabled.
 
It's not trivial to parse from the examples how to do this. 

You need experimental and etcd_tls_enabled set to false. 
The code implies from either flag a need to generate certs and use them for the etcd cluster regardless of whether we are using the external etcd code path.

You would then provide the client ca client cert and client key to terraform.
In 1.6.7 this functionality doesn't work because of the following lines: 

Terraform should then use these existing certs as client certs for the bootstrap and self hosted apiservers. 

The following PR will fix this so that it will work with client certs.

Bring up a provisioned etcd cluster and configure it with tls. 

Provide the client certs to terraform with: 

```
tectonic_etcd_servers = []
tectonic_etcd_ca_cert_path = ""
tectonic_etcd_client_cert_path = ""
tectonic_etcd_client_key_path = ""
```
Ensure that you are set these to false:
```
tectonic_etcd_tls_enabled = false
tectonic_experimental = false
```

Solutions is referenced below

colhom:
fixes #1489

### DIFF
--- a/modules/bootkube/assets.tf
+++ b/modules/bootkube/assets.tf
@@ -197,24 +197,24 @@ data "template_file" "etcd_client_crt" {
 data "template_file" "etcd_client_key" {
   template = "${var.experimental_enabled || var.etcd_tls_enabled
     ? join("", tls_private_key.etcd_client.*.private_key_pem)
-    : file(var.etcd_client_cert)
+    : file(var.etcd_client_key)
   }"
 }
 
 resource "local_file" "etcd_ca_crt" {
-  count    = "${var.experimental_enabled || var.etcd_tls_enabled ? 1 : 0}"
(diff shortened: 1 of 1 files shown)


colhom:
no review needed yet

colhom:
retest this please

colhom:
baremetal test flake, ref #1498

alexsomesan:
retest this please

PR #1499, opened 2017-07-25
Bump Tectonic console to v1.8.1.

ggreer:
Includes various bug fixes that we've made over the past 4 days.


ggreer:
Adding next-rc label because this should be cherry-picked into the installer release branch.

Issue #363, opened 2017-04-25
GUI: last screen should have additional explanations


sym3tri:
Some things we are missing. Maybe link to a website doc with all of these things:

- [ ] How to generate your kubeconfig
- [ ] How to shutdown the installer
- [ ] What to do with the assets you downloaded
- [ ] How to destroy your cluster

meghanschofield:
Current state:

<img width="526" alt="screen shot 2017-06-22 at 4 20 46 pm" src="https://user-images.githubusercontent.com/3373315/27460221-4fccbb6c-5767-11e7-9f3c-e27620507e8e.png">

I'll be taking a look at a revamp of this page! Updates coming soon.

rithujohn191:
Any updates on this @meghanschofield. We were trying to scope this out for the next sprint.

meghanschofield:
Ed and I caught up on this task just now. Two things are needed:

1. A new cleaner screen for Installation Complete. 
- with "go to console" primary button
- with "be sure to download assets" secondary button (and related reasons)
- removing: back, start over, tutorial and kubctl stuff. Tutorial and kubctl stuff will be moved to the first page in the console. 

2. Clicking "Go to my console" takes users to a Quick Start page to include:
- set up kubectl (new tectonic specific doc needed that explains setting up kubectl in the console and in general)
- set up your first application tutorial
- information on how to destroy a cluster
- a section for more advanced resources

I'll get work on this stuff asap!

meghanschofield:
Updated installer mock for the Installation Complete Page:

<img width="532" alt="screen shot 2017-07-19 at 4 10 58 pm" src="https://user-images.githubusercontent.com/3373315/28393674-f72e3c7e-6c9c-11e7-8f30-a9396e7ee46f.png">

##### Callouts and notes:

<img width="649" alt="screen shot 2017-07-19 at 4 11 07 pm" src="https://user-images.githubusercontent.com/3373315/28393675-f89c2ab2-6c9c-11e7-9386-b0e66b5153f0.png">

##### General Notes
- Users can still "go back" by clicking the link in the side nav. 
- This gives users a way to close the installer


##### Content for easy cut and paste
All set! Now you can access the Tectonic Console. Once there, you’ll be able to configure kubectl, and deploy your first application to the cluster.

The Tectonic Console gives you an easy-to-navigate view of your cluster. 

Cluster assets are important, save them now! 

[...]

PR #1502, opened 2017-07-26
frontend: Installation Complete screen assets download and UI changes

kyoto:
Partially implements #363.
  - Add assets download area
  - Text changes

The "Install kubectl" and "Deploy Your First Application" sections will be removed in a separate PR after that info has been moved to Console.

screenshot-1

### DIFF
--- a/installer/frontend/components/success.jsx
+++ b/installer/frontend/components/success.jsx
@@ -21,9 +21,8 @@ export const Success = connect(stateToProps)(
 ({navigatePrevious, tectonicConsole, platformType}) => <div>
   <div className="row">
     <div className="col-xs-12">
-      <p>
-        All set! Now you can access your Tectonic Console, configure kubectl, and deploy your first application to your cluster.
-      </p>
+      <p>All set! Now you can access the Tectonic Console. Once there, you’ll be able to configure kubectl, and deploy your first application to the cluster.</p>
+      <p>The Tectonic Console gives you an easy-to-navigate view of your cluster.</p>
     </div>
   </div>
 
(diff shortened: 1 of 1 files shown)


kyoto:
ok to test

alexsomesan:
cc: @kans @ggreer

kyoto:
RFR @tlwu2013 @squat

installer/frontend/components/success.jsx
```diff
@@ -32,10 +31,25 @@ export const Success = connect(stateToProps)(
       <a href={`https://${tectonicConsole}`} target="_blank">
         <button className="btn btn-primary wiz-giant-button"
                 style={{marginTop: 20}}
-                onClick={() => handleAllDone(platformType)}>Go to my Tectonic Console&nbsp;&nbsp;<i className="fa fa-external-link"></i></button>
+                onClick={() => handleAllDone(platformType)}>Go to my Tectonic Console</button>
```
ggreer: Clicking this opens a new window. Don't we want to keep the icon so the user knows what to expect?

kyoto:
@tlwu2013 Could you answer Geoff's questions regarding the UX choices?

tlwu2013:
Hi @kyoto , @ggreer , 

From our user testings, test participants mostly clicked around the side-nav when they were trying to go back. 

But I also think having a "Back" button at the bottom left is no harm (plus it's an existing pattern within installer too). So chat with more folks and update our final mock here.

I am still working on the Console "First time use" page. So I personally prefer we can wait until the Console changes are ready, and then merge this installer PR at the same time.

Thanks!

ggreer:
OK. I don't like long-lived PRs, so I think the best thing to do here would be to put the back button back and merge it.

kyoto:
@ggreer @tlwu2013 OK, I've reduced this PR so that it no longer removes the back link or button icon and I've updated the description / screenshot.

PR #1503, opened 2017-07-26
frontend: Fix Platform default to be AWS using GUI

kyoto:
Platform dropdown was defaulting to AWS manual install instead of AWS GUI install.

If AWS GUI install is disabled using the `-platforms` command line option, the default will be whichever option appears first in the dropdown.


alexsomesan:
Thanks for catching this.
I don't have enough background in that area of the UI code.
Deferring to @ggreer for assessment.

alexsomesan:
ok to test

Issue #1311, opened 2017-07-07
GUI: after delete, next button goes to "login to console" step


robszumski:
Once you successfully delete a cluster, the next step button is still active and goes to the "here's how you use your console" screen, but the is no more cluster.

Seems like the best place is to go back to the start of the installer? Ideally we remove it completely but we might have to keep there?

kyoto:
I think this can best be fixed by disabling the Next Step button if the cluster is destroyed: #1505.

PR #1505, opened 2017-07-26
Destroy cluster disable next step

kyoto:
Disables the Next Step button as soon as the destroy action begins (rather than waiting for the destroy action to complete). This avoids the situation where you could be viewing the Installation Complete page while the destroy is running and then be unexpectedly thrown back to the Start Installation page when the destroy action completes.

Also fixes a bug in the logic that should leave the Next Button disabled while `status` is `"Running" `.

Fixes: #1311 

screenshot-1


alexsomesan:
Good catch! Thanks a lot for contributing.

CC: @ggreer for confirmation.

alexsomesan:
ok to test

PR #1506, opened 2017-07-26
Expose module outputs in platform module

discordianfish:
Hi,

I propose to expose the platform outputs. This allows to easily integrate the tectonic-installer as module in other terraform deployments. For example you can add additional resources like SG within terraform managed subnets etc.
We're using this to create special worker nodes with public IP addresses attached for example. With these outputs I can just reuse the azure/worker module and pass in the existing parameters.

I believe this will be useful for most users of the tectonic-installer so hope you can merge it. For consistency I guess it should be done for all other platforms too. What do you think?

### DIFF
--- a/platforms/azure/output.tf
+++ b/platforms/azure/output.tf
@@ -1,0 +1,147 @@
+output "bootkube_ca_cert" {
+  value = "${module.bootkube.ca_cert}"
+}
+
+output "bootkube_ca_key" {
+  value = "${module.bootkube.ca_key}"
+}
+
+output "bootkube_ca_key_alg" {
+  value = "${module.bootkube.ca_key_alg}"
+}
+
+output "bootkube_etcd_ca_crt_pem" {
+  value = "${module.bootkube.etcd_ca_crt_pem}"
+}
+
+output "bootkube_etcd_client_crt_pem" {
+  value = "${module.bootkube.etcd_client_crt_pem}"
+}
+
+output "bootkube_etcd_client_key_pem" {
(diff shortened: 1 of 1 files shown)


alexsomesan:
I'm quite concerned about this change, to be honest.

Root module outputs like these will be printed out in the apply log. Apart from being a very verbose addition, it prints out sensitive information like private keys and credential fragments. This has implications for the contents of CI logs, forcing users to treat them like secret material.
More so, there is no way to selectively enable / disable which outputs get printed out.

Users unfamiliar with the code might not expect that the installer outputs secret materials to standard out. Doing public demos is one scenario where I would not want this to happen, for example.

There are other alternatives to integrating the installer with additional Terraform resources, such as referencing the state produced by the installer via [remote state data sources](https://www.terraform.io/docs/providers/terraform/d/remote_state.html).
Going this route avoids the side-effects I mentioned above.

I'm not sure this change is beneficial for the general use-case.

discordianfish:
I think remote_state also only works with root level outputs, so would require the same change, no?

> Only the root level outputs from the remote state are accessible. Outputs from modules within the state cannot be accessed. If you want a module output to be accessible via a remote state, you must thread the output through to a root output.

It would also mean I need to manage multiple terraform states.

I fully understand the concerns of leaking private information. Then again, this is a general problem with terraform and tracked in https://github.com/hashicorp/terraform/issues/516.

What about we hide the outputs selectively? I don't think this is yet supported natively but it should work to have a list of outputs that should be shown/hidden. Frankly, not sure it's worth it given that the TF output still can contain sensitive information but whatever makes you happy. :)

PR #1508, opened 2017-07-26
aws: create tectonic_aws_external_whitelisted_cidrs

tomwans:
This variable would allow users to specify a group of whitelisted IPs
that are allowed to connect to the API and Console ELB.

This was created so that tectonic clusters can be created without
exposing them to the internet outright, at a network level.


coresolve:
Thanks for the PR! Looks like a reasonable addition.
Ok to test

tomwans:
Thanks @coresolve — I see the continuous-integration/jenkins/pr-merge step failed; why is that? I can't log in since it seems to be a CoreOS-internal thing.

cpanato:
@tomwans please rebase and resolve the conflicts. If this not makes sense anymore please close the PR.

thanks!

tomwans:
@cpanato we're good; this isn't needed for our setup anymore.

PR #1511, opened 2017-07-26
docs: update rhel installer doc

kbrwn:


Documentation/install/rhel/installing-workers.md
```diff
@@ -124,6 +129,14 @@ $ firewall-cmd --add-port 4789/udp --permanent
 
 Note: These settings may not be all inclusive and will not represent relative node ports or other communications which may need to be performed. For more information consult the [Kubernetes Networking][k8s-networking] documentation.
 
+### SELinux
+
+To allow the Tectonic worker, run SELinux in Permissive mode. Running in Enforcing mode will block permissions for worker nodes.
+
```
zbwright: To allow the Tectonic worker to what?

Documentation/install/rhel/installing-workers.md
```diff
@@ -36,8 +36,15 @@ Once basic installation of a host is complete, ensure that the additional Red Ha
 $ subscription-manager repos --enable=rhel-7-server-extras-rpms
 ```
 
+This can also be accomplished with yum-config-manager:
+
```
zbwright: change to 'Or, use `yum-config-manager`:

Documentation/install/rhel/installing-workers.md
```diff
@@ -124,6 +129,14 @@ $ firewall-cmd --add-port 4789/udp --permanent
 
 Note: These settings may not be all inclusive and will not represent relative node ports or other communications which may need to be performed. For more information consult the [Kubernetes Networking][k8s-networking] documentation.
 
+### SELinux
```
zbwright: change to 'Configure SELinux' or 'Set SELinux to Permissive mode' or something descriptive.

Documentation/install/rhel/installing-workers.md
```diff
@@ -36,8 +36,15 @@ Once basic installation of a host is complete, ensure that the additional Red Ha
 $ subscription-manager repos --enable=rhel-7-server-extras-rpms
 ```
 
+This can also be accomplished with yum-config-manager:
+
+```
+$ yum-config-manager --enable rhel-7-server-extras-rpms
+```
+
 If `subscription-manager` is not in use, ensure that the correct URL for the mirror of extras that is to be used is placed in the corresponding file in `/etc/yum.repos.d` and set to `enabled`.
```
robszumski: does this line need to be removed, since (I assume) `yum-config-manager` is the case for when subscription manager isn't in use?

PR #1512, opened 2017-07-26
Update signing scripts

yifan-gu:
Enable signing a staging payload without yubikey.
Enable overwriting the staging payload when submitting the release payload.

### DIFF
--- a/modules/update-payload/sign-payload.sh
+++ b/modules/update-payload/sign-payload.sh
@@ -2,38 +2,95 @@
 
 set -e -o pipefail
 
+TEST_GPG_ID="AEDE60EDD314382875F4B2C279C51CC59F11DE78"
+TEST_GPG_KEY="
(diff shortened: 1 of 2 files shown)


yifan-gu:
Tested manually, by default it has the same effect as before unless `SIGN_WITH_TEST_KEY=true` is set when running `./sign-payload.sh`

yifan-gu:
I will update the QA / release doc to reflect the changes.

yifan-gu:
Changes in docs https://github.com/coreos-inc/tectonic/pull/1751

Quentin-M:
Thank you~

Quentin-M:
I'm sorry, I didn't see it failed with 

```
In ../modules/update-payload/sign-payload.sh line 52:
  trap "rm -rf '${GNUPGHOME}'" EXIT
                ^-- SC2064: Use single quotes, otherwise this expands now rather than when signalled.
```

Issue #1515, opened 2017-07-27
Make default prometheus spec.resources configurable (cpu memory requests & limits)


joshrosso:
## FEATURE REQUEST

Tectonic installations [provision a prometheus](https://github.com/coreos/tectonic-installer/blob/master/modules/tectonic/resources/manifests/monitoring/prometheus-k8s.yaml) used by the Tectonic console.

Based on observations in production scenarios, we've found this spec.resource.limits to be under what's feasible for how our Prometheus is tuned. Regardless, it seems desirable to be able to set these limits from the installer so that prometheus is provisioned in your desired state OOTB.

This feature request includes the desire to make the following fields configurable in the prometheus `prometheus-k8s.yaml`.

- `spec.resources.limits.*`
- `spec.resources.requests.*`

## Versions

* Tectonic version [1.6.7-tectonic.1](https://github.com/coreos/tectonic-installer/releases/tag/1.6.7-tectonic.1)
* Terraform version (`*`)
* Platform (*)

sym3tri:
this is now available

PR #1519, opened 2017-07-27
Make prometheus-k8s limits and requests configurable

joshrosso:
- Provides more flexibility in what cluster resource allocations are
provided for prometheus during a Tectonic cluster install.

Issue: https://github.com/coreos/tectonic-installer/issues/1515

### DIFF
--- a/Documentation/variables/config.md
+++ b/Documentation/variables/config.md
@@ -32,6 +32,10 @@ This document gives an overview of variables used in all platforms of the Tecton
 | tectonic_image_re | (internal) Regular expression used to extract repo and tag components | string | `/^([^/]+/[^/]+/[^/]+):(.*)$/` |
 | tectonic_license_path | The path to the tectonic licence file.<br><br>Note: This field MUST be set manually prior to creating the cluster unless `tectonic_vanilla_k8s` is set to `true`. | string | `` |
 | tectonic_master_count | The number of master nodes to be created. This applies only to cloud platforms. | string | `1` |
+| tectonic_prometheus_k8s_limit_cpu | (optional) Amount of cpu, enforce by cgroups, prometheus may consume. | string | `` |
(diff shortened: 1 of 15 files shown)


alexsomesan:
This looks quite useful to have configurable.

One thing I would suggest, to make it more future proof and also reduce clutter in the inputs, could we make this a generic `tectonic_resource_limits` variable that is actually a map of entries for any component.
I'm thinking something along the lines of `tectonic_container_images`, but with keys like `prometheus_k8s_limit_cpu`.
What do you think?

alexsomesan:
Also, cc-ing monitoring folks: @brancz @fabxc

brancz:
Actually these values will immediately be overridden by the TPO again. You need to also configure the TPO to pick up these values: https://github.com/coreos-inc/tectonic-prometheus-operator/blob/master/Documentation/user-guides/configuring-tectonic-monitoring.md#prometheusk8sconfig

joshrosso:
@alexsomesan 

The map does make sense and seems far more extensible, I'll add that.

@brancz Thanks. To be clear, do the resources in the initial `Prometheus` manifest still matter? Or should I revert that change and only set these in the configmap you linked above?

brancz:
@joshrosso Sorry I wasn't clear enough, yes you still need the resources in the intitial manifests, and additionally in the `tectonic-monitoring` config.

joshrosso:
I've addressed the configmap, @brancz if you can take a look and make sure I understood correctly. Seems to persist in my test cluster.

@alexsomesan I'll be refactoring into the more generic map soon. Just want to get/test this out into some clusters first.

brancz:
Generally lgtm, just make sure to know that we started this effort: #1656

cpanato:
@joshrosso please rebase and resolve the conflicts if this still makes sense. otherwise please close the PR.

sym3tri:
Hey @joshrosso thanks for the contribution, but we won't be able to merge this now that the manifests are under the control of the operator. I've added a note to jira to consider providing these as top-level config values to the operator config instead.

PR #1523, opened 2017-07-27
Configurable monitoring images

colhom:
Note: until https://github.com/coreos-inc/tectonic-prometheus-operator/pull/90 is merged and rolled into a release, `nodeExporter.baseImage` and `kubeStateMetrics.baseImage` will have no effect.

\cc @brancz

### DIFF
--- a/modules/tectonic/resources/manifests/updater/tectonic-monitoring-config.yaml
+++ b/modules/tectonic/resources/manifests/updater/tectonic-monitoring-config.yaml
@@ -19,3 +19,5 @@ data:
       baseImage: ${replace(tectonic_monitoring_auth_image,image_re,"$1")}
     nodeExporter:
       baseImage: ${replace(node_exporter_image,image_re,"$1")}
+    kubeStateMetrics:
+      baseImage: ${replace(kube_state_metrics_image,image_re,"$1")}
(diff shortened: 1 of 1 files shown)


brancz:
I'd like to hold off on this, until the next dev cycle since we're already past the code freeze.

squat:
@brancz we use a release branch so that development is not blocked while development for the release is frozen.

brancz:
Got it. In any case our workflow is that we merge functionality into the TPO first, which has not happened yet. As soon as that happens we can jump back to this though.

brancz:
We've now merged this functionality, but just to make sure we're not making a mistake here, this only works with latest master of the TPO, not v1.4.x which is going out with the next release.

colhom:
sgtm

cpanato:
@brancz can we merge this PR or keep it?

cpanato:
we need to run the smoke tests right?

alexsomesan:
Running smokes can't hurt, even though I don't think we have coverage for this case yet.

Let's run them though.

cpanato:
timeout in one test.

s-urbaniak:
closing this PR, as @brancz is addressing configurability of prom images differently now.

brancz:
(while making sure these config options will be available :wink: )

PR #1528, opened 2017-07-28
Makefile: add a structure check for the update payload

Quentin-M:
Yifan recently spotted inconsistencies between the state of the installer and the update payload, at a few different places, including in our most recent RC. 

This payload is a very critical part of Tectonic. I suggest we start considering the update payload as a first-class citizen by verifying its up to date status in every PR. Thus, every PR that would modify the update payload at the end of the chain, will be required to update it as part of the PR. Annoying conflicts might potentially appear, but it should be very limited as we rarely modify these managed manifests.

/cc @alexsomesan @s-urbaniak @sym3tri

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -77,6 +77,7 @@ pipeline {
               # TODO: Remove me.
               go get github.com/segmentio/terraform-docs
               go get github.com/s-urbaniak/terraform-examples
+              go get github.com/bronze1man/yaml2json
 
               cd $GO_PROJECT/
               make structure-check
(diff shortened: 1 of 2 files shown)


Quentin-M:
`Require yaml2json (https://github.com/bronze1man/yaml2json)`

The builder image is missing that dependency. I will follow up in a second.

Quentin-M:
ok to test

Quentin-M:
ok to test

Quentin-M:
Smoke tests experienced AWS time-out and usual Azure failures due to bad subscription. Not related to the added structure-check, which will simplify release work for the time being. Can we merge?

squat:
@Quentin-M merge away

Quentin-M:
Thank you~

PR #1535, opened 2017-07-31
frontend: Add progress bars to Start Installation page

### LINKED ISSUE #562: Add progress bars to progress screen
For the apply step, this will be timed to the average deploy time and stop at 80% if we run over. This is because Terraform gives us no status info.

For downloading and starting pods, we will use the embedded Kubernetes (https://github.com/coreos/tectonic-installer/issues/561) client to measure progress.

kyoto:
Adds progress bars for both the AWS Terraform apply step and for the Starting Tectonic step for both AWS and bare metal.

The Terraform progress is estimated by parsing the log output. This PR uses a very simple approach with the intention of improving the progress bar accuracy in a future PR.

The Tectonic progress is based on the sub-steps, except that the progress bar will not move backwards of a sub-step flips off again.

Makes some CSS changes to improve horizontal alignment of the page elements.

Adds back the install step "cancel" state, introduced by #1274, which seems to have been
accidentally removed and changes its text color to match the pending state.

Fixes #562

screenshot-1
screenshot-1
screenshot-3
screenshot-4

### DIFF
--- a/installer/frontend/components/tf-poweron.jsx
+++ b/installer/frontend/components/tf-poweron.jsx
@@ -39,7 +39,7 @@ export const TF_PowerOn = connect(stateToProps, dispatchToProps)(
 class TF_PowerOn extends React.Component {
   constructor (props) {
     super(props);
-    this.state = {showLogs: null};
+    this.state = {showLogs: null, xhrError: null};
   }
 
   componentDidMount () {
--- a/installer/frontend/components/tf-poweron.jsx
+++ b/installer/frontend/components/tf-poweron.jsx
@@ -209,35 +209,35 @@ class TF_PowerOn extends React.Component {
                 </div>
               }
             </WaitingLi>
-            <li style={{paddingLeft: 20, listStyle: 'none'}}>
-            { state.xhrError &&
-              <div className="row">
-                <div className="col-xs-12">
-                  <Alert severity="error">{state.xhrError}</Alert>
+            <li style={{paddingLeft: 22, listStyle: 'none'}}>
+              { state.xhrError &&
+                <div className="row">
+                  <div className="col-xs-12">
+                    <Alert severity="error">{state.xhrError}</Alert>
+                  </div>
                 </div>
-              </div>
-            }
-            { tfError && <Alert severity="error">{tfError.toString()}</Alert> }
-            { !terraformRunning && tfError &&
-              <Alert severity="error" noIcon>
-                <b>{_.startCase(action)} Failed</b>. Your installation is blocked. To continue:
-                <ol style={{ paddingLeft: 30, paddingTop: 10, paddingBottom: 10 }}>
-                  <li>Save your logs for debugging purposes.</li>
-                  <li>Destroy your cluster to clear anything that may have been created.</li>
-                  <li>Reapply Terraform.</li>
-                </ol>
-                {tfButtons}
-              </Alert>
-            }
-            { !terraformRunning && !tfError &&
-              <Alert severity="info" noIcon>
-                <b>{_.startCase(action)} Succeeded</b>.
-                <p>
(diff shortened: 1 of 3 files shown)


tlwu2013:
Hi @kyoto , 

Looking good, thanks!
Just want to paste our discussions on Slack here for tracking:

**[Tony]**
one quick feedback is that 
_for this “Start Installation” screen, “Starting Tectonic” step should be activated after “Terraform apply success”_.
that way, there should be only one progress bar on the screen

not sure if making ” _‘Starting Tectonic’ step should be activated after ‘Terraform apply success’_ ” should be separated out into a different PR or done in this PR.

But if it’s going to be a separate PR, then I feel it’ll be better if we merge that logic first, then merge this progress bar PR.

Thanks Andy!

**[Andy]**
ok, I’ll add that change to this PR
thanks!

kyoto:
Thanks @tlwu2013. I made that change and updated the screenshots.

kyoto:
@tlwu2013 Updated again. I also applied the "inactive" style to the "Resolving..." step, which I think better matches the logic of what the installation is doing.

kyoto:
ok to test

installer/frontend/components/tf-poweron.jsx
```diff
@@ -13,6 +13,20 @@ import { CLUSTER_NAME, PLATFORM_TYPE, getTectonicDomain } from '../cluster-confi
...
+const ProgressBar = ({progress}) => <div className="wiz-launch-progress__bar">
+  <div className="wiz-launch-progress__bar-inner" style={{width: `${progress * 100}%`}}></div>
+</div>;
+
+const estimateTerraformProgress = terraform => {
+  const {output = ''} = terraform;
+  const done = (output.match(/: Creation complete/g) || []).length;
+
+  // Approximate number of AWS resources we expect Terraform to create
+  const total = 145;
```
ggreer: Doesn't the number of resources to create depend on the size of the cluster? Would larger clusters claim to be completed far sooner than they actually are?

installer/frontend/components/ui.jsx
```diff
@@ -597,23 +597,26 @@ export const PrivateKeyArea = (props) => {
...
+export const WaitingLi = ({pending, done, error, cancel, children, substep}) => {
   const progressClasses = classNames({
     'wiz-launch-progress__step': !substep,
     'wiz-launch-progress__substep': substep,
+    'wiz-pending-fg': pending,
     'wiz-error-fg': error,
     'wiz-success-fg': done && !error,
-    'wiz-running-fg': !done && !error,
+    'wiz-cancel-fg': !done && !error && cancel,
+    'wiz-running-fg': !done && !error && !cancel && !pending,
```
ggreer: Since pending & cancel are styled the same and the cancel parameter isn't used anywhere besides this PR, would it make sense to subsume them both into a parameter & style called inactive? eg:

```jsx
 <WaitingLi inactive={terraformRunning || tfError} done={allDone} error={anyFailed} key="tectonicReady">
```

kyoto: Actually, the style is different. They have the same text color, but different icons.

ggreer: Oh. Nevermind then.

tlwu2013:
Hey @kyoto , 

Progress bar looks great in terms of styling, thanks!

kyoto:
Rebased on master.

installer/frontend/components/tf-poweron.jsx
kyoto: Yeah, I tried getting a more accurate number my looking at the number of instances, but that turned out to not be so helpful. It needs more experimentation, but I'd like to do that in a separate PR.

kyoto:
ok to test

kyoto:
ok to test

kyoto:
ok to test

PR #1542, opened 2017-08-01
azure: Add support for Azure managed disks

justaugustus:
* Set AS`managed = true` to align AS w/ storage fault domains
* Remove individual TF resources for:
    * `random_id`
    * storage accounts
    * storage containers
* Set VMs to use Azure managed disks for OS disk
* Add global `storage_id` for use in all Tectonic node VMs

@alexsomesan / @s-urbaniak / @metral PTAL when you have a chance.


modules/azure/etcd/etcd.tf
```diff
@@ -22,10 +23,11 @@ resource "azurerm_virtual_machine" "etcd_node" {
   }
 
   storage_os_disk {
-    name          = "etcd-disk"
-    vhd_uri       = "${azurerm_storage_account.etcd_storage.primary_blob_endpoint}${azurerm_storage_container.etcd_storage_container.name}/etcd-disk-${count.index}.vhd"
-    caching       = "ReadWrite"
-    create_option = "FromImage"
+    name              = "etcd-${count.index}-os-${var.storage_id}"
+    managed_disk_type = "Premium_LRS"
```
alexsomesan: Does this still make sense to be configurable? Can you have standard storage with managed disks?

modules/tectonic/assets.tf
```diff
@@ -3,6 +3,11 @@ resource "random_id" "cluster_id" {
   byte_length = 16
 }
 
+# Storage ID
+resource "random_id" "storage_id" {
```
alexsomesan: Small nit: this value isn't really used inside the tectonic module and it's also required when building vanilla K8S clusters (which would normally skip everything else in module tectonic. It's also required pretty early on. 

May I suggest we move this ID into the `modules/azure/resource-group` module? 
That module would eventually become some sort of boilerplate where we collect global stuff and we'd change it's name to something more appropriate.

alexsomesan:
This is super cool! Just two nits above, otherwise awesome stuff!

modules/azure/etcd/etcd.tf
justaugustus: @alexsomesan Yes, `managed_disk_type` is actually required to enable the use of managed disks for the OS disk.
`Premium_LRS` yields SSD, while `Standard_LRS` will provide HDD.

I considered parameterizing the value, but I'm not sure that we want to encourage HDD use.

justaugustus:
@alexsomesan Great point about `storage_id`! 
I moved it and rebase this PR, so we should be good to go.

modules/azure/etcd/etcd.tf
alexsomesan: We already have a parameter for the storage type today. I think it sets the storage account type. Was suggesting to repurpose it for here.
We can do that in a follow up as well.

justaugustus:
@alexsomesan Thanks for the review. Updated to parameterize the storage types, repurposing the old vars, albeit renamed. I've tested this successfully outside of CI. 

@squat Can you merge? I don't have the button anymore :/

alexsomesan:
Renaming existing config vars would normally require a config version bump (there's a var in there for that).
Not sure if we want to do this only for stable platforms or for all. Rob should have a say (can't quote him).
I'm going to approve and just merge in case he's fine with the version thing.

justaugustus:
@alexsomesan Thanks for the clarification on config version!

PR #1544, opened 2017-08-01
config.tf: update Console to 1.8.4

Quentin-M:
This put the Clair feature behind a feature flag, as necessary for Tectonic.


Quentin-M:
Shoot ... the payload sig was committed, not the actual payload.

PR #1566, opened 2017-08-02
modules/update-payload/sign-payload.sh: fix shellcheck issue

Quentin-M:
Follow up with https://github.com/coreos/tectonic-installer/pull/1512.


yifan-gu:
LGTM

Issue #1563, opened 2017-08-02
tests: Create public-pr-trigger Jenkins job via Jenkins job DSL


mxinden:
Right now the `public-pr-trigger` Jenkins job is still manually configured via the UI. We should move this into a Groovy job DSL script inside the `tests/jenkins-jobs` folder, so it can be created by our Seed job.

A good example on how to configure the *GitHub Pull Request Builder Plugin* in a Jenkins Job DSL can be found in the [Prometheus operator Github repo](https://github.com/coreos/prometheus-operator/blob/master/Jenkinsfile).

cpanato:
I will take this one

PR #1573, opened 2017-08-03
tests/ci: Create public-pr-trigger Jenkins job via Jenkins job DSL

cpanato:
fixes: https://github.com/coreos/tectonic-installer/issues/1563

also, move the job called `tectonic-installer-nightly-trigger` to the `triggers` folder

/cc @mxinden


tests/jenkins-jobs/tectonic_installer_nightly_trigger.groovy
```diff
@@ -1,6 +1,8 @@
 #!/bin/env groovy​
```
mxinden: Just to make sure the folder exists, can we add a `folder("triggers")` here? Or would that conflict with duplicate ones?

tectonic_installer_public_pr_trigger.groovy
```diff
@@ -0,0 +1,75 @@
...
+            failureThreshold("FAILURE")
+          }
+        }
+      }
+    }
+  }
+
+  publishers {
+    wsCleanup()
+    slackNotifier {
```
mxinden: Just to make sure: Does this notify us if a PR build fails, or if a PR trigger failed? The former could become quite noisy.

tectonic_installer_public_pr_trigger.groovy
```diff
@@ -0,0 +1,75 @@
+#!/bin/env groovy​
+
+folder("triggers")
+
+job("triggers/tectonic-installer-pr-trigger") {
+  description('Tectonic Installer PR Trigger. Changes here will be reverted automatically.')
+
+  logRotator(30, 100)
+  label("master")
```
mxinden: What does this label do? I am confused because we are testing PRs with this Jenkins job and not the master branch. Sorry if this is a stupid question.

cpanato: this was in the original job. It means this job will run only on master. since this is just a trigger and doesn't have any dependencies we can trigger that in the master and not in the slaves.

tectonic_installer_public_pr_trigger.groovy
cpanato: both cases.

tests/jenkins-jobs/tectonic_installer_nightly_trigger.groovy
cpanato: sure, will update

cpanato:
@mxinden already tested yesterday and works :)

tests/jenkins-jobs/tectonic_installer_nightly_trigger.groovy
cpanato: the folder is already there

mxinden:
Merged without CI testing. These scripts are not touched in any way by the CI.

mxinden:
Tested via https://github.com/coreos/tectonic-installer/pull/1604, thanks!

PR #1578, opened 2017-08-03
Bump calico to v2.4.0

abhinavdahiya:
* use official release image for calico/cni to use `SKIP_CNI_BINARIES` option
* update rbac rules to allow calico/node to run
* enable liveness & readiness probes on calico/node (felix agent)

/cc: @aaronlevy @s-urbaniak


abhinavdahiya:
closed in favor of #1637

Issue #1579, opened 2017-08-03
kube_dns: make HA when master_node count > 1


joshrosso:
## FEATURE REQUEST: kube_dns: make HA when master_node count > 1

When the quantity of master nodes in a Tectonic cluster exceed 1, Tectonic should deploy the same number of `kube_dns` pods to ensure high availability OOTB.

**Point to consider**:

- Today, we call out not to set the replica parameters as it won't be respected via the add on manager. However, I cannot see `kube-addon-manager` present in Tectonic clusters as of `1.6.7`. **Am I missing something?** A diff of what i'm describing follows.

    ```
    @@ -28,24 +28,29 @@ metadata:
         k8s-app: kube-dns
         kubernetes.io/cluster-service: "true"
    spec:
    -  # replicas: not specified here:
    -  # 1. In order to make Addon Manager do not reconcile this replicas parameter.
    -  # 2. Default is 1.
    -  # 3. Will be tuned in real time if DNS horizontal auto-scaling is turned on.
   +  replicas: ${master_count}
    ```

- When HA, we must ensure affinity works in the same respect as `kube-controller-manager` and `kube-scheduler`. Including the following behavior.
  - `kube-dns` pods exist on separate nodes.
  - updates still supported with affinity rules in place
    - specifically relevant for single-master clusters

- While the [dns-horizontal-autoscaler](https://kubernetes.io/docs/tasks/administer-cluster/dns-horizontal-autoscaling/) may still be the preferred method for scaling `kube-dns`, this change would:
  - ensure sensible default is init on cluster boostrap
  - introduce an HA-dns for those who won't use this autoscaler

## Versions

* Tectonic version (d7ce8cddc7906b87947a7dcb94718c537ad80e8d):
* Terraform version (`Terraform v0.9.7`):
* Platform (*):

/cc @aaronlevy : do you see issues with this approach? Especially regarding updates/bootkube?

aaronlevy:
Longer term I'd probably prefer that the autoscaler was used.

Often dns needs to be scaled for throughput not just HA -- so adding node based anti-affinity would restrict this to the number of nodes.

Additionally, DNS pods will get recreated if they die (even if there is only one copy).

Not necessarily opposed to this, but there are tradeoffs when dns-autoscaler is probably what you ultimately want.

aaronlevy:
I'd actually recommend that anti-affinity is not used. Scaling DNS for throughput is pretty common - and this seems too restrictive.

[...]

PR #1580, opened 2017-08-03
kube_dns: make HA when master_count is > 1

joshrosso:
- This commit automatically deploys the same number of kube-dns pods as
there are master nodes.

- To support HA scenarios, it adds affinity rules to kube-dns that
should ensure updates are still possible, even in the case of a
single-master Tectonic cluster.

issue: https://github.com/coreos/tectonic-installer/issues/1579

### DIFF
--- a/modules/bootkube/resources/manifests/kube-dns.yaml
+++ b/modules/bootkube/resources/manifests/kube-dns.yaml
@@ -28,24 +28,29 @@ metadata:
     k8s-app: kube-dns
     kubernetes.io/cluster-service: "true"
 spec:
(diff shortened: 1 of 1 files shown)


aaronlevy:
Commented in https://github.com/coreos/tectonic-installer/issues/1579 but dns is often scaled for throughput, not just HA -- adding node based anti-affinity rules greatly limits this ability. 

I think the better solution is to actually start using the dns-autoscaler

cpanato:
closing due to inactivity. thanks for your PR. if you think this is still useful, please re submit.

PR #1582, opened 2017-08-04
frontend: Add JS error reporting.

ggreer:
- Add sendError() to TectonicGA.
- Add window.onerror handler.
- Reduce nesting in send() & log error if ga is not a function.

### DIFF
--- a/installer/frontend/app.jsx
+++ b/installer/frontend/app.jsx
@@ -114,3 +114,17 @@ store.dispatch(validateAllFields(() => {
     document.getElementById('application')
   );
 }));
+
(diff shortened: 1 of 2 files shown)


installer/frontend/app.jsx
```diff
@@ -114,3 +114,17 @@ store.dispatch(validateAllFields(() => {
     document.getElementById('application')
   );
 }));
+
+window.onerror = function (message, source, lineno, colno, optError={}) {
```
kyoto: nit: Can we make this an ES6 arrow function?

ggreer: Weird. I have no idea why I did it the old-school way.

PR #1588, opened 2017-08-04
Added export TF_VAR_tectonic_azure_client_secret to azure install guide

sir-rob:
Added` export TF_VAR_tectonic_azure_client_secret="${ARM_CLIENT_SECRET}"` to azure install guide. Without this step the install pauses for the user to input the secret with:
```
var.tectonic_azure_client_secret
  The client secret to use.

  Enter a value:
```

### DIFF
--- a/Documentation/install/azure/azure-terraform.md
+++ b/Documentation/install/azure/azure-terraform.md
@@ -127,6 +127,7 @@ $ export ARM_SUBSCRIPTION_ID=abc-123-456
 $ export ARM_CLIENT_ID=generated-app-id
 $ export ARM_CLIENT_SECRET=generated-pass
 $ export ARM_TENANT_ID=generated-tenant
+$ export TF_VAR_tectonic_azure_client_secret="${ARM_CLIENT_SECRET}"
 ```
 
 Now we're ready to specify our cluster configuration.


squat:
@sir-rob could you please re-base this commit and push? Structure checking is failing because of something that was merged into master.

justaugustus:
Not sure that I agree with the approach... this assumes that the service principal (and by extension, `ARM_CLIENT_SECRET`) used for the installer is the same as the one we want to use for cloud provider config.
Is that what we want to do?

Documentation/install/azure/azure-terraform.md
```diff
@@ -127,6 +127,7 @@ $ export ARM_SUBSCRIPTION_ID=abc-123-456
 $ export ARM_CLIENT_ID=generated-app-id
 $ export ARM_CLIENT_SECRET=generated-pass
 $ export ARM_TENANT_ID=generated-tenant
+$ export TF_VAR_tectonic_azure_client_secret="${ARM_CLIENT_SECRET}"
```
justaugustus: Not sure that I agree with the approach... this assumes that the service principal (and by extension, ARM_CLIENT_SECRET) used for the installer is the same as the one we want to use for cloud provider config.
Is that what we want to do?

sir-rob: @metral I'll let Mike comment on this as the change came based on a recommendation Mike made to me. (Also I just realized I was asked to rebase commit & push again but I'll wait until after we answer your point @justaugustus )

cpanato:
@squat @justaugustus this still makes sense?

sym3tri:
@sir-rob sorry for the delay on this.
The docs have moved to https://github.com/coreos/tectonic-docs can you please submit the changes there?

justaugustus:
@sym3tri we've since added this to the docs, so good to go on that front! :)

PR #1586, opened 2017-08-04
modules/azure: disable TX checksum offloading

squat:
This commit disables TX checksum offloading as a workaround to fix #1171. We ran some iperf benchmarks to determine how different MTU and tx checksum offloading combinations would perform. From the results shown below, we found that disabling TX checksum offloading and keeping the default MTU does not result in a major performance penalty; it will be the most maintainable workaround going forward.

cc @crawford @philips @Quentin-M 

<details>
iperf.yaml:

```yaml
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: iperf
  namespace: default
  labels:
    app: iperf
spec:
  replicas: 1
  template:
    metadata:
      name: iperf
      labels:
        app: iperf
    spec:
      containers:
      - name: iperf-server
        image: networkstatic/iperf3
        args:
        - -s
        ports:
        - containerPort: 5201
          protocol: TCP
      nodeSelector:
        node-role.kubernetes.io/node: ""
---
apiVersion: v1
kind: Service
metadata:
  name: iperf
  namespace: default
  labels:
    app: iperf
spec:
  type: NodePort
  selector:
    app: iperf
  ports:
  - name: iperf
    protocol: TCP
    port: 5201
    nodePort: 31211
---
apiVersion: batch/v1
kind: Job
metadata:
  name: iperf
  namespace: default
  labels:
    app: iperf
spec:
  template:
    metadata:
      name: iperf
      labels:
        app: iperf
    spec:
      hostNetwork: true
      containers:
      - name: iperf
        image: networkstatic/iperf3
        args: ["-c", "green.westcentralus.cloudapp.azure.com", "-p", "31211", "-t", "30", "-V"]
      restartPolicy: Never
```

I got the following results:

MTU 1350 tx on
```
Time: Thu, 03 Aug 2017 23:21:12 GMT
Connecting to host green.westcentralus.cloudapp.azure.com, port 31211
      Cookie: green-worker-0.1501802472.198990.573
      TCP MSS: 1228 (default)
[  4] local 10.0.16.4 port 39306 connected to 52.161.110.8 port 31211
Starting Test: protocol: TCP, 1 streams, 131072 byte blocks, omitting 0 seconds, 30 second test
[ ID] Interval           Transfer     Bandwidth       Retr  Cwnd
[  4]   0.00-1.00   sec  88.3 MBytes   741 Mbits/sec  802    191 KBytes       
[...]

PR #1597, opened 2017-08-04
modules/azure: disable TX checksum offloading

Quentin-M:
Cherry-picks https://github.com/coreos/tectonic-installer/pull/1586


Quentin-M:
Last check is bare-metal, but Packet is offline. Also, Azure subscription. Merged.

Issue #1492, opened 2017-07-25
When using an existing etcd cluster the masters still start etcd-member locally.


coresolve:
## Versions

* Tectonic version ([release](https://github.com/coreos/tectonic-installer/releases) or commit hash):
* Terraform version (`terraform version`): 1.6.7
* Platform (aws|azure|openstack|metal): metal

## What happened?
Brought the cluster up using an existing etcd cluster.
and etcd-member service was started anyway.

## What you expected to happen?
No local etcd pod would be started.
the etcd-member service would not be started.

## How to reproduce it (as minimally and precisely as possible)?
Bring up tectonic with an existing external etcd cluster. 
Note that the etcd-member service is started.

## Anything else we need to know?

alexsomesan:
I'll look into this and post back the findings.

Quentin-M:
Have you found anything? :)

coresolve:
this is here:
https://github.com/coreos/tectonic-installer/blob/master/platforms/metal/cl/bootkube-controller.yaml.tmpl#L5

and here:
https://github.com/coreos/tectonic-installer/blob/master/platforms/metal/matchers.tf#L39

and the fact that in this case tectonic-experimental is false.

PR #1599, opened 2017-08-04
platforms/metal: disable etcd member when using external etcd

squat:
fixes: https://github.com/coreos/tectonic-installer/issues/1492

cc @coresolve @Quentin-M @alexsomesan


Quentin-M:
Oh wow ok. I didn't expect that one 👍

Quentin-M:
As we don't have smoke tests for that particular case, we can skip the wait.

PR #1605, opened 2017-08-07
azure: Remove Scale Sets implementation

justaugustus:
Not sure what everyone's thoughts are, but we're not using any of the Azure Scale Set code.
This PR removes it.

cc: @alexsomesan @s-urbaniak @mxinden @metral


alexsomesan:
There were some requests for keeping Scale Sets support, back when we switched to Availability Sets.
However, the current SS code is way unmaintained and I think I doesn’t even build anymore. I suspect it would take more effort to re-align it to the current reference than to actually delete it and back-port the AS version to SS.

So I guess we can get rid of it for now and plan to back-port from AS.

alexsomesan:
Can we keep the `-as` suffixes?
Like I mentioned, there were requests for the SS implementation so we need to bring that back at some point.

justaugustus:
@alexsomesan Fixed. Thanks for the review!

justaugustus:
@alexsomesan Rebased to fix `structure-check` failure on `master`.
Should be good to go now.

alexsomesan:
Setting do-not-test since this code is not exercised from anywhere.

PR #1609, opened 2017-08-07
tests: Enable concurrent PR builds

mxinden:
The Tectonic installer PR trigger job waits until the main pipeline
finishes. This enables mutliple trigger jobs to run concurrent,
otherwise we could never test more than one PR at once.

@cpanato What do you think? Is this a sane fix?


squat:
@mxinden, the main reason for not doing this before is that the resource limits on our cloud accounts were not high enough. Maybe now that Grafiti rubs in AWS on every pr this will work

mxinden:
@squat We have had concurrent build since quite a while. This just changed today, since instead of the manually configured PR-trigger job, we are now using the PR-trigger job created by a groovy script. See https://github.com/coreos/tectonic-installer/pull/1573

PR #1612, opened 2017-08-07
modules/tectonic: remove updater enabled statistic

squat:
This commit removes the `tectonicUpdaterEnabled` statistic from the
stats-extender and from the Tectonic Installer since this option is
always enabled. No changes are needed to the BigQuery schema thanks to
the design of the `extensions` schema. Furthermore, since the updater was
decoupled from the experimental features, the reported value for this metric
has been incorrect.

cc @alexsomesan @sym3tri


alexsomesan:
I think I'm lacking some context. What's the background for this change?

squat:
@alexsomesan there's two issues:
1. The updates are always enabled, regardless of whether the user chooses to enable experimental features or not. I spoke to Mackenzie and she agreed we should remove this stat because it's just noise now.
2. The stat is now reporting something orthogonal to the updated which is just introducing noise into the database. We could rename the statistic or just remove it.

alexsomesan:
Got it. All makes sense. Thanks for the clarification, @squat !

alexsomesan:
closing and reopening to run smokes.

squat:
Got one flake in one of the AWS tests. Merging.

Issue #1615, opened 2017-08-08
Ensure tectonic-console and tectonic-identity schedule on separate nodes


joshrosso:
## FEATURE REQUEST: Ensure tectonic-console and tectonic-identity schedule on separate nodes

At this time, we set the replica count of the `tectonic-console` and `tectonic-identity` to `2`. We should add affinity rules that ensure they're schedule on unique nodes.

My preference is [requiredDuringSchedulingIgnoredDuringExecution](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#node-affinity-beta-feature) in order to ensure pods exist on different nodes. Considering these pods land on worker nodes, it seems reasonable to assume most clusters will containe `> 1` workers.

However, I can understand if preference is to use `preferred`. I'm sure a case could be made around throughput. Although, again, I feel a stronger case could be made around HA-guarantees. 

## Versions

* Tectonic version (https://github.com/coreos/tectonic-installer/commit/9bd3835dde019c8c5a410c62596fb18e290e67ef)
* Terraform version (`Terraform v0.9.7`):
* Platform (*):

robszumski:
I believe there are blockers around supporting single node and/or master clusters because you can't execute rolling updates, as they would never get scheduled.

squat:
They will still get scheduled if we use
`preferedDuringSchedulingIgnoredDuringExecution`

joshrosso:
@robszumski See convo in PR for more context on that.

sym3tri:
We won't make this change b/c it's not relevant for the converged platform.

https://coreos.com/blog/coreos-tech-to-combine-with-red-hat-openshift

PR #1616, opened 2017-08-08
tectonic console and identity schedule on unique nodes.

joshrosso:
- Using anti-affinity, ensure tectonic-console and tectonic-identity
pods land on unique nodes ensuring HA.

- Resolves https://github.com/coreos/tectonic-installer/issues/1615

### DIFF
--- a/modules/tectonic/resources/manifests/console/deployment.yaml
+++ b/modules/tectonic/resources/manifests/console/deployment.yaml
@@ -22,9 +22,19 @@ spec:
       creationTimestamp: null
       labels:
         k8s-app: tectonic-console
+        pod-anti-affinity: tectonic-console-${tectonic_version}
         component: ui
       name: tectonic-console
     spec:
+      affinity:
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+          - labelSelector:
+              matchLabels:
+                pod-anti-affinity: tectonic-console-${tectonic_version}
+            namespaces:
+              - tectonic-system
+            topologyKey: kubernetes.io/hostname
       containers:
       - command:
         - /opt/bridge/bin/bridge
--- a/modules/tectonic/resources/manifests/identity/deployment.yaml
+++ b/modules/tectonic/resources/manifests/identity/deployment.yaml
@@ -21,8 +21,18 @@ spec:
       name: tectonic-identity
       labels:
         k8s-app: tectonic-identity
+        pod-anti-affinity: tectonic-identity-${tectonic_version}
         component: identity
     spec:
+      affinity:
+        podAntiAffinity:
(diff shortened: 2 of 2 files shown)


sym3tri:
@aaronlevy any objection to this?

alekssaul:
Hey @joshrosso , 

https://github.com/coreos/tectonic-installer/pull/1106 is also relevant.

modules/tectonic/resources/manifests/console/deployment.yaml
```diff
@@ -22,9 +22,19 @@ spec:
       creationTimestamp: null
       labels:
         k8s-app: tectonic-console
+        pod-anti-affinity: tectonic-console-${tectonic_version}
```
aaronlevy: This is unnecessary. This behavior was introduced specifically for supporting updates of core components where we must guarantee that there is always at least one copy running (even on single master). We don't need to ensure that consoles can be co-located on a single-master during an upgrade. If the console is down for a moment during an upgrade - it will be recreated soon enough (and not worth the complexity here).

If you're going this route I'd just say use a daemonset, you're essentially trying to re-create that behavior. But either way, any changes to these manifests need to have full upgrade plans before merging here. Moving these from a deployment to a daemonset would currently require a migration step.

aaronlevy:
This is trying to recreate the behavior of daemonsets. You should just use those instead - but the change of deployments to daemonsets is going to require some work during an upgrade.

diegs:
> ensuring HA.

Not sure I understand this statement. This doesn't guarantee HA, and can hurt HA if, let's say, the anti affinity can't be satisfied by available worker nodes. I'd rather have 2 pods on one node than 1 pod.

I agree with all of @aaronlevy's points and I also would like to understand which fundamental problem this is solving. For non-control plane items scheduling pods onto the same node is ok. If the node goes away the control plane will reschedule them.

modules/tectonic/resources/manifests/console/deployment.yaml
joshrosso: Thanks @aaronlevy, makes sense.

I don't think a daemonset is required either. I'll elaborate further on what i'm trying to solve in the comments of this PR.

aaronlevy:
I think pod disruption policy is a reasonable approach. That would mitigate issues during drains. Also a preferred-anti affinity I think is fine too. Regarding flannel issues, in that case the node would eventually start reporting as "NotReady" and the pods would be moved elsewhere (but yes, there could be a few minute delay).

cpanato:
@joshrosso thanks for this PR, can you please rebase and resolve the conflicts? thanks

Dominik-K:
@diegs Will the `tectonic-operators` repo be in the public or stay closed-source?

PR #1622, opened 2017-08-08
modules/update-payload: Make sure TCO always updates itself first.

yifan-gu:
Before 1.7.1, TCO updates itself by applying the TCO deployment manifest in the payload.
In order to make sure it updates itself first without changing other cluster state, we need to make TCO appears first in the list.

### DIFF
--- a/modules/update-payload/make-update-payload.sh
+++ b/modules/update-payload/make-update-payload.sh
@@ -42,8 +42,20 @@ VERSION=$(yaml2json < ${ASSETS_DIR}/app_versions/app-version-tectonic-cluster.ya
 # Don't include the meta app-version for sub-component's desired version.
 rm "${ASSETS_DIR}/app_versions/app-version-tectonic-cluster.yaml"
 
+tco_deployment="tectonic-channel-operator.yaml"
+
+# Get the TCO deployment first.
+f="${ASSETS_DIR}/operators/${tco_deployment}"
+tmpfile=$(mktemp /tmp/deployment.XXXXXX)
+# shellcheck disable=SC2086
+yaml2json < ${f} > ${tmpfile}
+tmpfiles+=(${tmpfile})
+
 # Get the deployments.
 for f in ${ASSETS_DIR}/operators/*.yaml; do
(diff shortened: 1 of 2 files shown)


modules/update-payload/make-update-payload.sh
```diff
@@ -42,8 +42,18 @@ VERSION=$(yaml2json < ${ASSETS_DIR}/app_versions/app-version-tectonic-cluster.ya
...
+# Get the TCO deployment first.
+f="${ASSETS_DIR}/operators/tectonic-channel-operator.yaml"
+tmpfile=$(mktemp /tmp/deployment.XXXXXX)
+# shellcheck disable=SC2086
+yaml2json < ${f} > ${tmpfile}
+tmpfiles+=(${tmpfile})
+
 # Get the deployments.
 for f in ${ASSETS_DIR}/operators/*.yaml; do
+  if [[ $(basename "${f}") == "tectonic-channel-operator.yaml" ]];then
```
aaronlevy: nit: might assign this name to a var because it's used as `f` above, and here.

modules/update-payload/payload.json
```diff
@@ -6,47 +6,76 @@
       "kind": "Deployment",
       "metadata": {
         "labels": {
-          "k8s-app": "container-linux-update-operator",
+          "k8s-app": "tectonic-channel-operator",
```
aaronlevy: I'm assuming this is just unfortunate git diff parsing?

yifan-gu: yes.

aaronlevy:
Overall this lgtm -- but I'm less familiar with this use-case -- maybe get a pass by one other person?

modules/update-payload/make-update-payload.sh
yifan-gu: Fixed

yifan-gu:
test upgrade from 1.6.7-tectonic.2

yifan-gu:
test upgrade from 1.6.7-tectonic.2

yifan-gu:
test upgrade from 1.6.7-tectonic.2 1.6.7-tectonic.2

PR #1623, opened 2017-08-08
README: add HTTP limit note.

zbwright:


README.md
```diff
@@ -122,5 +120,14 @@ make apply
 make destroy
 ```
 
+### Notes
+
+#### Limitations of HTTP headers
+
+Large group counts will not work with reverse proxies that limit HTTP headers to 4K. These counts will return an error similar to that returned for a cookie based approach. "Normal" HTTP header setups will support 1M header.
```
kans: | Large group counts will not work with reverse proxies that limit HTTP headers to 4K.

The Ingress Controller limits headers to around ~8 Kb by default.  `groups` refer to LDAP or github - something that you connect to dex.  I think these docs would make more sense with other auth docs somewhere.


| These counts will return an error similar to that returned for a cookie based approach. 

The error that end users will see is something like: `400: Request Header Or Cookie Too Large` after they log in to console.  Also, we no longer use cookies for auth.

| "Normal" HTTP header setups will support 1M header.
A "vanilla" Tectonic cluster will run into the Ingress Controller limit mentioned above.

README.md
```diff
@@ -122,5 +120,14 @@ make apply
...
+
+To work with large-client-header-buffers: set `tectonic-custom-error` on the Ingress Configmap to `4 64k` to increase the limit in Ingress to EG 64k. See the nginx docs on the flag.
```
kans: The configmap in question is actually called `tectonic-custom-error` in `tectonic-system`.

zbwright:
Rob agrees that this does not belong in the README.md. I'll take a look next week, see where it should live, and open a new PR. leaving this open as a reminder to me to do so.

sym3tri:
@zbwright I agree, README is not the best place.

cpanato:
@zbwright Can I close this PR?

zbwright:
yes - replaced by https://github.com/coreos/tectonic-docs/pull/20

PR #1633, opened 2017-08-09
tests: Do not block PR trigger job until triggered jobs are finished

mxinden:
In my opinion there is no need for this. In addition, whenever a PR
build fails, the failure will propagate to the trigger job otherwise,
which then sends a slack message.

@cpanato What do you think?

### DIFF
--- a/tests/jenkins-jobs/tectonic_installer_public_pr_trigger.groovy
+++ b/tests/jenkins-jobs/tectonic_installer_public_pr_trigger.groovy
@@ -47,22 +47,15 @@ job("triggers/tectonic-installer-pr-trigger") {
     }
   }
 
-  steps {
-    triggerBuilder {
+  publishers {
+    hudson.plugins.parameterizedtrigger.buildTrigger {
       configs {
-        blockableBuildTriggerConfig {
+        buildTriggerConfig {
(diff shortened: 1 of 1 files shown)


cpanato:
@mxinden did you test if removing this the job will work?

mxinden:
You are right, this is not working. Thanks for the catch!

I have tried using the `buildTriggerConfig` instead of the `blockableBuildTriggerConfig`. This is only available in the `publishers` block. But Jenkins complains with the following:
```
16:07:03 ERROR: Found multiple extensions which provide method buildTrigger with arguments [tectonic_installer_public_pr_trigger$_run_closure1$_closure5$_closure7@1be27fa8]: [hudson.tasks.BuildTrigger, hudson.plugins.parameterizedtrigger.BuildTrigger]
```

@cpanato Did you add the ParameterizedTrigger plugin just for this? In that case, can we remove it so there is only `hudson.tasks.BuildTrigger`, right?

I am trying to unblock this job, not only because of the Slack notifications, but also because it is taking an entire build slot on the master whenever it runs. I think we only have 4 of those slots at a given time.

mxinden:
@cpanato Feel free to push on my branch if you have any solution to this.

cpanato:
will close this and open another PR.

PR #1640, opened 2017-08-09
kube-flannel-cfg: enable hairpin mode

kbrwn:
Edit flannel configmap to enable hairpin mode

Hairpin mode is needed to allow pods to communicate with themselves via Service IP. This change is required as the kubelet --hairpin-mode flag is not supported when CNI is being used.

### DIFF
--- a/modules/net/canal/resources/manifests/kube-flannel.yaml
+++ b/modules/net/canal/resources/manifests/kube-flannel.yaml
@@ -12,7 +12,8 @@ data:
       "name": "cbr0",
       "type": "flannel",
       "delegate": {
-        "isDefaultGateway": true
+        "isDefaultGateway": true,
+        "hairpinMode": true
       }
     }
   net-conf.json: |
--- a/modules/net/flannel_vxlan/resources/manifests/kube-flannel.yaml
+++ b/modules/net/flannel_vxlan/resources/manifests/kube-flannel.yaml
@@ -12,7 +12,8 @@ data:
       "name": "cbr0",
       "type": "flannel",
       "delegate": {
-        "isDefaultGateway": true
+        "isDefaultGateway": true,
+        "hairpinMode": true
       }
     }
   net-conf.json: |


sym3tri:
> This change is required as the kubelet --hairpin-mode flag is not supported when CNI is being used

Link to supporting issue/documentation?

sym3tri:
Will also need to coordinate this change with the KVO.
/cc @aaronlevy

aaronlevy:
In general we wouldn't touch networking configuration during updates. Without a lot of testing it would be pretty risky (unknown the networking infra of various clusters).

Is this a value that should be set on existing clusters as well? If so, then we would need to write migration code to insert this value into existing configs (but we also need to know that it would be very safe to change this on live clusters). /cc @diegs

kbrwn:
> Link to supporting issue/documentation?

https://github.com/kubernetes/kubernetes/issues/43966#issuecomment-294968545
https://github.com/kubernetes/kubernetes/pull/29009
https://github.com/containernetworking/cni/issues/320

@sym3tri  FR-287 has more information as well.

aaronlevy:
@squeed any thoughts on trying to add this into existing clusters / the risk of doing so (without knowing what else they might have changed in their CNI config)?

Or do we only fix it on new clusters and leave it to the user to update on existing?

aaronlevy:
I wouldn't block this going into the installer. Just note that we are introducing configuration skew between existing and new clusters unless we were to add this change to the update process. I also don't know if that is reasonably safe to assume we can do (what other changes have people made to their CNI configuration that might be incompatible?)

If we want to upgrade this, we would need to parse the existing config and add this option if it doesn't exist yet.

That is not currently something that anyone is working on - and would not be able to by the next release.

I'll let the call of if this goes in or not to @sym3tri and @robszumski 

If it does go in, I would suggest that it is tracked as known skew - and what version it was introduced in.

robszumski:
@aaronlevy Two questions:

> what other changes have people made to their CNI configuration that might be incompatible?

Per our official guidelines, this would not have happened, correct?

> If we want to upgrade this, we would need to parse the existing config and add this option if it doesn't exist yet.

Will this be covered by the patchMerge epic in the future?

aaronlevy:
> Per our official guidelines, this would not have happened, correct?

Can you link to those guidelines? I don't know that this has been communicated sufficiently.

> Will this be covered by the patchMerge epic in the future?

Not initially. This is just an unstructured blob from the perspective of kubernetes objects. 
Options would be to add some kind of CNI config handling later where:
- our merge logic becomes aware of the CNI config struct
- we blindly stomp on the config no matter what (or fail if it mis-matches in anyway)
- we try and do a git-style 3-way-merge, but that would be somewhat imperfect.

So right now it would be an imperative migration step that someone would need to write to handle this in an update.

squeed:
A quick chime-in from the networking side: This change, if it were deployed to running clusters, happens to only affect new containers joining the bridge. Hairpin mode, in Linux, is a property of each bridge interface, not the bridge itself. This is definitely something we want.

In this case, roll-out isn't really anything to worry about. That's not true for the general case though. Kubernetes always uses the current CNI configuration on disk for the CNI ADD and DELETE operations. If we were to change configurations such that a DELETE of an existing pod failed, or disrupted existing pods, we'd be very sad. For example, renaming the flannel bridge would be a bad idea.

As for how we deploy CNI configurations:

Kubernetes doesn't really understand CNI config files, except as a blob of bytes to hand to the CNI plugin. The fact that our CNI configuration is in a ConfigMap doesn't change anything.

As for how we deploy the network configuration, it's installed as a configmap to the pod, then copied from the pod to the system at start-time by a bind mount. So, if this change is added to running clusters, as the DaemonSet changes were rolled out, the new CNI configuration would be installed.

So, if we want to roll out a CNI config change, we need to update the DaemonSet and roll that out somehow.

Newbie-level question: are ConfigMap changes rolled out independently? Perhaps it makes sense for the "install-cni" container to watch for changes.

aaronlevy:
@squeed so are you saying adding hairpin might affect deletions of existing bridges or just making note of the risk of other changes?

It is easier to control roll-out of config-map changes if we don't have the install-cni container watch for changes. Otherwise they all will update at the same time VS. enforcing this via rolling update of the daemonset pods. Or we would leave the configMaps as immutable and create a new configMap, then update the reference in the daemonset. This way rollbacks are easier.

Again, I'll let the call go to @sym3tri and @robszumski. In my mind the options are:

- Do not merge yet
- installer only (fix only on new clusters) + track the new skew we've introduced
- installer + update existing: someone needs to pick up the update work. Freeze for next release is Friday.

squeed:
@aaronlevy I was just noting that changing CNI configs on running kubeletes nodes can be risky. This particular change is fine.

We should absolutely get this in new clusters. How to roll out changes to running clusters does need careful thought; but that shouldn't block this change.

robszumski:
@squeed I think @aaronlevy's point is that accumulating that skew without a plan to realign is dangerous over time.

squeed:
I don't disagree. However, if possible, this particular change should be merged. Not having hairpin enabled is an actively bad situation.

robszumski:
Let's go with "installer only + track skew" so we can get this landed for new clusters.

diegs:
Please make sure an issue is filed about tracking the skew before merging this PR. Thanks!

squeed:
Created CORS-251.

squeed:
Filed CORS-257

sym3tri:
Spoke to @robszumski and he is ok merging for this release give this is a bug.

rubbish:
Hi, is this going to be part of a major release soon? I ran into this today on 1.7.5. Thanks!

diegs:
@kbrwn @sym3tri is this going into the installer for the 1.8 release? We need to know if the operator needs to be changed as well.

sym3tri:
@diegs I will merge if you're ok with it. Just ping me back here when you're ready.

diegs:
@sym3tri let's merge it, thanks!

squat:
ok to test

Issue #784, opened 2017-05-19
Tectonic Installer for AWS only returning first 100 Route 53 zones in drop-down


sym3tri:
When selecting Route 53 zones the drop-down is only showing the first 100 records. Our installer code does not handle REST pagination for AWS's ListHostedZonesByName API which states it returns a maximum of 100 per request. We need to fix the code so it can handle more then 100+ zones.

PR #1641, opened 2017-08-09
aws: Retrieve all route 53 hosted zones

lander2k2:
Currently, only the first 100 hosted zones are retrieved from AWS.
This is the default MaxItems. This change will make subsequent calls
if the response indicates there are more records to be listed.

### DIFF
--- a/installer/api/handlers_aws.go
+++ b/installer/api/handlers_aws.go
@@ -259,7 +259,7 @@ func awsGetZonesHandler(w http.ResponseWriter, req *http.Request, _ *Context) er
 		return nil
 	}
 
-	allZones := make([]*route53.HostedZone, 0)
+	var allZones []*route53.HostedZone
 	var marker string
 
 	for marker != "complete" {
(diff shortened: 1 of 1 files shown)


sym3tri:
Fix for https://github.com/coreos/tectonic-installer/issues/784

installer/api/handlers_aws.go
```diff
@@ -229,20 +229,49 @@ func awsGetKeyPairsHandler(w http.ResponseWriter, req *http.Request, _ *Context)
...
+		}
+
+		return nil
 	}
-	route53svc := route53.New(awsSession)
 
-	resp, err := route53svc.ListHostedZones(&route53.ListHostedZonesInput{})
-	if err != nil {
-		return fromAWSErr(err)
+	allZones := make([]*route53.HostedZone, 0)
```
squat: Looks good overall, but tests are failing because our linter is complaining that rather than making the hosted zone slice here, you should just declare the slice var:
```go
var allZones []*route53.HostedZone
```

squat:
@lander2k2 thanks for contributing this! It's been a common complaint for a little while now. Just one small issue and I think it's good to go.

lander2k2:
@squat Fixed the linter complaint.  Hope that does the trick.

squat:
@lander2k2 thanks, that did the tricks. All tests passed except for one of the AWS tests, which timed out while the cluster integrity tests were running. In the interest of resource-efficiency, im going to merge.

PR #1642, opened 2017-08-09
config.tf: update node-agent to c021a49

Quentin-M:
Getting https://github.com/coreos-inc/kube-version-operator/pull/216 in.


Quentin-M:
Every tests passed, except bare-metal because nodes are offline. Merging.

PR #1643, opened 2017-08-09
config.tf: update node-agent to c021a49

Quentin-M:
Cherry-picks https://github.com/coreos/tectonic-installer/pull/1642


Quentin-M:
Most tests passed, except Azure due to subscription and bare-metal because nodes are offline. Merging.

PR #1653, opened 2017-08-10
frontend: AWS access keys do not always begin with AK.

ggreer:
Fixes a bug where STS tokens are marked as invalid.


ggreer:
This should probably ship in a bugfix release for 1.7.

robszumski:
Can we make this smarter instead of removing it? This was added in response to a bunch of user testing were folks were trying to put in email addresses into this form.

PR #1658, opened 2017-08-11
tests: Add option to trigger PR with and without smoke tests

mxinden:
By adding an additional trigger job, we can differentiate the amount of
tests that should run per PR:

- `do-not-test` - run no tests at all
- no labels - run all tests except smoke tests
- `do-smoke-test` - run all tests

This is to temporally unblock Tectonic installer developers e.g. @ggreer which don't need to run the entire smoke test suite on their PRs all the time. In the long run I would like to implement more fine granular triggers (e.g. by cloud provider, ...).

### DIFF
--- a/tests/jenkins-jobs/tectonic_installer_public_pr_trigger_no_smoke.groovy
+++ b/tests/jenkins-jobs/tectonic_installer_public_pr_trigger_no_smoke.groovy
@@ -2,7 +2,7 @@
 
 folder("triggers")
 
-job("triggers/tectonic-installer-pr-trigger") {
+job("triggers/tectonic-installer-pr-trigger-no-smoke") {
   description('Tectonic Installer PR Trigger. Changes here will be reverted automatically.')
 
   concurrentBuild()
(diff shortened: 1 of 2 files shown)


tests/jenkins-jobs/tectonic_installer_public_pr_trigger_no_smoke.groovy
```diff
@@ -0,0 +1,72 @@
...
+        }
+      }
+    }
+  }
+
+  publishers {
+    wsCleanup()
+    slackNotifier {
+      authTokenCredentialId('tectonic-slack-token')
+      customMessage("Tectonic Installer PR Trigger")
```
cpanato: Just a minor thing: Update the message to be `Tectonic Installer PR Trigger No Smoke Tests` or something like that

mxinden: Good catch. Thanks. Fixed.

mxinden:
The last build of this PR was run with the smoke tests even though the `do-smoke-test` label is not set. This is because these updates are yet present on the Jenkins machine. I did test it out though before.

PR #1660, opened 2017-08-11
Installer: make backend compatible with Terraform 0.10.x

alexsomesan:
This change makes the the backend compatible with Terraform beyond 0.9.x

It adds logic to determine the version of Terraform being used and decides based on version if it needs to run `terraform get` in case of 0.9.x or `terraform init` for 0.10.x and beyond.

Note that the new `init` command also includes the previous functionality of `get` so there is no need to run both.

/cc: @sym3tri

### DIFF
--- a/installer/api/handlers_terraform.go
+++ b/installer/api/handlers_terraform.go
@@ -4,7 +4,9 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"os/exec"
 	"path/filepath"
+	"regexp"
 	"time"
 
 	"github.com/dghubble/sessions"
@@ -58,12 +60,23 @@ func terraformApplyHandler(w http.ResponseWriter, req *http.Request, ctx *Contex
 		return newInternalServerError("could not write TerraForm templates: %s", err)
 	}
 
-	// Execute TerraForm get and wait for it to finish.
-	_, getDone, err := ex.Execute("get", "-no-color", tfMainDir)
+	// Choose to run 'get' or 'init' based on Terraform version.
+	verRx := regexp.MustCompile("^Terraform v0\\.[0-9]\\.[0-9]+")
+	out, err := exec.Command("terraform", "version").Output()
 	if err != nil {
-		return newInternalServerError("Failed to run TerraForm (get): %s", err)
+		return newInternalServerError("Failed to determine Terraform version: %s", err)
 	}
-	<-getDone
+	var prepCommand = "init"
+	if verRx.Match(out) {
+		prepCommand = "get"
+	}
+
+	// Execute TerraForm get or init and wait for it to finish.
+	_, prepDone, err := ex.Execute(prepCommand, "-no-color", tfMainDir)
+	if err != nil {
+		return newInternalServerError("Failed to run TerraForm (%s): %s", prepCommand, err)
+	}
+	<-prepDone
 
 	// Store both the path to the Executor and the ID of the execution so that
 	// the status can be read later on.
--- a/installer/api/handlers_terraform.go
+++ b/installer/api/handlers_terraform.go
@@ -47,7 +47,7 @@ func terraformApplyHandler(w http.ResponseWriter, req *http.Request, ctx *Contex
 		// Restore the execution environment from the session.
 		_, ex, _, err = restoreExecutionFromSession(req, ctx.Sessions, &input.Credentials)
 	} else {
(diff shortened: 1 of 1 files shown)


alexsomesan:
One test-case failed due to Terraform timeout issues.
IMHO it's fine, we shouldn't need to re-run the whole suite.

sym3tri:
I would prefer we re-run tests and only merge once green.

alexsomesan:
re-test this please

installer/api/handlers_terraform.go
```diff
@@ -58,12 +60,23 @@ func terraformApplyHandler(w http.ResponseWriter, req *http.Request, ctx *Contex
 		return newInternalServerError("could not write TerraForm templates: %s", err)
 	}
 
-	// Execute TerraForm get and wait for it to finish.
-	_, getDone, err := ex.Execute("get", "-no-color", tfMainDir)
+	// Choose to run 'get' or 'init' based on Terraform version.
+	verRx := regexp.MustCompile("^Terraform v0\\.[0-9]\\.[0-9]+")
```
squat: nit: this regular expression only checks if a string is version <0.10, not if it is a valid version in general, so consider renaming for clarity and maintainability, e.g. `sub10Rx`

installer/api/handlers_terraform.go
```diff
@@ -58,12 +60,23 @@ func terraformApplyHandler(w http.ResponseWriter, req *http.Request, ctx *Contex
...
-		return newInternalServerError("Failed to run TerraForm (get): %s", err)
+		return newInternalServerError("Failed to determine Terraform version: %s", err)
 	}
-	<-getDone
+	var prepCommand = "init"
+	if verRx.Match(out) {
+		prepCommand = "get"
+	}
+
+	// Execute TerraForm get or init and wait for it to finish.
```
squat: s/TerraForm/Terraform/

installer/api/handlers_terraform.go
```diff
@@ -58,12 +60,23 @@ func terraformApplyHandler(w http.ResponseWriter, req *http.Request, ctx *Contex
...
+	_, prepDone, err := ex.Execute(prepCommand, "-no-color", tfMainDir)
+	if err != nil {
+		return newInternalServerError("Failed to run TerraForm (%s): %s", prepCommand, err)
```
squat: s/TerraForm/Terraform/

installer/api/handlers_terraform.go
```diff
@@ -58,12 +60,23 @@ func terraformApplyHandler(w http.ResponseWriter, req *http.Request, ctx *Contex
...
-	_, getDone, err := ex.Execute("get", "-no-color", tfMainDir)
+	// Choose to run 'get' or 'init' based on Terraform version.
+	verRx := regexp.MustCompile("^Terraform v0\\.[0-9]\\.[0-9]+")
+	out, err := exec.Command("terraform", "version").Output()
 	if err != nil {
-		return newInternalServerError("Failed to run TerraForm (get): %s", err)
+		return newInternalServerError("Failed to determine Terraform version: %s", err)
 	}
-	<-getDone
+	var prepCommand = "init"
```
squat: typically, we don't use `var` within func blocks. Consider rewriting as `prepCommand := "init"`

installer/api/handlers_terraform.go
alexsomesan: Good spot!
This is all over the file, and thus got copy-pasted in here.
I'll fix it everywhere.

installer/api/handlers_terraform.go
alexsomesan: copy-paste. will fix.

installer/api/handlers_terraform.go
alexsomesan: Don't have any preference here. I'll rename it.

installer/api/handlers_terraform.go
alexsomesan: Make sense. I'll adapt it.

alexsomesan:
@squat addressed all the nits.

Looks like the 'TerraForm' spelling is all over the place, also in other areas of the code. I'll follow-up with another PR to align it, just to keep this one concise.

squat:
All tests passed except for one of the AWS tests, which timed out during the cluster integrity tests. No need to re-run tests IMO. Go ahead and merge it.

cpanato:
Agree, the test that failed passed before in another test execution.

PR #1666, opened 2017-08-11
frontend: AWS access keys do not always begin with AK.

Quentin-M:
Cherry-picks https://github.com/coreos/tectonic-installer/pull/1653


Quentin-M:
Most tests passed, the rest are typical Azure failure.

PR #1667, opened 2017-08-12
Use kubernetes version for anti-affinity.

diegs:
We don't always upgrade the control plane when bumping tectonic versions
(e.g. from 1.6.7-tectonic.1 to 1.6.7-tectonic.2). I think it only makes
sense to enforce the anti-affinity when we are actually bumping
kubernetes.

Right now we have a divergence where 1.6.7-tectonic.2 clusters that
were installed from that version will have '1.6.7-tectonic.2'
anti-affinity in their manifests, whereas anything prior will have
'1.6.7-tectonic.1'.

### DIFF
--- a/modules/bootkube/assets.tf
+++ b/modules/bootkube/assets.tf
@@ -106,7 +106,7 @@ resource "template_dir" "bootkube" {
     etcd_client_cert = "${base64encode(data.template_file.etcd_client_crt.rendered)}"
     etcd_client_key  = "${base64encode(data.template_file.etcd_client_key.rendered)}"
 
-    tectonic_version = "${var.versions["tectonic"]}"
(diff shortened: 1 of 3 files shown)


squat:
@diegs these test are failing because the templates need a `kubernetes_version` variable that is not being specified in the terraform config: https://github.com/coreos/tectonic-installer/blob/master/modules/bootkube/assets.tf#L109

diegs:
@squat thanks, updated the variable.

@yifan-gu I also added explicit selectors to the manifests, can you PTAL?

Thanks!

squat:
ok to test

diegs:
@squat

yifan-gu:
Sorry for the delay, LGTM.

PR #1673, opened 2017-08-14
tests: Refactor cluster logic to use inheritance for different platforms

mxinden:
I soon want to add Azure to our RSpec code base. This will make it as simple as adding  a child class to the `Cluster` parent class.


cpanato:
looking...

PR #1598, opened 2017-08-04
modules/bootkube: always deploy etcd operator

squat:
The etcd operator should always be deployed, even if k8s etcd is *not*
self-hosted. This was cluster users can still manage hosted clusters.

fixes: https://github.com/coreos/tectonic-installer/issues/1549

cc @Quentin-M @alexsomesan @robszumski

PR #1683, opened 2017-08-15
*: separate etcd-operators, self-hosted vs as a service

hasbro17:
As discussed offline the etcd-operator used for `etcd-as-a-service` needs to be separated from the etcd-operator used for self-hosted etcd. 

Partly reverts https://github.com/coreos/tectonic-installer/pull/1598 to ensure that:
- The etcd-operator in kube-system for self-hosted is going to be created via a bootkube manifest as was the case before the above PR.
- The etcd-operator used for etcd-as-a-service will be created in tectonic-system from a tectonic manifest by the tectonic.sh script. 

@xiang90 @squat @hongchaodeng @robszumski @Quentin-M


hasbro17:
The console does not show the `Operators->etcd Clusters` tab with this PR. I'm assuming that's probably since the console thinks the etcd-operator is not installed. 
Can somebody from the console team confirm if it checks for the etcd-operator specifically in the kube-system namespace? If so that might need to be changed to look at tectonic-system instead.
/cc @amrutac

squat:
ok to test

squat:
test this please

squat:
ok to test

squat:
test this

squat:
ok to test

squat:
The bootkube module is failing for the Azure platform when experimental features are enabled. Looks like a variable is missing there. That is causing the `null_resource.tectonic` to fail

hongchaodeng:
@squat @Quentin-M 
Can we get this merge before code freeze? Thanks!

hasbro17:
Ready for a review.

Quentin-M:
This would need to get LGTM from the installer team before the freeze. Not sure if anyone will have the bandwidth?

/cc @sym3tri

PR #1702, opened 2017-08-16
modules/bootkube: remove critical pod annotation

squat:
Remove the critical pod annotation and toleration from the
controller-manager and scheduler, since this can result in dangerous
behavior as discussed in
https://github.com/kubernetes-incubator/bootkube/issues/519.

fixes: #734 

cc @aaronlevy

### DIFF
--- a/modules/bootkube/resources/manifests/kube-controller-manager.yaml
+++ b/modules/bootkube/resources/manifests/kube-controller-manager.yaml
@@ -24,8 +24,6 @@ spec:
         k8s-app: kube-controller-manager
         pod-anti-affinity: kube-controller-manager-${kubernetes_version}
         tectonic-operators.coreos.com/managed-by: kube-version-operator
-      annotations:
-        scheduler.alpha.kubernetes.io/critical-pod: ""
     spec:
       affinity:
         podAntiAffinity:
@@ -74,8 +72,6 @@ spec:
         runAsNonRoot: true
         runAsUser: 65534
       tolerations:
-      - key: "CriticalAddonsOnly"
-        operator: "Exists"
       - key: "node-role.kubernetes.io/master"
         operator: "Exists"
         effect: "NoSchedule"
--- a/modules/bootkube/resources/manifests/kube-apiserver.yaml
+++ b/modules/bootkube/resources/manifests/kube-apiserver.yaml
@@ -20,7 +20,6 @@ spec:
         tectonic-operators.coreos.com/managed-by: kube-version-operator
       annotations:
         checkpointer.alpha.coreos.com/checkpoint: "true"
-        scheduler.alpha.kubernetes.io/critical-pod: ""
     spec:
       containers:
       - name: kube-apiserver
@@ -81,8 +80,6 @@ spec:
           readOnly: false
       hostNetwork: true
       tolerations:
-      - key: "CriticalAddonsOnly"
-        operator: "Exists"
       - key: "node-role.kubernetes.io/master"
         operator: "Exists"
         effect: "NoSchedule"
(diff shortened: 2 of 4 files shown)


modules/bootkube/resources/manifests/kube-controller-manager.yaml
```diff
@@ -73,8 +71,6 @@ spec:
           pod-anti-affinity: kube-controller-manager-${kubernetes_version}
           tier: control-plane
       tolerations:
-      - key: "CriticalAddonsOnly"
```
aaronlevy: I'm unsure if we want to remove this toleration though. Might need to look into the upstream behaviors this ties into.

aaronlevy:
@robszumski If this is done without a the same being done in an update payload, please track the skew that this introduces and the version it is introduced in.

To update this currently it would require custom migration code (annotations are not handled for scheduler/controller-manager until the patch support is finalized).

modules/bootkube/resources/manifests/kube-controller-manager.yaml
squat: If we leave the toleration then it is possible that a controller-manager pod could get scheduled into a slot that was designated for a different critical pod. Maybe this is not a big issue.

squat:
ok to test

modules/bootkube/resources/manifests/kube-controller-manager.yaml
aaronlevy: Oh, right. Well we probably do want the controller-manager to have this toleration. it is critical it's just unsafe with the behavior of the rescheduler. But I'm not immediately sure what the interaction would be like. The rescheduler might just keep kicking it off, and then it keeps getting rescheduled.

It's probably just not safe to run the rescheduler on the master nodes at all -- maybe that's the end goal we get to (and then we also remove the annotation / taint). And priority / preemption will deprecate rescheduler anyway).

aaronlevy: This might be less of a big deal now that I'm thinking of it. We should just say "do not run the rescheduler on master nodes, your master nodes should be provisioned such that they have a minimum of X resources and nothing else is scheduled there".

s-urbaniak:
ping @squat, is this ready to merge?

squat:
@s-urbaniak, this fell off my radar. I need to check back in with KVO team to see if patch support was finalized; otherwise I need to write a migration to fix the skew.

diegs:
@s-urbaniak @squat Patch support is merged in so we can trivially replicate this in the KVO for this release.

diegs:
Should we also remove it from the `apiserver`? It sounds like the rescheduler is going away anyway.

aaronlevy:
ref (priority/preemption): https://github.com/kubernetes/features/issues/268

squat:
@diegs I'll make the adjustments in apiserver and line proxy as well

diegs:
@squat is this in or out for `1.7.5`? Just want to know if I should land my KVO PR or not. Thanks!

squat:
@diegs PTAL

squat:
@diegs sorry, I just rebased on master to fix a test failure. PTAL once more :sweat_smile:

aaronlevy:
Sorry to be late to the review - but are we sure we want to remove the toleration?

aaronlevy:
Hmm. Maybe ignore me. The behavior with just the toleration would be weird too (pod would be deleted, then re-created). I guess my concern was if the taint would be used elsewhere (not just re-scheduler), and we actually don't want these pods messed with.

I skimmed the priority / preemption discussion and I don't think this taint will be re-used (we can always revisit if we need to later).

diegs:
@aaronlevy that taint is added by the rescheduler, which this is removing support for (and apparently the rescheduler is deprecated anyway, right)?

https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/

squat:
ok to test

aaronlevy:
Yeah, I retract my concern

squat:
One AWS flake. Merging

PR #1707, opened 2017-08-17
make-update-payload.sh: Include AppVersion's metadata in desired versions.

yifan-gu:
This copies the appversion's metadata in the payload so we can
decide whether to create an appversion by looking at its annotation
in the payload.


yifan-gu:
/cc @abhinavdahiya @dghubble 

Ref https://github.com/coreos-inc/tectonic-channel-operator/pull/173

PR #1708, opened 2017-08-17
modules/tectonic/resources: pin components to masters

squat:
This commit pins the identity, ingress, and console components to master
nodes and configures them with `preferred` anti-affinity to promote HA
configurations.

Tomorrow, when @derekparker is in, we can create the update-spec and add these new manifests to allow clusters to update to this configuration.


alexsomesan:
retest this please

squat:
Closing in favor of #1722 because Jenkins is not picking up the changes to this branch so tests keep failing.

PR #1709, opened 2017-08-17
docs: update the RHEL package version

dm0-:
There is a newer `tectonic-release` package with some minor fixes that should be preferred in the documentation.


dm0-:
Yes, that is the correct format.

PR #1711, opened 2017-08-17
frontend: Add warning for small pod IP address ranges

### LINKED ISSUE #791: Network sizing requirements are ambiguous
In the current Tectonic installer it requests network ranges for both pods and services.  

As there is no further description about the expectations for network sizes, it is quite possible that a user will choose a network size that is incompatible with  their environment.

As an example:

A user requests a network allocation from their network administrator.  The network administrator asks "what size do you need?".  As the systems person cannot give an accurate answer they receive the subnet `10.7.80.0/22` to begin with.  This will lead the systems administrator to know that they will have 1024 IPs to begin working with for assignment.

While walking through the installer, the user will be prompted for the pod and service networks.  Knowing that both networks need to be carved out from the same range (and that each must be a simple subnet, not an aggregation of differently sized networks) it is not unreasonable that they would simple divide the `/22` into 2 networks of size `/23`.

In doing this they assign as follows:

```
Pod Network: 10.7.80.0/23
Service Network: 10.7.82.0/23
```

As Flannel assigns a minimum network size of `/24` this means that a user will (at most) be able to configure a single worker and controller.  It would be ideal if the GUI mechanism could validate and possibly warn of the concerns when using small network sizes (e.g. over 75% utilization of networks vs nodes for example).

Currently if a user attempts to deploy in this fashion with a single controller and two workers the installer will cheerfully allow this to occur and then only one of the workers will have a `flannel.1` interface configured leading to a failed install.

Affects #783

kyoto:
Shows an error box if the number of configured nodes is greater than the pod range will allow. (But does not disable the Next Step button.)

Shows an info box if over 75% of the range will be used by the configured nodes.

This PR only checks for AWS. A future PR will add the warning for bare metal.

See #791

screenshot-1
screenshot-2

### DIFF
--- a/installer/frontend/components/k8s-cidrs.jsx
+++ b/installer/frontend/components/k8s-cidrs.jsx
@@ -1,15 +1,59 @@
+import _ from 'lodash';
+import pluralize from 'pluralize';
 import React from 'react';
+import { connect } from 'react-redux';
 
-import { CIDR } from './cidr';
+import { AWS_TF } from '../platforms';
+import { Alert } from './alert';
+import { CIDR, cidrSize } from './cidr';
 import {
+  AWS_CONTROLLERS,
+  AWS_WORKERS,
+  NUMBER_OF_INSTANCES,
+  PLATFORM_TYPE,
   POD_CIDR,
   SERVICE_CIDR,
 } from '../cluster-config';
+
+const PodRangeWarning = connect(
+  ({clusterConfig}) => ({clusterConfig})
+)(({clusterConfig}) => {
+  const size = cidrSize(_.get(clusterConfig, POD_CIDR));
+
+  // Currently, we only expect to have the node count for AWS because of the wizard screen order
+  if (!size || clusterConfig[PLATFORM_TYPE] !== AWS_TF) {
+    return null;
+  }
+
+  // Flannel assigns a minimum network size of /24 (256 IP addresses)
+  const maxNodes = Math.floor(size / 256);
+
+  const controllers = _.get(clusterConfig, `${AWS_CONTROLLERS}-${NUMBER_OF_INSTANCES}`, 0);
+  const workers = _.get(clusterConfig, `${AWS_WORKERS}-${NUMBER_OF_INSTANCES}`, 0);
+  const nodes = controllers + workers;
+  const utilization = nodes / maxNodes;
+
+  if (utilization < 0.75) {
+    return null;
+  }
+
+  if (utilization > 1) {
+    return <Alert severity="error">
+      <b>Pod Range Too Small</b><br />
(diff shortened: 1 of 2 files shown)


installer/frontend/components/k8s-cidrs.jsx
```diff
@@ -1,15 +1,56 @@
...
+
+  // For AWS, we know the number of nodes, but for bare metal, we probably don't yet know the number of nodes because
+  // that wizard screen comes later
+  if (clusterConfig[PLATFORM_TYPE] === AWS_TF) {
+    const controllers = _.get(clusterConfig, `${AWS_CONTROLLERS}-${NUMBER_OF_INSTANCES}`, 0);
+    const workers = _.get(clusterConfig, `${AWS_WORKERS}-${NUMBER_OF_INSTANCES}`, 0);
+    utilization = (controllers + workers) / maxNodes;
+  }
+
+  if (maxNodes >= 6 && utilization < 0.75) {
```
squat: why this specific exception? what if you have 5 nodes but only 50% utilization? that will still cause an alert. as mentioned below, utilization is derived from maxNodes, so you can actually get away with just saying:
```js
if (utilization < 0.75) {
  return null;
}
```

installer/frontend/components/k8s-cidrs.jsx
```diff
@@ -1,15 +1,56 @@
...
+    return null;
+  }
+
+  const isError = maxNodes < 2 || utilization > 1;
```
squat: What about running a single master node with pod range /24? This would cause an error when it should work just fine.

squat: This test is a little funny IMO; you are using two metrics, one of which is a derivative of the first. Are you only including `maxNodes` because there are times we do not know the real utilization?

I think it would be more correct to do:
```js
const isError = utilization > 1;
```

installer/frontend/components/k8s-cidrs.jsx
```diff
@@ -1,15 +1,56 @@
...
+  }
+
+  // Flannel assigns a minimum network size of /24 (256 IP addresses)
+  const maxNodes = Math.floor(size / 256);
+
+  let utilization = 0;
+
+  // For AWS, we know the number of nodes, but for bare metal, we probably don't yet know the number of nodes because
+  // that wizard screen comes later
+  if (clusterConfig[PLATFORM_TYPE] === AWS_TF) {
```
squat: what about Azure?

kyoto: GUI does not support Azure. I changed the comment to be more general.

installer/frontend/components/k8s-cidrs.jsx
kyoto: Removed bare metal support to make this PR simpler. Will add bare metal support in a future PR when it can be done properly.

installer/frontend/components/k8s-cidrs.jsx
kyoto: Removed bare metal support to that this could be simplified.

PR #1713, opened 2017-08-17
modules/bootkube: fix disruption

squat:
Pull request #1667 introduced a regression in the kube-scheduler and
kube-controller-manager deployments that prevented them from being
created by bootkube start because the pod specs were invalid. The specs
contained a `selector` field, which should have been in the disruption
instead.

This is a very urgent issue. PTAL @diegs @yifan-gu @Quentin-M.

### DIFF
--- a/modules/bootkube/resources/manifests/kube-controller-manager-disruption.yaml
+++ b/modules/bootkube/resources/manifests/kube-controller-manager-disruption.yaml
@@ -7,5 +7,6 @@ spec:
   minAvailable: 1
   selector:
     matchLabels:
-      tier: control-plane
       k8s-app: kube-controller-manager
+      pod-anti-affinity: kube-controller-manager-${kubernetes_version}
+      tier: control-plane
--- a/modules/bootkube/resources/manifests/kube-scheduler-disruption.yaml
+++ b/modules/bootkube/resources/manifests/kube-scheduler-disruption.yaml
@@ -7,5 +7,6 @@ spec:
   minAvailable: 1
   selector:
     matchLabels:
-      tier: control-plane
       k8s-app: kube-scheduler
+      pod-anti-affinity: kube-scheduler-${kubernetes_version}
+      tier: control-plane
--- a/modules/bootkube/assets.tf
+++ b/modules/bootkube/assets.tf
@@ -106,7 +106,7 @@ resource "template_dir" "bootkube" {
     etcd_client_cert = "${base64encode(data.template_file.etcd_client_crt.rendered)}"
     etcd_client_key  = "${base64encode(data.template_file.etcd_client_key.rendered)}"
 
-    kubernetes_version = "${var.versions["kubernetes"]}"
+    kubernetes_version = "${replace(var.versions["kubernetes"], "+", "-")}"
 
     master_count              = "${var.master_count}"
     node_monitor_grace_period = "${var.node_monitor_grace_period}"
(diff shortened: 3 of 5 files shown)


squat:
@diegs it's not your fault; my eyes didn't catch it either. looking back at it it seems like the tests never actually passed :( we just switched to a new way of running the tests that requires adding a label to the PR; it seems that despite adding the label, the smoke tests were not executed.

squat:
@diegs deployment is failing to create because labels are invalid:
```
The Deployment "kube-scheduler" is invalid:
* spec.selector.matchLabels: Invalid value: "kube-scheduler-1.7.1+tectonic.1": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')
* spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"k8s-app":"kube-scheduler", "pod-anti-affinity":"kube-scheduler-1.7.1+tectonic.1", "tier":"control-plane"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: invalid label selector.
```
we need to either transform the `kubernetes_version` before setting it on the manifest or use the `tectonic_version` or something else.

squat:
scheduler and controller manager are now created again. The tests fail at the very end because of the issue in #1714.

modules/bootkube/resources/manifests/kube-controller-manager-disruption.yaml
```diff
@@ -7,5 +7,6 @@ spec:
   minAvailable: 1
   selector:
     matchLabels:
-      tier: control-plane
       k8s-app: kube-controller-manager
+      pod-anti-affinity: kube-controller-manager-${kubernetes_version}
```
yifan-gu: Why is this needed for disruption budget?
Without this, the budget is able to select both old and new CM.
With this, then it only selects the new CM and old CM's can all be evicted, what's the purpose of doing this? @squat

modules/bootkube/resources/manifests/kube-scheduler-disruption.yaml
```diff
@@ -7,5 +7,6 @@ spec:
   minAvailable: 1
   selector:
     matchLabels:
-      tier: control-plane
       k8s-app: kube-scheduler
+      pod-anti-affinity: kube-scheduler-${kubernetes_version}
```
yifan-gu: same here.

modules/bootkube/resources/manifests/kube-controller-manager-disruption.yaml
squat: @yifan-gu these are changes that @diegs added, i'm just putting them in the right place. Ask him. If this is not right, then I am just going to revert the changes and we can let @diegs submit the right PR.

squat:
closing in favor of #1718

modules/bootkube/resources/manifests/kube-controller-manager-disruption.yaml
yifan-gu: @squat I looked at @diegs 's PR, his didn't add the selector in the *disruption budget*. Am I missing something?

PR #1704, opened 2017-08-16
modules/tectonic/resources/manifests/updater: Change all AppVersion TPRs to CRDs.

rithujohn191:

PR #1714, opened 2017-08-17
modules/tectonic: fix CRD API group

squat:
This commit fixes the API group of the Migration Status CRD. The correct
API group is `kvo.coreos.com`. The incorrect group was causing the
Tectonic systemd unit to fail to bring up all the components.

Fixes a regression introduced in: #1704 

cc @yifan-gu @rithujohn191


rithujohn191:
Thanks for this

squat:
tests are broken because of #1713. we need both of these fixes for tests to pass again. Merging this; rebasing 1713, then tests should pass there.

PR #1715, opened 2017-08-17
make-update-payload.sh Fix the payload for kubernetes desired version.

yifan-gu:
Thanks @abhinavdahiya for catching this!


Quentin-M:
@yifan-gu For which release do we need that particularly?

yifan-gu:
@Quentin-M This is for 173.

dghubble:
LGTM

PR #1718, opened 2017-08-17
Revert "Use kubernetes version for anti-affinity."

squat:
Pull request #1667 introduced a regression in the kube-scheduler and
kube-controller-manager deployments that prevented them from being
created by bootkube start because the pod specs were invalid. The specs
contained a selector field, which should have been in the disruption
instead.

This reverts commit 6a37f2e365094165254679d3cac100ab7e20f143.

@yifan-gu @diegs


squat:
One of the GUI tests flaked. Everything else is looking good. Merging in the interest of time so we can get back to a working state on master.

dghubble:
SGTM

PR #1719, opened 2017-08-18
module/bootkube: Fix anti-affinity labels / selectors for controller manager and scheduler.

yifan-gu:
The anti-affinity label should match the kubernetes version instead
of the tectonic version.

The selector of the controller manager and scheduler should not
contain the anti-affinity label.

### DIFF
--- a/modules/bootkube/resources/manifests/kube-scheduler.yaml
+++ b/modules/bootkube/resources/manifests/kube-scheduler.yaml
@@ -12,12 +12,16 @@ spec:
     type: RollingUpdate
     rollingUpdate:
       maxUnavailable: 0
+  selector:
+    matchLabels:
+      tier: control-plane
+      k8s-app: kube-scheduler
   template:
     metadata:
       labels:
         tier: control-plane
         k8s-app: kube-scheduler
-        pod-anti-affinity: kube-scheduler-${tectonic_version}
+        pod-anti-affinity: kube-scheduler-${kubernetes_version}
       annotations:
         scheduler.alpha.kubernetes.io/critical-pod: ""
     spec:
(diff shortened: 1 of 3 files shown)


squat:
ok to test

sym3tri:
retest this

alexsomesan:
Closed and re-opened to run full test suite.

alexsomesan:
retest this please

alexsomesan:
@yifan-gu Would you be so kind to rebase this?

I'm seeing test failures that should be resolved by recently merged stuff (the CIDR validation in the UI tests).

yifan-gu:
Rebased @alexsomesan @Quentin-M

modules/bootkube/resources/manifests/kube-scheduler.yaml
```diff
@@ -12,12 +12,16 @@ spec:
     type: RollingUpdate
     rollingUpdate:
       maxUnavailable: 0
+  selector:
+    matchLabels:
+      tier: control-plane
+      k8s-app: kube-scheduler
```
diegs: I'm pretty sure we'll need to add the anti affinity here eventually for patch merges to work.

yifan-gu: @diegs Could you elaborate? Will there be a conflict? Despite that, ideally we don't need the anti-affinity in the selector right?

diegs: It gets defaulted into the selector. So then when you update it later, unless you update in both places you can hit conflicts.

See: https://github.com/kubernetes/kubernetes/issues/26202#issuecomment-228558649

yifan-gu: After OOB discussed with @diegs , we think this should work since we now specify both labels and selectors, so selectors will not be default to the label value. Someone also mentioned this in https://github.com/kubernetes/kubernetes/issues/26202#issuecomment-299164110
@diegs will manually test this if he has a chance.

Issue #1731, opened 2017-08-18
Not using custom Flannel cni plugin binary


squat:
When we added support for Calico, we introduced a skew where we do not use the cni binaries from the `flannel_cni` image but rather default to the cni binaries from hyperkube. Flannel went from installing the cni binaries in `/opt/cni/bin`, the default cni plugin directory, to `/var/lib/cni/bin` [0][1], however, the kubelet only ever looks for cni binaries in that directory if Calico is enabled [2][3][4]. This means that we cannot fine-tune the version of Flannel used in Tectonic by applying a new Flannel deployment and we introduce more skew between deployments of Tectonic.

xref #1723

[0] https://github.com/coreos/tectonic-installer/commit/977ca19ce3c6c9add9ba0224884efaaf81898140#diff-1003a3172fb7f5ad31291f349661ac38L103
[1] https://github.com/coreos/tectonic-installer/blob/master/modules/net/flannel-vxlan/assets.tf#L8
[2] https://github.com/coreos/tectonic-installer/blob/master/platforms/aws/main.tf#L89
[3] https://github.com/coreos/tectonic-installer/blob/master/modules/aws/ignition/ignition.tf#L44
[4] https://github.com/coreos/tectonic-installer/blob/master/modules/aws/ignition/resources/services/kubelet.service#L32

PR #1732, opened 2017-08-18
modules/*,platforms/*: use custom flannel cni bin

squat:
This commit fixes issue #1731, which describes a regression introduced
when we added support for Calico network policies. At this point,
whenever Calico is not enabled, we do not use the cni plugin
binaries installed by Flannel and default to using the binaries in
hyperkube. This means that we cannot update our cni plugins
just by updating the Flannel deployment and introduces more skew
between deployments of Tectonic.

cc @abhinavdahiya @s-urbaniak

### DIFF
--- a/modules/aws/ignition/ignition.tf
+++ b/modules/aws/ignition/ignition.tf
@@ -41,7 +41,6 @@ data "template_file" "kubelet" {
     cluster_dns_ip         = "${var.kube_dns_service_ip}"
     node_label             = "${var.kubelet_node_label}"
     node_taints_param      = "${var.kubelet_node_taints != "" ? "--register-with-taints=${var.kubelet_node_taints}" : ""}"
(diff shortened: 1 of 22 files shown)


abhinavdahiya:
@aaronlevy

aaronlevy:
We did this on purpose originally to keep all cluster configuration the same unless you turned on the "experimental" network policy feature.

Not opposed to this as this gives us more control of CNI version for all new clusters, but unfortunately we don't have a good way of changing this on existing clusters - so we were mitigating that skew until we had some plan. Short of being able to modify a the cloud-provider user-data though, there isn't a plan at this point.

If this is merged, @sym3tri @robszumski please track this configuration skew and the version it was introduced in (old clusters will continue using CNI binaries we ship with hyperkube, new clusters will use CNI binaries shipped with CNI daemonsets).

squat:
closing since we've changed approaches

PR #1734, opened 2017-08-21
WIP: Updating Admission Controllers for apiserver

coresolve:
- Adding a config flag for admission controllers.
- Updatig the default admission controller list based on
  https://kubernetes.io/docs/admin/admission-controllers/#is-there-a-recommended-set-of-plug-ins-to-use
- Also adding PodSecurityPolicy to the default set.


cpanato:
Hello @coresolve Thanks for the PR, This is still in WIP or can we start the review ?

thanks

s-urbaniak:
raising the flag here, as this is touching manifests.

ping @robszumski @coresolve how critical is this to have configurable at all?
also /cc @aaronlevy @diegs for further discussion.

robszumski:
Folks have been requesting things and we should probably be a bit more proactive about some of these. Adding the defaults in this PR sound ok to me, but my hesitation is whether this should be exposed for customization. Interested to hear how the migration would go.

diegs:
The litmus test we are working on regarding whether it should be a configurable parameter is how often will it be customized? 80%+ => add a parameter. Otherwise, operators can edit the DaemonSet manually (and we will carry the customized flag forward through upgrades automatically).

Migrations are relatively easy for each, I think that's a secondary concern vs the product decision of config parameter creep.

@robszumski @aaronlevy

aaronlevy:
This is somewhat of a special one - as people likely will want to have individual control of this, and it might be perfectly valid to let them (e.g. alwaysPullImage).

If we do set this as a parameter, we're essentially saying "we now control this field - if you change it, it will be a conflict during upgrade".

From our end, however, we also should be enforcing a subset of these (if they got turned off - some assumptions we make would be broken - e.g. service account). So this makes me feel like we might want to do some more nuanced handling of this field specifically (e.g. we provide a required list, make sure those exist - but if user adds more to the slice - we're not going to overwrite them).

In the short term - we could enforce this field and see if it causes issue (e.g. "it says conflict when I try and upgrade"). Or we could wait until we have better handling.

Also, as a side-note - pod-security-policy is not one of the default recommended admission controllers. Rather it looks to be `NamespaceLifecycle,LimitRanger,ServiceAccount,PersistentVolumeLabel,DefaultStorageClass,ResourceQuota,DefaultTolerationSeconds`. Also just turning it on here isn't enough, I believe you also need to enable a new api endpoint.

If we want to explore enabling pod-security-policy I think that probably should be a separate discussion.

aaronlevy:
My initial thought would be that we do not add this as a parameter (rather it just defaults to the recommended list at install time). Then we enforce our "required" list in the update business logic. A user is then free to add more admission controllers, but a removal of our required would cause an update conflict (and we're not maintaining a new install-time parameter)

diegs:
Another addendum, after @aaronlevy and I synced offline:

- Regarding the parameter vs. not, I stand by my original comment (is this something that is often customized, or do we just set recommended defaults that work for most customers?)
- If we are changing the defaults, this will work fine with patch merging. We will be able to update existing clusters that have *not* changed these values automatically. If a user *did* customize this value we will flag that as a conflict and then they can decide what to set as the new value (ours, theirs, or some combination)

coresolve:
@diegs @aaronlevy I think there is a case to be made that customers would want to change the admission controllers. In the case of the psp admission controllers this one is interesting as the k8s api already supports writing pod security policies but without the admission controller they are effectively useless.

sym3tri:
@coresolve Thanks for the contribution, but unfortunately it looks like this one got stale. We won't be able to merge it now that the manifests have moved into the KVO. If one doesn't exist can you please just file an issue with the desired behavior/use-case?

diegs:
This is what we have for the apiserver in track 2:

https://github.com/coreos-inc/tectonic-operators/blob/master/operator/kube-core/spec/manifests/kube-apiserver.yaml#L27

coresolve:
Thanks @diegs that looks great.

PR #1741, opened 2017-08-22
frontend/tests: GUI Tests README fixes

kyoto:
Google Chrome v60 is now in the stable channel. I think this means there should be no need to specify the Chrome binary path.


kyoto:
ok to test

PR #1755, opened 2017-08-23
Azure: delete os disk on termination

alexsomesan:
I seems the current installer can leak OS disks when deleting instances on Azure (observed in the CI environment).

This change attempts to address that problem.

/cc: @mxinden @cpanato


mxinden:
@alexsomesan Thanks for addressing this!

PR #1760, opened 2017-08-23
identity: cast tectonic_admin_email to lower case

alekssaul:
Per conversation with @rithujohn191, Tectonic-Identity expects `tectonic_admin_email` to be all lower case. Use of upper & lower case characters for Admin email (ex: `tectonic_admin_email = Admin@Example.com`) causes unexpected behavior on tectonic-console.

This change modifies Tectonic module to make sure that Tectonic assets generated uses admin_email in all lower case format. Change also include updates documentation for the `tectonic_admin_email` variable to make a note of the requirements.


config.tf
```diff
@@ -287,7 +287,7 @@ The e-mail address used to:
 1. login as the admin user to the Tectonic Console.
 2. generate DNS zones for some providers.
 
-Note: This field MUST be set manually prior to creating the cluster.
+Note: This field MUST be in all lower-case e-mail address format and set manually prior to creating the cluster.
```
squat: I don't think we need to specify the lower-case format here. That is an implementation detail that is taken care of in Terraform.

alekssaul: Thanks Lucas. 

Agreed, the idea was to "train" the user as they may enter mixed characters but provisioning changes it based on the requirement. Happy to remove the docs bits if the team feels like that's the way to go.

rithujohn191: I actually think the documentation would be helpful...it shouldn't seem like we are changing user inputted values just to fit our back end needs

PR #1767, opened 2017-08-24
frontend: Better validation of email addresses.

ggreer:
Previously, we were using a regex that only allowed a subset of valid emails.
Now this will work with email addrs that are non-ascii.

### DIFF
--- a/installer/frontend/validate.js
+++ b/installer/frontend/validate.js
@@ -38,15 +38,16 @@ export const validate = {
     return 'Invalid private key. Check your key for copy/paste errors';
   },
 
-  email: (s) => {
-    // from: https://www.w3.org/TR/html5/forms.html#valid-e-mail-address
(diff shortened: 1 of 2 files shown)


installer/frontend/validate.js
```diff
@@ -38,15 +38,23 @@ export const validate = {
...
-    if (EMAIL_RE.test(s)) {
-      return;
+  email: (s = '') => {
+    let errMsg = validate.nonEmpty(s);
+    if (errMsg) {
+      return errMsg;
+    }
+    errMsg = 'Invalid email address.';
+    const [name, domain] = s.split('@');
+    if (!domain) {
```
kyoto: Nit: Maybe abbreviate to
```
if (validate.nonEmpty(name) || /\s/g.test(name) || !domain || validate.domainName(domain)) {
  return 'Invalid email address.';
}
```
or similar?

ggreer: Good call on the code golf. Fix pushed.

Ideally I'd like to chain validators. eg: pass an array of validators into a form & have them all required to pass. That way we could avoid an explosion of similar validators (domainOptional, domainRequired).

PR #1776, opened 2017-08-25
monitoring: bump monitoring to 1.5.2, CRD compatiblity

colhom:
This will need to be cherry-picked on to 1.7.3 branch as well.

\cc @brancz

PR #1656, opened 2017-08-11
tectonic: remove monitoring assets

brancz:
As discussed with @sym3tri and @mxinden, we're removing the monitoring assets in favor of them being fully managed by the Tectonic Prometheus Operator.

/cc @fabxc 

This is missing some tests that @mxinden will take care of so until then, hold off on merging.

PR #1779, opened 2017-08-25
monitoring: 1.7.3 release - bump monitoring to 1.5.2, CRD compatiblity

colhom:
cherry-pick of #1776 on to 1.7.3 release branch

\cc @Quentin-M


Quentin-M:
Unfortunately, it looks like we can't get that in.

colhom:
Ack- we should definitely wait until the monitoring folks say the new versions are ready :)

brancz:
We actually have a story of removing the monitoring manifests all together from the tectonic installer, and instead just let them be provisioned by the TPO, so I think we can actually just close this.

https://github.com/coreos/tectonic-installer/pull/1656

s-urbaniak:
as per comment https://github.com/coreos/tectonic-installer/pull/1779#issuecomment-325579900, closing this.

PR #1791, opened 2017-08-29
Rspec: cluster platform detection

alexsomesan:
Autodetection of the cloud provider platform for the given `.tfvars` file.

This enables the 'withCluster' shared example to create the cluster object for the right platform based on the content of the tfvars file.

It looks for variables called `tectonic_<cloud-provider>_*` in the `.tfvars` file to establish the target platform for the cluster.


tests/rspec/lib/cluster_factory.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require 'aws_cluster'
+require 'azure_cluster'
+
+# Creates a platform specific Cluster object based on the provided tfvars file.
+#
+module ClusterFactory
+  def self.from_tf_vars(tf_vars_file)
+    cluster_class = "#{tf_vars_file.platform.capitalize}Cluster"
```
mxinden: `capitalize` works for Azure clusters, but not for AWS clusters, because `aws` capitalized turns into `Aws`, but the class is called `AWSCluster`. Instead of capitalizing here, can we have the cluster type enum written correctly (`Azure`, `AWS`) and lower case the strings in line 43 of tests/rspec/lib/tfvars_file.rb?

tests/rspec/lib/tfvars_file.rb
```diff
@@ -27,6 +31,20 @@ def region
...
+  def prefix
+    prefix = File.basename(path).split('.').first
+    raise 'could not extract prefix from tfvars file name' if prefix == ''
+    prefix
+  end
+
+  def platform
+    PLATFORMS.each do |platform|
+      return platform if data.keys.any? do |key|
+        key.start_with?("tectonic_#{platform}")
```
mxinden: s/platform/platform.downcase
(See comment above)

tests/rspec/lib/tfvars_file.rb
```diff
@@ -1,5 +1,9 @@
+# frozen_string_literal: true
+
 require 'json'
 
+PLATFORMS = %w[aws azure metal vmware].freeze
```
mxinden: s/aws azure metal vmware/AWS Azure Metal VMWare/

tests/rspec/lib/tfvars_file.rb
alexsomesan: I fixed it the other way around: I adapted the AWSCluster class to follow the camel-case convention.
The reason is that I would like to keep the data from the `.tfvars` as the source of truth (as I was mentioning earlier on in the project). Seems like a simpler rule to follow: the tfvars file is always right.

mxinden: Sounds good to me. Thanks for the changes.

mxinden:
@alexsomesan Can you rebase?

alexsomesan:
retest this please

mxinden:
@alexsomesan Thanks for the help!

PR #1792, opened 2017-08-29
frontend: Move BM network config page after nodes config and add the CIDR size warning for BM

kyoto:
Also fixes a JS error when the CIDR is invalid.


installer/frontend/components/cidr.jsx
```diff
@@ -20,6 +20,9 @@ const CIDRTooltip = connect(
   ({clusterConfig}, {field}) => ({clusterConfig: clusterConfig, value: _.get(clusterConfig, field)})
 )(({value}) => {
   const addresses = cidrSize(value);
+  if (addresses === undefined) {
+    return null;
+  }
```
ggreer: Why not `_.isInteger(addresses)`?

PR #1795, opened 2017-08-29
[WIP] tests: reduce machine type and count

cpanato:


cpanato:
I guess is ready, ran several times and no new issue was observed.

@alexsomesan @sym3tri when you guys have time for review

tests/smoke/azure/vars/basic.tfvars
```diff
@@ -1,17 +1,17 @@
-tectonic_azure_etcd_storage_account_type = "Premium_LRS"
-tectonic_azure_etcd_vm_size = "Standard_DS2_v2"
-tectonic_azure_master_storage_account_type = "Premium_LRS"
-tectonic_azure_master_vm_size = "Standard_DS2_v2"
-tectonic_azure_worker_storage_account_type = "Premium_LRS"
-tectonic_azure_worker_vm_size = "Standard_DS2_v2"
+tectonic_azure_etcd_storage_type = "Standard_LRS"
+tectonic_azure_etcd_vm_size = "Standard_D2"
```
alexsomesan: Just one small nit: Is there a reason for using the first generation D1 / D2?

There are newer generations of the D1 / D2, called D1_v2 or even D1_v3 depending on the region.
These usually provide better performance for the same or less of the price.
Also, when using less than the current generation we run the risk of them being taken out of service in the datacenters and resulting in errors about not enough capacity.

cpanato: makes a lot of sense, will change! thanks for pointing out.

cpanato:
changes are not related to AWS only to azure and this test `IntegrationTest AWS Installer Gui` is failing. merging...

PR #1803, opened 2017-08-30
frontend: Add Start Over button back to install screen

kyoto:
Also moves canReset to be an attribute of the trail pages.

screenshot-1


squat:
What happens if you start over from the install screen? Does the installer destroy the half-created resources?

kyoto:
@squat No, it won't destroy any half-created resources, which was actually the motivation for removing it. Unfortunately, there are one or two issues that need resolving first, then we will remove it again.

PR #1804, opened 2017-08-30
Replace aws test implemented in bash

mxinden:
This PR removes our main AWS test that was implemented in bash. In order to replace it this PR introduces in addition in RSpec:
- Grafiti to run after failed destroy
- Assume AWS tectonic-installer role


mxinden:
Merging here. Failures are not related to RSpec. Only RSpec failure is due to earlier leftover files which trigger a permission denied.

PR #1805, opened 2017-08-30
frontend/tests: GUI test fix

kyoto:
Make `setField()` more robust, which hopefully fixes the occasional GUI test flakes like: `Expected element <#podCIDR> to have value that equals: "10.2.0.0/16" - expected "equals '10.2.0.0/16'" but got: "10"`.

Fix for Bare Metal tests.

Also speeds up manual boot page tests by waiting until the download button is present instead of just waiting for 10 seconds.


squat:
One of the azure tests flaked. Merging

PR #1812, opened 2017-08-31
tests: Always run terraform_destroy in aws vpc test

mxinden:
Before, whenever the vpn connection was never established, `@vpn_connection.stop` would fail, which meant that `terraform_destroy` was never executed.

### DIFF
--- a/tests/rspec/lib/aws_vpc.rb
+++ b/tests/rspec/lib/aws_vpc.rb
@@ -74,8 +74,11 @@ class AWSVPC
 
   def destroy
     @vpn_connection.stop
+  rescue
+    raise 'could not disconnect from vpn'
+  ensure
     terraform_destroy
-    FileUtils.cp '/etc/resolv.conf.bak', '/etc/resolv.conf'
+    recover_etc_resolv
   end
 
   def terraform_destroy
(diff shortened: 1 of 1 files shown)


mxinden:
@cpanato I think this is quite an important fix so that we are not leaking resources anymore.

mxinden:
All green except one unrelated Azure failure.

cpanato:
this is very important! great job!! 🎉

tests/rspec/lib/aws_vpc.rb
```diff
@@ -73,9 +73,17 @@ def parse_subnets(tf_out)
...
+    rescue
+      # Make sure to always run terraform_destroy
+      terraform_destroy
+      recover_etc_resolv
+      raise 'could not disconnect from vpn'
+    end
+
     terraform_destroy
-    FileUtils.cp '/etc/resolv.conf.bak', '/etc/resolv.conf'
+    recover_etc_resolv
```
cpanato: can we use `ensure` then we can avoid this code duplication. `ensure` will always run no matter what, then in the `rescue` we get the error and log it

what do you think?

mxinden: Thanks for the hint!

mxinden:
@cpanato could you have another look?

mxinden:
All green except unrelated Bare Metal tests. I will merge here. @cpanato thanks for review.

PR #1816, opened 2017-08-31
Fix legacy release scripts

Quentin-M:
Cherry-picks https://github.com/coreos/tectonic-installer/pull/1809

- Fixes an issue where the tag was never used as part of the version, as initially intended (intro in #987) 
- Fixes an issue where the commit SHA was used for the release instead of the provided tag (intro in #1613) - therefore creating invalid tarball names and breaking the Github release creation (`422 Unprocessable Entity` because no tag with that name)
- Fixes a issue that existed for too long where the Github release was created without a name provided - letting Github pick a default name for the release, a.k.a. the git commit message, and thus forcing the release manager to rename the release manually after the fact

/cc @sudhaponnaganti

### DIFF
--- a/git-version
+++ b/git-version
@@ -8,8 +8,8 @@ VERSION=$(git rev-parse HEAD)
 # check if the current commit has a matching tag,
 # and prepend the matching tag to the version, if available.
 TAG=$(git describe --exact-match --abbrev=0 --tags ${VERSION} 2> /dev/null || true)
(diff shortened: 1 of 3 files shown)


Quentin-M:
Tested by re-creating the 1.7.3-tectonic.1-rc.2.

Quentin-M:
Thanks

PR #1818, opened 2017-09-01
Ingress certs

squat:
Enable user-provided signed certificates for ingress.
fixes INST-148
cc @s-urbaniak @alexsomesan @sym3tri @kyoto


squat:
closing an

modules/tls/ingress/outputs.tf
```diff
@@ -0,0 +1,11 @@
+output "ca_cert_pem" {
+  value = "${data.null_data_source.ingress_certs.inputs["ca_cert_pem"]}"
```
squat: I don't think this value is actually needed anywhere but I added it to align with the API in #1818. We can always keep this just in case for future use.

installer/frontend/components/certificate-authority.jsx
```diff
@@ -87,6 +87,55 @@ export const CertificateAuthority = connect(
...
+                I'll provide a CA-signed certificate, certificate key, and CA certificate.
+              </label>
+              <p className="text-muted wiz-help-text">Your certificate will be used by the Tectonic Console.</p>
+            </div>
+            <div className="wiz-radio-group__body">
+              {
+                caType === 'ca-signed' && <div>
+                  <div className="row form-group">
+                    <div className="col-xs-12">
+                      <WithClusterConfig field={INGRESS_CERTIFICATE}>
```
kyoto: I was told by @kans that `WithClusterConfig` is legacy and we should be switching over to `Connect` instead.

squat: 100% agree. Im using it in this PR for consistency. We'll migrate this component over in a separate PR.

modules/tectonic/resources/manifests/console/deployment.yaml
```diff
@@ -98,6 +98,8 @@ spec:
           value: /etc/tectonic-ca-cert-secret/ca-cert
         - name: BRIDGE_LICENSE_FILE
           value: /etc/tectonic/licenses/license
+        - name: BRIDGE_DEX_CLIENT_CA_FILE
+          value: /etc/tectonic-identity-grpc-client-secret/ca-cert
```
squat: thanks for the heads up, @s-urbaniak

installer/frontend/components/certificate-authority.jsx
kyoto: 👍

modules/tectonic/resources/manifests/console/deployment.yaml
s-urbaniak: no problem 👍

s-urbaniak:
@sym3tri @squat just a logistical heads-up: As discussed OOB #1811 introduces TLS modules in a fashion that modules can be swapped manually, as described in [modules/tls/ingress/user-provided/README.md](https://github.com/s-urbaniak/tectonic-installer/blob/40da8e2a840df0254c15c5b90947e561812bc563/modules/tls/ingress/user-provided/README.md).

This PR on the other hand uses ternaries and a null data source inside the ingress TLS module to switch module functionality (user-provided vs self-signed).

Once swappable module land (INST-143), #1811 can be integrated in the UI.

I'm leaving the final call to @sym3tri to decide which PR lands first. I am leaning towards #1811, because in conjunction with INST-143 it will be the cleaner approach from a TF perspective.

squat:
@s-urbaniak yes sounds good. If swappable modules (e.g. via symlinks rather than by hand) will land in the short term then we may be able to delay/replace this PR. Do we have a projection for when it will be ready? It would definitely be simpler/flatter from a TF point of view, though arguably more complex in another respect. It all comes down to how urgent ca-signed ingress certs in the GUI are. I tried to keep the API for this PR as close to #1811 as possible to facilitate a rebase on that branch. We could of course also merge a combination of the two and then add swappable modules down the line.

s-urbaniak:
@squat INST-143 is planned for the next sprint. I set the do-not-merge label tentatively.

adarshaj:
I am interested in this, I tried to manually overwrite the ingress tls secrets with letsencrypt certificates and locked myself out of dashboard, because login calls fail with invalid_code (-_-)! (I was seeing this error on the default nginx backend logs - `OCSP_basic_verify() failed (SSL: error:27069076:OCSP routines:OCSP_basic_verify:signer certificate not found)` with an HTTP error of 499). I hope this PR would make it possible to set CA for ingress and thereby not cause anyone to lock themselves out of dashboard!

s-urbaniak:
@adarshaj you can follow the instructions at https://github.com/coreos/tectonic-installer/tree/master/modules/tls/ingress/user-provided to set up user provided ingress certs. We are working on a way to automate this process.

adarshaj:
@s-urbaniak Thank you, I followed the steps there and got it to working state (on bare-metal)! :tada:

sym3tri:
@squat I'm going to close this out for now since it's stale. Please keep your branch around & let's revisit once we have a swappable module solution.

PR #1824, opened 2017-09-01
platforms/metal: unify kubelet, install as enabled

lucab:
This updates kubelet setup on bare-metal to unify with all other
platforms. It gets rid of path activation and enables kubelet.service
directly from ignition.


lucab:
NB: not saying that a crash-looping kubelet.service is better than a path-activated one (global approach can be revisited later by the team if they wish). This just unifies all platforms so that it's easier to reason about the bootstrapping flow.

lucab:
Gentle ping for a review.

s-urbaniak:
ok to test

s-urbaniak:
ok to test

s-urbaniak:
CI failure is unrelated (Azure), hence merging.

PR #1397, opened 2017-07-14
Changed hardcoded value of IOPS to 100

simnalamburt:
This is to prevent Terraform from re-creating the master and worker nodes right away after a deployment where root_volume_type is not io1.

Issue #1035 is reproduced exactly the same for master and worker nodes.

###### References
- https://github.com/coreos/tectonic-installer/issues/1035
- https://github.com/coreos/tectonic-installer/pull/1166
- https://github.com/coreos/tectonic-installer/commit/93d485d9

PR #1826, opened 2017-09-01
Fix terraform resource churn

mrwacky42:
Possibly related to this issue with the Terraform AWS provider:
terraform-providers/terraform-provider-aws#1521

Terraform or the AWS API report `0` IOPS for `gp2` volumes after creation.
If I create a `gp2` volume with these modules, every subsequent run of
Terraform will attempt to destroy and recreate the LC in order to make
IOPs match. This change sets IOPS to 0 for non-`io1` type volumes.

Since the number of IOPS for `gp2` volume is based on the size, Terraform
arguably should only accept the `iops` parameter for `io1` volumes.

Interestingly, for the etcd nodes, this problem does not occur because
the Terraform state shows 100 IOPS even for `gp2` volumes.

### DIFF
--- a/Documentation/variables/aws.md
+++ b/Documentation/variables/aws.md
@@ -10,7 +10,7 @@ This document gives an overview of variables used in the AWS platform of the Tec
 | tectonic_aws_config_version | (internal) This declares the version of the AWS configuration variables. It has no impact on generated assets but declares the version contract of the configuration. | string | `1.0` |
 | tectonic_aws_etcd_ec2_type | Instance size for the etcd node(s). Example: `t2.medium`. Read the [etcd recommended hardware](https://coreos.com/etcd/docs/latest/op-guide/hardware.html) guide for best performance | string | `t2.medium` |
 | tectonic_aws_etcd_extra_sg_ids | (optional) List of additional security group IDs for etcd nodes.<br><br>Example: `["sg-51530134", "sg-b253d7cc"]` | list | `<list>` |
-| tectonic_aws_etcd_root_volume_iops | The amount of provisioned IOPS for the root block device of etcd nodes. | string | `100` |
+| tectonic_aws_etcd_root_volume_iops | The amount of provisioned IOPS for the root block device of etcd nodes. Ignored if the volume type is not io1. | string | `100` |
 | tectonic_aws_etcd_root_volume_size | The size of the volume in gigabytes for the root block device of etcd nodes. | string | `30` |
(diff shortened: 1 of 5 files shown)


squat:
@mrwacky42 this generally looks pretty good but seems fragile. Why not update the etcd node launch config as well? And what happens when that node's IOPs in Terraform state is no longer 100?

mrwacky42:
@squat - I based this change on the current behavior of Terraform and AWS.

The etcd nodes are not using autoscaling, so there is no launch configuration, just `aws_instance` resources. 

For `aws_instance` resources, Terraform state currently reports `100` iops (for non-`io1` EBS volumes). For `aws_launch_configuration` resources, Terraform state currently reports `0` iops.

There is a PR attached to the original issue I linked above to fix the `aws_instance` to behave consistent with `aws_launch_configuration`: https://github.com/terraform-providers/terraform-provider-aws/pull/1573

squat:
thank I see now. From the issue you linked, it looks like even settings iops to 100 for gp2 volumes will not always work. @fraenkel writes that for large volume sizes the iops can be reported to be 150, so we may still have resource churn.

squat:
As @mrwacky42 correctly, notes, iops defaults to 0 for gp2 volumes in launch configurations, but defaults to size/3 for instance resources. This rcplsind why @simnalamburt was seeing the issue in https://github.com/coreos/tectonic-installer/pull/1397#issuecomment-317329260 for only the volumes used in launch configurations but not the volumes used etcd nodes. This PR partially undoes https://github.com/coreos/tectonic-installer/pull/1397. We should merge https://github.com/coreos/tectonic-installer/pull/953 next to ensure that iops is set correctly for volumes in instance resources.

squat:
closing and reopening to trigger tests

squat:
ok to test

squat:
@mrwacky42 This is failing with:
```
misformatted files (run 'terraform fmt .' to fix): platforms/aws/variables.tf
```
can you please fix the formatting?

mrwacky42:
*pushes a fixed commit*

squat:
@mrwacky42 no, rebasing was good :+1:. Linting is still failing. Can you please `make docs && make examples` and commit the result?

mrwacky42:
I cheated and ran `make docs examples`, then rebased again.

😀

squat:
@mrwacky42 had an unrelated flake in one of the azure tests. Merging!

PR #1829, opened 2017-09-04
Double dash flag consistency.

enxebre:
All etcd flags are preceded by double dash except for "-peer-client-cert-auth" which can be found with either single and double dash (modules/vmware/etcd/ignition.tf, modules/bootkube/resources/experimental/bootstrap-manifests/bootstrap-etcd.yaml)

Not sure if there's any reason behind this, so just adding consistency here by setting always double dash.


squat:
ok to test

Issue #1567, opened 2017-08-02
Azure console LB incorrectly proxies to master pool


squat:
instead of proxying to worker pool, the console LB proxies to the master pool.

PR #1834, opened 2017-09-05
[WIP] tectonic/azure: Move Tectonic components to worker nodes

justaugustus:
* azure/vnet: Creates a second LB for Ingress traffic e.g., Tectonic Console, which is backed by the worker nodes
* tectonic: Moves Tectonic components to worker nodes, removes tolerations

Fixes: #1567


justaugustus:
WIP, as this would need to happen on other cloud providers.
I'd like to make sure the approach is okay before I clean this up in other places.

PR #1842, opened 2017-09-05
modules/aws: fix AWS iops default

squat:
This commit fixes INST-96, which notes that users cannot customize the
IOPS of any AWS nodes unless the nodes are of type `io1`.

cc @alexsomesan


mrwacky42:
See also #1826 which conflicts with this. I don't know what INST-96 says, but I don't think iops are configurable for EBS volumes that are not `io1`.

squat:
Closed in favor of coordinating #1826 and #953.

PR #1846, opened 2017-09-06
tests/ci: Keep Jenkins build logs for at least one week

cpanato:


mxinden:
@cpanato What does the "-1" mean?

cpanato:
@mxinden means no value. we will keep the build in days only and not in # of builds

mxinden:
@cpanato How about we set it to 1000? Just in case we have a messed up trigger that loops.

cpanato:
@mxinden done

PR #1851, opened 2017-09-06
Tests: pin Ruby version for tools like rbenv

alexsomesan:
Had this lying around for my environment.

It's brings a better degree of confidence about which version of Ruby will be used to run the tests (when tools like rbenv are in use).


mxinden:
@alexsomesan Is there a way to enforce this version in `images/tectonic-smoke-test-env/Dockerfile`?

alexsomesan:
@mxinden 
We could, if we used rbenv to install ruby in the Docker image instead of the distro package. 

I have some experimental alternative image based on Alpine that installs ruby from rbenv, but haven't tested it enough yet to PR it.

Without a version manager in the docker image, this setting only applies to desktop development. Which is where you're likely to have more than one version anyway.

mxinden:
@alexsomesan Or we read the `ruby-version` file during `docker build` to make sure we install the right version?

PR #1853, opened 2017-09-06
frontend: Add id attributes to the session token radio inputs

kyoto:
Otherwise they end up with duplicate ids.

Also makes it easier to test with Nightwatch.


cpanato:
merging, the error was in Azure to destroy, however, the cleanup get it

PR #1858, opened 2017-09-07
tests/ci: fix dsl issue

cpanato:


tests/jenkins-jobs/tectonic_installer_multibranch_job.groovy
```diff
@@ -3,31 +3,31 @@
...
+      buildOriginPRHead(false)
+      buildOriginPRMerge(true)
+    }
+  }
+
+  configure { project ->
+    project / 'sources' / 'data' / 'jenkins.branch.BranchSource'/ strategy(class: 'jenkins.branch.DefaultBranchPropertyStrategy') {
+      properties(class: 'java.util.Arrays$ArrayList') {
+        a(class: 'jenkins.branch.BranchProperty-array'){
+          'jenkins.branch.NoTriggerBranchProperty'()
```
s-urbaniak: this is extreme wizardry ;-)

on a serious note, let's put a reference here (as a comment) on the upstream disucssion with the maintainers so we can backtrack any potential fixes.

We use hardcoded class names and use reflection to resemble deprecated behavior here, so I am afraid this would break again, once we do another jenkins update round.

cpanato: 👍

PR #1861, opened 2017-09-07
tests: Enfore consistent name for local clusters

mxinden:
When someone runs the smoke tests locally this PR enforces that a cluster name is set. This ensures, that we can identify a left over cluster with a person instead of seeing random cluster names.


cpanato:
Fail to delete the folder, not related. we can merge

PR #1864, opened 2017-09-07
RSpec: recognize AWS profile and session token types of credentials

alexsomesan:


cpanato:
@alexsomesan rubocop complains :)

PR #1877, opened 2017-09-08
Fix v1.7.3

ggreer:
Regression introduced in #1660, specifically https://github.com/coreos/tectonic-installer/pull/1660/files#diff-28c6dee8dd102cf55d7b010e938c9d6aR65

### DIFF
--- a/installer/pkg/terraform/executor.go
+++ b/installer/pkg/terraform/executor.go
@@ -237,6 +237,28 @@ func (ex *Executor) Execute(args ...string) (int, chan struct{}, error) {
 	return cmd.Process.Pid, done, nil
 }
 
+// ExecuteSync Like Execute, but synchronous. TODO: clean up duplicate cmd setup code
+func (ex *Executor) ExecuteSync(args ...string) ([]byte, error) {
+	// Prepare TerraForm command by setting up the command, configuration,
+	// working directory (so the files such as terraform.tfstate are stored at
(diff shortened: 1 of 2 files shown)


ggreer:
After merging into the 1.7.3 branch, this needs to be cherry-picked into master. There are merge conflicts with master though, as some of the terraform executor code has changed.

installer/pkg/terraform/executor.go
```diff
@@ -237,6 +237,28 @@ func (ex *Executor) Execute(args ...string) (int, chan struct{}, error) {
...
+	for _, v := range os.Environ() {
+		if strings.HasPrefix(v, "SSH_") {
+			cmd.Env = append(cmd.Env, v)
+		}
+	}
+	for k, v := range ex.envVariables {
+		cmd.Env = append(cmd.Env, fmt.Sprintf("%s=%s", strings.ToUpper(k), v))
+	}
+	cmd.Dir = ex.executionPath
+	return cmd.Output()
```
Quentin-M: nit: CombinedOutput should be used. But as stated in the header comment, the function needs some rework (i.e. the header comment itself should state that the main difference of that command is that the output is returned rather than written to the log file.

sudhaponnaganti:
Thanks for the fix. Should we wait till this is merged in to Master to release patch for 1.7.3. Not sure who is going to resolve the conflicts with master branch

PR #1885, opened 2017-09-12
Builder image - bump Go to 1.9

alexsomesan:
Keep up with times.


alexsomesan:
smoke env doesn't install Go.

PR #1890, opened 2017-09-12
ci/maintenance: Add more date formats to filter when using grafiti to clean aws resources

cpanato:


cpanato:
my understanding of grafiti is it takes the tag you defined to use when deleting resources and match if the tag in the resource.
I added this others formats because I saw some resources in our AWS account with that format in the tag, and then when grafiti run will not delete those because does not match with the filter.

I ran one grafiti delete using this config and at least it no break anything :)

@squat

cpanato:
@squat grafiti is consistent in tagging the resources. What I saw in AWS was other users tagging the `expirationDate` with other formats like the one I added. eg. `M-D-Y`

squat:
Sure, the questions is: can Grafiti interpret all of these expiration date formats? My suspicion is "no".

cpanato:
@squat I ran this in one region which I I had free tier machines and each of one had this specific date format in the `expirationDate` tag and grafiti delete all machines.
I will send thru slack the log from the execution.

squat:
Ok, sorry I misunderstood how we were leveraging this date in Grafiti. This tag has no special meaning, e.g. time, in Grafiti other than being a string. We are just using this as the set of strings to select resources to delete.

cpanato:
thanks! also was good because I did a real testing :)

PR #1897, opened 2017-09-12
Add kvo config configmap

yifan-gu:
Create kvo config configmap when the cluster starts.
The kvo config contains install time options, it will be consumed by kvo during upgrade.

### DIFF
--- a/modules/bootkube/assets.tf
+++ b/modules/bootkube/assets.tf
@@ -108,6 +108,9 @@ resource "template_dir" "bootkube" {
     master_count              = "${var.master_count}"
     node_monitor_grace_period = "${var.node_monitor_grace_period}"
     pod_eviction_timeout      = "${var.pod_eviction_timeout}"
+
+    cloud_provider_profile = "${var.cloud_provider != "" ? "${var.cloud_provider}" : "metal"}"
+    cloud_config_path      = "${var.cloud_config_path}"
   }
 }
 
--- a/modules/bootkube/resources/manifests/kvo-config.yaml
+++ b/modules/bootkube/resources/manifests/kvo-config.yaml
@@ -20,3 +20,5 @@ data:
       cluster_cidr: ${cluster_cidr}
       etcd_servers: ${etcd_servers}
       service_cidr: ${service_cidr}
+    inititalConfig: |
+      initial_master_count: ${master_count}
--- a/modules/bootkube/resources/manifests/kvo-config.yaml
+++ b/modules/bootkube/resources/manifests/kvo-config.yaml
@@ -1,0 +1,22 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kvo-config
+  namespace: tectonic-system
+data:
+  kvo-config: |
+    apiVersion: v1
+    kind: KubeVersionOperatorConfig
+    authConfig: |
+      oidc_client_id: ${oidc_client_id}
+      oidc_issuer_url: ${oidc_issuer_url}
+      oidc_groups_claim: ${oidc_groups_claim}
+      oidc_username_claim: ${oidc_username_claim}
+    cloudProviderConfig: |
+      cloud_config_path: ${cloud_config_path}
+      cloud_provider_profile: ${cloud_provider_profile}
+    networkConfig: |
+      advertise_address: ${advertise_address}
+      cluster_cidr: ${cluster_cidr}
+      etcd_servers: ${etcd_servers}
+      service_cidr: ${service_cidr}
--- a/modules/bootkube/variables.tf
+++ b/modules/bootkube/variables.tf
@@ -129,3 +129,8 @@ variable "pod_eviction_timeout" {
   type        = "string"
   default     = "5m"
 }
+
+variable "cloud_config_path" {
+  description = "The path to the secret file that contains the cloud config contents. Either be empty ('') or ('/etc/kubernetes/cloud/config')."
+  type        = "string"
+}
--- a/platforms/aws/tectonic.tf
+++ b/platforms/aws/tectonic.tf
@@ -66,6 +66,8 @@ module "bootkube" {
   node_monitor_grace_period = "2m"
 
   pod_eviction_timeout = "220s"
+
+  cloud_config_path = ""
 }
 
 module "tectonic" {
--- a/modules/bootkube/resources/manifests/01-tectonic-namespace.yaml
+++ b/modules/bootkube/resources/manifests/01-tectonic-namespace.yaml
@@ -1,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: tectonic-system # Create the namespace first.
--- a/modules/tectonic/resources/manifests/namespace.yaml
+++ b/modules/tectonic/resources/manifests/namespace.yaml
@@ -1,4 +1,0 @@
-apiVersion: v1
-kind: Namespace
-metadata:
-  name: tectonic-system
--- a/modules/tectonic/resources/tectonic.sh
+++ b/modules/tectonic/resources/tectonic.sh
@@ -119,10 +119,6 @@ set -e
 # wait for Kubernetes pods
 wait_for_pods kube-system
 
-# Creating resources
-echo "Creating Tectonic Namespace"
-kubectl create -f namespace.yaml
-
 echo "Creating Initial Roles"
 kubectl delete -f rbac/role-admin.yaml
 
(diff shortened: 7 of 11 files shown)


yifan-gu:
The next plan will be, remove all the kubernetes manifests and just move in one kvo manifest, let the kvo to create all the objects with the given kvo-config.

diegs:
Re: the description, it will be consumed by KVO during install time and upgrade time, right?

modules/bootkube/resources/manifests/kvo-config.yaml
```diff
@@ -0,0 +1,29 @@
...
+metadata:
+  name: tectonic-system
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kvo-config
+  namespace: tectonic-system
+data:
+  kvo-config: |
```
diegs: nit: I would alphabetize all the keys to make it easier to find stuff as this grows.

modules/bootkube/resources/manifests/kvo-config.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: v1
```
diegs: I see why we need to create the namespace sooner now, but I would keep it in its own file rather than combining it with the kvo config.

yifan-gu:
@diegs Yes.
So the steps are:
1. bootkube creates bootstrap control plane.
2. bootkube creates the kvo-config and kvo
3. kvo creates the selfhosted control plane based on kvo-config.
4. selfhosted control plane replaces the bootstrap control plane.
5. tectonic continues to create other manifests.

platforms/aws/tectonic.tf
```diff
@@ -66,6 +66,10 @@ module "bootkube" {
   node_monitor_grace_period = "2m"
 
   pod_eviction_timeout = "220s"
+
+  cni_profile       = "${var.tectonic_calico_network_policy ? "canel" : "flannel"}"
```
diegs: `s/canel/canal/g` (in all the files?)

modules/bootkube/resources/manifests/kvo-config.yaml
yifan-gu: Yeah, but @aaronlevy said bootkube can't handle order requirement.

config.tf
```diff
@@ -408,3 +408,12 @@ WARNING: Enabling an alpha feature means that future updates may become unsuppor
 This should only be enabled on clusters that are meant to be short-lived to begin validating the alpha feature.
 EOF
 }
+
+variable "tectonic_ingress_profile" {
```
aaronlevy: Lets drop this for now, it won't actually end up under KVO management long-term so it's not immediately needed in this config

modules/bootkube/assets.tf
```diff
@@ -108,6 +108,11 @@ resource "template_dir" "bootkube" {
     master_count              = "${var.master_count}"
     node_monitor_grace_period = "${var.node_monitor_grace_period}"
     pod_eviction_timeout      = "${var.pod_eviction_timeout}"
+
+    cloud_provider_profile = "${var.cloud_provider != "" ? "${var.cloud_provider}" : "metal"}"
+    cni_profile            = "${var.cni_profile}"
```
aaronlevy: drop - not immediately needed for core k8s components

modules/bootkube/assets.tf
```diff
@@ -108,6 +108,11 @@ resource "template_dir" "bootkube" {
...
+    ingress_profile        = "${var.ingress_profile}"
```
aaronlevy: also drop for now

modules/bootkube/assets.tf
```diff
@@ -108,6 +108,11 @@ resource "template_dir" "bootkube" {
...
+    cloud_config_path      = "${var.cloud_config_path}"
```
aaronlevy: We should discuss if this is the best option -- but fine to leave it in for now.

modules/bootkube/resources/manifests/kvo-config.yaml
```diff
@@ -0,0 +1,29 @@
...
+data:
+  kvo-config: |
+    apiVersion: v1
+    kind: KubeVersionOperatorConfig
+    networkConfig: |
+      service_cidr: ${service_cidr}
+      cluster_cidr: ${cluster_cidr}
+      advertise_address: ${advertise_address}
+      etcd_servers: ${etcd_servers}
+      cni_profile: ${cni_profile}
```
aaronlevy: drop

modules/bootkube/resources/manifests/kvo-config.yaml
```diff
@@ -0,0 +1,29 @@
...
+    authConfig: |
+      oidc_issuer_url: ${oidc_issuer_url}
+      oidc_client_id: ${oidc_client_id}
+      oidc_username_claim: ${oidc_username_claim}
+      oidc_groups_claim: ${oidc_groups_claim}
+    cloudProviderConfig: |
+      cloud_provider_profile: ${cloud_provider_profile}
+      ingress_profile: ${ingress_profile}
```
aaronlevy: drop

modules/bootkube/variables.tf
```diff
@@ -129,3 +129,18 @@ variable "pod_eviction_timeout" {
   type        = "string"
   default     = "5m"
 }
+
+variable "cni_profile" {
+  description = "The installation time choice of CNI pluggins. Valid options are 'flannel', 'canal', etc."
+  type        = "string"
+}
```
aaronlevy: drop

modules/bootkube/variables.tf
```diff
@@ -129,3 +129,18 @@ variable "pod_eviction_timeout" {
...
+
+variable "ingress_profile" {
+  description = "The installation time choice of ingress setup. Value options are 'daemonset-hostPort', 'deployment-nodeport', etc."
+  type        = "string"
+}
```
aaronlevy: drop

modules/bootkube/resources/manifests/kvo-config.yaml
yifan-gu: done

platforms/aws/tectonic.tf
yifan-gu: done

modules/bootkube/variables.tf
yifan-gu: done

modules/bootkube/variables.tf
yifan-gu: done

modules/bootkube/assets.tf
yifan-gu: done

yifan-gu:
@alexsomesan @s-urbaniak PTAL.

modules/bootkube/resources/manifests/kvo-config.yaml
```diff
@@ -0,0 +1,22 @@
...
+      oidc_groups_claim: ${oidc_groups_claim}
+      oidc_username_claim: ${oidc_username_claim}
+    cloudProviderConfig: |
+      cloud_config_path: ${cloud_config_path}
+      cloud_provider_profile: ${cloud_provider_profile}
+    networkConfig: |
+      advertise_address: ${advertise_address}
+      cluster_cidr: ${cluster_cidr}
+      etcd_servers: ${etcd_servers}
+      service_cidr: ${service_cidr}
```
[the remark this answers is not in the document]
s-urbaniak: another suggestion is to name it `initial_master_count` to make the semantics more explicit.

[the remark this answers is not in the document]
diegs: There are a couple of other alternatives we've thought about:

- Use the cluster-proportional autoscaler to set the replicas: https://github.com/kubernetes-incubator/cluster-proportional-autoscaler
- Migrate the scheduler and controller-manager to daemonsets once this feature lands in 1.9: https://github.com/kubernetes/community/pull/977
- In the near term, just set the replicas to something fixed (like 2).

For this iteration I think passing the "initial_master_count" through is probably fine.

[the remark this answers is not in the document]
yifan-gu: I just stored the `initial_master_count` under the `networkConfig` key since I can't think of a better name and it will be removed later anyway.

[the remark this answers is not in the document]
diegs: I suggest perhaps adding `InitialConfig` in case we have other parameters like this.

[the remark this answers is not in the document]
yifan-gu: Done.

PR #1913, opened 2017-09-14
tests/frontend: Add more Nightwatch tests for bare metal

kyoto:
Bare metal tests now run at least some basic tests for every screen. (Edit: Except for the Starting Installation and Installation Complete screens, which currently have no Nightwatch tests.)


kyoto:
ok to test

cpanato:
UI tests passed, we have some flaky in other

PR #1917, opened 2017-09-15
frontend: Fix AWS Terraform apply progress bar heuristic

kyoto:
Reflects changes introduced in #1910.

Switch from a list of modules to ignore, to a list of modules to include.

Now also ignores the 2 `aws_s3_bucket_object` resources.

### DIFF
--- a/installer/frontend/components/tf-poweron.jsx
+++ b/installer/frontend/components/tf-poweron.jsx
@@ -27,11 +27,11 @@ const estimateTerraformProgress = terraform => {
 
   let done = output.match(/.*: Creation complete/g) || [];
 
-  // Ignore resources from the bootkube, flannel-vxlan and tectonic modules because they all complete very quickly
-  done = done.filter(c => !/module\.(bootkube|flannel-vxlan|tectonic)/.test(c));
+  // Ignore resources that complete very quickly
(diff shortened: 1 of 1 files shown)


installer/frontend/components/tf-poweron.jsx
```diff
@@ -27,11 +27,11 @@ const estimateTerraformProgress = terraform => {
...
   let done = output.match(/.*: Creation complete/g) || [];
 
-  // Ignore resources from the bootkube, flannel-vxlan and tectonic modules because they all complete very quickly
-  done = done.filter(c => !/module\.(bootkube|flannel-vxlan|tectonic)/.test(c));
+  // Ignore resources that complete very quickly
+  done = done.filter(c => /(aws_route53_zone|aws_s3_bucket|module\.masters|module\.vpc|module\.workers)\./.test(c));
 
   // Approximate number of AWS resources we expect Terraform to create
-  const total = 82;
+  const total = 80;
```
ggreer: I'm guessing this could be made more accurate by looking at the number of nodes in the cluster. eg: `total = 40 + (nodes * 20);` Some experimenting would have to be done to figure out the right constants.

kyoto: Actually, the number of nodes doesn't change the number of created resources reported in the Terraform logs. Provisioning a cluster for etcd does though, so I will create a separate PR for that. Thanks.

Issue #1754, opened 2017-08-23
Azure - NSG for etcd 


mwieczorek:
## What keywords did you search in tectonic-installer issues before filing this one?
```azure subnet```, ```azure etcd```

## Is this a BUG REPORT or FEATURE REQUEST?
BUG REPORT (?)

## Versions

* Tectonic version ([release](https://github.com/coreos/tectonic-installer/releases) or commit hash): 1.7.1 v2
* Terraform version (`terraform version`): v0.9.11
* Platform (aws|azure|openstack|metal): azure

## What happened?
It looks like network security group for etcd has 'wide' rules. It's possible to connect to etcd cluster even outside of vnet (in case we have some peerings between vnets). It's because of rule:
```
resource "azurerm_network_security_rule" "etcd_ingress_client_master" {
(...)
  access                 = "Allow"
  protocol               = "tcp"
  source_port_range      = "*"
  destination_port_range = "2379"
  source_address_prefix       = "*"
  destination_address_prefix  = "*"
(...)
}
```
https://github.com/coreos/tectonic-installer/blob/master/modules/azure/vnet/nsg-etcd.tf#L105

## What you expected to happen?
Access to etcd is restricted only to master and other etcd nodes (like for AWS)
https://github.com/coreos/tectonic-installer/blob/master/modules/aws/vpc/sg-etcd.tf#L35

## Anything else we need to know?

Maybe it should be restricted only to master subnet?

PR #1929, opened 2017-09-18
azure/vnet: NSG improvements

justaugustus:
* Update rule names
* Fix rule priorities
* Remove extraneous rules
* Tighten access
* Move API / Console `AzureLoadBalancer` rules to master subnet
* Add rule descriptions
* Remove references to `external_nsg_api_id`

Fixes: #1754 

/cc @alexsomesan @colemickens @metral


alexsomesan:
retest this please

modules/azure/vnet/nsg-etcd.tf
```diff
@@ -6,149 +6,127 @@ resource "azurerm_network_security_group" "etcd" {
...
+  direction              = "Inbound"
+  access                 = "Allow"
+  protocol               = "TCP"
+  source_port_range      = "*"
+  destination_port_range = "22"
+
+  # TODO: Reference subnet
   source_address_prefix       = "${var.ssh_network_internal}"
-  destination_address_prefix  = "*"
+  destination_address_prefix  = "${var.vnet_cidr_block}"
```
alexsomesan: I think this can be tightened even more to just the etcd subnet CIDR, right?

modules/azure/vnet/nsg-etcd.tf
```diff
@@ -6,149 +6,127 @@ resource "azurerm_network_security_group" "etcd" {
...
+  protocol               = "TCP"
   source_port_range      = "*"
   destination_port_range = "22"
 
-  # TODO: Need to allow traffic from self
-  source_address_prefix       = "${var.vnet_cidr_block}"
-  destination_address_prefix  = "*"
+  # TODO: Reference subnet
+  source_address_prefix       = "${var.ssh_network_external}"
+  destination_address_prefix  = "${var.vnet_cidr_block}"
```
alexsomesan: same thing, can just be the etcd CIDR block here.

modules/azure/vnet/nsg-etcd.tf
```diff
@@ -6,149 +6,127 @@ resource "azurerm_network_security_group" "etcd" {
...
   direction              = "Inbound"
   access                 = "Allow"
-  protocol               = "tcp"
+  protocol               = "TCP"
   source_port_range      = "*"
   destination_port_range = "22"
 
-  # TODO: Need to allow traffic from master
+  # TODO: Reference subnet
   source_address_prefix       = "${var.vnet_cidr_block}"
```
alexsomesan: master subnet CIDR block

modules/azure/vnet/nsg-etcd.tf
```diff
@@ -6,149 +6,127 @@ resource "azurerm_network_security_group" "etcd" {
...
-  destination_address_prefix  = "*"
+  destination_address_prefix  = "${var.vnet_cidr_block}"
```
alexsomesan: etcd CIDR block.

modules/azure/vnet/nsg-etcd.tf
```diff
@@ -6,149 +6,127 @@ resource "azurerm_network_security_group" "etcd" {
...
   direction              = "Inbound"
   access                 = "Allow"
-  protocol               = "tcp"
+  protocol               = "TCP"
   source_port_range      = "*"
   destination_port_range = "2379"
 
-  # TODO: Need to allow traffic from self
+  # TODO: Reference subnet
   source_address_prefix       = "${var.vnet_cidr_block}"
```
alexsomesan: master subnet CIDR

modules/azure/vnet/nsg-etcd.tf
```diff
@@ -6,149 +6,127 @@ resource "azurerm_network_security_group" "etcd" {
...
-  destination_address_prefix  = "*"
+  destination_address_prefix  = "${var.vnet_cidr_block}"
```
alexsomesan: etcd CIDR

modules/azure/vnet/nsg-etcd.tf
```diff
@@ -6,149 +6,127 @@ resource "azurerm_network_security_group" "etcd" {
...
   direction              = "Inbound"
   access                 = "Allow"
-  protocol               = "tcp"
+  protocol               = "TCP"
   source_port_range      = "*"
   destination_port_range = "2379"
 
-  # TODO: Need to allow traffic from workers
+  # TODO: Reference subnet
   source_address_prefix       = "${var.vnet_cidr_block}"
```
alexsomesan: same as above

modules/azure/vnet/nsg-etcd.tf
```diff
@@ -6,149 +6,127 @@ resource "azurerm_network_security_group" "etcd" {
...
   direction              = "Inbound"
   access                 = "Allow"
-  protocol               = "tcp"
+  protocol               = "TCP"
   source_port_range      = "*"
   destination_port_range = "2380"
 
-  # TODO: Need to allow traffic from self
+  # TODO: Reference subnet
   source_address_prefix       = "${var.vnet_cidr_block}"
```
alexsomesan: etcd CIDR

modules/azure/vnet/nsg-etcd.tf
```diff
@@ -6,149 +6,127 @@ resource "azurerm_network_security_group" "etcd" {
...
-  destination_address_prefix  = "*"
+  destination_address_prefix  = "${var.vnet_cidr_block}"
```
alexsomesan: etcd CIDR

modules/azure/vnet/nsg-master.tf
```diff
@@ -1,342 +1,260 @@
...
+  name                        = "${var.cluster_name}-alb-probe"
+  description                 = "${var.cluster_name} master - Azure Load Balancer probe"
+  priority                    = 295
   direction                   = "Inbound"
   access                      = "Allow"
-  protocol                    = "tcp"
+  protocol                    = "TCP"
   source_port_range           = "*"
-  destination_port_range      = "22"
-  source_address_prefix       = "${var.ssh_network_internal}"
```
alexsomesan: vnet CIDR

modules/azure/vnet/nsg-master.tf
```diff
@@ -1,342 +1,260 @@
...
-  direction                   = "Inbound"
-  access                      = "Allow"
-  protocol                    = "tcp"
-  source_port_range           = "*"
-  destination_port_range      = "22"
-  source_address_prefix       = "${var.ssh_network_external}"
-  destination_address_prefix  = "*"
+  destination_port_range      = "*"
+  source_address_prefix       = "AzureLoadBalancer"
+  destination_address_prefix  = "${var.vnet_cidr_block}"
```
alexsomesan: master subnet CIDR

modules/azure/vnet/nsg-master.tf
```diff
@@ -1,342 +1,260 @@
...
-  destination_address_prefix  = "*"
+  # TODO: Reference subnet
+  source_address_prefix       = "${var.ssh_network_external}"
+  destination_address_prefix  = "${var.vnet_cidr_block}"
   resource_group_name         = "${var.resource_group_name}"
   network_security_group_name = "${azurerm_network_security_group.master.name}"
 }
 
-resource "azurerm_network_security_rule" "master_ingress_etcd_self" {
+resource "azurerm_network_security_rule" "master_ingress_flannel_from_master" {
```
alexsomesan: I think this can just be one rule shared by all the nodes in the cluster, no?

modules/azure/vnet/nsg-master.tf
```diff
@@ -1,342 +1,260 @@
...
+  # TODO: Reference subnet
   source_address_prefix       = "${var.vnet_cidr_block}"
-  destination_address_prefix  = "*"
+  destination_address_prefix  = "${var.vnet_cidr_block}"
   resource_group_name         = "${var.resource_group_name}"
   network_security_group_name = "${azurerm_network_security_group.master.name}"
 }
 
-resource "azurerm_network_security_rule" "master_ingress_bootstrap_etcd" {
+resource "azurerm_network_security_rule" "master_ingress_flannel_from_worker" {
```
alexsomesan: this seems to have identical rules to the one above.
see my comment about sharing it.

justaugustus:
@alexsomesan Thanks for the review!

Each of the `# TODO: Reference subnet` basically meant "this should be swapped to reference the relevant subnet", so we had the same idea :)

In the instance of installer-created subnets, it's easy to grab the output of the resource and use that.

With external subnets, it would be necessary to introduce another set of variables, `tectonic_azure_external_*_cidr`, since we normally reference the subnet by resource ID.

I wanted to get your thoughts on that, before adding more variables into the mix.

alexsomesan:
You're right about the need to determine CIDRs for the external subnets. I thought we had a data source in Terraform we could use to query the subnets' attributes. Turns out it's not available. We should wait for that to happen and keep the current vnet CIDR and the TODOs.
I'd rather not introduce another set of inputs for those CIDRs since it's fairly easy to make that datasource.

justaugustus:
@alexsomesan Additional notes on your review:
* etcd CIDR block would be master subnet
* The idea behind `ssh_network_internal` and `ssh_network_external` was to provide a means to allow an admin to add SSH access for different Ops networks e.g., `internal --> Ops subnet`, `external --> Office IP`
* for flannel, there are a few options:

1. `VirtualNetwork`
2. `vnet_cidr_block`
3. create two rules for the master and the worker subnets (I went with this one as it's the tightest, although it's dependent on being able to query for the master / worker CIDR ranges)

PR #1921, opened 2017-09-15
Azure rspec smoke tests

alexsomesan:
Just porting verbatim from previous smoke.sh
Was a low hanging fruit after implementing most of the logic for Azure private.

PR #1930, opened 2017-09-18
Disables the Azure private test while I fix it

alexsomesan:


alexsomesan:
@s-urbaniak Definitely! I'm fixing it as part of https://github.com/coreos/tectonic-installer/pull/1921

PR #1933, opened 2017-09-18
modules/tectonic: Add cluster-config.yaml

yifan-gu:


yifan-gu:
These are the flags we are using today in ingress, identity, stats-emitter, console. But I am not sure everything here is allowed to be changed by the users, so maybe we will remove some of those here.
cc @aaronlevy @diegs @derekparker @rithujohn191 @amrutac @squat

modules/tectonic/resources/manifests/utility-config.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: utility-config
```
diegs: I thought we were putting all of these in one `ConfigMap`, which is why you renamed it in #1928?

yifan-gu: @diegs That's good question. The reason is that then I realized that the bootkube module doesn't have all these required terraform variables defined.
So I couldn't decide if we want to put all those tectonic related terraform variables into bookube module.

diegs: Hm, maybe we should make this a `cluster-config` in the `tectonic-system` namespace and make the bootkube module `cluster-config` in the `kube-system` namespace then?

That way we can try and limit our number of config objects to one per module.

yifan-gu: @diegs I am ok with that, wdty @aaronlevy ?

modules/tectonic/resources/manifests/utility-config.yaml
```diff
@@ -0,0 +1,35 @@
...
+      grafana_callback: ${grafana_callback}
+      kubectl_client_id: ${kubectl_client_id}
+      kubectl_secret: ${kubectl_secret}
+      oidc_issuer_url: ${oidc_issuer_url}
+      prometheus_callback: ${prometheus_callback}
+    ingressConfig: |
+      console_base_host: ${console_base_host}
+    statsEmitterConfig: |
+      stats_url: ${stats_url}
+    tectonicConfigMap: |
```
yifan-gu: I feel we can actually dissolve key-values under this key into the above sub-keys.
I can do an investigation on who is using these values. But if someone knows this already, that will be helpful.
/cc @amrutac @squat @rithujohn191

modules/tectonic/resources/manifests/utility-config.yaml
aaronlevy: Hmmm interesting thought. I agree that components should not need to cross namespace boundaries for their config (ideally). I'd be fine starting with this - and during the next few weeks of burn-in we can decide if we like that pattern.

modules/tectonic/resources/manifests/utility-config.yaml
```diff
@@ -0,0 +1,35 @@
...
+metadata:
+  name: utility-config
+  namespace: tectonic-system
+data:
+  tuo-config: |
+    apiVersion: v1
+    kind: UtilityOperatorConfig
+    identityConfig: |
+      admin_email: ${admin_email}
+      admin_password_hash: ${admin_password_hash}
```
aaronlevy: Do we need this? I'd prefer that we don't keep anything sensitive in this config (also why do we need this - what component consumes?)

yifan-gu: The identity configmap already contains it: https://github.com/coreos/tectonic-installer/blob/master/modules/tectonic/resources/manifests/identity/configmap.yaml

modules/tectonic/resources/manifests/cluster-config.yaml
```diff
@@ -0,0 +1,35 @@
...
+      alertmanager_callback: ${alertmanager_callback}
+      console_callback: ${console_callback}
+      console_client_id: ${console_client_id}
+      console_secret: ${console_secret}
+      grafana_callback: ${grafana_callback}
+      kubectl_client_id: ${kubectl_client_id}
+      kubectl_secret: ${kubectl_secret}
+      oidc_issuer_url: ${oidc_issuer_url}
+      prometheus_callback: ${prometheus_callback}
+    ingressConfig: |
```
aaronlevy: I'm on fence if this should be kube-addon or tectonic-utility. It's in the tectonic-system namespace - so I guess it makes sense here.

modules/bootkube/resources/manifests/cluster-config.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: cluster-config
-  namespace: tectonic-system
+  namespace: kube-system
```
aaronlevy: wouldn't this be tectonic-system if it's managing components in that namespace?

diegs: this one is in the `bootkube` module

aaronlevy: nm. mixing up files

diegs:
Can you remove `modules/bootkube/resources/manifests/01-tectonic-namespace.yaml` ?

modules/tectonic/resources/manifests/utility-config.yaml
```diff
@@ -0,0 +1,36 @@
...
+      admin_user_id: ${admin_user_id}
+      alertmanager_callback: ${alertmanager_callback}
+      console_callback: ${console_callback}
+      console_client_id: ${console_client_id}
+      console_secret: ${console_secret}
+      grafana_callback: ${grafana_callback}
+      kubectl_client_id: ${kubectl_client_id}
+      kubectl_secret: ${kubectl_secret}
+      oidc_issuer_url: ${oidc_issuer_url}
+      prometheus_callback: ${prometheus_callback}
```
rithujohn191: If we plan to remove all the `callbacks` from the [IdentityConfig](https://github.com/coreos-inc/tectonic-utility-operator/pull/8/files) we should remove it from here as well

yifan-gu: will do.

modules/bootkube/resources/manifests/cluster-config.yaml
rithujohn191: So if we plan to make this ConfigMap general to all, how can I include my tectonic network config that's meant to be in the tectonic-system namespace?

modules/tectonic/resources/tectonic.sh
```diff
@@ -127,6 +127,9 @@ kubectl create -f rbac/role-user.yaml
 kubectl create -f rbac/binding-admin.yaml
 kubectl create -f rbac/binding-discovery.yaml
 
+echo "Creating Tectonic Utility Operator Config"
+kubectl create -f utility-config.yaml
```
diegs: `cluster-config.yaml` and update `echo` text?

yifan-gu:
@rithujohn191 @diegs Addressed comments.

modules/tectonic/assets.tf
```diff
@@ -58,6 +58,7 @@ resource "template_dir" "tectonic" {
     admin_email         = "${lower(var.admin_email)}"
     admin_password_hash = "${var.admin_password_hash}"
 
+    base_address         = "${var.base_address}"
     console_base_address = "${var.base_address}"
     console_base_host    = "${element(split(":", var.base_address), 0)}"
```
yifan-gu: @alexsomesan @s-urbaniak Does anyone know why this is not just `"${var.base_address}"`?

yifan-gu:
Let me close and reopen the PR again to see if the smoke tests get running.

yifan-gu:
Merging for now to unblock other works, we can address https://github.com/coreos/tectonic-installer/pull/1933#discussion_r141467260 later.

PR #1934, opened 2017-09-18
modules/bootkube: add managed-by annotations.

diegs:
This marks the objects that are (and will continue to be) managed by the
KVO as such. This uses the standard `managed-by` annotation that is
declared in the x-operator.


yifan-gu:
@diegs How about the kube-dns and pod checkpointer? Maybe we don't have to add labels to the later one.

diegs:
@yifan-gu those aren't part of the initial 4 that are managed by KVO. I think we'll manage those with the "addons" operator.

yifan-gu:
@diegs ACK.

PR #1943, opened 2017-09-19
Adds Wants=rpc-statd.service to kubelet

coresolve:
Picked up from https://github.com/coreos/bugs/issues/2074

Need to add a wants here so that using nfs persistent volumes work as
intended on tectonic.

Without this change you see an error like:

```
Mounting arguments: 192.168.1.2:/mnt/user/tectonic /var/lib/kubelet/pods/6961c181-9d79-11e7-abdc-005056a6ea29/volumes/kubernetes.io~nfs/nfs nfs []
Output: mount.nfs: rpc.statd is not running but is required for remote locking.
mount.nfs: Either use '-o nolock' to keep locks local, or start statd.
mount.nfs: an incorrect mount option was specified
```

### DIFF
--- a/modules/ignition/resources/services/kubelet.service
+++ b/modules/ignition/resources/services/kubelet.service
@@ -1,6 +1,6 @@
 [Unit]
 Description=Kubelet via Hyperkube ACI
-
+Wants=rpc-statd.service
 [Service]
 EnvironmentFile=/etc/kubernetes/kubelet.env
 Environment="RKT_RUN_ARGS=--uuid-file-save=/var/cache/kubelet-pod.uuid \


euank:
I suspect this only works for new installs, not existing. cc @aaronlevy for whether kubelet.service can be updated yet.

sym3tri:
How would this be updated?

coresolve:
@crawford

aaronlevy:
It can't be easily updated right now - so it would only be for new clusters. If we need it updated on existing clusters this is part of a larger discussion.

crawford:
This looks fine to me. We really need to figure out the story around updating `kubelet.service`.

coresolve:
ok to test

s-urbaniak:
ok to test

s-urbaniak:
@cpanato @mxinden builds 4 and 5 have failures, but combined they are green. Please PTAL and confirm, this LGTM and we should merge it as it is part of the release.

PR #1950, opened 2017-09-20
[WIP] frontend: Prevent Starting Tectonic substeps flipping on / off

kyoto:
This PR makes 2 changes that together prevent the install page's Start Tectonic substeps from flipping between done and pending state in the UI.

The backend initially reports `isEtcdSelfHosted = false`, then switches it to `true` after a while, which caused the Etcd substep to suddenly appear. This changes the UI to use `clusterConfig` to determine whether to display the Etcd substep instead, so we can correctly show or hide it from the beginning.

The backend API will sometimes report a substep as `success = true`, but in a later call report the same substep as `success = false`, which caused substeps to flip from "done" back to "pending". This changes the UI to keep substeps as "done" in this case.

### DIFF
--- a/installer/frontend/components/tf-poweron.jsx
+++ b/installer/frontend/components/tf-poweron.jsx
@@ -9,7 +9,7 @@ import { DropdownInline } from './ui';
 import { AWS_DomainValidation } from './aws-domain-validation';
 import { commitPhases } from '../actions';
 import { TFDestroy } from '../aws-actions';
-import { CLUSTER_NAME, PLATFORM_TYPE, getTectonicDomain } from '../cluster-config';
+import { CLUSTER_NAME, ETCD_OPTION, ETCD_OPTIONS, PLATFORM_TYPE, getTectonicDomain } from '../cluster-config';
 import { AWS_TF, BARE_METAL_TF } from '../platforms';
 import { commitToServer, observeClusterStatus } from '../server';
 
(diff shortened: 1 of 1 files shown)


installer/frontend/components/tf-poweron.jsx
```diff
@@ -121,22 +132,18 @@ export const TF_PowerOn = connect(stateToProps, dispatchToProps)(
...
+        _.each(services, ({key}, i) => {
+          const {failed, message, success} = tectonic[key];
 
-        // Don't let progress bars go backwards (e.g. if a service flips back to incomplete)
-        if (tectonicProgress > this.state.tectonicProgress) {
-          this.setState({tectonicProgress});
-        }
+          // The backend sometimes flips a service from success equals true back to success = false. This looks strange,
+          // so once a service has been reported as successful, keep it true.
+          services[i] = _.assign(services[i], {failed, message, success: services[i].success || success});
```
kans: My fuzzy recollection of this behavior is that we have 1/1 of some resource, then 1/2, then 2/2, etc.  Assuming this is accurate, I'm not sure locking the results to `true` is what we want.

kyoto:
We now plan to remove the self-hosted etcd option, so this PR will need changing.

kyoto:
Closing because the self-hosted etcd option has now been removed and because we should probably fix the backend API to report progress correctly rather than accommodating its faulty behavior.

PR #1951, opened 2017-09-20
frontend: Increase GUI test timeout to allow time for generating assets

kyoto:
Because we now include some Terraform plugins in the build dir / assets.zip, this step now takes longer.


sym3tri:
Is there a way to loop and wait for something to be available without requiring a timeout?

ggreer:
I think if it takes 30 seconds to generate assets, users will assume the page is broken. This seems like a reasonable timeout to me.

kyoto:
@sym3tri Like @ggreer mentioned, I'm thinking that if it takes longer than 30 seconds, that is a real problem. To be clear, as soon as the condition is met, the test will pass. It won't block for the whole 30 seconds.

sym3tri:
ok

PR #1953, opened 2017-09-20
Temporarily disable bare metal smokes

alexsomesan:
Temporarily disabled for consolidation
Fails very often due to Packet flakiness

/cc: @sym3tri

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -359,6 +359,9 @@ pipeline {
               }
             }
           },
+          /* Temporarily disabled for consolidation
+           * Fails very often due to Packet flakiness
+           *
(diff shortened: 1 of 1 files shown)


alexsomesan:
A note to add some more clarity here:

We have a lot of flaky tests, but we've recently made good progress to reduce those and we can now start trusting the tests more and upholding the rule for green CI on every PR.

On cloud provider tests when a flake shows up one can just type `retest this please` to have another go at curbing the flake. However, in the metal case, when the underlying machines go out of service there is no remediation with a mere `retests this` and manual intervention is required to re-align the infra.

This is why I opted to disable this until these tests are self contained enough to be recoverable with a `retest`.

PR #1958, opened 2017-09-20
platforms/aws: Use separate ELB attachment resources

bison:
This uses standalone resources to attach extra ELBs to workers rather
than doing it inline on the ASG resource.  Attaching the ELBs inline
means it isn't possible to configure other attachments outside the ASG
definition.

I haven't actually had time to test this yet. Will do that soon.


alexsomesan:
retest this please

PR #1962, opened 2017-09-21
frontend: Migrate CertificateAuthority from WithClusterConfig to Connect and use Radio component

kyoto:
This change causes the following to be added to the `.progress` file when the CA Certificate and CA Private Key fields are not set.

```
"clusterConfig": {
  ...
  "ignore": {
     ...
     "caCertificate": true,
     "caPrivateKey": true
```

### DIFF
--- a/installer/frontend/cluster-config.js
+++ b/installer/frontend/cluster-config.js
@@ -38,6 +38,7 @@ export const BM_WORKERS = 'workers';
 export const CA_CERTIFICATE = 'caCertificate';
 export const CA_PRIVATE_KEY = 'caPrivateKey';
 export const CA_TYPE = 'caType';
+export const CA_TYPES = {SELF_SIGNED: 'self-signed', OWNED: 'owned'};
 export const CLUSTER_NAME = 'clusterName';
 export const CLUSTER_SUBDOMAIN = 'clusterSubdomain';
 export const CONTROLLER_DOMAIN = 'controllerDomain';
@@ -170,7 +171,7 @@ export const DEFAULT_CLUSTER_CONFIG = {
   [BM_TECTONIC_DOMAIN]: '',
   [CA_CERTIFICATE]: '',
   [CA_PRIVATE_KEY]: '',
-  [CA_TYPE]: 'self-signed',
+  [CA_TYPE]: '',
   [CLUSTER_NAME]: '',
   [CONTROLLER_DOMAIN]: '',
   [DRY_RUN]: false,
@@ -186,7 +187,6 @@ export const DEFAULT_CLUSTER_CONFIG = {
   [POD_CIDR]: '10.2.0.0/16',
   [SERVICE_CIDR]: '10.3.0.0/16',
 };
-
 
 export const toAWS_TF = (cc, FORMS) => {
   const controllers = FORMS[AWS_CONTROLLERS].getData(cc);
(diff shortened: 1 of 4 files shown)


installer/frontend/cluster-config.js
```diff
@@ -176,7 +177,7 @@ export const DEFAULT_CLUSTER_CONFIG = {
   [BM_TECTONIC_DOMAIN]: '',
   [CA_CERTIFICATE]: '',
   [CA_PRIVATE_KEY]: '',
-  [CA_TYPE]: 'self-signed',
+  [CA_TYPE]: '',
```
kans: I don't think you need this at all anymore.

installer/frontend/components/certificate-authority.jsx
```diff
@@ -1,102 +1,70 @@
...
+import { CA_CERTIFICATE, CA_PRIVATE_KEY, CA_TYPE, CA_TYPES } from '../cluster-config';
+import { Field, Form } from '../form';
 import { validate } from '../validate';
-import { CA_TYPE, CA_CERTIFICATE, CA_PRIVATE_KEY } from '../cluster-config';
-import { WithClusterConfig, CertArea, PrivateKeyArea } from './ui';
+import { CertArea, Connect, PrivateKeyArea, Radio } from './ui';
+
+const form = new Form('CERTIFICATE_AUTHORITY', [
+  new Field(CA_TYPE, {default: CA_TYPES.SELF_SIGNED}),
+  new Field(CA_CERTIFICATE, {default: '', validator: validate.certificate}),
```
kans: There is a way to tell `CA_CERTIFICATE` to look at `CA_TYPE` and optionally enable it.

installer/frontend/cluster-config.js
kyoto: Apparently it's still needed for the unit tests.

installer/frontend/components/certificate-authority.jsx
kyoto: Thanks, I added `ignoreWhen` callbacks for the cert and key fields and updated the PR description.

installer/frontend/cluster-config.js
kans: Oh, I think you just need to import the file in the test suite.

installer/frontend/components/certificate-authority.jsx
```diff
@@ -1,102 +1,78 @@
...
-  if (clusterConfig[CA_TYPE] === 'self-signed') {
-    return true;
-  }
+  </div>
+</div>);
 
-  return (!validate.certificate(clusterConfig[CA_CERTIFICATE]) &&
-          !validate.privateKey(clusterConfig[CA_PRIVATE_KEY]));
-};
+CertificateAuthority.canNavigateForward = state => state.clusterConfig[CA_TYPE] === CA_TYPES.SELF_SIGNED || form.canNavigateForward(state);
```
kans: You should be able to remove this logic now.

installer/frontend/cluster-config.js
kyoto: Actually the tests do pass, but they start to output "Removed clusterConfig.caType because it's not in defaults." with `console.error`. Maybe the tests should be changed? I'm thinking that shouldn't be part of this PR though.

installer/frontend/components/certificate-authority.jsx
kyoto: OK, I changed it so that this is now just `CertificateAuthority.canNavigateForward = form.canNavigateForward` and all the validation is done in the form level `validator`.

kyoto:
@kans Addressed your review comments.

kyoto:
ok to test

installer/frontend/components/etcd.jsx
```diff
@@ -76,7 +76,7 @@ export const Etcd = connect(({clusterConfig}) => ({
               <Connect field={ETCD_OPTION}>
                 <Radio name={ETCD_OPTION} value={ETCD_OPTIONS.EXTERNAL} id={ETCD_OPTIONS.EXTERNAL} />
               </Connect>
-            I have my own v3 etcd cluster
+              I have my own v3 etcd cluster
```
squat: Not part of this pr but I am just noticing that this should really be "etcd v3" rather than the other way around.

kyoto: Fixed: #2019

PR #1971, opened 2017-09-22
frontend: Remove self-hosted etcd option from GUI

kyoto:
cc @robszumski


kyoto:
@ggreer @kans @squat Could one of you review? We want to include this in the 1.7.5 release.

PR #1973, opened 2017-09-22
monitoring: bump PO and TPO

brancz:
In order to ship the PO memory leak to clusters that are on 1.6.x we released a new version of the TPO, and these are the respective version bumps for the installer.

I just tested the upgrade, and will shortly also launch a new cluster off of head of this PR.

@Quentin-M 

/cc @s-urbaniak @alexsomesan


s-urbaniak:
@brancz is this supposed to get into the 1.7.5 release? If so I suggest to add a redundant cherry-pick PR against https://github.com/coreos/tectonic-installer/tree/1.7.5.

brancz:
@s-urbaniak It's supposed to go into the 1.6.10 release.

alexsomesan:
smokes can't run here, 1.6.x is not properly setup for the current CI

brancz:
I tested the upgrade from a 1.6.7 cluster as well as a fresh cluster from the HEAD of this PR, both resulted in a functioning cluster. Should we leave the final call up to @Quentin-M ?

Quentin-M:
Thank you guys~

PR #1977, opened 2017-09-23
Increase Memory Limits

dmyerscough:
Today we ran  into an issue where the `tectonic-cluo-operator` was requesting `80Mi` and kept OOMing due to memory limits. Increasing the limit to `100Mi`allowed the `tectonic-cluo-operator` pod to start and after a few moments the memory limit dropped down to `40Mi`.

It might be worth increasing the hard limits just to account for any overhead.

### DIFF
--- a/modules/tectonic/resources/manifests/updater/operators/tectonic-cluo-operator.yaml
+++ b/modules/tectonic/resources/manifests/updater/operators/tectonic-cluo-operator.yaml
@@ -28,7 +28,7 @@ spec:
         resources:
           limits:
             cpu: 20m
-            memory: 50Mi
+            memory: 100Mi
           requests:
             cpu: 20m
             memory: 50Mi


lucab:
Forward dispatching:
 * @dghubble I see the limits/requests here but I don't know where those numbers are coming from (they are not in tectonic-x-op nor tectonic-cluo-op examples)
 * @yifan-gu this operator is a just plain manifest-overlay on top of tectonic-x-op. I don't have idea about resource allocation for that, and I was not able to find any example or other operators with them.

dmyerscough:
Just as a reference; we encountered this on Friday:-

cluo

The limits were increased to prevent a crash loop as OOM-killer was triggered.

dghubble:
@lucab yeah, tectonic-cluo-operator is a layer that adds manifests to the tectonic-x-operator image. Both are internal so we should maybe discuss there instead. As for the resource requests, I left them at the defaults, unaware tectonic-x-operator can (sometimes?) consume more than that. I haven't seen in OOM in my tests as a consumer, but maybe there is something to investigate.

Did this fix the issue and usage stabilized or did it just buy more time?

dmyerscough:
@dghubble this stabilized the pod. I have not seen a crash since upping the limit and the pods usage is running around 40Mi.

dghubble:
sgtm

euank:
Do note that this is implicitly changing the pod from being in the "guaranteed" QoS class to the "Burstable" class due to making `limits` != `requests`.

I assume it was "guaranteed" for a reason before?

dmyerscough:
This still gives the pod in a "guaranteed" state for 50Mi of memory. I feel the pod should have the ability to burst if space is available otherwise crash loop until space is available. This is better than just having a pod crash loop until an administrator has to intervene.

euank:
@dmyerscough I'm talking specifically about the kubernetes concept of a "Quality of Service class" which have a couple subtle impacts on scheduling and eviction.
It has a request of 50Mi of memory which isn't quite the same as being in the guaranteed QoS class.

One obvious difference is the oom score adjust behavior, which uses a different value for guaranteed than for best effort or burstable.

dmyerscough:
Any movement on this? We have had two upgrades that reset this value and cause Cluo to crashloop.

squat:
@dghubble @yifan-gu any further opinions on this PR?

euank:
@coreosbot test this please

/retest

euank:
ok to test

squat:
ok to test

squat:
Some flakes. Running again.

squat:
ok to test

PR #1978, opened 2017-09-25
frontend: Remove experimental features and AWS entitlements code

kyoto:
This functionality is currently unused and not expected to be reinstated.


ggreer:
A follow-up PR to this should probably remove `tectonic_experimental` and friends from the terraform scripts and docs.

kyoto:
@ggreer Thanks. We have a ticket for that.

PR #1843, opened 2017-09-06
[WIP] Create generic getter for CL version

squat:

PR #1979, opened 2017-09-25
Create generic getter for CL version

squat:
Opening in favor of #1843 because jenkins got in a weird state:tm:.
Fixes: INST-120

Please TAL @alexsomesan @s-urbaniak. This is the starting point: https://github.com/coreos/tectonic-installer/pull/1979/commits/82735fc6e75323586b9a7b0e0f057c4a7e0b4aa3#diff-b2934582bf0e217862c4d729e6779af2
There was one flake in the smoke tests, documented and reported here: https://coreosdev.atlassian.net/browse/INST-470


alexsomesan:
retest this please

modules/container_linux/main.tf
```diff
@@ -0,0 +1,4 @@
+data "external" "version" {
+  count   = "${var.version == "latest" ? 1 : 0}"
+  program = ["sh", "-c", "curl https://${var.channel}.release.core-os.net/amd64-usr/current/version.txt | sed -n 's/COREOS_VERSION=\\(.*\\)$/{\"version\": \"\\1\"}/p'"]
```
lucab: One detail that I would like to see documented somewhere: when is this external value determined and would it be re-computed as a side-effect of some terraform invocations?

My understanding is that the goal for this is to be computed exactly once per cluster and then never change, but I would like to have this confirmed and noted down.

squat: This value is computed *exactly once* per platform per `terraform <operation>` and used for the entire operation. 

Note that this does create the following dempotency issue:
if a user sets `tectonic_container_linux_version = "latest"`:
* runs `terraform apply` and creates a cluster today; then
* runs `terraform apply` next week
they will see a diff caused by the newly calculated version of CL.

I think that this lack of idempotency is strictly better than what we have now: currently, users will see the same idemptency issues in AWS caused by the lookup of the latest container linux AMI. This PR at least unifies the behavior of the installer. We could add declarations to all created instances to ignore the `ami` field (and appropriate field for other clouds) if the container linux version was set to "latest", however I think that is fair to tackle in a separate issue.

lucab: What about noting these details somewhere under `Documentation/`?

squat: Definitely. I'll add another commit to that end.

alexsomesan:
A new smoke test for this scenario would truly be outstanding :)

squat:
@alexsomesan you mean a test for the scenarios where `apply` is run after a new version of CL is released?

squat:
@alexsomesan what test were you imagining?

alexsomesan:
I guess we can start off with just the simple case where we boot a specific
older version of CL and validate that all nodes come up with that one. Just
to have baseline for validating this holds true future changes.

We can then build a more elaborate node upgrade on top should that fall in
installer test territory.

squat:
ok to test

squat:
ok to test

squat:
ok to test

squat:
Tests passed. Running again for confidence.

modules/container_linux/main.tf
```diff
@@ -0,0 +1,3 @@
+data "external" "version" {
+  program = ["sh", "-c", "curl https://${var.channel}.release.core-os.net/amd64-usr/current/version.txt | sed -n 's/COREOS_VERSION=\\(.*\\)$/{\"version\": \"\\1\"}/p'"]
```
squat: This has no `count` and is instead always run, otherwise Terraform's issue with evaluating both sides of ternary expressions causes problems in the output.

s-urbaniak: as discussed OOB: we should write a custom datasource for this as this code might break platform compat (looking at you, Windows).

tests/smoke/aws/vars/aws.tfvars.json
```diff
@@ -7,7 +7,9 @@
     "tectonic_base_domain": "tectonic.dev.coreos.systems",
     "tectonic_ca_cert": "",
     "tectonic_ca_key": "",
-    "tectonic_cl_channel": "stable",
+    "tectonic_container_linux_channel": "stable",
+    "tectonic_container_linux_version": "1465.6.0",
+    "tectonic_bootstrap_upgrade_cl": "false",
```
squat: This is necessary to prevent torx from rebooting the node when it finds that the version of CL is not the latest.

tests/smoke/bare-metal/vars/metal.tfvars
```diff
@@ -93,8 +93,8 @@ kA==
 -----END CERTIFICATE-----
 EOF
 
-tectonic_cl_channel = "stable"
-tectonic_metal_cl_version = "1298.7.0"
+tectonic_container_linux_channel = "stable"
+tectonic_container_linux_version = "1298.7.0"
```
s-urbaniak: could we bump this while being here?

modules/container_linux/main.tf
squat: Correct. We are considering writing a small binary to provide cross-OS compatibility for this request. we could make a Tectonic resource provider that exposes this and other needed datasources.

squat:
Tests passed twice in a row. Currently running a third time. Will run a fourth time (with the updated Container Linux version for bare metal that @s-urbaniak requested).

squat:
@alexsomesan the reason is that the datasource is non-trivial and execs several binaries and implementing it several times could lead to divergence or could be problematic to maintain. Instead, it is nice to abstract the logic into its own module. If we write a custom resource provider for this and other Tectonic-specific data sources then we can remove the module and add the new datasource directly to each platform.

squat:
Tests passed three times in a row. Running a fourth time with a bumped CL version for baremetal smoke tests (even though these are not currently exercised). Side-note, these tests have gotten very stable, props to @cpanato and @mxinden.

tests/rspec/lib/shared_examples/k8s.rb
```diff
@@ -32,6 +33,16 @@
     expect { SmokeTest.run(@cluster) }.to_not raise_error
   end
 
+  it 'installs the correct Container Linux version' do
+    version = @cluster.tf_var('tectonic_container_linux_version')
+    if version == 'latest' then version = @cluster.tf_value('module.container_linux.version') end
```
alexsomesan: This kind of construct is more idiomatically expressed in Ruby with the condition at the end. 
Like this:
```
version = @cluster.tf_value('module.container_linux.version') if version == 'latest'
```

squat: yes. good catch. this is ruby convention (though I disagree that it increases legibility). updating this now

squat: updated!

alexsomesan: I'm not super keen on staying dead-on idiomatic with Ruby because there's a lot of weird conventions about it, I agree.
Although they take time to get used to, this kind of conditionals read more like natural language.
Not the best place to debate this, but I think their intention was to have code read like a phrase: "Do eat IF you are (==) hungry".

squat:
Tests passed 5 times. Can I get one last sign off and then merge?

mxinden:
@squat Thanks for the feedback!

PR #1985, opened 2017-09-25
Added kubernetes addon operator config

thorfour:
ConfigMap for the kubernetes-addon-operator


diegs:
cc @yifan-gu 

I think we want to group all the configs into the `cluster-config.yaml` objects that exist one per namespace:

https://github.com/coreos/tectonic-installer/pull/1933

Can we finish that PR first and then move forward with these?

thorfour:
Will be merged into single config map https://github.com/coreos/tectonic-installer/pull/1933

PR #1988, opened 2017-09-25
Bump Tectonic Console to v2.0.3

ggreer:
Includes:

- Fix styling for cluo update.
- More accurate queries for some gauge graphs on the front page.

This should get merged into the release branch.

### DIFF
--- a/config.tf
+++ b/config.tf
@@ -58,7 +58,7 @@ variable "tectonic_container_images" {
     bootkube                     = "quay.io/coreos/bootkube:v0.6.2"
     calico                       = "quay.io/calico/node:v2.4.1"
     calico_cni                   = "quay.io/calico/cni:v1.10.0"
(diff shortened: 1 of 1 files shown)


Quentin-M:
Unless this is a critical bug fix or a deal-breaker for product requirements, I would argue that this does not have to be in the release branch (and require yet another KVO bump).

/cc @sudhaponnaganti @robszumski

ggreer:
This doesn't need a new release of the installer, KVO, etc, but I think it should get in the release eventually. It's just some minor bug fixes.

alexsomesan:
running the smokes

sudhaponnaganti:
@ggreer - I will make RC5 tomorrow. So can you cherry-pick this in to 1.7.5 if applicable. I understood from notes above that this doens't have any other dependencies and can be merged cleanly in to 1.7.5. As we are close to 1.7.5, wanted to reduce impact.

Issue #1164, opened 2017-06-21
Feature Request - AWS: Parameterize S3 Bucket Name


alekssaul:
## Versions

* Tectonic version ([release](https://github.com/coreos/tectonic-installer/releases) or commit hash):
1.6.X
* Terraform version (`terraform version`):
0.9.6
* Platform (aws|azure|openstack|metal):
aws

## What happened?
User evaluating tectonic requested ability to specify S3 bucket name due to their Security controls. 
This seem to be set in; https://github.com/coreos/tectonic-installer/blob/master/platforms/aws/s3.tf#L9

sym3tri:
Fixed in #1991

PR #1991, opened 2017-09-26
Addressing feature request #1164

jasmingacic:
This is addressing issue #1164 in the way if parameter `tectonic_s3_bucket_name` isn't passed the name will be constructed using existing logic. https://github.com/coreos/tectonic-installer/blob/master/platforms/aws/s3.tf#L9


Documentation/variables/config.md
```diff
@@ -6,6 +6,7 @@ This document gives an overview of variables used in all platforms of the Tecton
 
 | Name | Description | Type | Default |
 |------|-------------|:----:|:-----:|
+| tectonic_S3_bucket_name | (optional) Unique name of the S3 bucket<br><br>If name is not provided the installer will construct the name using "tectonic_cluster_name", current AWS region and "tectonic_base_domain" | string | `` |
```
alexsomesan: Can we have S3 in lowercase here. We actually don't use capitals in variable names anywhere in the project.

config.tf
```diff
@@ -119,6 +119,17 @@ variable "tectonic_versions" {
   }
 }
 
+variable "tectonic_S3_bucket_name" {
+  type    = "string"
+  default = ""
+
+  description = <<EOF
+(optional) Unique name of the S3 bucket
```
alexsomesan: It's not obvious from this text what this parameter does.
The description should clearly explain what the input parameter is used for.

jasmingacic: @alexsomesan I've got this now:
```
(optional) Unique name under which the Amazon S3 bucket will be created. Bucket name must start with a lower case name and is limited to 63 characters

If name is not provided the installer will construct the name using "tectonic_cluster_name", current AWS region and "tectonic_base_domain"
```

Is this what you had in mind?

alexsomesan: Not really actually, but mentioning it's naming restriction is useful too.
I was referring to the purpose of this parameter. What the installer is doing with this value.

jasmingacic: @alexsomesan how does this sound?
```
(optional) Unique name under which the Amazon S3 bucket will be created. Bucket name must start with a lower case name and is limited to 63 characters. The Tectonic Installer uses the bucket to store tectonic assets and kubeconfig.
```

alexsomesan: Sounds good, thanks for addressing it.

Documentation/variables/config.md
```diff
@@ -35,6 +35,7 @@ This document gives an overview of variables used in all platforms of the Tecton
...
 | tectonic_master_count | The number of master nodes to be created. This applies only to cloud platforms. | string | `1` |
 | tectonic_pull_secret_path | The path the pull secret file in JSON format. This is known to be a "Docker pull secret" as produced by the docker login [1] command. A sample JSON content is shown in [2]. You can download the pull secret from your Account overview page at [3].<br><br>[1] https://docs.docker.com/engine/reference/commandline/login/<br><br>[2] https://coreos.com/os/docs/latest/registry-authentication.html#manual-registry-auth-setup<br><br>[3] https://account.coreos.com/overview<br><br>Note: This field MUST be set manually prior to creating the cluster unless `tectonic_vanilla_k8s` is set to `true`. | string | `` |
+| tectonic_s3_bucket_name | (optional) Unique name under which the Amazon S3 bucket will be created. Bucket name must start with a lower case name and is limited to 63 characters<br><br>If name is not provided the installer will construct the name using "tectonic_cluster_name", current AWS region and "tectonic_base_domain" | string | `` |
```
alexsomesan: Two more nits: 
1. This input is platform specific so it needs to have the `_aws_` infix in the variable name like the rest of them.
2. Can we reflect a little bit of the purpose in the name of the variable?
I'm thinking something like renaming it to `tectonic_aws_assets_s3_bucket_name`

jasmingacic: maybe if we keep it shorter `tectonic_aws_s3_bucket_name` since the purpose of the parameter is already given in description.

jasmingacic: And to make sure we should imply in description that the parameter is AWS specific

alexsomesan: Since this change touches the user facing API, I'm trying to be a little future-proof here. There might be a case in the future when we need to add a second s3 bucket for something else.
Imagine that one of them is called `tectonic_aws_s3_bucket_name` and the other one is called `tectonic_aws_s3_bucket_for_something_else`. It's going to become confusing then.

sym3tri: We should stick to the naming convention already established & used by all other variables as @alexsomesan suggests.

jasmingacic: gotcha

jasmingacic:
@alexsomesan All remarks addressed and conflicts resolved. Is there anything else that needs to be done.

PR #1994, opened 2017-09-26
frontend: Add the ability to set IAM role for master / worker nodes

kyoto:
Also removes the `extra` data from the .progress file because that data will be dynamically loaded again anyway when the .progress file is restored.


alexsomesan:
I'll run the smokes here, since there are small template changes.

kyoto:
Test failure is an Azure test, so presumably unrelated.

cpanato:
if it is just the frontend and we are not spinning clusters we don't need to set the label `run-smoke-tests` because the frontend tests will run always

cpanato:
yes, it an flaky

installer/frontend/components/base.jsx
```diff
@@ -19,6 +19,10 @@ import { Footer } from './footer';
 
 const downloadState = (state) => {
   const toSave = savable(state);
+
+  // Extra field data doesn't need to be saved
+  _.unset(toSave, 'clusterConfig.extra');
```
kans: I guess this is OK... it goes in localstorage but not progress files.

kyoto:
@squat Could you review?

installer/api/handlers_aws.go
```diff
@@ -227,6 +228,28 @@ func awsGetKeyPairsHandler(w http.ResponseWriter, req *http.Request, _ *Context)
...
 }
 
+// awsGetIamRolesHandler returns the list of IAM roles.
+func awsGetIamRolesHandler(w http.ResponseWriter, req *http.Request, _ *Context) error {
+	awsSession, err := awsSessionFromRequest(req)
+	if err != nil {
+		return fromAWSErr(err)
+	}
+	iamSvc := iam.New(awsSession)
+
```
squat: total nit: this newline should be above the iamSvc declaration, right?

installer/frontend/components/aws-define-nodes.jsx
```diff
@@ -40,13 +41,27 @@ const IOPs = connect(
   </Row>
 );
 
+const IamRoles = connect(
+  ({clusterConfig}) => ({roles: _.get(clusterConfig, ['extra', IAM_ROLE], [])})
+)(
+  ({roles, type}) => <Row htmlFor={`${type}--iam-role`} label="IAM Role">
+    <Connect field={toKey(type, IAM_ROLE)}>
```
squat: This seems like a funny construct to be `connect`-ing a connected component. Why not get it all from the stateToProps?

installer/frontend/components/aws-define-nodes.jsx
```diff
@@ -46,56 +46,45 @@ const Errors = connect(
...
-      <Connect field={toKey(type, STORAGE_TYPE)}>
-        <Select id={`${name}--storage-type`}>
-          <option value="" disabled>Please select storage type</option>
-          <option value="gp2" key="gp2">General Purpose SSD (GP2)</option>
-          <option value="io1" key="io1">Provisioned IOPS SSD (IO1)</option>
-          <option value="standard" key="standard">Magnetic</option>
-        </Select>
-      </Connect>
-    </Row>
+export const DefineNode = ({type, max}) => <div>
```
squat: Is this commit necessary for this feature? Could we break this into a separate PR?

installer/frontend/components/aws-define-nodes.jsx
kyoto: We actually use `<Connect>` inside a `connect`-ed component in a few places. This component case is copying the pattern of `IOPs` above.

squat: Yes this is what I suspected, it just seems funny/inefficient to need to bind twice. I understand why it is done but not why it needs to be this way vs just putting everything in the `connect`. In any case, we do not need to fix this today but we should consider refactoring it away in the future to improve performance and simplify our code.

installer/frontend/components/aws-define-nodes.jsx
kyoto: OK, I moved it to a separate PR: https://github.com/coreos/tectonic-installer/pull/2026, but it will need to be merged before this PR.

installer/api/handlers_aws.go
kyoto: Fixed

installer/frontend/components/aws-define-nodes.jsx
squat: Sorry, that's why I was asking if this commit was necessary for this PR. If it needs to be merged before this feature then it *is* necessary. In any case, I am LGTMing #2026.

squat:
All tests passed.

PR #1997, opened 2017-09-27
frontend: Migrate BM_Matchbox from WithClusterConfig to Connect

kyoto:


installer/frontend/components/bm-matchbox.jsx
```diff
@@ -3,18 +3,27 @@ import classNames from 'classnames';
...
   BM_MATCHBOX_HTTP,
   BM_MATCHBOX_RPC,
   BM_OS_TO_USE,
 } from '../cluster-config';
 
+const form = new Form('BM_MATCHBOX_ADDRESS', [
+  new Field(BM_MATCHBOX_RPC, {default: '', validator: validate.hostPort}),
+], {
+  validator: cc => !validate.hostPort(cc[BM_MATCHBOX_HTTP]) &&
+      !validate.nonEmpty(cc[BM_OS_TO_USE]) &&
```
kans: This validator will be triggered when BM_MATCHBOX_RPC changes but it looks like BM_OS_TO_USE is set from a GET.  Do the timings work out here?  You may need to convert BM_MATCHBOX_HTTP as well.

kyoto:
ok to test

kyoto:
ok to test

installer/frontend/components/bm-matchbox.jsx
kyoto: I ended up moving this logic back to `BM_Matchbox.canNavigateForward()` to work around the problem.

PR #1999, opened 2017-09-27
config.tf: bump hyperkube to 1.7.6

squat:
@Quentin-M @diegs


squat:
Got one flake in the network policy smoke test.

squat:
ok to test

diegs:
This is not going into any release though, right?

PR #2007, opened 2017-09-28
config.tf: also bump dns sidecar and nanny

squat:
We need to bump all three DNS images as upstream has done: https://github.com/kubernetes/kubernetes/pull/53085/files

cc @alexsomesan


squat:
Tests are failing because something in master broke the examples. Addressed in this PR: #2008

alexsomesan:
@squat can you try rebasing this? Carlos thinks that might fix the tests.

cpanato:
I got the same issue in another PR and I did a rebase

PR #2013, opened 2017-09-28
Bump Tectonic Console to v2.0.3 (#1988)

ggreer:


amrutac:
Needs to updated in KVO as well

sudhaponnaganti:
@diegs - Can you update KVO and push it to master and 1.7.5 - Need to make RC5

PR #2016, opened 2017-09-28
config.tf: updated version string for 1.7.3 patch release

Quentin-M:


config.tf
```diff
@@ -75,7 +75,7 @@ variable "tectonic_versions" {
     kubernetes    = "1.7.3+tectonic.1"
```
diegs: bump to `1.7.3+tectonic.3`

Quentin-M: Do we need to do that usually? I think this is to indicate hyperkube bumps if I remember clearly.

diegs: this tells the KVO which version payload to use

Quentin-M: Ack. Thanks!

diegs: also the KVO version will be `quay.io/coreos/kube-version-operator:v1.7.3-kvo.4`

Quentin-M:
.4 is the KVO image but all the rest of the strings are .3 :)

Quentin-M:
test upgrade from 1.7.3-tectonic.2 1.7.3-tectonic.2

PR #2023, opened 2017-09-29
config.tf: kvo to v1.7.5-kvo.10

diegs:
Master branch bump, no rush.


diegs:
@squat when you have a change this is the final version that went into 1.7.5 so we can put it into master too. Thanks!

PR #2027, opened 2017-10-02
tests/frontend: Increase timeout to account for slow test environment

kyoto:


squat:
Unrelated flake. merging.

PR #2028, opened 2017-10-02
examples: run `make examples`

squat:
cc @alexsomesan here's a quick one to re-sync the docs and examples directories.


squat:
static tests passed. merging.

PR #1389, opened 2017-07-13
aws: use existing security groups for masters/workers

hhoover:
This change allows a user to specify custom security groups for
master and worker nodes, which need to exist in advance before
install. If a custom security group is not provided, the groups
will be created. This also assumes you are using an existing VPC.

This is based on the Azure work for custom NSG and makes heavy
use of Terraform's interpolation. This change also preserves the
use of `tectonic_aws_master_extra_sg_ids` and
`tectonic_aws_worker_extra_sg_ids` for users looking to use more
than one security group.

Closes #1163

Issue #1163, opened 2017-06-21
AWS: Allow Customization of Security Group Ingress/Egress


ScottAmes:
## Is this a BUG REPORT or FEATURE REQUEST?

FEATURE REQUEST

## Versions

* Tectonic version: [1.6.4-tectonic.1](https://github.com/coreos/tectonic-installer/tree/1.6.4-tectonic.1)
* Terraform version: `0.9.4`
* Platform (aws|azure|openstack|metal): aws

## What happened?

Security Groups created via the [vpc module](https://github.com/coreos/tectonic-installer/tree/1.6.4-tectonic.1/modules/aws/vpc) hard code egress & ingress cidr blocks to `0.0.0.0/0`. Example:

```
resource "aws_security_group_rule" "worker_ingress_ssh" {
  type              = "ingress"
  security_group_id = "${aws_security_group.worker.id}"

  protocol    = "tcp"
  cidr_blocks = ["0.0.0.0/0"]
  from_port   = 22
  to_port     = 22
```

## What you expected to happen?

Optimally, similar behavior used to provide additional tags ([`variable "tectonic_aws_extra_tags` & `variable "tectonic_autoscaling_group_extra_tags"`](https://github.com/coreos/tectonic-installer/blob/1.6.4-tectonic.1/platforms/aws/variables.tf)) could be used to provide either security group ID's (preferred), using the `source_security_group_id` argument ([example](https://github.com/coreos/tectonic-installer/blob/1.6.4-tectonic.1/modules/aws/vpc/sg-worker.tf#L71-L79)) or an input cidr (i.e. `10.0.0.0/16`) to allow communication to/from. Separating Ingress from Egress, instead of one blanket argument, would likely be necessary to allow customization of one versus the other.

## How to reproduce it (as minimally and precisely as possible)?

n/a

## Anything else we need to know?

n/a

alekssaul:
Two prospects asked about this as well /cc: @sir-rob

dcowden:
Related to this but different, we would like to to be able to supply our own SG's entirely. This could be simply a new flag added that allows suppressing creation of the existing ones, and applying the ones we specify only.

robszumski:
[...]

PR #2029, opened 2017-10-02
aws: use existing security groups for masters/workers [previously #1163]

alexsomesan:
This is just to test rebasing of #1389


alexsomesan:
PR hooks not working on the original PR

sym3tri:
@alexsomesan Can we close out #1389 now that this replaces it?

alexsomesan:
Yes, I would say so. I'll close that one out.

alexsomesan:
ok to test

alexsomesan:
On closer inspection, it looks like this change duplicates functionality already present in the installer.

We already have input variables `tectonic_aws_*_extra_sg_ids` that allow injecting additional external security groups.
This change aims to introduce new inputs that provide the same functionality.

PR #2039, opened 2017-10-04
modules/ignition: introduce runtime-mappings.yaml

lucab:
This introduces a runtime-mappings.yaml to be used by tectonic-torcx.

In particular, this file is consumed by k8s-node-bootstrap.service
as a fallback to determine the docker version to install when the
api-server is unreachable (e.g. while installing a brand new cluster).
If an api-server is available, an up-to-date version of this data
will be sourced from a config-map instead as a primary source.

Part of OST-79

/cc @squeed


alexsomesan:
run smokes

lucab:
@robszumski this is the seconday source of version-mappings for the bootstrapper. The primary source is at https://github.com/coreos-inc/tectonic-cluo-operator/pull/13. Both of them will need to be owned by somebody involved in the release process and sync'd.

/cc @Quentin-M @diegs

lucab:
Bikeshedding question: is the name specific enough? This thing used to be called `version-manifest` but it felt a bit too generic, hence the current name. I'm happy as it is, but still open to better proposals.

diegs:
cc @derekparker

derekparker:
One thing we have discussed offline is using this configmap to express both Docker and Kubelet versions, since those are so tightly coupled. This means that the KVO Node Agent will also read from this configmap, and KVO will write to this configmap with the versions of docker / kubelet that are defined in the KVO upgrade spec.

The upgrade flow for Nodes then becomes:

1. KVO updates this configmap with correct versions
2. KVO applies `needs-reboot` to all Nodes
3. CLUO applies `before-update` label on those Nodes
4. Torcx and Node-agent run as before reboot hooks to do their work
5. KVO waits for all Nodes to reboot, then proceeds

cc @diegs @lucab @squeed @crawford @aaronlevy

squeed:
The flow makes sense - it's definitely an improvement - but I'm not sure that it's necessary, given that most kubelet upgrades will be non-disruptive and not involve tectonic-torcx.

modules/ignition/resources/kubernetes/runtime-mappings.yaml
```diff
@@ -0,0 +1,11 @@
...
+versions:
+  k8s:
+    1.6:
+        docker: [ "1.12"]
+    1.7:
+        docker: [ "1.12" ]
+    1.8:
+        # As per k8s issue 42926, 1.11, 1.12, 1.13, and 17.03 are supported
+        # see https://github.com/kubernetes/kubernetes/issues/42926#issuecomment-325733231
+        docker: [ "1.13", "1.12"]
```
crawford: We don't ship 1.13. We should instead just jump straight to 17.03.

lucab: This was just an optimistic forecast made at 1.7 time. This file is not yet consumed by the bootstrapper anyway, so I'm keeping this in sync with the old one for the moment, will bump them all at the same time once testing is green. We'll also need a forecast for 1.9.

lucab:
@diegs so it looks like this can be already merged as is? (I'll decouple and take care of the new version followup)

mxinden:

Quentin-M:
@lucab
> Both of them will need to be owned by somebody involved in the release process and sync'd.

May I understand why/how? 
Does not look like it's changing anything to the way Tectonic is released?

diegs:
@Quentin-M he means that this needs to be sync'd with one that is shipped in the KVO or TCLUO

modules/gcp/worker-igm/variables-ignition.tf
```diff
@@ -1,25 +0,0 @@
-# This file is supposed to be symlinked in consuming modules
```
lucab: @enxebre it looks you copied this file instead of symlinking it. I'm fixing GCP here as it's making my PR fail. Thanks to @squat for quickly figuring this out.

lucab:
Rebased, PTAL. It is ready for review and supposed to get in Helium as per https://github.com/coreos-inc/kube-version-operator/blob/master/Documentation/tectonic_torcx.md#tectonic-install-flow. I think this needs additional labels in order to be tested on all platforms.

squat:
ok to test

config.tf
```diff
@@ -86,7 +86,7 @@ variable "tectonic_container_images" {
     tectonic_etcd_operator       = "quay.io/coreos/tectonic-etcd-operator:v0.0.2"
     tectonic_prometheus_operator = "quay.io/coreos/tectonic-prometheus-operator:v1.7.1"
     tectonic_cluo_operator       = "quay.io/coreos/tectonic-cluo-operator:v0.2.4"
-    tectonic_torcx               = "quay.io/coreos/tectonic-torcx:installer-latest"
+    tectonic_torcx               = "quay.io/coreos/tectonic-torcx:v0.2.0"
```
lucab: @s-urbaniak @sym3tri this gets rid of the mutable tag.

meghnagala:
@cpanato What is the ETA on these tests passing?

lucab:
@meghnagala I'm re-spawning my local metal cluster to double-check I didn't miss anything stpuid. I've also re-pushed this PR, re-kicking the CI to ensure I didn't hit some odd flakes. If neither of those helps, I'm going to grab @cpanato tomorrow morning to dig further.

cpanato:
@lucab looks like all tests failed due timeout during the bootstrap process, we have all the logs from docker and journals we can dig into tomorrow if you don't mind.

lucab:
This is still failing due to an unrelated failure in metal testing environment. We tracked this down to broken bridge connectivity across VMs on the Packet machine, which should be hopefully resolved soon. I'm going to rebase and push this once metal testing environment is fixed.

squat:
ok to test

squat:
Some azure tests failed; running those selectively so we can run just the bare-metal tests once that CI infrastructure is fixed.

cpanato:
retest this please

cpanato:
now all azure are green

cpanato:
retest this please

cpanato:
@lucab tests passed

PR #2044, opened 2017-10-05
modules/bootkube: Fix cluster-config object

derekparker:
There was a typo in "initialConfig" and other objects should not be
marked as strings.


derekparker:
Note this should make it into the 1.7.5 release. @Quentin-M (or whoever I should CC about this).

PR #2049, opened 2017-10-05
Fix Makefile target e2e-docker-image for linux

mxinden:
The `e2e-docker-image` make target does not work with Linux and GNU
Make 4.2.1 because `E2E_IMAGE` variable is not passed to both following
commands. This patch exports it, before using it later.

Not sure if this is a idiomatic approach. Happy for any suggestions. 

@ethernetdan @alexsomesan Could you try this out? As far as I know you are both running on Mac?!


mxinden:
@alexsomesan Merci!

PR #2052, opened 2017-10-05
updater: Add toleration and node selector for tco to schedule it to master

yifan-gu:
This is because tco replies on ssl certs path that doesn't exist on rhel nodes.

### DIFF
--- a/modules/tectonic/resources/manifests/updater/operators/tectonic-channel-operator.yaml
+++ b/modules/tectonic/resources/manifests/updater/operators/tectonic-channel-operator.yaml
@@ -17,6 +17,12 @@ spec:
         k8s-app: tectonic-channel-operator
         tectonic-app-version-name: tectonic-cluster
     spec:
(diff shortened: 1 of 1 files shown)


yifan-gu:
test upgrade from 1.7.3-tectonic.2 1.7.3-tectonic.2

yifan-gu:
test upgrade from 1.7.3-tectonic.2 1.7.3-tectonic.2

yifan-gu:
test upgrade from 1.7.3-tectonic.2 1.7.3-tectonic.2

yifan-gu:
Bump'd the terraform builder image in the upgrade test.

lucab:
I guess this is/should be part of INST-227?

yifan-gu:
@lucab INST-227 is not the direct reason we did this for tco today, it's because of an issue happens on rhel worker nodes.
But I will leave a comment in the jira story, thanks for pointing!

PR #2053, opened 2017-10-05
Fix Security Groups to correct node port range

coresolve:
Correct default node port range is 30000 - 32767 per:
https://git.io/vd0v3

This fixes the range we probably need to think about how to address this
for existing clusters.


coresolve:
@sym3tri @alexsomesan

squat:
cc @s-urbaniak

s-urbaniak:
forgot what I just said, port 32000-32002 is only declared as a MUST for ingress, the corrected port range here is correctly declared in the spec from 30000-32767.

PR #2057, opened 2017-10-06
Bump console to v2.2.1.

ggreer:
A few bug fixes and styling tweaks. Also allows editing service monitors, filtering more resource types, and a few other UI niceties.


ggreer:
I'm not sure why these smoke tests failed. I booted a cluster with this branch, and the console came up fine.

PR #2063, opened 2017-10-09
tests: fix updated css in the console

cpanato:


tests/rspec/lib/pages/login_page.rb
```diff
@@ -13,7 +13,7 @@ class Login < BasePage
   PASSWORD_INPUT          = { id: 'password' }.freeze
   SUBMIT_INPUT            = { css: 'body > div.dex-container > div > form > button' }.freeze
   CLUSTER_STATUS_LABEL    = {
-    css: '#content > div > div.co-p-cluster__body > div.row.co-m-nav-title > div > h1 > span'
```
squat: we should consider adding an ID or class to this element so we can select it more easily and not need to rely on fragile nested selectors that are likely to continue to evolve

cpanato: @squat yes, I'm cloning the console repo to add those :)

cpanato: but lets fix this quickly :) @squat

PR #2076, opened 2017-10-10
tests: Clean workspace with root before any test execution

mxinden:
We have been facing a lot of `permission denied` issues due to tests
executed as root that did not clean up after themselves before exiting.
Instead of relying on other tests to clean up, tests should clean up before
each execution to make sure this doesn't block them anymore.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -82,6 +82,7 @@ pipeline {
       }
       steps {
         node('worker && ec2') {
+          forcefullyCleanWorkspace()
           withDockerContainer(params.builder_image) {
             ansiColor('xterm') {
               checkout scm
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -267,10 +271,26 @@ pipeline {
...
 }
 
+def forcefullyCleanWorkspace() {
+  return withDockerContainer(
+    image: tectonic_smoke_test_env_image,
+    args: '-u root'
+  ) {
+    ansiColor('xterm') {
+      sh """#!/bin/bash -ex
+        if [ "\$(pwd)" == "\$WORKSPACE" ]
```
alexsomesan: Why not just `rm -rfv \$WORKSPACE` ?

mxinden: I want to make sure $WORKSPACE is not corrupted and e.g. set to `/`.

PR #2078, opened 2017-10-10
azure: make vnet subnet cidr extend bits configurable

rubbish:
I don't really know if this is useful to anyone else or not. But when I spun up a cluster in Azure on a smaller CIDR than a `/16`, this ran into problems because it expect to just extend it by 4 by default. It's not an even split. So I made it configurable. 

However, I probably will go down the route of making my own subnets instead.

### DIFF
--- a/platforms/azure/variables.tf
+++ b/platforms/azure/variables.tf
@@ -111,6 +111,15 @@ service range or a private datacenter connected via ExpressRoute."
 EOF
 }
 
+variable "tectonic_azure_vnet_cidr_block" {
+  type    = "string"
+  default = "4"
+
+  description = <<EOF
+(optional) Bits to extend the given Virtual Network CIDR block prefix to split into the master and worker subnets.
+EOF
+}
+
 variable "tectonic_azure_external_vnet_id" {
   type    = "string"
   default = ""
--- a/modules/azure/vnet/variables.tf
+++ b/modules/azure/vnet/variables.tf
@@ -28,6 +28,10 @@ variable "vnet_cidr_block" {
   type = "string"
 }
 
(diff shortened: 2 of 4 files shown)


platforms/azure/variables.tf
```diff
@@ -111,6 +111,15 @@ service range or a private datacenter connected via ExpressRoute."
 EOF
 }
 
+variable "tectonic_azure_vnet_cidr_block" {
```
squat: It looks like this directive re-declares the `tectonic_azure_vnet_cidr_block` variable.

squat:
@rubbish overall this looks fine, however, I'd like to keep the platforms implementations consistent with one another, so it would be nice to make the AWS subnet sizes configurable as well. I am trying to think about the best UX for this. Currently this path asks a user to supply:
1. cluster CIDR block
2. number of subnets
3. subnet extension bits

then, using 1 and 3 we calculate the mask for each subnet.

I think it would be a nicer experience to provide:
1. cluster CIDR block
2. number of subnets
3. subnet mask

This is subjects the infrastructure to the same constraints but seems like a more natural way to express the requirements.

cc @robszumski @alexsomesan

rubbish:
@squat What I ended up instead of doing this was just creating the vnet separately and using the azure external terraform variables. After doing that, I think instead of making it so the subnet extension bits an input, if we could be able to supply the master & worker CIDR blocks and default to splitting the cluster CIDR block if they aren't supplied.

sym3tri:
@rubbish thanks for the contribution and sorry about the delay. Unfortunately we're not adding any new features to this repo since we're converging Tectonic into Open Shift. See our blog for more details: https://coreos.com/blog/coreos-tech-to-combine-with-red-hat-openshift

PR #2082, opened 2017-10-11
tests: Report specific smoke test status to GitHub

mxinden:


mxinden:
ok to test

mxinden:
@cpanato Could you review this?

cpanato:
retest this please

cpanato:
@mxinden when you click in the `details` for each feedback it redirects to `https://example.com/build/status` and not to our Jenkins. Can you please take a look on that?

mxinden:
ok to test

mxinden:
@sym3tri In the status field of this PR you can see the detailed Jenkins status feedback. Does this look good to you?

mxinden:
@cpanato Could you give this another review?

sym3tri:
@mxinden This is great. 

As a follow-up, it'd be even better if you can override the test title to be a short description instead of just the file path.

squat:
@mxinden looks great so far. Do you want to address Ed's comment in a follow up PR or in this one?

mxinden:
@squat I will address it here. There is an additional feature discussed offline to report the status of the basic tests separately as well and make that one required instead of the overall status.

mxinden:
ok to test

mxinden:
ok to test

mxinden:
ok to test

mxinden:
ok to test

mxinden:
ok to test

mxinden:
@kyoto @ggreer With this PR I would like to introduce running the "GUI Tests" stage only when the `run-gui-tests` Github label is set. This way we don't need to run the tests on pure backend changes. Does this sound reasonable to you two or are there any objections?

kyoto:
@mxinden Adding a `run-gui-tests` label seems reasonable to me, especially as they are likely to get slower over time as we add more checks.

mxinden:
@cpanato Could you give this another review? I fixed the stashing issue by reducing the scope of the stashing to only those files necessary (See https://github.com/coreos/tectonic-installer/pull/2082/commits/50b31d0c60dd56fbe5296d5c0d542dd6e48106ef). Let me know if this looks good to you.

cpanato:
retest this please

mxinden:
@cpanato What is the concern? 

In case you are worried about the delay introduced by this change: The last test execution of this PR was 45 minutes. Last master execution was 48 minutes.

cpanato:
@mxinden more in terms of doing the checkout every time during the build. but this will be fixed when we implement the new builder :)

PR #2087, opened 2017-10-11
modules/ignition: ensure --cloud-provider is set

squat:
The default behavior of the kubelet's `--cloud-provider` flag is
changing in the upcomint k8s releases:

* v1.8 - auto detection of cloud provider using cAdvisor is deprecated but
still remains the default.
* v1.9 - auto detection of cloudprovider is still supported but no longer
the default, the new default will be no provider.
* v1.10 - auto detecting cloud provider is removed

For consistent non-autodetection behavior we need to ensure that we
always set the cloud-provider flag, even if it is explicitly set to no
provider.

cc @s-urbaniak @aaronlevy

### DIFF
--- a/modules/ignition/assets.tf
+++ b/modules/ignition/assets.tf
@@ -32,7 +32,7 @@ data "template_file" "kubelet" {
   template = "${file("${path.module}/resources/services/kubelet.service")}"
 
   vars {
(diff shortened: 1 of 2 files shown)


s-urbaniak:
one final small red-flag: this changes manifests, i.e. introducing possible skew.

Issue #2101, opened 2017-10-11
Enable PodSecurityPolicy by default in Tectonic installations


joshrosso:
### Description: PodSecurityPolicy (PSP)

PSP is an admission controller that ensures pods scheduled in the cluster match a policy defined by the cluster administrator. For example, this can prevent pods from running with host networking or in privileged mode.

### Enabling PSPs

Adding PodSecurityPolicy to the list of admission controllers via `kube-apiserver` will enable it. The example below shows the PodSecurityPolicy added to the admission-controller flag.

```yaml
- --admission-control=NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass,ResourceQuota,PodSecurityPolicy
```

### Possible Approach

I understand this change needs to not break existing clusters and also not impede on the bootstrapping process. The following is a **suggestion** on how it could be approached and what my initial PR will contain. I also have an internal (google) doc going with greater detail.

1. Update the `modules/bootkube/resources/manifests/kube-apiserver.yaml` to contain the `PodSecurityPolicy` admission controller. 
2. Add a permissive PSP (full rights / match all) to be used by all cluster-admin users and similar service accounts.
    ```yaml
    apiVersion: extensions/v1beta1
    kind: PodSecurityPolicy
    metadata:
      name: system-permissive
    spec:
      seLinux:
        rule: RunAsAny
      supplementalGroups:
        rule: RunAsAny
      runAsUser:
        rule: RunAsAny
      fsGroup:
        rule: RunAsAny
      hostPorts:
      - min: 0
        max: 65535
      volumes:
      - '*'
      allowedCapabilities:
      - '*'
    ```
    > Without this, or a similar policy, bootstrapping will fail to deploy any tectonic pods.

3. Document how to create PSPs and roles that are able to use those PSPs. 
    - The ability to specify which PSP a role snippet that can **use** the PSP `limited-psp-workload-rule` is shown below.
        ```yaml
        - apiGroups:
           - extensions
          resources:
          - podsecuritypolicies
          resourceNames:
          - limited-psp-workload-rule
          verbs:
          - use
        ```

## Versions

* Tectonic version ([release](https://github.com/coreos/tectonic-installer/releases) or commit hash): 647ce84912142a0cd84b7f85427bbe5c619df57a
* Terraform version (`terraform version`): `Terraform v0.10.7`
* Platform (aws|azure|openstack|metal): aws

joshrosso:
[...]

PR #2102, opened 2017-10-11
Enable PodSecurityPolicy by default

joshrosso:
This commit enables the PodSecurityPolicy (PSP) admission controller in
Tectonic by amending it to the admission-controller flag of the
kube-apiserver. The --use-service-account-credentials flag has been added to the
kube-controller-manager, ensuring it runs with all individual
ServiceAccounts provided by k8s [0].

A 'permissive' and 'limited' PSP have been added for use by the
kube-controller-manager. The permissive PSP has no restrictions on
pods to create. The limited policy is the same as permissive with
the following added restrictions.

  - Pods may not run as privileged [1]
  - Pods may not run using host networking [2]

A ClusterRole allowing the verb 'use' for each PSP resource ('permissive',
'limited') has been created for binding to. Binding to means the
ServiceAccount will be able to resolve against that PSP when attempting
to create a pod. A ClusterRoleBinding has been created for the
limited PSP ClusterRole. Making all pod creations in all namespaces
require to match the rules set in the limited PSP. RoleBindings have
been created in kube-system and tectonic-system for the permissive PSP.
Making all pods created in the kube-system or tectonic-system namespace
able to resolve against the permissive PSP, containing no restrictions.
All mentioned bindings include the Subjects for every ServiceAccount
that uses the 'create' verb on the 'pod' resource. The following
ServiceAccounts (from the kube-system namespace) are bound.

  - replicaset-controller
  - replication-controller
  - job-controller
  - daemon-set-controller
  - statefulset-controller

This change will prevent users from being able to create pods in
namespaces as their user account won't have access to 'use' any PSP.
They will have to rely on high-level constructs, which they normally
would, to create pods such as StatefulSets or Deployments. When rights
are desired beyond that of the limited PSP, admins will need to add new
PSPs and likely change RBAC rules [3].

Resolves: #2101

[0]: https://kubernetes.io/docs/admin/authorization/rbac/#controller-roles
[1]: https://kubernetes.io/docs/concepts/workloads/pods/pod/#privileged-mode-for-pod-containers
[2]: https://kubernetes.io/docs/concepts/policy/pod-security-policy/#host-network
[3]: https://github.com/kubernetes/examples/blob/master/staging/podsecuritypolicy/rbac/README.md

## TODO:

- Documentation

### DIFF
--- a/modules/bootkube/resources/manifests/02-permissive-policy.yaml
+++ b/modules/bootkube/resources/manifests/02-permissive-policy.yaml
@@ -1,0 +1,158 @@
+---
+# Permissive policy allowing for any pod to run.
+apiVersion: extensions/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: system-tectonic-permissive
+spec:
+  privileged: true
+  hostNetwork: true
+  hostPID: true
+  hostIPC: true
+  seLinux:
+    rule: RunAsAny
+  supplementalGroups:
+    rule: RunAsAny
+  runAsUser:
+    rule: RunAsAny
+  fsGroup:
+    rule: RunAsAny
+  hostPorts:
+  - min: 0
+    max: 65535
+  volumes:
+  - '*'
+  allowedCapabilities:
+  - '*'
+
+---
+# Limited policy notably preventing privileged containers and containers wish to use host networking.
+apiVersion: extensions/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: system-tectonic-limited
+spec:
+  privileged: false
+  hostNetwork: false
+  seLinux:
+    rule: RunAsAny
+  supplementalGroups:
+    rule: RunAsAny
+  runAsUser:
+    rule: RunAsAny
+  fsGroup:
+    rule: RunAsAny
+  volumes:
+  - '*' 
+  allowedCapabilities: 
+  - '*'
+
+---
+# Role allowing the usage of full permissive PSP
+# Should be bound by a RoleBinding inside of kube-system and tectonic-system
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: permissive-psp-access
+rules:
+  - apiGroups: ["extensions"]
+    resources: ["podsecuritypolicies"]
+    resourceNames: ["system-tectonic-permissive"]
+    verbs: ["use"]
+
+---
+# Role allowing the usage of limited PSP
+# Should be bound by a ClusterRoleBinding for the controller-manager to ensure it has this PSP
+# to resolve to by default.
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: limited-psp-access
+rules:
+  - apiGroups: ["extensions"]
+    resources: ["podsecuritypolicies"]
+    resourceNames: ["system-tectonic-limited"]
+    verbs: ["use"]
+
+---
+# Allow kube-controller-manager to have permissive PSP access in tectonic-system namespace
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: permissive-psp-access
+  namespace: tectonic-system
+subjects:
+- kind: ServiceAccount
+  name: replicaset-controller
+  namespace: kube-system
+- kind: ServiceAccount
+  name: replication-controller
+  namespace: kube-system
+- kind: ServiceAccount
+  name: job-controller
+  namespace: kube-system
+- kind: ServiceAccount
+  name: daemonset-controller
+  namespace: kube-system
+- kind: ServiceAccount
+  name: statefulset-controller
+  namespace: kube-system
+roleRef:
+  kind: ClusterRole
+  name: permissive-psp-access
+  apiGroup: rbac.authorization.k8s.io
+
+---
+# Allow kube-controller-manager to have permissive PSP access in kube-system namespace
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: permissive-psp-access
+  namespace: kube-system
+subjects:
+- kind: ServiceAccount
+  name: replicaset-controller
+  namespace: kube-system
+- kind: ServiceAccount
+  name: replication-controller
+  namespace: kube-system
+- kind: ServiceAccount
+  name: job-controller
+  namespace: kube-system
+- kind: ServiceAccount
+  name: daemon-set-controller
+  namespace: kube-system
+- kind: ServiceAccount
+  name: statefulset-controller
+  namespace: kube-system
+roleRef:
+  kind: ClusterRole
+  name: permissive-psp-access
+  apiGroup: rbac.authorization.k8s.io
+
+---
+# Allow kube-controller-manager to have limit PSP access in all namespaces
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: limited-psp-access
+subjects:
+- kind: ServiceAccount
+  name: replicaset-controller
+  namespace: kube-system
+- kind: ServiceAccount
+  name: replication-controller
+  namespace: kube-system
+- kind: ServiceAccount
+  name: job-controller
+  namespace: kube-system
+- kind: ServiceAccount
+  name: daemon-set-controller
+  namespace: kube-system
+- kind: ServiceAccount
+  name: statefulset-controller
+  namespace: kube-system
+roleRef:
+  kind: ClusterRole
+  name: limited-psp-access
+  apiGroup: rbac.authorization.k8s.io
--- a/modules/bootkube/resources/manifests/kube-apiserver.yaml
+++ b/modules/bootkube/resources/manifests/kube-apiserver.yaml
@@ -41,7 +41,7 @@ spec:
         - --storage-backend=etcd3
         - --allow-privileged=true
         - --service-cluster-ip-range=${service_cidr}
-        - --admission-control=NamespaceLifecycle,LimitRanger,ServiceAccount,PersistentVolumeLabel,DefaultStorageClass,ResourceQuota,DefaultTolerationSeconds
+        - --admission-control=NamespaceLifecycle,LimitRanger,ServiceAccount,PersistentVolumeLabel,DefaultStorageClass,ResourceQuota,DefaultTolerationSeconds,PodSecurityPolicy
         - --tls-ca-file=/etc/kubernetes/secrets/ca.crt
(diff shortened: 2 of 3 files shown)


squat:
@joshrosso looks great. cc ing for any last comments before merging: @s-urbaniak @alexsomesan @robszumski

ericchiang:
There are some concerns upstream with the current state of pod security policies as shipped in 1.8. If you have multiple policies matching the same request, the result is non-deterministic, which makes PSP really hard to use as you get into more complex cases.

This was enough for upstream to not enable PSPs for e2e testing and the example GCP scripts https://github.com/kubernetes/kubernetes/pull/52367#issuecomment-331135292

ericchiang:
As it stands, this PR won't let us restrict any pods created on behalf of a deployment, statefulset, replicaset, etc.

We're going to want to setup RBAC access to ensure that we limit certain pod security policies on the controller manager. The controller manager should only have the "use" verb for namespaces we want to be extremely permissive. A restricted rule needs to be the default.

see https://github.com/kubernetes/examples/tree/master/staging/podsecuritypolicy/rbac

For example, the following roles let a user use "permissive" or "restrictive" pod security policy. 

```
kind: ClusterRole
apiVersion: rbac.kubernetes.io/v1beta1
metadata:
  name: psp-permissive
rules:
- apiGroups: ["extensions"]
  resources: ["podsecuritypolicies"]
  resourceNames: ["permissive"]
  verbs: ["use"]
---
kind: ClusterRole
apiVersion: rbac.kubernetes.io/v1beta1
metadata:
  name: psp-restricted
rules:
- apiGroups: ["extensions"]
  resources: ["podsecuritypolicies"]
  resourceNames: ["restricted"]
  verbs: ["use"]
```

By default, we let the controller manager use the "restricted" PSP:

```
# Controller manager can use the restricted pod security policy in any namespace.
kind: ClusterRoleBinding
apiVersion: rbac.kubernetes.io/v1beta1
metadata:
  name: psp-restricted
subjects:
- kind: ServiceAccount
  namespace: kube-system
  name: controller-manager
roleRef:
  kind: ClusterRole
  name: psp-restricted
```

Then in privileged namespace ("kube-system", "tectonic-system") we let it use the permissive one:

```
# controller manager can use the permissive policy in kube-system
kind: RoleBinding
apiVersion: rbac.kubernetes.io/v1beta1
metadata:
  name: psp-permissive
  namespace: kube-system
subjects:
- kind: ServiceAccount
  namespace: kube-system
  name: controller-manager
roleRef:
  kind: ClusterRole
  name: psp-permissive
```

This is going to be problematic because we don't run the controller-manager with a unique service account, and because it's run as admin. The controller manager will be able to use any pod security policy we define, and as such we're not restricting anything.

joshrosso:
@ericchiang : Thanks for the feedback. As we discussed in our internal JIRA, this was an oversight on my part.

After some review of how the controller manager service account works, I've updated this PR with how I believe the RBAC structure and PSPs should look. I'm going to update the testing steps in our internal JIRA to reflect this.

When you have a moment, could you take a look at the updated commit and let me know your thoughts?

Thanks in advanced!

modules/bootkube/resources/manifests/02-permissive-policy.yaml
```diff
@@ -0,0 +1,156 @@
...
+  hostPorts:
+  - min: 0
+    max: 65535
+  volumes:
+  - '*'
+  allowedCapabilities:
+  - '*'
+
+---
+# Limited policy notably preventing privileged containers and containers wish to use host networking.
```
ericchiang: This policy isn't actually very limited. You can still mount arbitrary host volumes, for example.

modules/bootkube/resources/manifests/02-permissive-policy.yaml
```diff
@@ -0,0 +1,156 @@
...
+  allowedCapabilities: 
+  - '*'
+
+---
+# Role allowing the usage of full permissive PSP
+# Should be bound by a RoleBinding inside of kube-system and tectonic-system
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: permissive-psp-access
```
ericchiang: nit: maybe namespace them with `tectonic:psp-permissive`? It's clearer that this is a custom policy for tectonic and this will be easier to make sense of in an alphabetized list.

modules/bootkube/resources/manifests/kube-controller-manager.yaml
```diff
@@ -49,6 +49,7 @@ spec:
         - --node-monitor-grace-period=${node_monitor_grace_period}
         - --pod-eviction-timeout=${pod_eviction_timeout}
         - --cloud-provider=${cloud_provider}
+        - --use-service-account-credentials
```
ericchiang: :+1: that's the flag I was looking for.

s-urbaniak: just raising a small red-flag that this introduces manifest skew we need to fix in KVO

modules/bootkube/resources/manifests/kube-apiserver.yaml
```diff
@@ -41,7 +41,7 @@ spec:
         - --storage-backend=etcd3
         - --allow-privileged=true
         - --service-cluster-ip-range=${service_cidr}
-        - --admission-control=NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass,ResourceQuota
+        - --admission-control=NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass,ResourceQuota,PodSecurityPolicy
```
s-urbaniak: same as below, this introduces manifest skew we should fix in the KVO

modules/bootkube/resources/manifests/02-permissive-policy.yaml
```diff
@@ -0,0 +1,156 @@
...
+
+---
+# Allow kube-controller-manager to have permissive PSP access in kube-system namespace
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: permissive-psp-access
+  namespace: kube-system
+subjects:
+- kind: ServiceAccount
```
ericchiang: Hmmm, sucks that there's not a common group we can target here...

s-urbaniak:
Thanks a lot @joshrosso for the contribution! 👍 Nevertheless, as this introduces manifest skew, we should implement migration in KVO itself, hence set `do-not-merge` tentatively until we resolve this "upstream".

/cc @aaronlevy @diegs

ericchiang:
@joshrosso left a couple notes. Out of office today so I'll try to take a more in depth look tomorrow.

Might also want to take a look at the following PRs:

https://github.com/kubernetes/kubernetes/pull/52367 (PSP for GCP)
https://github.com/kubernetes/kubernetes/pull/54029 (Porting PSP fix to 1.8.3)

We should probably wait for 1.8.3 to make PSP a default. Also as the same with RBAC, we need to improve docs for users if we're going to force them to use PSP to use tectonic (I can help with that).

modules/bootkube/resources/manifests/02-permissive-policy.yaml
joshrosso: Agreed. We need to come up with _how_ restrictive this _default_ policy should be.

modules/bootkube/resources/manifests/02-permissive-policy.yaml
joshrosso: sgtm

joshrosso:
@s-urbaniak @ericchiang : Thanks for the feedback.

While determining how to handle this in the `kubernetes-version-operator` (KVO), i'll take on the following.
  
  - Prefix `tectonic:` to PSP names.
  - Come up with a better proposal for our _default_ limited PSP.

Regarding

> Also as the same with RBAC, we need to improve docs for users if we're going to force them to use PSP to use tectonic (I can help with that).

Sounds good. Docs **will be** a large portion of this work. I'd like to get commitment from product first that this *is* part of our roadmap before investing all the time in documenting PSP+RBAC in Tectonic. I've requested PM review in our internal JIRA.

joshrosso:
Additionally, after reviewing the GCP PR and docs for the `ResourceQuota` admission controller. It appears `PodSecurityPolicy` should come **before** `ResourceQuota.

As per the docs:

> It is strongly encouraged that this plug-in is configured last in the sequence of admission control plug-ins. This is so that quota is not prematurely incremented only for the request to be rejected later in admission control.

erasmus74:
Wondering if there has been anything more to track on this. If there's something I can contribute, let me know.

modules/bootkube/resources/manifests/02-permissive-policy.yaml
```diff
@@ -0,0 +1,158 @@
...
+  name: statefulset-controller
+  namespace: kube-system
+roleRef:
+  kind: ClusterRole
+  name: permissive-psp-access
+  apiGroup: rbac.authorization.k8s.io
+
+---
+# Allow kube-controller-manager to have limit PSP access in all namespaces
```
colemickens: This looks like the same bindings (kube-system) as the permissive grant just above. The subjects all still say `namespace: kube-system`.

Am I reading this incorrectly?

modules/bootkube/resources/manifests/02-permissive-policy.yaml
```diff
@@ -0,0 +1,158 @@
...
+  name: replicaset-controller
+  namespace: kube-system
+- kind: ServiceAccount
+  name: replication-controller
+  namespace: kube-system
+- kind: ServiceAccount
+  name: job-controller
+  namespace: kube-system
+- kind: ServiceAccount
+  name: daemon-set-controller
```
yifan-gu: @joshrosso Should this be daemonset-controller or not?

modules/bootkube/resources/manifests/02-permissive-policy.yaml
```diff
@@ -0,0 +1,158 @@
...
```
yifan-gu: same question here.

sym3tri:
@joshrosso thanks for the contribution. This is very useful for tracking purposes, but unfortunately we can't track this with the track-2 changes. I've created a Jira issue to reference this and we will provide docs on how to achieve this behavior going forward.

modules/bootkube/resources/manifests/02-permissive-policy.yaml
yifan-gu: Figured out, `daemon-set-controller` is correct... https://kubernetes.io/docs/admin/authorization/rbac/#controller-roles

modules/bootkube/resources/manifests/02-permissive-policy.yaml
yifan-gu: https://kubernetes.io/docs/admin/authorization/rbac/#controller-roles

PR #2103, opened 2017-10-12
aws: add s3 vpc endpoint resource, default enabled

colhom:
This should do great things for our download speeds and bandwidth charges for internet gateways and nat gateways.

An encouraging  preliminary result:

Here's rkt downloading the Kubelet container image in **under 10 seconds** from quay.io on an m3.medium worker node. 

This should greatly accelerate any download from s3 in general, outside of just container image pulls. 

```                                    
Oct 12 01:39:02 localhost kubelet-wrapper[808]: Added key for prefix "quay.io/coreos/hyperkube" at "/etc/rkt/trustedkeys/prefix.d/quay.io/coreos/hyperkube/bff313cdaa560b16a8987b8f72abf5f6799d33bc"               
Oct 12 01:39:02 localhost kubelet-wrapper[808]: Downloading signature:  0 B/473 B                        
Oct 12 01:39:02 localhost kubelet-wrapper[808]: Downloading signature:  473 B/473 B                      
Oct 12 01:39:03 ip-10-110-36-38 kubelet-wrapper[808]: Downloading ACI:  0 B/241 MB                       
Oct 12 01:39:03 ip-10-110-36-38 kubelet-wrapper[808]: Downloading ACI:  8.19 KB/241 MB                   
Oct 12 01:39:04 ip-10-110-36-38 kubelet-wrapper[808]: Downloading ACI:  87 MB/241 MB                     
Oct 12 01:39:05 ip-10-110-36-38 kubelet-wrapper[808]: Downloading ACI:  185 MB/241 MB                    
Oct 12 01:39:05 ip-10-110-36-38 kubelet-wrapper[808]: Downloading ACI:  241 MB/241 MB                    
Oct 12 01:39:10 ip-10-110-36-38 kubelet-wrapper[808]: image: signature verified: 
```
\cc @squat


platforms/aws/variables.tf
```diff
@@ -146,6 +146,29 @@ EOF
   default = []
 }
 
+variable "tectonic_aws_external_vpc_s3_endpoint" {
+  default = false
+
+  description = <<EOF
+(optional) Should the installer create a an s3 vpc endpoint within the existing VPC and associate
```
squat: s/create a an s3/create an S3/

platforms/aws/variables.tf
```diff
@@ -146,6 +146,29 @@ EOF
...
+regardless of the value set here.
+
+Example:
+
+  tectonic_aws_external_vpc_id          = "vpc-xxxxx"
+  tectonic_aws_external_vpc_s3_endpoint = true
+
+An S3 VPC endpoint allows your ec2 instances to communicate with the S3 service without leaving
+the AWS network. This will mean you do not incur bandwidth charges for Internet Gateways or NAT Gateways
+when downloading or uploading to S3. There is no additional cost associated with and S3 VPC endpoint.
```
squat: s/with and S3/with an S3/

platforms/aws/variables.tf
```diff
@@ -146,6 +146,29 @@ EOF
...
 }
 
+variable "tectonic_aws_external_vpc_s3_endpoint" {
+  default = false
+
+  description = <<EOF
+(optional) Should the installer create a an s3 vpc endpoint within the existing VPC and associate
+it with the route tables for the external master and worker subnets.
+
+If you are not using an existing VPC, the VPC created by the installer will have an S3 VPC endpoint
```
squat: Consider rephrasing to simply:
The installer will always create an S3 VPC endpoint for new VPCs.

platforms/aws/variables.tf
```diff
@@ -146,6 +146,29 @@ EOF
...
+regardless of the value set here.
+
+Example:
+
+  tectonic_aws_external_vpc_id          = "vpc-xxxxx"
+  tectonic_aws_external_vpc_s3_endpoint = true
+
+An S3 VPC endpoint allows your ec2 instances to communicate with the S3 service without leaving
```
squat: Consider rephrasing without "you"/"your":
S3 VPC endpoints allow EC2 instances to communicate with the S3 service without leaving the AWS network and without incurring bandwidth charges for Internet Gateways or NAT Gateways when uploading to or downloading from S3.

platforms/aws/main.tf
```diff
@@ -57,6 +57,10 @@ module "vpc" {
     join("|", keys(var.tectonic_aws_worker_custom_subnets)) :
     join("|", data.aws_availability_zones.azs.names)
   )}"
+  # if new vpc- always create endpoint
```
squat: nit: add space between "vpc" and "-"

modules/aws/vpc/vpc.tf
```diff
@@ -23,3 +27,14 @@ data "aws_vpc" "cluster_vpc" {
   #
   id = "${var.external_vpc_id == "" ? join(" ", aws_vpc.new_vpc.*.id) : var.external_vpc_id }"
 }
+
+resource "aws_vpc_endpoint" "s3_vpc_endpoint" {
+  count        = "${var.create_s3_endpoint ? 1 : 0}"
+  vpc_id       = "${data.aws_vpc.cluster_vpc.id}"
+  service_name = "${format("com.amazonaws.%s.s3",data.aws_region.current.name)}"
+
+  route_table_ids = ["${concat(
```
squat: Terraform lists are finicky. This may in some conditions need a `compact` around the `concat`

squat:
ok to test

colhom:
Seems the test failed due to the coreos stable release changing while the AWS smoke test ran:
```
Failures:

  1) aws-standard installs the correct Container Linux version
     Failure/Error: expect(ContainerLinux.version(@cluster)).to eq(version)

       expected: "1465.6.0"
            got: "1520.6.0"
```
How should I go about kicking off a re-test?

## conditional splat references to resource lists / data lists == bad news. 
 
Also- the vpc module had some structuring issues around dealing with existing vs. new VPC that was making it difficult for me to reliably reference route tables attached to subnets in a configuration-agnostic way. 

The pattern of conditionally reading from `data.aws_subnet.external_worker/master` sources **or** the `aws_subnet.master/worker_subnet` resources was one of the main culprits in causing weird hiccups w/ how terraform computed variables. 

I'm fairly certain conditionally referencing list splats for data sources or resources is universally a bad idea- any conditionals should operate on `_id` string lists which then are feed into the data source metadata unconditionally. 

Given this, it's also nice that terraform doesn't allow conditionals for expressions that evaluate to a list, because assuming you also stick to the "no conditional references to data source or resource lists" paradigm- things should stay pretty sane.

As a general rule- you should instead trust resource/data list counts and contents to be correct and do strictly non-conditional array operations like `coalescelist()`, `concat()`, `compact()`, etc when referencing resource/data arrays (`data.aws_subnet.worker.*.id`). This goes inductively as well for extending the data reference chain. (data.subnet.worker.*.id --> data.route_table.worker.*.id --> data.route_table_entry.worker.*.id)

So- I think I've got this general problem in the `aws/vpc` module ironed out by merging the `vpc_id` and `subnet_id` string arrays, non-conditionally, in local variables before passing along (again non-conditionally) to the `data.aws_subnet.master/worker` sources, which can then be used unconditionally as well. The only conditional references allowed are to the local count variables for these lists, which are defined canonically in locals blocks.

## subnet creation logic does not allow for set of master subnets/AZs changing, results in subnet cidr conflicts on update.
 
On another note, there was also a pretty annoying bug with how subnet cidr offsets were being computed, which I've filed in Jira and is fixed in the first commit here as well. https://coreosdev.atlassian.net/browse/INST-539

I'm wasn't planning filing the conditional data source access causing weird dependency timing issues as an issue- I think that's just part of "living with terraform" we should just avoid doing that going forward , fix what we have now and call it a day.

## changes to s3 vpc endpoint functionality

Now that the generic route table selection and subnet selection works reliably for any configuration and the state variables are consolidated and consolidated- I have gone ahead and made the s3-vpc-endpoint default for both existing and new VPC cases, with single boolean option to explicitly disable (default to false).

My rationale is that there can be many s3 vpc endpoints per vpc, and they don't cost money so there is little harm in creating one per cluster as long as clusters have different route tables, even if the clusters share a VPC. 

The upside is somewhere from 2-10x bandwidth to s3  and complete elimination of gateway bandwidth charges for and s3 traffic, so definitely worth a small subset customers potentially seeing IAM "permission denied" errors on the first apply. Failure to create the S3 endpoint or partial creation does not impair cluster functionality, and no other components depend on it.

In the case that there are not permissions or the route table already has an s3 vpc endpoint, the user should get a pretty obvious error and have enough of a hint to look through the tfvars and flip the boolean,  and then re-apply with no problems.

\cc @squat

squat:
ok to test

squat:
That failure was due to a race that has been eliminated since you first made the PR. If we see that again I'll ask for you to rebase.

cpanato:
restest this please

modules/aws/vpc/outputs.tf
```diff
@@ -2,14 +2,12 @@ output "vpc_id" {
   value = "${data.aws_vpc.cluster_vpc.id}"
 }
 
-# We have to do this join() & split() 'trick' because null_data_source and
-# the ternary operator can't output lists or maps
 output "master_subnet_ids" {
-  value = ["${split(",", var.external_vpc_id == "" ? join(",", aws_subnet.master_subnet.*.id) :  join(",", data.aws_subnet.external_master.*.id))}"]
+  value = ["${data.aws_subnet.master.*.id}"]
```
squat: this output is now distinctly different. Shouldn't this now be `["${local.master_subnet_ids}"]`?

modules/aws/vpc/outputs.tf
```diff
@@ -2,14 +2,12 @@ output "vpc_id" {
...
 }
 
 output "worker_subnet_ids" {
-  value = ["${split(",", var.external_vpc_id == "" ? join(",", aws_subnet.worker_subnet.*.id) :  join(",", data.aws_subnet.external_worker.*.id))}"]
+  value = ["${data.aws_subnet.worker.*.id}"]
```
squat: same as above s/master/worker/

squat:
cc @alexsomesan @s-urbaniak for extra eyes on the terraform

sym3tri:
@colhom thanks for this, but it's gotten stale. If you'd like to get this in can you please rebase and address the comments? Otherwise we've added to our backlog and will revisit when we have time.

PR #2122, opened 2017-10-13
frontend: Upgrade some dependencies

kyoto:
Upgrade Jest, immutable, eslint and eslint-plugin-react. Includes Jest and immutable switching to the MIT license.

Also add eslint rule to disallow direct mutation of a React component's state.


squat:
ok to test

PR #2131, opened 2017-10-14
examples/terraform.tfvars.azure

tylerauerbeck:
examples/terraform.tfvars.azure: remove aws variable

Removed aws variables inside of azure as example as this is confusing/misleading


squat:
@tylerauerbeck, that file is auto-generated so that AWS-specific documentation will keep coming back unless the source is changed. It looks like some AWS-specific variable has creeped into the config.tf file. We should move that to platforms/aws/variables.tf.

tylerauerbeck:
@squat Ahh, you're right. It appears in all of the other examples as well. Looks like the issue lies on lines 126-136 in config.tf. Not sure if it's appropriate to make the changes in this same PR or if it's best to close this one and open up a new one. Regardless, I've made the changes noted above and moved them to platforms/aws/variables.tf. If it's better to submit this as a new PR let me know.

squat:
@tylerauerbeck thanks for this! This looks good. We just need to clean up the git commits a little bit; please
* remove the first commit;
* squash the second and third commits; and
* run `make docs examples` to generate the .tfvars files and commit the result

Once that's ready, tests will pass and we can merge this contribution.

tylerauerbeck:
@squat Alright I think everything _should_ be in order now. Let me know if there's anything else that needs to be done.

squat:
ok to test

squat:
ok to test

squat:
Note for us: looks like two test flakes due to the rkt --dns=host issue. Retesting

squat:
Looks good! Merging.

PR #2132, opened 2017-10-16
frontend: Unify the AWS and BM cluster name inputs

kyoto:
Combine Cluster Name input code for AWS and bare metal.

This changes the AWS UI to match the bare metal UI (the bare metal UI is unchanged).

Before | After
--- | ---
screenshot-3 | screenshot-2

Rename `TectonicLicense` to `ClusterInfo` since it now includes the cluster name and rename the file too.

Remove `BM_ClusterInfo` since it would otherwise just wrap `ClusterInfo`.


kyoto:
cc @robszumski @tlwu2013

PR #2133, opened 2017-10-16
frontend: Clean up AWS regions

kyoto:
We only actually use the region names from the "amis" API call result, so we can simplify the code a bit by storing just the names in redux.


kyoto:
ok to test

kyoto:
ok to test

PR #2138, opened 2017-10-16
frontend: Remove "undefined" CSS class

kyoto:
This removes the CSS class `undefined` that was being output for certificate and private key inputs and also cleans up the code a bit.


kyoto:
ok to test

PR #2142, opened 2017-10-16
tests: Bump wait_for_bootstrapping timeout

mxinden:
The bootkube and tectonic systemd services can take up to 10 minutes to
start. This patch bumps the retry value to prevent timeouts in our CI.

@squat We have seen multiple timeouts like [this](https://jenkins-tectonic-installer.prod.coreos.systems/blue/organizations/jenkins/tectonic-installer/detail/PR-2137/10/pipeline/). I guess the machines are restarting during `wait_for_bootstrapping`. Bumping the timeout might be a quick fix. What do you think?

### DIFF
--- a/tests/rspec/lib/cluster.rb
+++ b/tests/rspec/lib/cluster.rb
@@ -144,8 +144,8 @@ class Cluster
   def wait_for_bootstrapping
     from = Time.now
     loop do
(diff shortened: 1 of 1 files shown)


mxinden:
I will merge here. It seems to fix the issue. @squat let me know if you can think of a better solution. Happy to revert at a later point in time.

PR #2143, opened 2017-10-16
tests: Remove PHONY flag from bin/smoke Make target

mxinden:
`/bin/smoke` represents an actual file which is compiled with this
recipe. Removing the `PHONY` flag enables Make to decide on its own when
to rebuild. If the source files are less recent than the compiled
binary, there is no need to rebuild the binary, thereby saving time for
local and CI test execution.

@cpanato @squat What do you think?

See
https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html


squat:
LGTM. Looks like it was unnecessarily added in https://github.com/coreos/tectonic-installer/pull/1386

mxinden:
ok to test

PR #2145, opened 2017-10-16
CORS-326 update cluster config names to contain version

thorfour:


thorfour:
@diegs fixed

rithujohn191:
Should we rename the filenames as well to contain the version?

diegs:
@rithujohn191 personally I'd keep the filenames constant so we don't have to worry about updating other things like the tectonic.sh script that installs the config.

PR #2147, opened 2017-10-16
modules/aws/vpc - Better security for master nodes

mrwacky42:
For a VPC with master nodes in a public subnet, they are reachable
from the entire internet. Usually, one will communicate with them via
an ELB, and not directly. This change limits access to 80/443 TCP and
ICMP to only nodes within the VPC and cluster.


squat:
@mrwacky42 IMO this is actually not a good idea. Connections via an ELB are understandably loadbalanced across all the pool nodes. Thus, SSHing through a loadbalancer typically results in broken SSH connections after a few minutes. You should always SSH either:
1. directly into a master node; or
2. via a designated, non-loadbalanced jumpbox/bastion host.

We intend to add support for the second option very soon to address your concerns.

Closing for now.

mrwacky42:
@squat  - I think you misunderstand the purpose of this PR, or I misunderstand your response.

Currently, any IP on the entire internet can connect to the master public IP addresses. We see lots of attempts at SSH brute forcing, and I can assume others are attempting to attack the API servers on port 80/443.

As far as I can tell, nothing external to the cluster talks directly to port 80/443 on the master nodes. This change limits the security group to only the VPC CIDR network. This grants access to:
* AWS for the purpose of load balancing
* Things inside the cluster

We have configured our test cluster this way to reduce our attack surface.

squat:
Ah sorry rereading this now I see that I misunderstood this PR as limiting SSH, rather than http/https. Sure this could definitely make sense. However, the ELB forwards all requests to the API servers, so how much better is this really?

mrwacky42:
It's better because we limit access to the API with security groups attached to the ELB!
😀

mxinden:

mrwacky42:
*rebased*

enxebre:
retest this please

enxebre:
retest this please

enxebre:
retest this please

PR #2157, opened 2017-10-17
tests: Log network interfaces before and after destroy

mxinden:
For better upstream Terraform debugging we would like to log the network
interfaces related to the VPC. For further details see:
https://github.com/terraform-providers/terraform-provider-aws/pull/1051

@erikkn This change will add something along the lines of this to the test log:
``` json
{
    "NetworkInterfaces": 
           ...
}
```
The goal would be to be able to analyse this with the Elastic search project. Should I log any particular keyword before or after so we can extract it from the logs?

### DIFF
--- a/tests/rspec/lib/aws_cluster.rb
+++ b/tests/rspec/lib/aws_cluster.rb
@@ -8,6 +8,7 @@ require 'grafiti'
 require 'env_var'
 require 'aws_iam'
 require 'aws_support'
+require 'tfstate_file'
 
 # AWSCluster represents a k8s cluster on AWS cloud provider
 class AwsCluster < Cluster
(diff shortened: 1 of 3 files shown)


tests/rspec/lib/aws_cluster.rb
```diff
@@ -107,4 +100,23 @@ def tectonic_console_url
...
+  private
+
+  def destroy
+    # For debugging purposes (see: https://github.com/terraform-providers/terraform-provider-aws/pull/1051)
+    describe_network_interfaces
+    super
+    describe_network_interfaces
+  end
+
+  def describe_network_interfaces
```
alexsomesan: I'm afraid such debug sections will be forgotten around the code once the issues get fixed.
Do you think it makes sense to gate these sections with an environment variable flag so that they only run while that var is set? That way we can also grep the tests code for that var name to discover any stale debug statements.

mxinden:
It fails on the second go, as in the best case there are no network interfaces no more. I will look into it.

tests/rspec/lib/aws_cluster.rb
mxinden: @alexsomesan Would a comment with a keyword like `DEBUG` be enough? This way we can grep for it in the code base.

squat:
@alexsomesan @mxinden is this ready?

mxinden:
@squat Added one small refactoring. I will request for review once ready. 

@alexsomesan Any opinion on https://github.com/coreos/tectonic-installer/pull/2157#discussion_r146851825 ?

mxinden:
@squat @alexsomesan Could you give this another review?

s-urbaniak:
@mxinden taking over, as INST-288 has been priorized

mxinden:
ok to test

mxinden:
@s-urbaniak @alexsomesan Tests are all green. Can you approve once again, so I can merge?

PR #2161, opened 2017-10-17
frontend: Remove unused UPDATER field

kyoto:
Not sure what this was being used for originally, but it appears to no longer be used.


kyoto:
ok to test

kyoto:
ok to test

PR #2166, opened 2017-10-17
frontend: Remove most DEFAULT_CLUSTER_CONFIG initial entries

kyoto:
The `Field` constructor adds the field's default value to `DEFAULT_CLUSTER_CONFIG`, so we don't need to also define those defaults when `DEFAULT_CLUSTER_CONFIG` is initialized.

Some fields remain for several different reasons. They should be refactored away over time.


kyoto:
ok to test

kyoto:
ok to test

PR #2176, opened 2017-10-18
tests: Improve wait_for_bootstrapping

mxinden:
The Tectonic installer CI had repeatedly failed with timeouts waiting
for the bootkube and tectonic service to finish.

This patch includes the following improvements:
- The bootkube and tectonic systemd unit is only deployed on one master
node. Instead of relying on the fact, that we choose the right master
node to check if the services finished, we check all of them and return
true if the services succeeded on one of them
- Instead of checking for success via:
`systemctl is-active <service>`, we check whether the
`init_<service>.done` file exists. `is-active` returns non zero when the
bootstrap succeeded but the node restarted in the meantime. The
`init_<service>.done` file is only written when the service exits with a
non-zero exit code [1]

[1] See `ExecStartPost`
https://www.freedesktop.org/software/systemd/man/systemd.service.html

### DIFF
--- a/tests/rspec/lib/azure_cluster.rb
+++ b/tests/rspec/lib/azure_cluster.rb
@@ -2,6 +2,7 @@
 
 require 'cluster'
 require 'azure_support'
+require 'json'
 
 # AzureCluster represents a k8s cluster on Azure cloud provider
 #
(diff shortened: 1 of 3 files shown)


tests/rspec/lib/azure_cluster.rb
```diff
@@ -14,12 +15,17 @@ def initialize(tfvars_file)
     super(tfvars_file)
   end
 
-  def master_ip_address
+  def master_ip_addresses
```
alexsomesan: This method no longer returns master IP addresses so the name is not representative anymore.
We should rename it to be descriptive of its current role.

mxinden:
Will take care of @alexsomesan  comment in a follow up PR as discussed offline.

PR #2182, opened 2017-10-19
frontend: Migrate CIDR inputs to Connect

kyoto:
**PR #2166 should be merged first.**

This PR also removes the `BM_NetworkConfig` component since it is now the same as `KubernetesCIDRs`.

Also fixes the validation for the "CIDR block" filed and the Masters and Workers CIDR fields to use `validate.AWSsubnetCIDR` instead of `validate.CIDR`.


squat:
@kyoto let's rebase now that #2166 is merged

kyoto:
@squat Rebased

PR #2185, opened 2017-10-19
Use filepath on user provided ingress

alexouzounis:
Instead of having to append the contents of the certificates, use file paths

A separate PR would be required to updates the docs; https://github.com/coreos/tectonic-docs/blob/master/Documentation/reference/tls-certificates.md

### DIFF
--- a/platforms/aws/tectonic.tf
+++ b/platforms/aws/tectonic.tf
@@ -33,6 +33,14 @@ module "ingress_certs" {
   ca_key_pem   = "${module.kube_certs.ca_key_pem}"
 }
 
+//module "ingress_certs" {
+//  source = "../../modules/tls/ingress/user-provided"
+//
+//  ca_cert_pem = "${file(var.tectonic_ingress_ca_cert_pem_path)}"
+//  cert_pem    = "${file(var.tectonic_ingress_cert_pem_path)}"
+//  key_pem     = "${file(var.tectonic_ingress_key_pem_path)}"
+//}
+
 module "identity_certs" {
   source = "../../modules/tls/identity/self-signed"
 
--- a/platforms/aws/variables.tf
+++ b/platforms/aws/variables.tf
@@ -316,39 +316,3 @@ Example:
  * `["ingress-nginx"]`
 EOF
 }
-
-variable "tectonic_ingress_ca_cert_pem_path" {
-  type  = "string"
-  default = ""
-
-  description = <<EOF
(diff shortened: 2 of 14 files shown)


alexouzounis:
I tried to do something smart with count or interpolation to have one module but the source filed cannot be interpolated in TF. If anyone has any other ideas, please let me know. 

The solution I see is to merge the `modules/tls/ingress/self-signed` and `modules/tls/ingress/user-provided`

platforms/aws/variables.tf
```diff
@@ -316,3 +316,39 @@ Example:
  * `["ingress-nginx"]`
 EOF
 }
+
+variable "tectonic_ingress_ca_cert_pem_path" {
+  type  = "string"
+  default = ""
+
+  description = <<EOF
+(optional) File with contents the public CA certificate in PEM format
```
squat: These comments are a little confusing. They say to provide a path but then show the contents of the file. Please make these comments match the descriptions for the other PEM path variables, e.g. https://github.com/coreos/tectonic-installer/blob/master/config.tf#L202

platforms/aws/variables.tf
```diff
@@ -316,3 +316,39 @@ Example:
  * `["ingress-nginx"]`
 EOF
 }
+
+variable "tectonic_ingress_ca_cert_pem_path" {
```
squat: These variables are not specific to AWS. Please move them out of the AWS-specific variables.tf and into config.tf.

platforms/aws/tectonic.tf
```diff
@@ -33,6 +33,14 @@ module "ingress_certs" {
   ca_key_pem   = "${module.kube_certs.ca_key_pem}"
 }
 
+//module "ingress_certs" {
```
squat: Since this is not AWS-specific, please add this into the Azure, Openstack, bare-metal, and GCP Terraform as well.

alexouzounis:
Sure, yeah. Thanks for having a look so quickly. Will come back to you end of next week.

alexouzounis:
Hey @squat - let me know what you think about the latest changes when you get some time :)

mxinden:

enxebre:
hey @alexouzounis thanks a lot for this contribution. Part of the reasoning behind embedding was ease of content sharing within the tf space. Let's close this for now as we are currently under considerably refactor of master and the way we generate/consume certificates will be impacted as well

PR #2187, opened 2017-10-19
azure/vnet: custom vnet error when using module

rossedman:
Prevents issue where when providing an external vnet id errors are thrown


rossedman:
If providing custom VNet ID this error is provided:

```
* module.kubernetes.module.vnet.azurerm_virtual_network.tectonic_vnet: azurerm_virtual_network.tectonic_vnet: value of 'count' cannot be computed
```

squat:
ok to test

squat:
@rossedman does this really fix the error?

squat:
Regardless, it is a good change to get in.

rossedman:
@squat blah didn't fix. will look into it more. 😢

squat:
That’s unfortunately what I expected. We’ll have to dig into this

PR #2190, opened 2017-10-20
frontend: Fix Tectonic license and pull secret errors to be strings

kyoto:
These can't be React elements because they get written to state, which meant the whole React object would be dumped into the .progress file if these fields failed validation.


kyoto:
ok to test

kyoto:
@squat Rebased

PR #2192, opened 2017-10-20
tests: do not check the host key when ssh 

cpanato:
sometimes we receive the same IP for the machines we created and if it runs in the same slave if can throw this exception: Net::SSH::HostKeyMismatch to avoid that we are not checking the key
 
solution: clean the know_hosts in the end of the test

Jira: https://jira.prod.coreos.systems/browse/INST-552

### DIFF
--- a/tests/rspec/lib/ssh.rb
+++ b/tests/rspec/lib/ssh.rb
@@ -18,7 +18,8 @@ def ssh_exec(ip_address, command, max_retries = 5)
   stdout = ''
   stderr = ''
   begin
-    Net::SSH.start(ip_address, 'core', forward_agent: true, use_agent: true) do |ssh|
+    Net::SSH.start(ip_address, 'core', forward_agent: true, use_agent: true,
+                                       verify_host_key: Net::SSH::Verifiers::Null.new) do |ssh|
       ssh.exec! command, status: status do |_ch, stream, data|
         if stream == :stdout
           stdout = data


cpanato:
ok to test

cpanato:
retest this please

squat:
what about just deleting the known_hosts file entirely? We should also consider enabling SSH options during the execs so that the keys are not added to the file in the first place.

cpanato:
@squat did not apply that because other tests execution from other tests might be running in the same slave and I don't know what will be the impact if you delete the file. If there is no impact I will update the code to delete the whole file

mxinden:
Is this happening on all platforms or only on bare metal?

On AWS and Azure we are running the smoke tests inside a docker container. As far as I can tell the `known_hosts` file is not mounted from the host into the container, so it should be reset on every test execution anyways, right?

In case this is correct, can you add this only to the `metal-cluster.rb` file?

Example docker run command:
```
docker run -t -d -u 1000:1000 -u root -w /home/ubuntu/workspace/tectonic-installer_PR-2192-SHRMFZGCYZED3B6GQUA5SISU6UDIS2GMBJJZDJOCXOHE2MT2MUCA -v /home/ubuntu/workspace/tectonic-installer_PR-2192-SHRMFZGCYZED3B6GQUA5SISU6UDIS2GMBJJZDJOCXOHE2MT2MUCA:/home/ubuntu/workspace/tectonic-installer_PR-2192-SHRMFZGCYZED3B6GQUA5SISU6UDIS2GMBJJZDJOCXOHE2MT2MUCA:rw,z -v /home/ubuntu/workspace/tectonic-installer_PR-2192-SHRMFZGCYZED3B6GQUA5SISU6UDIS2GMBJJZDJOCXOHE2MT2MUCA@tmp:/home/ubuntu/workspace/tectonic-installer_PR-2192-SHRMFZGCYZED3B6GQUA5SISU6UDIS2GMBJJZDJOCXOHE2MT2MUCA@tmp:rw,z  --entrypoint cat quay.io/coreos/tectonic-smoke-test-env:v5.7
```

cpanato:
@mxinden I saw happening in Azure

cpanato:
retest this please

PR #2195, opened 2017-10-20
Removed the CriticalAddOn toleration as it is not used

thorfour:
Removing the CriticalAddonsOnly toleration as it isn't being used.

### DIFF
--- a/modules/bootkube/resources/manifests/kube-dns.yaml
+++ b/modules/bootkube/resources/manifests/kube-dns.yaml
@@ -43,8 +43,6 @@ spec:
     metadata:
       labels:
         k8s-app: kube-dns
-      annotations:
-        scheduler.alpha.kubernetes.io/critical-pod: ''
     spec:
       containers:
       - name: kubedns
@@ -162,8 +160,6 @@ spec:
             cpu: 10m
       dnsPolicy: Default  # Don't use cluster DNS.
       tolerations:
-      - key: "CriticalAddonsOnly"
-        operator: "Exists"
       - key: "node-role.kubernetes.io/master"
         operator: "Exists"
         effect: "NoSchedule"
--- a/modules/net/canal/resources/manifests/kube-calico.yaml
+++ b/modules/net/canal/resources/manifests/kube-calico.yaml
@@ -125,12 +125,6 @@ spec:
     metadata:
       labels:
         k8s-app: kube-calico
-      annotations:
-        # This, along with the CriticalAddonsOnly toleration below,
-        # marks the pod as a critical add-on, ensuring it gets
-        # priority scheduling and that its resources are reserved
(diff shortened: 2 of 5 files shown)


modules/net/canal/resources/manifests/kube-calico.yaml
```diff
@@ -126,8 +126,7 @@ spec:
       labels:
         k8s-app: kube-calico
       annotations:
-        # This, along with the CriticalAddonsOnly toleration below,
-        # marks the pod as a critical add-on, ensuring it gets
+        # This marks the pod as a critical add-on, ensuring it gets
         # priority scheduling and that its resources are reserved
         # if it ever gets evicted.
         scheduler.alpha.kubernetes.io/critical-pod: ''
```
squat: @thorfour if the pod has the critical-pod annotation then it is required to have the criticaladdonsonly toleration to be guaranteed space. Without the toleration:
1. the node will get the CriticalAddonsOnly taint
2. the scheduler will try to schedule the kube-calico pod but since it doesn't tolerate the correct taint, it will not be schedulable
3. so the rescheduler will keep evicting pods from the nodes until the kube-calico pod can be scheduled (which it will never be)
4. go to 2

TL;DR: if we are removing criticaladdonsonly tolerations then we must remove this annotation as well.

More info: https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/

modules/tectonic/resources/manifests/heapster/deployment.yaml
```diff
@@ -58,6 +58,3 @@ spec:
             requests:
               cpu: 50m
               memory: 90Mi
-      tolerations:
```
squat: we need to remove the annotation in this manifest as well.

modules/bootkube/resources/manifests/kube-dns.yaml
```diff
@@ -162,8 +162,6 @@ spec:
             cpu: 10m
       dnsPolicy: Default  # Don't use cluster DNS.
       tolerations:
-      - key: "CriticalAddonsOnly"
```
squat: we need to remove the annotation in this manifest as well.

modules/net/canal/resources/manifests/kube-calico.yaml
thorfour: Good to know, thanks for that information. I've removed the remaining annotations.

yifan-gu:
We want to get this in for 179 right? Cuz I saw all other critical pod annotations are removed in 179.

squat:
ok to test

yifan-gu:
PR merged, I also rebased 1.7.9 branch to master to pick up this one.

PR #2199, opened 2017-10-23
frontend: Fix Cluster Info page's next button after Start Over

kyoto:
Fixes the Cluster Info page's next button enabled / disabled state.

The problem only occurred if you completed the install, then clicked Start Over and went through the wizard again. In that case, the Next Step button would stay disabled even after filling in the name field. You needed to upload the license and pull secret again to enable the button.


kyoto:
ok to test

PR #2206, opened 2017-10-24
frontend: Remove 3 unused "validator" attributes

kyoto:
With the `Connect` component, `validator` is specified when creating the `Field`, not passed as an attribute to `Connect`.


squat:
ok to test

PR #2207, opened 2017-10-24
tests: enable TF_LOG when retrying apply or destroy

cpanato:
JIRA: https://jira.prod.coreos.systems/browse/INST-231

### DIFF
--- a/tests/rspec/lib/cluster.rb
+++ b/tests/rspec/lib/cluster.rb
@@ -99,15 +99,19 @@ class Cluster
   end
 
   def apply
-    3.times do
-      return if system(env_variables, 'make -C ../.. apply')
+    3.times do |idx|
+      env = env_variables
+      env['TF_LOG'] = 'TRACE' if idx.positive?
+      return true if system(env, 'make -C ../.. apply')
     end
     raise 'Applying cluster failed'
   end
 
   def destroy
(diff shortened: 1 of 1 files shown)


tests/rspec/lib/cluster.rb
```diff
@@ -101,13 +101,15 @@ def localconfig
   def apply
     3.times do
       return if system(env_variables, 'make -C ../.. apply')
+      env_variables.merge('TF_LOG' => 'DEBUG') unless ENV.key?('TF_LOG')
```
alexsomesan: In Ruby, the `merge()` method of a Hash doesn't modify that actual Hash instance. Rather, it returns a new copy of a hash, with the contents of the argument hash merged into the old one.
If you want to modify the env_variables in place, you need to use the `merge!()`(with an exclamation mark) method instead. 
Documented here: https://ruby-doc.org/core-2.4.1/Hash.html#method-i-merge

Also, I think the check after unless might not be necessary since the subsequent `merge!` would just overwrite the value of TF_LOG in place in the hash.

cpanato:
@s-urbaniak @alexsomesan did a small refactor and changed TF_LOG to TRACE per @alexsomesan recommendation

alexsomesan:
retest this please

PR #2208, opened 2017-10-24
Terraform providers: bump versions to latests releases

alexsomesan:
We were lagging quite a few versions behind on some of the providers.
This means we're not incorporating potentially critical fixes from upstream. 
This is especially relevant in the light of the recently occurring Azure DNS API errors.

/cc: @sym3tri

### DIFF
--- a/config.tf
+++ b/config.tf
@@ -3,7 +3,7 @@ terraform {
 }
 
 provider "archive" {
-  version = "0.1.0"
+  version = "1.0.0"
 }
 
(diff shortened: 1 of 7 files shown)


alexsomesan:
@squat I didn't check, but I have the same hunch. Some of 1.0.0s seem ceremonial, to mark the transition to standalone providers, I guess.

squat:
Very good catch!

PR #2215, opened 2017-10-25
frontend: Improve bare metal Matchbox Address page spinner

kyoto:
Removes `bootCfgInfly` hack.

Clears `osToUse` (`BM_OS_TO_USE`) and `matchboxHTTP` at the same time, which I think is better to ensure they are consistent.

This causes the spinner to start when you being typing rather than when the API call actually starts, which I think is better since otherwise the call can be so fast that the spinner doesn't visibly animate, giving no visual feedback.


kyoto:
ok to test

PR #2219, opened 2017-10-25
tests: using tf_value and new function tf_state to avoid duplication

cpanato:
cc @alexsomesan @mxinden

### DIFF
--- a/tests/rspec/lib/aws_cluster.rb
+++ b/tests/rspec/lib/aws_cluster.rb
@@ -47,14 +47,12 @@ class AwsCluster < Cluster
 
   def master_ip_addresses
     ssh_master_ips = []
-    Dir.chdir(@build_path) do
-      terraform_state = `terraform state show module.masters.aws_autoscaling_group.masters`.chomp.split("\n")
-      terraform_state.each do |value|
-        attributes = value.split('=')
-        next unless attributes[0].strip.eql?('id')
-        instances_id = AwsSupport.sorted_auto_scaling_instances(attributes[1].strip.chomp, @aws_region)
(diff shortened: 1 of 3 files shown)


tests/rspec/lib/aws_cluster.rb
```diff
@@ -47,14 +47,12 @@ def stop
...
-    Dir.chdir(@build_path) do
-      terraform_state = `terraform state show module.masters.aws_autoscaling_group.masters`.chomp.split("\n")
-      terraform_state.each do |value|
-        attributes = value.split('=')
-        next unless attributes[0].strip.eql?('id')
-        instances_id = AwsSupport.sorted_auto_scaling_instances(attributes[1].strip.chomp, @aws_region)
-        ssh_master_ips.push AwsSupport.preferred_instance_ip_address(instances_id[0], @aws_region)
-      end
+    terraform_state = tf_state('module.masters.aws_autoscaling_group.masters').split("\n")
+    terraform_state.each do |value|
```
mxinden: @cpanato I have build similar logic in #2157 (https://github.com/coreos/tectonic-installer/pull/2157/files#diff-783c678aa2ad2e25fb996138dce80d93R4) in a separate `TFStateFile` class. If you don't mind, I will finish up the PR, get it merged and then you can use it here. What do you think?

cpanato: ok

cpanato: will close this one

mxinden: Sorry, I did a bad job at explaining my changes. They will only introduce the `TFStateFile` class, which your changes could use, they will not do this clean-up.

I would still very much like to get your clean-up changes merged! Let's adjust this PR once #2157 is merged.

PR #2220, opened 2017-10-25
test: use popen3 to run shell commands

cpanato:


tests/rspec/lib/metal_support.rb
```diff
@@ -128,4 +156,16 @@ def self.env_variables_setup
...
   end
+
+  def self.execute_command(cmd)
+    Open3.popen3(cmd) do |_stdin, stdout, _stderr, wait_thr|
+      exit_status = wait_thr.value
+      unless exit_status.success?
+        while (line = stdout.gets)
+          puts line
+        end
+        raise "Comand execution FAILED !!! #{cmd}"
```
mxinden: s/comand/command

cpanato: thanks!

cpanato:
@mxinden you can have another look when you have time 😄

tests/rspec/lib/metal_support.rb
```diff
@@ -24,13 +25,13 @@
...
-    `sudo mv /tmp/terraform /usr/local/bin/`
-    `sudo chmod +x /usr/local/bin/terraform`
-    `(cd #{root}/ && rm -rf matchbox && git clone https://github.com/coreos/matchbox)`
-    `(cd #{root}/matchbox && git checkout #{MATCHBOX_VERSION})`
+    execute_command("sudo curl -L -o /usr/local/bin/kubectl #{KUBECTL_URL}")
+    execute_command('sudo chmod +x /usr/local/bin/kubectl')
+    execute_command("curl #{TERRAFORM_URL} | funzip > /tmp/terraform")
+    execute_command('sudo mv /tmp/terraform /usr/local/bin/')
+    execute_command('sudo chmod +x /usr/local/bin/terraform')
+    execute_command("(cd #{root}/ && rm -rf matchbox && git clone https://github.com/coreos/matchbox)")
```
mxinden: Are the brackets inside the `"` needed?

tests/rspec/lib/metal_support.rb
```diff
@@ -128,4 +156,16 @@ def self.env_variables_setup
       'ASSETS_DIR' => '/tmp/matchbox/assets'
     }
   end
+
+  def self.execute_command(cmd)
+    Open3.popen3(cmd) do |_stdin, stdout, _stderr, wait_thr|
+      exit_status = wait_thr.value
+      unless exit_status.success?
```
mxinden: If the exit status is not `0`, why not print out `stderr` as well?

tests/rspec/lib/metal_support.rb
cpanato: maybe not

tests/rspec/lib/metal_support.rb
cpanato: will remove

mxinden: @cpanato "Remove"? I though about adding something like:

```Ruby
while (line = stderr.gets)
         puts line
end
```

What do you think? Am I missing something here?

cpanato: ahh, got it, ok

tests/rspec/lib/metal_support.rb
cpanato: done

tests/rspec/lib/metal_support.rb
cpanato: done

PR #2221, opened 2017-10-25
Build system using Google's Bazel

alexsomesan:
With this change, we aim to adopt the Bazel build system as the de-facto build tool for the project.
We want the development and testing processes to test against the exact set of artefacts that users will ultimately see and use.

As we do this, we will transition to running the installer tool exclusively from the built output rather than directly  out of the checked-out repository. This will remove skew between the experience of a user running the installer and the one of a developer on the project.

The transition to the new build system will observe the following milestones:

- [x] Initial configuration files for WORKSPACE and top-level BUILD file
- [x] Readme on how to get Bazel and run the main build process
- [x] Configure Jenkins to run the build process and run tests against the built artefacts
- [x] Copy the Terraform templates to build output
- [x] Download Terraform runtime binary and copy it to output
- [x] Download Matchbox provider plugin and copy it to output
- [ ] Build the frontend JS code with Bazel
- [x] Build the Go backend code with Bazel and release it to output
- [ ] Run the existing syntax checkers and validations at build time with Bazel
- [ ] Auto-generate the example configurations and variable documentation with Bazel


WORKSPACE
```diff
@@ -0,0 +1,32 @@
...
+    url = "https://releases.hashicorp.com/terraform/0.10.7/terraform_0.10.7_darwin_amd64.zip",
+)
+
+new_http_archive(
+    name = "terraform_provider_matchbox",
+    build_file_content = """exports_files(
+["terraform-provider-matchbox-v0.2.2-darwin-amd64/terraform-provider-matchbox"],
+visibility = ["//visibility:public"]
+)""",
+    url = "https://github.com/coreos/terraform-provider-matchbox/releases/download/v0.2.2/terraform-provider-matchbox-v0.2.2-darwin-amd64.tar.gz",
```
s-urbaniak: `v.0.2.2` appears 4 times, does bazel have some methods to define a central constant?

Makefile
```diff
@@ -1,199 +0,0 @@
...
-	-e TF_VAR_tectonic_license_path \
-	-e TF_VAR_tectonic_pull_secret_path \
-	-e TF_VAR_tectonic_base_domain \
-	-e TF_VAR_tectonic_admin_email \
-	-e TF_VAR_tectonic_admin_password \
-	-e TECTONIC_TESTS_DONT_CLEAN_UP \
-	--cap-add NET_ADMIN \
-	--device /dev/net/tun \
-	quay.io/coreos/tectonic-smoke-test-env \
-	$(TEST_COMMAND)
```
s-urbaniak: The removal of this file is 👍 🎉 :-)

WORKSPACE
```diff
@@ -0,0 +1,32 @@
...
+go_register_toolchains()
+
+# Runtime binary dependencies follow.
+# These will be fetched and included in the build output verbatim.
+#
+new_http_archive(
+    name = "terraform_runtime",
+    build_file_content = """exports_files(["terraform"], visibility = ["//visibility:public"])""",
+    type = "zip",
+    url = "https://releases.hashicorp.com/terraform/0.10.7/terraform_0.10.7_darwin_amd64.zip",
```
s-urbaniak: this is darwin, I don't expect this to run on my machine ;-)

alexsomesan: This is hardcoded for me now. I'll soon have it do autodetection for the arch and OS.

WORKSPACE
alexsomesan: It does. All versions will eventually be based on variables configurable in a single place.

alexsomesan:
ok to test

sym3tri:
Please update readme to remove any unnecessary make references etc.

tests/rspec/lib/cluster.rb
```diff
@@ -67,13 +68,22 @@ def env_variables
     }
   end
 
+  def terraform_cmd(action)
+    templates_top = "#{@root_path}/templates/platforms/#{env_variables['PLATFORM']}"
+    "terraform #{action} #{templates_top}"
+  end
+
+  def terraform_action(action, envvars = env_variables)
+    Dir.chdir(@build_path) { system(envvars, terraform_cmd(action)) }
```
cpanato: do you need do two time Dir.chdrir(@build_path) ? it is here and in the line 36

alexsomesan: Not really. That's a good point, I'll clean that up.

Quentin-M:
Hey,

Just discovered this PR, how does that interact with RA - more specifically to the [building script](https://github.com/coreos-inc/tectonic-release-automation/blob/master/scripts/build-tectonic-tarball/build_tectonic_tarball.sh)? The build logic is extremely simple there and essentially dependency free: it can be built on Container Linux natively without the need of maintaining any builder image.

Note that if you intend to download Terraform / matchbox yourself and give that to RA, their signatures must be verified (see script above).

Quentin-M:
Hey,

Just discovered this PR, how does that interact with RA - more specifically to the [building script](https://github.com/coreos-inc/tectonic-release-automation/blob/master/scripts/build-tectonic-tarball/build_tectonic_tarball.sh)?

The build logic is extremely simple there and essentially dependency free: it can be built on Container Linux natively without the need of maintaining any builder image or anything else. The script runs `make release-bins`, download/verify matchbox+terraform, put the right files at the right places, make a tarball and signs it. This tarball+signature is the one released. Later steps take that tarball+signature to create the release docker image, payload, etc.

You can find the job there: https://jenkins-tectonic-installer.prod.coreos.systems/job/tectonic-release-automation/job/build-tectonic/build?delay=0sec 

Would the Installer be responsible for maintaining the logic creating the release tarballs? Or would RA still be? Note that if you intend to download Terraform / matchbox yourself and give that to RA, their signatures must be verified (see script above).

Quentin-M:
I was also talking to @mxinden this morning. The Installer team needs to provision various cluster for smoke tests (among others), Core Services to run upgrade tests, and RA to run integration/conformance/readiness. I am almost sure a few other teams need to do that too.

Right now, everybody pretty much do it differently with more or less success/reliability. With the tremendous amount of work you have done around that over the past months, generalizing and stabilizing the whole thing, I think it'd be extremely beneficial to allow other teams to provision/destroy clusters based on a few inputs (such as a tarball + tfvars). This could be a Jenkins script (or job) that people would use in their own pipelines. If the Installer produces a tarball, that tarball could definitely be used as an input - and be common to everybody.

I am tasked to implement tests into RA but definitely would like to avoid duplicating all that existing logic and reproducing all the pitfalls for weeks. I am convinced this would also save significant amount of time for other teams.

Do you think it'd be doable and prioritizable?

alexsomesan:
Quentin,

I think we're thinking the same thing about the boundaries between the installer build step and RA. Essentially, when the Bazel logic is done, it would produce the exact contents of the tarball you're talking about. From that point onwards it hands over to RA for packaging the actual releasable package, with all the signing and validations that RA already does. We will stop short of producing the actual tarball for now, because I think RA is better positioned to handle that step. If you want to however, we can also consider building the tarball as part of the build step.

What would change in RA is it would invoke a Bazel command instead of a make one. And it would not have to care about fetching any additional dependencies like Terraform or bundled providers because the built installer output would already have them.

As you'll realise from my explanation below, this actually makes things simpler for everyone, especially end- and intermediate users, like other CoreOS teams up the stack.

Now a bit of summary about why we do this. I'll try to just touch on each point, the more elaborate reasoning is in the build logic design document.

Most importantly, it's about dogfooding. We as the developers of the Tectonic (should) want to run the installer in exactly the same manner as our users get to run it. That means moving away from invoking the installer directly from the repository. A build step is necessary to ensure that things are the same as in a release package every time a change is made. That way, when a developer makes a change, they would test it the same way a user would see it. Right now this is not happening. The same applies for running CI. We want our pipeline to run our tests against the same artefacts the users would see, not against the code repository directly.

Second important aspect is consolidating dependency management. We want to be able to precisely determine which version of a dependency we test with and be confident that we ship the same as we tested. This again ties back to the developer workflow - for example developers need to know that the version of Terraform they test with is actually the one being shipped. This means it needs to be fetched much sooner, during the build process.

There are more aspects (some logistical) to this change. I'm not going to touch on those here to keep this to a reasonable size.

Reach out to me internally and I'm happy to run you through all the practical aspects of this change.

Quentin-M:
ok to test

PR #2223, opened 2017-10-25
platforms/vmware: Add ability to span virtual datacenters

lander2k2:
Currently, Tectonic clusters must reside entirely within a virtual
datacenter.  This change will allow the virtual machines to be mapped to
the virtual datacenters.

Jira issue: https://jira.prod.coreos.systems/browse/FR-361


lander2k2:
@squat Rebased.  Once this is merged I'll rebase the resource pool commit [1] as I suspect it will conflict with this commit, too.
[1] https://github.com/coreos/tectonic-installer/pull/2250

PR #2224, opened 2017-10-25
config.tf: Bump tectonic version to 1.8.

yifan-gu:
This is a placeholder so that the [release automation job](https://jenkins-tectonic-installer.prod.coreos.systems/job/tectonic-release-automation/) can generate
a new payload, which will be used by [nightly upgrade test job](https://jenkins-tectonic-installer.prod.coreos.systems/job/upgrade-test/).

Ideally this could be done automatically when we cut a release branch.

### DIFF
--- a/config.tf
+++ b/config.tf
@@ -117,7 +117,7 @@ variable "tectonic_versions" {
     etcd          = "3.1.8"
     kubernetes    = "1.7.5+tectonic.1"
     monitoring    = "1.7.0"
-    tectonic      = "1.7.5-tectonic.1"
+    tectonic      = "1.8.2-tectonic.1"
     tectonic-etcd = "0.0.1"
     cluo          = "0.2.3"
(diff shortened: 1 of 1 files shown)


config.tf
```diff
@@ -117,7 +117,7 @@ variable "tectonic_versions" {
     etcd          = "3.1.8"
     kubernetes    = "1.7.5+tectonic.1"
```
diegs: We should bump this to 1.7.9 right?

yifan-gu: We can leave as 175 until the real 179 is out.
I did this PR just for the sake of upgrade tests, it expects a payload with newer version than the latest release.

yifan-gu:
I think the azure failures are unrelated. 
Besides it's just a placeholder anyway to enable nightly upgrade tests, so merging for now.

PR #2233, opened 2017-10-26
tests: added the env variables for flake project

erikkn:
Added the required variables for the flaked logstash project.

### DIFF
--- a/tests/rspec/lib/cluster.rb
+++ b/tests/rspec/lib/cluster.rb
@@ -102,6 +102,7 @@ class Cluster
   def apply
     ::Timeout.timeout(30 * 60) do # 30 minutes
       3.times do |idx|
+        print "tf_apply_attempts=#{idx}"
         env = env_variables
         env['TF_LOG'] = 'TRACE' if idx.positive?
         return true if system(env, 'make -C ../.. apply')
@@ -113,6 +114,7 @@ class Cluster
   def destroy
     ::Timeout.timeout(30 * 60) do # 30 minutes
       3.times do |idx|
+        print "tf_destroy_attempts=#{idx}"
         env = env_variables
         env['TF_LOG'] = 'TRACE' if idx.positive?
         return true if system(env, 'make -C ../.. destroy')
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -268,10 +268,6 @@ pipeline {
         script {
           def builds = [:]
           def aws = [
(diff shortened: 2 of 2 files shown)


tests/rspec/lib/cluster.rb
```diff
@@ -99,7 +99,8 @@ def localconfig
   end
 
   def apply
-    3.times do |idx|
+    3.times do |idx||count|
```
cpanato: you can use idx which is the counter

tests/rspec/lib/cluster.rb
```diff
@@ -100,6 +100,7 @@ def localconfig
 
   def apply
     3.times do |idx|
+      print "tf_apply_attempts=#{idx}"
```
cpanato: print "tf_apply_attempts=#{idx}\n"

cpanato: or use puts

cpanato:
@erikkn there are some lint errors, please check the test results

tests/rspec/spec/aws_ca_spec.rb
```diff
@@ -3,6 +3,8 @@
 require 'shared_examples/k8s'
 
 RSpec.describe 'aws-custom-ca' do
+  puts "TestCase=#{self.description.to_s}"
+  puts "platform=aws"
```
mxinden: Instead of printing the platform on every spec file, why not adding it once to e.g. the cluster factory https://github.com/coreos/tectonic-installer/blob/master/tests/rspec/lib/cluster_factory.rb#L12 ?

tests/rspec/spec/aws_ca_spec.rb
```diff
@@ -3,6 +3,8 @@
 require 'shared_examples/k8s'
 
 RSpec.describe 'aws-custom-ca' do
+  puts "TestCase=#{self.description.to_s}"
```
mxinden: Instead of adding this to every spec file we could instead change the RSpec output format to e.g. `documentation`. @cpanato What do you think?

cpanato: sounds good to me.

cpanato:
@erikkn  if this is still needed, please rebase

erikkn:
ok to test

erikkn:
ok to test

erikkn:
ok to test

erikkn:
closed, not necessary anymore with the new workflow.

PR #2241, opened 2017-10-26
Addresses FR-425

coresolve:
This change makes init_assets.service dependant on
network-online.target.

This will ensure that the service waits until resolv.conf on the node is
present before trying to fetch assets.


mxinden:

PR #2244, opened 2017-10-27
modules/tectonic: change rkt->docker

squat:
This commit modifies all of the Systemd services to use docker rather
than rkt to run their containers. The motivation for this is:

1. bootstrap services run with rkt often experience failures due to
transient network issues that cause image download failures; docker
auto-retries image downloads;
2. bootstrap services run with rkt sometimes fail due to a race
condition in double-bind-mounting /etc/resolv.conf and changes caused by
systemd-resolvd; docker does not double-bind-mount /etc/resolv.conf;

Thanks to @s-urbaniak and @lucab for the deep digging on issue 2.

cc @s-urbaniak


squat:
ok to test

s-urbaniak:
also /cc @aaronlevy @diegs for any concerns regarding dockerizing bootkube

PR #2245, opened 2017-10-27
tests: add with_retry and small refactor

cpanato:
fixes: https://jira.prod.coreos.systems/browse/INST-544

### DIFF
--- a/tests/rspec/lib/aws_support.rb
+++ b/tests/rspec/lib/aws_support.rb
@@ -1,21 +1,28 @@
 # frozen_string_literal: true
 
 require 'aws-sdk'
+require 'with_retries'
 
 # Shared support code for AWS-based operations
 #
 module AwsSupport
   def self.sorted_auto_scaling_instances(aws_autoscaling_group_id, aws_region)
     aws = Aws::AutoScaling::Client.new(region: aws_region)
-    resp = aws.describe_auto_scaling_groups(auto_scaling_group_names: [
-                                              aws_autoscaling_group_id.to_s
-                                            ])
+    resp = ''
+    Retriable.with_retries(Aws::EC2::Errors::RequestLimitExceeded, limit: 5, sleep: 10) do
+      resp = aws.describe_auto_scaling_groups(auto_scaling_group_names: [
+                                                aws_autoscaling_group_id.to_s
+                                              ])
+    end
     resp.auto_scaling_groups[0].instances.map(&:instance_id).sort
   end
 
   def self.preferred_instance_ip_address(instance_id, aws_region)
     aws = Aws::EC2::Client.new(region: aws_region)
-    resp = aws.describe_instances(instance_ids: [instance_id.to_s])
+    resp = ''
+    Retriable.with_retries(Aws::EC2::Errors::RequestLimitExceeded, limit: 5, sleep: 10) do
+      resp = aws.describe_instances(instance_ids: [instance_id.to_s])
+    end
     ssh_master_ip = if resp.reservations[0].instances[0].network_interfaces[0].association.nil?
                       resp.reservations[0].instances[0].network_interfaces[0].private_ip_address
                     else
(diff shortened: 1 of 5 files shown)


tests/rspec/lib/aws_support.rb
```diff
@@ -31,18 +31,34 @@ def self.create_aws_key_pairs(aws_region)
...
-                                  public_key_material: ssh_pub_key)
-    raise 'Error to import AWS key pair' unless resp.successful?
+    retries = 0
+    begin
+      resp = client.import_key_pair(dry_run: false,
+                                    key_name: key_pair_name,
+                                    public_key_material: ssh_pub_key)
+    rescue Aws::EC2::Errors::RequestLimitExceeded
+      raise 'failed to exec aws import_key_pair in 5 retries' if retries >= 5
+      retries += 1
```
squat: The first implementation of this retry logic is in the ssh lib and now we have two more copies. We should now break this into a little function so that the implementations don’t drift or get written incorrectly. Let’s make a little retry lib

tests/rspec/lib/aws_support.rb
```diff
@@ -31,18 +31,34 @@ def self.create_aws_key_pairs(aws_region)
     dir_home = `echo ${HOME}`.chomp
     ssh_pub_key = File.read("#{dir_home}/.ssh/id_rsa.pub")
 
-    resp = client.import_key_pair(dry_run: false,
-                                  key_name: key_pair_name,
-                                  public_key_material: ssh_pub_key)
-    raise 'Error to import AWS key pair' unless resp.successful?
+    retries = 0
+    begin
+      resp = client.import_key_pair(dry_run: false,
```
mxinden: If I am not missing something here, `import_key_pair` is only executed once per test. The logic that exceeds the limit is: https://github.com/coreos/tectonic-installer/blob/master/tests/rspec/lib/cluster.rb#L162 in the bootstrap phase. So adding the `rescue` here will only fix the result, not the root cause.

The reason why we are refreshing the master ips so often in [`wait_for_service`](https://github.com/coreos/tectonic-installer/blob/master/tests/rspec/lib/cluster.rb#L162) is because we can't be sure, that at the first point in time, all master machines are returned by the `master_ip_addresses` function.

I guess in this [build](https://jenkins-tectonic-installer.prod.coreos.systems/blue/organizations/jenkins/tectonic-installer/detail/PR-2219/3/pipeline/143/) it exceeded in `import_key_paris`, because other parallel builds were in the bootstrapping loop at that moment. @cpanato is that a reasonable theory?

If you think `wait_for_service` is the source of this issue as well, @cpanato, I will add better back off into `wait_for_service`, and thereby prevent the request limit exeeded error. (@squat I will extract it into a separate retry function.)

cpanato:
@squat @mxinden I will rewrite the retry logic and refactor the master_ip thingy. bear with me. 🐻

cpanato:
can you guys give another review, please? @squat @mxinden

cpanato:
retest this please

cpanato:
retest this please

tests/rspec/lib/with_retries.rb
```diff
@@ -0,0 +1,42 @@
...
+    exceptions = [Exception] if exceptions.empty?
+
+    retried = 0
+    begin
+      yield
+    rescue *exceptions => e
+      raise e if retried + 1 > options[:limit]
+      retried += 1
+      sleep options[:sleep] * retried
+      puts "Error #{e}"
```
squat: Can you include something that says how many times this has been retried, like it used to do? e.g.
```ruby
puts "Error #{e}; retrying in #{sleep_time} seconds"
```

tests/rspec/lib/with_retries.rb
```diff
@@ -0,0 +1,42 @@
...
+      retry
+    end
+  end
+end
+
+# Add module to the Object class
+class Object
```
squat: Does anything actually use this class?

tests/rspec/lib/ssh.rb
```diff
@@ -13,32 +13,26 @@ def ssh_agent_has_key?
...
-  Net::SSH.start(
-    ip_address, 'core', forward_agent: true, use_agent: true, verify_host_key: Net::SSH::Verifiers::Null.new
-  ) do |ssh|
-    ssh.exec! command, status: status do |_ch, stream, data|
-      case stream
-      when :stdout
-        stdout << data
-      when :stderr
-        stderr << data
+  with_retries(Errno::ECONNREFUSED, Errno::ECONNRESET, Errno::ETIMEDOUT,
```
squat: How come we are not importing the `with_retries` module in this lib?

tests/rspec/lib/with_retries.rb
cpanato: done

tests/rspec/lib/ssh.rb
cpanato: discussed offline

tests/rspec/lib/with_retries.rb
cpanato: discussed offline

tests/rspec/lib/ssh.rb
mxinden: 👍 for importing it here as well.

PR #2246, opened 2017-10-27
test: remove hard coded CL version

cpanato:
cc @s-urbaniak


cpanato:
ok to test

squat:
Rather than set the defaultValue to ‘’ can we remove that block for the variable entirely?

cpanato:
@squat we can remove, but since we have that, we can reuse in future for other tests.

cpanato:
@squat if want to remove I will remove, let me know

squat:
I prefer to remove it to avoid any confusion

cpanato:
@squat done

PR #2248, opened 2017-10-27
1.7.9: restore tectonic-namespace.yaml.

diegs:
I erroneously assumed that there was still another one in the tectonic
module, and merged it because the smoke tests passed.


mxinden:
@yifan-gu @diegs Please make sure to not only add the `run-smoke-tests` label, but also specify which platform to run them on by setting the `platform/<xxx>` label. Otherwise, no smoke tests will be run.

For further details see: https://github.com/coreos/tectonic-installer/tree/master/tests#running-smoke-tests-on-prs

PR #2250, opened 2017-10-27
platforms/vmware: Add ability to assign VMs to resource pools

lander2k2:
Currently, all VMs in a cluster must be assigned to a single VMware
resource pool.  This change allows VMs to be assigned to different
resource pools so that, for example, master nodes won't run into a
resource contention with worker nodes.

Jira issue: https://jira.prod.coreos.systems/projects/FR/issues/FR-364


lander2k2:
@squat Rebased.  Should be ready to merge.

PR #2257, opened 2017-10-30
frontend: Migrate AWS VPC page from WithClusterConfig to Connect

kyoto:
Removes the last remaining use of `WithClusterConfig`.

Fixes the AWS VPC validation, which had 2 bugs:
1. It only ran when the VPC dropdown changed. Since it also validates the subnets and the pod range and service range CIDRs, it should run whenever any of those inputs changes.
2. The `validate` API endpoint was call incorrectly. `privateSubnets` and `publicSubnets` were sent as a list of CIDR strings, but `validate` expects an object for each subnet.

There are still 2 issues with the VPC validation to fix in future PRs:
- Clicking the deselect checkbox doesn't trigger `asyncValidator()`.
- Unnecessary calls to `validate` when `Field` validation has failed.

Also changes `Form` to pass `oldCC` to `asyncValidator()` so that the VPC form can access the previous `POD_CIDR` and `SERVICE_CIDR` values, which are not available via `oldData` because they are in a different `Form`.


kyoto:
ok to test

PR #2259, opened 2017-10-30
tests: Remove unused param reference

mxinden:
The container linux version parameter was removed recently. This
reference has to be removed as well. Otherwise, the wanted container
linux version will be an empty string which results in:
```
module.workers.data.aws_ami.coreos_ami: data.aws_ami.coreos_ami: Your
query returned no results. Please change your search criteria and try
again.
```


mxinden:
@squat Thanks for the quick review here.

PR #2273, opened 2017-11-01
Adds support for tectonic_etcd_scheme

coresolve:
This adds support for specifying the scheme that will be used for the
external etcd cluster. it retains the default `https`

I have tested bare metal azure and aws. @squat @kyoto


config.tf
```diff
@@ -184,6 +184,15 @@ EOF
   default = []
 }
 
+variable "tectonic_etcd_scheme" {
+  type    = "string"
+  default = "https"
+
+  description = <<EOF
+(optional) Can be either "http" or "https" When set this scheme will be used for all provided etcd endpoints.
```
squat: Please add a “.” between the two sentences and add a “,” after “When set” so it reads better.

platforms/gcp/main.tf
```diff
@@ -61,6 +61,8 @@ module "etcd" {
   master_subnetwork_name = "${module.network.master_subnetwork_name}"
   external_endpoints     = ["${compact(var.tectonic_etcd_servers)}"]
 
+  etcd_scheme = "${var.tectonic_etcd_scheme}"
```
squat: We need to declare “etcd” in platforms/gcp/etcd otherwise GCP will fail to build.

platforms/vmware/main.tf
```diff
@@ -7,6 +7,7 @@ module "etcd" {
   container_image    = "${var.tectonic_container_images["etcd"]}"
   base_domain        = "${var.tectonic_base_domain}"
   external_endpoints = ["${compact(var.tectonic_etcd_servers)}"]
+  etcd_scheme        = "${var.tectonic_etcd_scheme}"
```
squat: Same here, let’s declare etcd in platforms/vmware/etcd so as to not break this platform.

mxinden:

coresolve:
We can't infer https only from certs. There are a few ways to configure etcd.
1. etcd non tls.
2. etcd tls but no client certs and no ca crt.
    This means that we the apiservers will be able to trust the etcd endpoints with the ca cert bundle they already have (/usr/share/ca-certificates)
3. etcd tls with etcd ca crt but not mtls
    This means the etcd cluster is signed by a CA that isn't represented in /usr/share/ca-certificates
    For example an Corp CA.
4. etcd with mtls
    This means that we provide client cert, key and optionally ca cert. In this case it's expected that the etcd servers are configured for mtls (and will assert that the client is signed by a cert that it knows)

platforms/gcp/main.tf
coresolve: fixed

platforms/vmware/main.tf
coresolve: fixed

coresolve:
closed in favor of https://github.com/coreos/tectonic-installer/pull/2288

PR #2282, opened 2017-11-01
Add ability to map VMs to network gateways

lander2k2:
Currently, all nodes of a given role (etcd, master, worker) need to
share a common network gateway.  This change will allow network gateways
to be assigned to nodes individually to support nodes being spread
across subnets with different gateways in a multi-az kind of approach.

Jira issue: https://jira.prod.coreos.systems/projects/FR/issues/FR-369


squat:
All basic tests have passed. Stopping smoke tests from running and merging.

PR #2284, opened 2017-11-01
platform/gcp: add kubenet and cloud-provider optional support

enxebre:
This adds support and test coverage for running the platform on GCP using ```kubenet``` and ```--cloud-provider=gce```

When kubenet is enabled:
- Set ```--cloud-provider=gce```
- Set ```--cloud-config```
- Set ```--hostname-override``` to the nodename expected by gce provider i.e machine name
- Set ```--network-plugin=kubenet```
- Set ```--configure-cloud-routes=true```

Why this flags are needed:
- At the moment because of https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet_network.go#L85
The following scenario happens when running the Kubelet with ```--cloud-provider=gce```:
The node always set NetworkUnavailable property to true.
This flag is only cleared by the cloud-provider route controller which is run when the flag ```--configure-cloud-routes=true``` is set in the controller manager (default).
- A mismatch between the (kubelet) nodeName which is the hostname by default (for CL the fqdn) and the gcp machine name. See https://github.com/kubernetes/kubernetes/pull/54504

This also adds ```--cluster-name``` to the controller manager
Adding do-not-merge label as there's a non solved problem in this PR yet:
When using  ```--configure-cloud-routes=true``` the route controller will generate network routes not known by terraform. Hence when running ```terraform destroy``` it can’t destroy the network because it has dependencies with the routes on the google API.

### DIFF
--- a/modules/bootkube/resources/bootstrap-manifests/bootstrap-controller-manager.yaml
+++ b/modules/bootkube/resources/bootstrap-manifests/bootstrap-controller-manager.yaml
@@ -10,11 +10,12 @@ spec:
     command:
     - ./hyperkube
     - controller-manager
+    - --cluster-name=${cluster_name}
     - --allocate-node-cidrs=true
     - --cluster-cidr=${cluster_cidr}
     - --cloud-provider=${cloud_provider}
     ${cloud_provider_config_flag}
-    - --configure-cloud-routes=false
+    - --configure-cloud-routes=${configure_cloud_routes}
     - --leader-elect=true
     - --kubeconfig=/etc/kubernetes/kubeconfig
     - --root-ca-file=/etc/kubernetes/bootstrap-secrets/ca.crt
--- a/modules/bootkube/resources/manifests/kube-controller-manager.yaml
+++ b/modules/bootkube/resources/manifests/kube-controller-manager.yaml
@@ -40,8 +40,9 @@ spec:
         command:
         - ./hyperkube
         - controller-manager
+        - --cluster-name=${cluster_name}
         - --allocate-node-cidrs=true
-        - --configure-cloud-routes=false
+        - --configure-cloud-routes=${configure_cloud_routes}
         - --cluster-cidr=${cluster_cidr}
         - --root-ca-file=/etc/kubernetes/secrets/ca.crt
         - --service-account-private-key-file=/etc/kubernetes/secrets/service-account.key
--- a/modules/gcp/master-igm/resources/services/node-name.service
+++ b/modules/gcp/master-igm/resources/services/node-name.service
@@ -1,0 +1,14 @@
+[Unit]
+Description=Set NODENAME for kubelet as expected by GCE cloud Provider
+ConditionPathExists=/etc/kubernetes/kubelet.env
+Requires=systemd-hostnamed.service systemd-networkd.service systemd-networkd-wait-online.service
+Before=kubelet.service
+After=systemd-hostnamed.service systemd-networkd.service systemd-networkd-wait-online.service
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/bin/sh -c "/bin/echo NODENAME=$(hostname -s) >> /etc/kubernetes/kubelet.env"
+
+[Install]
+WantedBy=kubelet.service
--- a/modules/gcp/network/firewall-master.tf
+++ b/modules/gcp/network/firewall-master.tf
@@ -93,3 +93,23 @@ resource "google_compute_firewall" "master-ingress-services" {
   source_tags = ["tectonic-masters"]
   target_tags = ["tectonic-masters"]
 }
+
+resource "google_compute_firewall" "ingress-pods" {
+  name    = "${var.cluster_name}-ingress-pods"
+  network = "${google_compute_network.network.name}"
+
+  allow {
+    protocol = "udp"
+  }
+
+  allow {
+    protocol = "tcp"
+  }
+
+  allow {
+    protocol = "icmp"
+  }
+
+  source_ranges = ["10.0.0.0/8", "10.2.0.0/16"]
+  target_tags   = ["tectonic-masters", "tectonic-workers"]
+}
--- a/modules/ignition/resources/services/kubelet.service
+++ b/modules/ignition/resources/services/kubelet.service
@@ -24,7 +24,7 @@ ExecStart=/usr/lib/coreos/kubelet-wrapper \
   --kubeconfig=/etc/kubernetes/kubeconfig \
   --require-kubeconfig \
   --cni-conf-dir=/etc/kubernetes/cni/net.d \
-  --network-plugin=cni \
+  --network-plugin=${network_plugin} \
   --lock-file=/var/run/lock/kubelet.lock \
   --exit-on-lock-contention \
   --pod-manifest-path=/etc/kubernetes/manifests \
@@ -38,7 +38,8 @@ ExecStart=/usr/lib/coreos/kubelet-wrapper \
   --client-ca-file=/etc/kubernetes/ca.crt \
   --cloud-provider=${cloud_provider} \
   ${cloud_provider_config} \
-  --anonymous-auth=false
+  --anonymous-auth=false \
+  ${hostname_override_cmd} \
 
 ExecStop=-/usr/bin/rkt stop --uuid-file=/var/cache/kubelet-pod.uuid
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -265,6 +265,7 @@ pipeline {
           def gcp = [
             [file: 'basic_spec.rb', args: ''],
             [file: 'ha_spec.rb', args: ''],
+            [file: 'kubenet_ha_spec.rb', args: ''],
           ]
 
           if (params."PLATFORM/AWS") {
--- a/modules/bootkube/assets.tf
+++ b/modules/bootkube/assets.tf
@@ -66,6 +66,8 @@ resource "template_dir" "bootkube" {
 
     cloud_provider_profile = "${var.cloud_provider != "" ? "${var.cloud_provider}" : "metal"}"
     cloud_config_path      = "${var.cloud_config_path}"
+    cluster_name           = "${var.cluster_name}"
+    configure_cloud_routes = "${var.configure_cloud_routes}"
   }
 }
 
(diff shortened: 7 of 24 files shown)


modules/bootkube/resources/bootstrap-manifests/bootstrap-controller-manager.yaml
```diff
@@ -10,11 +10,12 @@ spec:
     command:
     - ./hyperkube
     - controller-manager
+    - --cluster-name=${cluster_name}
```
s-urbaniak: this most probably needs a migration for KVO. Do we really need this setting now?

/cc @diegs

modules/bootkube/resources/bootstrap-manifests/bootstrap-controller-manager.yaml
```diff
@@ -10,11 +10,12 @@ spec:
...
     - --allocate-node-cidrs=true
     - --cluster-cidr=${cluster_cidr}
     - --cloud-provider=${cloud_provider}
     ${cloud_provider_config_flag}
-    - --configure-cloud-routes=false
+    - --configure-cloud-routes=${configure_cloud_routes}
```
s-urbaniak: This would need to go into a config map for the potential tectonic network operator, /cc @rithujohn191

modules/bootkube/resources/manifests/kube-controller-manager.yaml
```diff
@@ -40,8 +40,9 @@ spec:
         command:
         - ./hyperkube
         - controller-manager
+        - --cluster-name=${cluster_name}
```
s-urbaniak: see above, we potentially need a migration for this setting for all other platforms, if this is a showstopper for GCP.

modules/bootkube/resources/manifests/kube-controller-manager.yaml
```diff
@@ -40,8 +40,9 @@ spec:
...
         - --allocate-node-cidrs=true
-        - --configure-cloud-routes=false
+        - --configure-cloud-routes=${configure_cloud_routes}
```
s-urbaniak: see above

modules/gcp/master-igm/resources/services/node-name.service
```diff
@@ -0,0 +1,14 @@
...
+Description=Set NODENAME for kubelet as expected by GCE cloud Provider
+ConditionPathExists=/etc/kubernetes/kubelet.env
+Requires=systemd-hostnamed.service systemd-networkd.service systemd-networkd-wait-online.service
+Before=kubelet.service
+After=systemd-hostnamed.service systemd-networkd.service systemd-networkd-wait-online.service
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/bin/sh -c "/bin/echo NODENAME=$(hostname -s) >> /etc/kubernetes/kubelet.env"
```
s-urbaniak: is there any possibility to do this via ignition and some datasource in terraform itself? If possible it would be great to leave resolving the hostname on GCP to intrinsic CL methods, /cc @lucab @crawford

modules/gcp/network/firewall-master.tf
```diff
@@ -93,3 +93,23 @@ resource "google_compute_firewall" "master-ingress-services" {
...
+
+  allow {
+    protocol = "tcp"
+  }
+
+  allow {
+    protocol = "icmp"
+  }
+
+  source_ranges = ["10.0.0.0/8", "10.2.0.0/16"]
```
s-urbaniak: Isn't `10.2.0.0./16` dependent on the `tectonic_cluster_cidr` value?

modules/ignition/resources/services/kubelet.service
```diff
@@ -38,7 +38,8 @@ ExecStart=/usr/lib/coreos/kubelet-wrapper \
   --client-ca-file=/etc/kubernetes/ca.crt \
   --cloud-provider=${cloud_provider} \
   ${cloud_provider_config} \
-  --anonymous-auth=false
+  --anonymous-auth=false \
+  ${hostname_override_cmd} \
```
s-urbaniak: this is tricky, as we have no upgrade path for the kubelet for now. Changing the manifest with dynamic values leaves an open door here. Is it possible to get along without these settings? /cc @diegs @aaronlevy

modules/ignition/resources/services/kubelet.service
```diff
@@ -24,7 +24,7 @@ ExecStart=/usr/lib/coreos/kubelet-wrapper \
   --kubeconfig=/etc/kubernetes/kubeconfig \
   --require-kubeconfig \
   --cni-conf-dir=/etc/kubernetes/cni/net.d \
-  --network-plugin=cni \
+  --network-plugin=${network_plugin} \
```
s-urbaniak: see above. for the kubelet it is even trickier, since no upgrade path exists on the KVO side.

modules/bootkube/resources/bootstrap-manifests/bootstrap-controller-manager.yaml
enxebre: It's not strictly necessary at the moment. It basically prefixes the resources created by the controllers so it might help to solve the problem of cleaning up the resources. It seems right to have it

modules/gcp/master-igm/resources/services/node-name.service
enxebre: The hostname will still resolve to intrinsic CL methods. This will just set a env variable then used by the kubelet on ```--hostname-override```, haven't found a cleaner way so far. This is needed because of this https://github.com/kubernetes/kubernetes/pull/54504

modules/gcp/network/firewall-master.tf
enxebre: yup, needs to be injected

modules/ignition/resources/services/kubelet.service
enxebre: See https://github.com/coreos/tectonic-installer/pull/2284/files#r148466487

modules/gcp/master-igm/resources/services/node-name.service
lucab: coreos-metadata has `COREOS_GCE_HOSTNAME`, but I don't know if it matches with @enxebre requirements.

In any case, please don't use `kubelet.env` to store this, as it has very specific semantics and ordering during bootstrap and is also manipulated by operators. Just use your own cloud-specific dropin.

/cc @squeed I think we may want have an env-dir at some point for such cases.

lucab:
It looks like this is de-touring into hostname-related mismatches. I don't have an exhaustive understanding of GCP-specific details, but I would recommend to:
 * research and document what is the expected/recommended hostname for a tectonic node (FQDN vs plain name) and how that gets set (`/etc/hostname` vs `/etc/hosts` vs DHCP)
 * research why/how this is different in GCP, why/how other clouds are doing it differently and try to align them

I fear that painting over the symptoms with cloud-specific CLI flags will just result in more troubles down the road, and forgotten corner-cases in future migration logic. It may still be needed in the end, but at the moment it sounds like this could be adjusted with proper hostname-setting on the host.

modules/bootkube/resources/bootstrap-manifests/bootstrap-controller-manager.yaml
squeed: Also annoyingly, there is a limit of 300 routes, globally, across a GCP account. This can be raised, but it requires a manual request.

squeed:
We do need to be careful if we want to formally support kubenet. It excludes us from ever supporting things like Calico policy.

enxebre:
So from my understanding and at high level:
There’s a discrepancy between GCP and CL:
- A Compute Engine image https://github.com/GoogleCloudPlatform/compute-image-packages retrieves the hostname by querying metadata https://cloud.google.com/compute/docs/storing-retrieving-metadata at boot time.
- The hostname property coming here is the fqdn and is stored so it can be used my daemons inside the machine
- Then GCP intentionally sets the hostname to be the unqualified domain name by default (which is the same as the machine name). https://github.com/GoogleCloudPlatform/compute-image-packages/blob/master/google_config/bin/google_set_hostname#L29
Other distros run google_set_hostname script as a dhcpd exit hook

However the behaviour with CL is a bit different:
- The machine gets the values from the metadata via coreos-metadata.service / oem-cloudinit.service
oem-gce.service run some google_* scripts via rkt but google_set_hostname is not among them https://github.com/coreos/coreos-overlay/blob/master/coreos-base/oem-gce/files/units/oem-gce.service
https://github.com/coreos/coreos-overlay/blob/c76f4b716b3966e7dd668082becffcaaf42743d7/coreos-base/coreos-oem-gce/files/manglefs.sh
Then it relies on https://github.com/coreos/systemd/blob/master/src/hostname/hostnamed.c to set the hostname that will use whatever has been stored coming from the metadata (so fqdn).

The kubernetes cloud provider assumes the first case is the only one possible
https://github.com/kubernetes/kubernetes/blob/57f6c9dd39b8c82e42b9e041f8475a9171c5ba91/pkg/cloudprovider/providers/gce/gce_routes.go#L63

Potential solutions:
- Remove the assumption upstream https://github.com/kubernetes/kubernetes/pull/54504
- Change the wide system hostname by running ```hostnamectl``` at boot time to set the hostname to be the machine-name/hostname -s. But even if there’s a nice way to do this I’m not sure it’s a good idea from CL perspective, e.g there might be other consumers expecting the hostname to be the fqdn so I was hoping to get some light from someone with deeper CL knowledge
- Use kubelet —hostname-override to the machine-name=hostname -s

But aside from the hostname thing my main concern is about introducing inter platform skew cc  @aaronlevy @diegs

modules/ignition/resources/services/kubelet.service
diegs: For kubelet changes, for GCP purposes we can accept a one-time change because we don't manage it. There will be skew but we'll have to track it. However, we need to guarantee that for non-GCP then the resulting kubelet.service file is the exact same as it has been (forever).

Also, in the future there will be no way to change it under the current paradigm, so this is your one chance to make GCP-related changes.

cc @aaronlevy

modules/ignition/resources/services/kubelet.service
diegs: Same comment as above; if there are GCP-only changes that have to go in I'm of the opinion that it's ok to introduce skew under the current paradigm as a one-shot. But you have to know that (a) this can never change again for GCP either and (b) the `hostname_override_command` can only be used for GCP. All other platforms must have identical kubelet.service files as before.

PR #2290, opened 2017-11-02
tests: get overview of all env variables

erikkn:
Need an overview of all te environment variables for the logstash project.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -101,7 +101,7 @@ pipeline {
           script {
             def err = null
             try {
-              PrintLogstashAttributes()
+              printLogstashAttributes()
               forcefullyCleanWorkspace()
               withDockerContainer(params.builder_image) {
                 ansiColor('xterm') {
@@ -409,7 +409,7 @@ def reportStatusToGithub(status, context) {
   }
 }
 
-def PrintLogstashAttributes() {
+def printLogstashAttributes() {
   withCredentials(creds) {
     echo  "Build_Number="+ env.BUILD_NUMBER
     echo  "Build_ID="+ env.BUILD_ID
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -410,9 +410,7 @@ def reportStatusToGithub(status, context) {
 }
 //Function used to print environment variables that are used by Logstash as attributes for the log file.
 def printLogstashAttributes() {
-    sh "env"
     echo  "Build_Number=" + env.BUILD_NUMBER
-    echo  "Build_ID=" + env.BUILD_ID
     echo  "JOB_NAME=" + env.JOB_NAME
     echo  "USER_REQUEST=" +  env.CHANGE_AUTHOR
     echo  "Build_Status=" + env.BUILD_STATUS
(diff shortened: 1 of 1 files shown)


mxinden:
@erikkn You will have to wait a little for Jenkins to trigger.

Jenkinsfile
```diff
@@ -407,3 +408,15 @@ def reportStatusToGithub(status, context) {
...
   }
 }
+
+def printLogstashAttributes() {
+  withCredentials(creds) {
+    echo  "Build_Number="+ env.BUILD_NUMBER
+    echo  "Build_ID="+ env.BUILD_ID
+    echo  "JOB_NAME="+ env.JOB_NAME
+    echo  "USER_REQUEST="+ env.CHANGE_AUTHOR
+    echo  "Build_Status="+ env.BUILD_STATUS
```
mxinden: As this function is executed at the beginning of the build, `Build_Status` will always be empty, right?

Jenkinsfile
```diff
@@ -407,3 +408,15 @@ def reportStatusToGithub(status, context) {
     """
   }
 }
+
+def printLogstashAttributes() {
```
mxinden: Can you add a comment above this function. Something along the lines of: `Function prints meta data used to aggregate and analyse build logs in long term storage.`

Jenkinsfile
```diff
@@ -407,3 +408,15 @@ def reportStatusToGithub(status, context) {
...
+  withCredentials(creds) {
+    echo  "Build_Number="+ env.BUILD_NUMBER
```
mxinden: Can you add a space before the `+`? Same for following lines.

Jenkinsfile
```diff
@@ -407,3 +408,15 @@ def reportStatusToGithub(status, context) {
     """
   }
 }
+
+def printLogstashAttributes() {
+  withCredentials(creds) {
```
mxinden: Do you need this line? This will only inject credentials into the environment, which I would prefer to not save in any long term storage system.

Jenkinsfile
erikkn: @mxinden I prefer not too, because then I would have to change the logic of my Logstash and write a custom pattern.

Jenkinsfile
erikkn: Can we create a separate function and call that one at the end of the build?

mxinden: To get the build status? Jenkins prints it by default at the end of the build. (e.g. "Finished: SUCCESS")

Jenkinsfile
```diff
@@ -407,3 +408,12 @@ def reportStatusToGithub(status, context) {
     """
   }
 }
+//Function used to print environment variables that are used by Logstash as attributes for the log file.
+def printLogstashAttributes() {
+    echo  "Build_Number=" + env.BUILD_NUMBER
+    echo  "JOB_NAME=" + env.JOB_NAME
+    echo  "USER_REQUEST=" +  env.CHANGE_AUTHOR
+    echo  "Build_Status=" + env.BUILD_STATUS
```
cpanato: build status will be null because the job did not finish

cpanato:
@mxinden can you review again please?

PR #2291, opened 2017-11-02
tests: Run conformance tests with ruby logic 

mxinden:
This patch adds the ability to run the k8s upstream conformance tests on any PR by adding the `run-conformance-tests` Github label.

Thinks to do that I would like to defer to follow up PRs:
- Add k8s conformance nightly trigger
- Choose conformance image dependent on the hyperkube image


cpanato:
please remove the conformance jenkins job as well : tests/jenkins-jobs/tectonic_coformance_tests_job.groovy

mxinden:
@cpanato adjusted the timings.

tests/rspec/lib/shared_examples/k8s.rb
```diff
@@ -72,6 +73,13 @@
       expect(@login.fail_to_login?).to be_truthy
     end
   end
+
+  it 'passes the k8s conformance tests' do
```
cpanato: what do you think to we use labels in the tests? because if the environment variable is not set the test will always report passed, and in the report will say conformance test passed even if this never run

https://stackoverflow.com/questions/5069677/how-do-i-run-only-specific-tests-in-rspec

tests/rspec/lib/shared_examples/k8s.rb
```diff
@@ -72,6 +73,13 @@
...
+    if ENV['RUN_CONFORMANCE_TESTS'] == 'true'
```
cpanato: then we can remove this

tests/rspec/lib/shared_examples/k8s.rb
mxinden: Addressed in idiomatic way. Thanks for the tip @cpanato!

mxinden:
@cpanato Added the nightly trigger, updated the timing, and made conformance and smoke tests optional in idiomatic RSpec manner. In addition the documentation is updated.

Can you give this another review?

mxinden:
ok to test

cpanato:
@mxinden some tests timeout, i guess we need give some more time.

PR #2295, opened 2017-11-02
Send credentials when checking release version

ethernetdan:
[Pod-proxy](https://github.com/coreos-inc/pod-proxy/) assigns a Pod per user session with a cookie. 

Sending cookies with the modified request to prevents an unnecessary Pods from being spawned with Pod Proxy.


squat:
@ethernetdan is for an installer instance hosted on k8s?

ethernetdan:
@squat yup! 

Having the proxy manage the Pods seems to be working so I'm moving through getting the installation flow to work.

PR #2305, opened 2017-11-03
test: add tests to check scale worker node

cpanato:


tests/rspec/lib/cluster.rb
```diff
@@ -41,6 +41,12 @@ def start
     wait_til_ready
   end
 
+  def update_cluster
+    prepare_assets
```
mxinden: `prepare_assets` overrides the `terraform.tfvars` file. Doesn't that then undo your worker count changes?

tests/rspec/lib/cluster.rb
```diff
@@ -41,6 +41,12 @@ def start
...
+    start
+    wait_nodes_ready
```
mxinden: Why not make `wait_nodes_ready` part of `wait_til_ready` (append at the very end)? Waiting for all workers to become ready should be part of our normal startup tests as well. Then the entire `update_cluster` function is not needed anymore. Instead you can call `start` from the `_spec.rb` file.

tests/rspec/lib/cluster.rb
```diff
@@ -141,6 +147,35 @@ def wait_til_ready
     end
   end
 
+  def wait_nodes_ready
+    puts 'Waiting for nodes become in ready state after an update'
+    from = Time.now
+    loop do
```
mxinden: Why not use the `with_retries` logic here?

tests/rspec/lib/cluster.rb
```diff
@@ -141,6 +147,35 @@ def wait_til_ready
...
+              nodes_ready[index] = true
+            end
+          end
+        end
+
+        break if nodes_ready.uniq.length == 1
+
+        puts 'One or more nodes are not ready yet or missing nodes. Waiting...'
+        puts "# of returned nodes #{nodes['items'].size}. Expected #{@tfvars_file.node_count}"
+        raise StandardError
```
mxinden: Raising an error without an error message seems not to be a good practice to me. Can you either throw a custom error message (See https://github.com/coreos/tectonic-installer/blob/master/tests/rspec/lib/kubectl_helpers.rb#L11), or add an error message to the `StandardError`?

tests/rspec/lib/cluster.rb
```diff
@@ -141,6 +147,35 @@ def wait_til_ready
...
+      rescue KubeCTL::KubeCTLCmdError, StandardError
```
mxinden: `StandardError` can be a lot of different issues, not only your `raise StandardError` from before. Thereby if anything else fails here, debugging it is very difficult, as this rescue just swallows the error message.

tests/rspec/lib/cluster.rb
mxinden: How about making the above two `puts` statements part of the error message instead?

tests/rspec/lib/shared_examples/k8s.rb
```diff
@@ -72,6 +72,30 @@
...
+
+      expect(nodes['items']).not_to be_empty
+      expect(nodes['items'].size).to eq(@cluster.tfvars_file.node_count)
+    end
+
+    it 'can scale up nodes by 1 worker' do
+      @cluster.tfvars_file.configure_worker(@cluster.tfvars_file.worker_count + 1)
+      @cluster.update_cluster
+
+      nodes = KubeCTL.run_and_parse(@cluster.kubeconfig, 'get nodes')
```
mxinden: I would prefer `wait_nodes_ready` not only to check whether all present nodes are ready, but also that the amount of ready nodes matches the amount of nodes specified in the `terraform.tfvars` file. That way we don't need the check below anymore. What do you think?

tests/rspec/lib/tfvars_file.rb
```diff
@@ -26,6 +26,19 @@ def node_count
...
 
+  def master_count
+    data['tectonic_master_count'].to_i
+  end
+
+  def worker_count
+    data['tectonic_worker_count'].to_i
+  end
+
+  def configure_worker(node_count)
```
mxinden: This doesn't really configure the worker, but set the worker count. How about `set_worker_count`?

mxinden:
Let's make sure we add the `platform/azure` label before merging.

tests/rspec/lib/cluster.rb
cpanato: no, because we read the once and the values are in the `@data` variable.

tests/rspec/lib/cluster.rb
cpanato: sounds good I will update that

tests/rspec/lib/cluster.rb
cpanato: ok

tests/rspec/lib/cluster.rb
cpanato: ok, will refactor

tests/rspec/lib/cluster.rb
mxinden: Ok, but there is no need to override the file itself with the initial tfvars file during an `update_cluster`, right?

cpanato:
retest this please

tests/rspec/lib/cluster.rb
```diff
@@ -186,6 +178,32 @@ def wait_til_ready
     end
   end
 
+  def wait_nodes_ready
```
mxinden: How about making this part of `wait_til_ready` at the end of the function?

tests/rspec/lib/cluster.rb
```diff
@@ -48,6 +47,11 @@ def start
     wait_til_ready
   end
 
+  def update_cluster
+    start
+    wait_nodes_ready
```
mxinden: See comment below: If we make `wait_nodes_ready` part of `wait_til_ready`, this function call is not needed anymore.

tests/rspec/lib/cluster.rb
```diff
@@ -48,6 +47,11 @@ def start
     wait_til_ready
   end
 
+  def update_cluster
```
mxinden: See comment below:  If we make `wait_nodes_ready` part of `wait_til_ready`, instead of introducing this function, one could just call `start` right?

tests/rspec/lib/cluster.rb
```diff
@@ -186,6 +178,32 @@ def wait_til_ready
...
+        nodes = KubeCTL.run_and_parse(@kubeconfig, 'get nodes')
+        nodes_ready = Array.new(@tfvars_file.node_count, false)
+        nodes['items'].each_with_index do |item, index|
+          item['status']['conditions'].each do |condition|
+            if condition['type'] == 'Ready' && condition['status'] == 'True'
+              nodes_ready[index] = true
+            end
+          end
+        end
+        if nodes_ready.uniq.length == 1
```
mxinden: If `nodes_ready` is only containing `false`, `nodes_ready.uniq.length` will be `1` as well, right?

tests/rspec/lib/shared_examples/k8s.rb
```diff
@@ -117,6 +123,32 @@
     end
   end
 
+  describe 'scale up worker cluster' do
+    before(:all) do
+      platform = @cluster.env_variables['PLATFORM']
+      skip "This test is not ready to run in #{platform}" if platform == 'metal'
```
mxinden: How about `skip "This test is not ready to run in #{platform}" if (platform == 'metal') || (platform == 'azure'`)

tests/rspec/lib/shared_examples/k8s.rb
```diff
@@ -117,6 +123,32 @@
...
+      # remove this when the JIRA https://jira.prod.coreos.systems/browse/INST-619 is fixed
+      skip "This test is not ready to run in #{platform}" if platform == 'azure'
+    end
+
+    it 'has the right number of nodes in the initial configuration' do
+      nodes = KubeCTL.run_and_parse(@cluster.kubeconfig, 'get nodes')
```
mxinden: I think `wait_nodes_ready` already checks not only, that all nodes are up, but also that all nodes exist as described in the `tfvars_file`. If that is checked in the function, this logic doesn't need to be repeated here.

tests/rspec/lib/shared_examples/k8s.rb
```diff
@@ -117,6 +123,32 @@
...
+
+      expect(nodes['items']).not_to be_empty
+      expect(nodes['items'].size).to eq(@cluster.tfvars_file.node_count)
+    end
+
+    it 'can scale up nodes by 1 worker' do
+      @cluster.tfvars_file.add_worker_node(@cluster.tfvars_file.worker_count + 1)
+      @cluster.update_cluster
+
+      nodes = KubeCTL.run_and_parse(@cluster.kubeconfig, 'get nodes')
```
mxinden: Again, the check whether enough nodes exist is part of `wait_nodes_ready`. It can be removed here.

tests/rspec/spec/aws/basic_spec.rb
```diff
@@ -3,5 +3,6 @@
 require 'shared_examples/k8s'
 
 RSpec.describe 'aws-standard' do
-  include_examples('withRunningCluster', '../smoke/aws/vars/aws.tfvars.json')
+  include_examples('withRunningCluster',
```
mxinden: Is this and the following linebreaks needed or by accident?

tests/rspec/lib/cluster.rb
```diff
@@ -124,18 +128,6 @@ def license_and_pull_secret_defined?
     EnvVar.set?([license_path, pull_secret_path])
   end
 
-  def prepare_assets
```
mxinden: 👍

tests/rspec/lib/cluster.rb
```diff
@@ -186,6 +178,32 @@ def wait_til_ready
     end
   end
 
+  def wait_nodes_ready
+    puts 'Waiting for nodes become in ready state after an update'
+    180.times do
+      nodes = ''
+      Retriable.with_retries(KubeCTL::KubeCTLCmdError, limit: 5, sleep: 10) do
+        nodes = KubeCTL.run_and_parse(@kubeconfig, 'get nodes')
```
mxinden: Getting nodes of a cluster can be extracted into a function `get_nodes` of the `cluster.rb` class, what do you think?

tests/rspec/lib/cluster.rb
```diff
@@ -186,6 +178,32 @@ def wait_til_ready
     end
   end
 
+  def wait_nodes_ready
+    puts 'Waiting for nodes become in ready state after an update'
+    180.times do
```
mxinden: Am I right that this waits for 1h? To me that sounds too long. If a cluster needs 1h to come up, after the api-server is available and the `bootkube` and `tectonic` service is done, it's time to fail. What do you think?

tests/rspec/lib/cluster.rb
```diff
@@ -186,6 +178,32 @@ def wait_til_ready
...
+      nodes = ''
```
mxinden: Why setting `nodes` to empty string?

tests/rspec/lib/cluster.rb
```diff
@@ -186,6 +178,32 @@ def wait_til_ready
...
+        if nodes_ready.uniq.length == 1
+          puts '**All nodes are Ready!**'
+          return true
+        end
+        puts "One or more nodes are not ready yet or missing nodes. Waiting...\n \
+              # of returned nodes #{nodes['items'].size}. Expected #{@tfvars_file.node_count}"
+      end
+      sleep 20
+    end
+    raise 'kubectl get nodes never returned with successful error code'
```
mxinden: Am I right, that this line can also be executed, when `kubectl` works fine, but doesn't return all nodes as ready? If so, this error message is misleading. How about something like `waiting for all nodes to become ready timed out`.

tests/rspec/lib/cluster.rb
mxinden: In addition, can't this be replaced with `with_retries` instead, including the `sleep 20` below?

tests/rspec/lib/cluster.rb
cpanato: we can make, but I also would like to have the `update_cluster` to be more explicit in the action we are doing. We are not starting a cluster, we are updating that. make sense?

tests/rspec/lib/shared_examples/k8s.rb
cpanato: the other test I agree to remove, but this tests this check need to be here, otherwise what should I test?

cpanato:
@mxinden can you please do another look? thanks

tests/rspec/lib/shared_examples/k8s.rb
mxinden: You can wrap `@cluster.update_cluster` like this:

```ruby
expect { @cluster.update_cluster }.to_not raise_error
```

tests/rspec/lib/cluster.rb
mxinden: @cpanato What do you think of this idea?

tests/rspec/lib/cluster.rb
```diff
@@ -46,6 +45,11 @@ def plan
   def start
     apply
     wait_til_ready
+    wait_nodes_ready
+  end
+
+  def update_cluster
+    start
```
mxinden: 👍

tests/rspec/spec/aws/vpc_internal_spec.rb
```diff
@@ -7,6 +7,9 @@
 require 'aws_iam'
 
 RSpec.describe 'aws-vpc' do
+  include_examples('withBuildFolderSetup',
```
mxinden: Is this and the following linebreaks needed or by accident?

tests/rspec/lib/shared_examples/k8s.rb

tests/rspec/lib/cluster.rb
cpanato: 🌮

tests/rspec/spec/aws/vpc_internal_spec.rb
cpanato: removed

tests/rspec/lib/cluster.rb

cpanato:
@mxinden can you please review this again when you have some free time? thanks

tests/rspec/spec/metal/basic_spec.rb
```diff
@@ -3,14 +3,13 @@
 require 'shared_examples/k8s'
 require 'metal_support'
 
-TEST_CLUSTER_CONFIG_FILE = '../smoke/bare-metal/vars/metal.tfvars.json'
-
 RSpec.describe 'bare-metal-standard' do
+  include_examples('withBuildFolderSetup',
```
mxinden: Is this and the following linebreaks needed or by accident?

tests/rspec/spec/azure/self_hosted_etcd_spec.rb
```diff
@@ -3,5 +3,6 @@
 require 'shared_examples/k8s'
 
 RSpec.describe 'azure-self-hosted-etcd' do
-  include_examples('withRunningCluster', '../smoke/azure/vars/self-hosted-etcd.tfvars')
+  include_examples('withRunningCluster',
```
mxinden: Is this and the following linebreaks needed or by accident?

tests/rspec/spec/azure/external_self_hosted_etcd_spec.rb
```diff
@@ -3,5 +3,6 @@
 require 'shared_examples/k8s'
 
 RSpec.describe 'azure-external-self-hosted-etcd' do
-  include_examples('withRunningCluster', '../smoke/azure/vars/external-self-hosted-etcd.tfvars')
+  include_examples('withRunningCluster',
```
mxinden: Is this and the following linebreaks needed or by accident?

tests/rspec/spec/azure/basic_spec.rb
```diff
@@ -3,5 +3,6 @@
 require 'shared_examples/k8s'
 
 RSpec.describe 'azure-basic' do
-  include_examples('withRunningCluster', '../smoke/azure/vars/basic.tfvars')
+  include_examples('withRunningCluster',
```
mxinden: Is this and the following linebreaks needed or by accident?

tests/rspec/lib/cluster.rb
mxinden: @cpanato As discussed, the `.done` files will stay. So why replacing the simple alias here?
```ruby
def update_cluster
  start
end
```

tests/rspec/spec/azure/basic_spec.rb
cpanato: done

tests/rspec/spec/azure/external_self_hosted_etcd_spec.rb
cpanato: done

tests/rspec/spec/azure/self_hosted_etcd_spec.rb
cpanato: done

tests/rspec/spec/metal/basic_spec.rb
cpanato: done

tests/rspec/lib/cluster.rb

PR #2308, opened 2017-11-03
modules/ignition: move metadata service to dropins

crawford:
The coreos-metadata unit is actually a drop-in; not a stand-alone
service unit.

This has not been tested at all.


mxinden:
Adding the `run-smoke-tests` and the `platform/aws` label for now, for basic smoke testing.

mxinden:
ok to test

PR #2309, opened 2017-11-04
fix external etcd cluster support.

coresolve:
With this change:
If you set `tectonic_etcd_tls_enabled`
- We will use the `https` scheme.
- The default remains true

If you set `tectonic_etcd_servers`
- We will assume we are not generating certs for the etcd cluster
- Optionally, you can provide `tectonic_etcd_ca_cert_path` if needed.


squat:
ok to test

squat:
ok to test

coresolve:
Got it will address

coresolve:
ok to test

coresolve:
Looks like this flaked on the `azure_private_external` test. Thoughts @squat?

coresolve:
ok to test

PR #2319, opened 2017-11-06
platform/vmware: Disable coreos-metadata

lander2k2:
This issue came up in testing 1.7.9 on vmware platform.  VMware is not a supported platform in coreos-metadata.  This change will turn off using that utility.


lander2k2:
This fix was included in: https://github.com/coreos/tectonic-installer/pull/2328/files

PR #2320, opened 2017-11-07
Allow specifying of AWS profile

aknuds1:
Allow specifying the AWS credentials profile to use with the AWS platform. As far as I can tell this should work fine, it's useful to me as I juggle different AWS profiles.

### DIFF
--- a/platforms/aws/variables.tf
+++ b/platforms/aws/variables.tf
@@ -7,6 +7,15 @@ EOF
   default = "1.0"
 }
 
+variable "tectonic_aws_profile" {
+  description = <<EOF
+(optional) This declares the AWS credentials profile to use.
+EOF
+
+  type    = "string"
+  default = "default"
+}
+
 variable "tectonic_aws_ssh_key" {
   type        = "string"
   description = "Name of an SSH key located within the AWS region. Example: coreos-user."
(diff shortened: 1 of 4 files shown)


aknuds1:
Thanks @squat. How does it look after my most recent changes?

platforms/aws/variables.tf
```diff
@@ -7,6 +7,14 @@ EOF
   default = "1.0"
 }
 
+variable "tectonic_aws_profile" {
+  description =  <<EOF
+(optional) This declares the AWS credentials profile to use.
+EOF
```
squat: lines 12 and 13 need to be indented. Please run `terraform fmt platforms/aws/variables.tf`. Also, we will need to regenerate docs. Please run `make docs examples` and commit the result as a separate commit.

aknuds1: Are you sure the heredoc should be indented? If you look at `tectonic_aws_config_version`, the heredoc lines are not indented. It makes sense to me, as the indentation wouldn't only be stylistic but also be part of the documentation string.

aknuds1:
@squat I just ran `terraform fmt` btw and it made some minor whitespace changes, which I checked in.

squat:
Looks indeed like we have a mix of indentation; `tectonic_aws_etcd_ec2_type` is indented but most of the other vars are not. Please keep you original indentation and add any new files that change after `make docs examples`

aknuds1:
@squat I made and updated docs!

squat:
ok to test

aknuds1:
Thanks @squat! BTW, I would appreciate very much if you had the time to test my DigitalOcean PR, it currently works both as self hosted and not! :)

I have to apply my other PR to force overlay2 Docker storage driver though, for mysterious reasons this isn't applied by default.

PR #2326, opened 2017-11-07
config.tf: Update tectonic-cluo-operator to v0.2.4

dghubble:
* Updates tectonic-torcx which sets the docker versions to install for Kubernetes 1.8 and 1.9


squeed:
This is to take in coreos/tectonic-torcx#53.

diegs:
@dghubble @squeed @lucab I don't believe we need to get this into 1.7.9. As discussed in the shared doc, we are going to make sure that TCLUO (and therefore tectonic-torcx) is updated before KVO in the 1.8 release. So these changes will be deployed up *before* KVO starts updating the control plane and kubelets to 1.8.

lucab:
@diegs right, if in the meanwhile that ordering is sorted out and committed, this won't be order sensitive anymore. As both me and @dghubble will be away a few days, this is already up and can be merged "just in case".

diegs:
@lucab certainly fine to merge now, just wanted to clarify that it does not need to be cherry-picked into the 1.7.9 release.

PR #2327, opened 2017-11-07
Terraform vars: Make secrets internal

joshix:
Make secrets (keys, location/regions, admin name/passwd) `(internal)` tf vars in config.tf and platforms/azure/variables.tf.
    
Ref: https://jira.prod.coreos.systems/browse/TECREL-126
https://jira.prod.coreos.systems/browse/DOCS-209


examples/terraform.tfvars.aws
```diff
@@ -24,12 +11,17 @@ tectonic_admin_password = ""
...
+//   default = "t2.medium"
+// }
+// 
+// variable "tectonic_aws_etcd_extra_sg_ids" {
+//   description = <<EOF
 // (optional) List of additional security group IDs for etcd nodes.
 // 
 // Example: `["sg-51530134", "sg-b253d7cc"]`
-// tectonic_aws_etcd_extra_sg_ids = ""
+tectonic_aws_etcd_ec2_type = ""
```
robszumski: is this missing a default value?

joshix: Seems like it -- the changes in this stanza result from https://github.com/coreos/tectonic-installer/pull/2322, I think.

s-urbaniak: @joshix please update your local copy of `terraform-examples` to the newest version. There used to be an early bug which caused variables to be rendered in a wrong way. I re-executed `make docs examples` and have a clean state, as is also reported by CI.

I will update this PR accordingly.

Issue #2262, opened 2017-10-30
Azure Installation fail


jbucar:
## What keywords did you search in tectonic-installer issues before filing this one?

azure terraform platform_fault_domain_count

## Is this a BUG REPORT or FEATURE REQUEST?

BUG REPORT

## Versions

* Tectonic version ([release](https://github.com/coreos/tectonic-installer/releases) or commit hash):
tectonic-1.7.5.1
* Terraform version (`terraform version`):
* Platform (aws|azure|openstack|metal):
azure

## What happened?

Installation fail on azure zone westus2

## What you expected to happen?

Installation ok

## How to reproduce it (as minimally and precisely as possible)?

Install a new cluster on westus2 zone

## Anything else we need to know?

This change fix the installation:

--- a/modules/azure/master-as/master.tf
+++ b/modules/azure/master-as/master.tf
@@ -3,7 +3,8 @@ resource "azurerm_availability_set" "tectonic_masters" {
   location            = "${var.location}"
   resource_group_name = "${var.resource_group_name}"
   managed             = true
-
+  platform_fault_domain_count = "2"
+  
   tags = "${merge(map(
     "Name", "${var.cluster_name}-masters",
     "tectonicClusterID", "${var.cluster_id}"),
diff --git a/modules/azure/worker-as/workers.tf b/modules/azure/worker-as/workers.tf
--- a/modules/azure/worker-as/workers.tf
+++ b/modules/azure/worker-as/workers.tf
@@ -3,6 +3,7 @@ resource "azurerm_availability_set" "tectonic_workers" {
   location            = "${var.location}"
   resource_group_name = "${var.resource_group_name}"
   managed             = true
+  platform_fault_domain_count = "2"
 
   tags = "${merge(map(
     "Name", "${var.cluster_name}-workers",

mxinden:
@jbucar Would you mind putting your git diff in a proper [markdown code block](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#code)? With the interpreted version your changes are difficult to understand.

jbucar:
Yes, sorry. The fix is hardcoded, but work on westus2

```
diff --git a/modules/azure/master-as/master.tf b/modules/azure/master-as/master.tf
--- a/modules/azure/master-as/master.tf
+++ b/modules/azure/master-as/master.tf
@@ -3,7 +3,8 @@ resource "azurerm_availability_set" "tectonic_masters" {
   location            = "${var.location}"
[...]

PR #2333, opened 2017-11-08
Azure: fix build failures in regions with just 2 fault domains

alexsomesan:
We've seen users report build failures on Azure in certain locations. This was due to mismatched number of fault domains between the default in Availability Set resource of Terraform and the actual number of fault domains each region has. The Terraform resource always defaults to 3.

This change introduces a (rather simplistic) mechanism to determine the number of fault domains for each region. It just looks it up in a static map, because the Azure API doesn't provide a way to query for this.


mxinden:
Fixes #2262

alexsomesan:
*@topiaruss*

Both northeurope and westeurope employ 3 failure domains so from the point
of view of this change it doesn’t actually make a difference in which one
we test.

Northeurope seems busier so we switched to westeurope to give our CI a
better chance of succeeding.

topiaruss:
@alexsomesan 
Thanks for your response. Makes total sense.

PR #2335, opened 2017-11-08
tests: Make sure `create_resource` raises error on failure

mxinden:
The index in rubys `each` iterator is zero indexed. Previously
`create_resource` would never raise an error if the last retry fails,
because `MAX_RETRIES > count` was always true. This patch makes sure
`create_resource` properly raises an error in case the last iteration
fails.


mxinden:
ok to test

mxinden:
ok to test

mxinden:
@cpanato can you review this again? CI went all green.

PR #2338, opened 2017-11-08
Azure: fix build failures in regions with just 2 fault domains (#2333)

alexsomesan:
Cherrypick of https://github.com/coreos/tectonic-installer/pull/2333

We've seen users report build failures on Azure in certain locations. This was due to mismatched number of fault domains between the default in Availability Set resource of Terraform and the actual number of fault domains each region has. The Terraform resource always defaults to 3.

This change introduces a (rather simplistic) mechanism to determine the number of fault domains for each region. It just looks it up in a static map, because the Azure API doesn't provide a way to query for this.

Additional test improvements included here:
* include regions with 2 fault domains
* don't run in "northcentralus" region

/cc: @sym3tri @sudhaponnaganti @Quentin-M


alexsomesan:
ok to test

alexsomesan:
ok to test

PR #2342, opened 2017-11-09
frontend: AWS Credentials page validation fixes

kyoto:
3 related fixes for validation on the AWS Credentials page.
- Fix `getExtraStuff()` to always return a promise. Looks like this was an old bug that only recently became visible due to AWS Credentials page changes.
- Add `asyncValidator()` for AWS regions field. Needed so that the containing form
notices when the API call fails and the field is therefore invalid.
- Delay `/aws/iam-roles` API call until AWS Credentials page form has been completed. We don't need a region to call `/aws/iam-roles`, so the call we being made as soon as the AWS id and secret key were provided. Unfortunately, that could result in lots of unnecessary API calls as the values are entered. To avoid this, wait until the whole form (include the region dropdown) is completed.


kyoto:
@squat No need to cherry pick as it was PR #2312 that introduced / uncovered these bugs.

PR #2315, opened 2017-11-06
platform/gcp: add ssh tweaks to upload assets after reboot

enxebre:

PR #2349, opened 2017-11-09
platform/gcp: add move-assets service

enxebre:
Problem:
- There is currently a potential race issue between ssh terraform provisioners and reboot triggered by torcx
- Side effects produced by two step terraform file provisioner + terraform remote-exec + reboot

Potential solution:
- This removes remote-exec
- This adds a dependency for the bootstraper node with move-assets while keeping the other masters functional - "such as ConditionPathExists=, ConditionPathIsSymbolicLink=, … — see below) do not cause the start job of a unit with a Requires= dependency on it to fail”
- This would supersede https://github.com/coreos/tectonic-installer/pull/2315

Adding this here for testing purposes with do not merge label as we are still working on a better solution

### DIFF
--- a/modules/gcp/master-igm/ignition.tf
+++ b/modules/gcp/master-igm/ignition.tf
@@ -13,7 +13,8 @@ data "ignition_config" "main" {
     var.ign_bootkube_service_id,
     var.ign_tectonic_service_id,
     var.ign_bootkube_path_unit_id,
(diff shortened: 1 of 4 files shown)


enxebre:
closing in favor of https://github.com/coreos/tectonic-installer/pull/2364

PR #2352, opened 2017-11-10
tests: Stash clean repository instead of repeated checkout

mxinden:


mxinden:
ok to test

Jenkinsfile
```diff
@@ -117,10 +117,11 @@ pipeline {
             try {
               timeout(time: 10, unit: 'MINUTES') {
                 forcefullyCleanWorkspace()
-                // First run `checkout scm` and then `printLogstashAttributes`
-                // as `printLogstashAttributes` accesses scm related variables
-                // like `env.CHANGE_ID`.
                 checkout scm
+                stash name: 'clean-repo', excludes: 'installer/vendor/**,tests/smoke/vendor/**,images/tectonic-stats-extender/vendor/**'
```
mxinden: @cpanato This reduces the stash size from ~5500 files down to 725 files. Hope this brings more stability and speed. We need to keep in mind, that the vendor folders are not included in the `clean-repo` stash.

cpanato: do we need to have the vendor folders for our tests?

PR #2357, opened 2017-11-10
tests: add docker logs

cpanato:
fixes: https://jira.prod.coreos.systems/browse/INST-622

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -55,7 +55,7 @@ pipeline {
     // Individual steps have stricter timeouts. 300 minutes should be never reached.
     timeout(time:5, unit:'HOURS')
     timestamps()
-    buildDiscarder(logRotator(numToKeepStr:'100'))
+    buildDiscarder(logRotator(numToKeepStr:'20', artifactNumToKeepStr: '20'))
   }
   parameters {
     string(
@@ -264,6 +264,7 @@ pipeline {
         TECTONIC_INSTALLER_ROLE = 'tectonic-installer'
         GRAFITI_DELETER_ROLE = 'grafiti-deleter'
         TF_VAR_tectonic_container_images = "${params.hyperkube_image}"
+        TF_VAR_tectonic_kubelet_debug_config = "--minimum-container-ttl-duration=8h --maximum-dead-containers-per-container=9999 --maximum-dead-containers=9999"
         GOOGLE_PROJECT = "tectonic-installer"
       }
       steps {
@@ -343,6 +344,7 @@ pipeline {
                   throw error
                 } finally {
                   reportStatusToGithub((err == null) ? 'success' : 'failure', specFile)
+                  archiveArtifacts allowEmptyArchive: true, artifacts: 'build/**/logs/**/*'
                   cleanWs notFailBuild: true
                 }
               }
@@ -439,6 +441,7 @@ def runRSpecTest(testFilePath, dockerArgs) {
         throw error
       } finally {
         reportStatusToGithub((err == null) ? 'success' : 'failure', testFilePath)
+        archiveArtifacts allowEmptyArchive: true, artifacts: 'build/**/logs/**/*'
         cleanWs notFailBuild: true
       }
 
--- a/tests/rspec/lib/cluster.rb
+++ b/tests/rspec/lib/cluster.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require 'cluster_support'
 require 'fileutils'
 require 'jenkins'
 require 'kubectl_helpers'
@@ -190,13 +191,15 @@ class Cluster
     wait_for_service('bootkube', ips)
     wait_for_service('tectonic', ips)
     puts 'HOORAY! The cluster is up'
-  rescue
+  rescue => e
     ips.each do |master_ip|
-      ['bootkube', 'tectonic', 'kubelet', 'k8s-node-bootstrap'].each do |s|
-        print_service_logs(master_ip, s)
+      save_docker_logs(master_ip, @name)
+
+      ['bootkube', 'tectonic', 'kubelet', 'k8s-node-bootstrap'].each do |service|
+        print_service_logs(master_ip, service, @name)
       end
     end
-    raise
+    raise "Error in wait_for_bootstrapping: #{e}"
   end
 
   def wait_for_service(service, ips)
@@ -234,18 +237,5 @@ class Cluster
     end
 
     false
-  end
-
-  def print_service_logs(ip, service)
-    command = "journalctl --no-pager -u '#{service}'"
-    begin
-      stdout, stderr, exitcode = ssh_exec(ip, command)
-      puts "Journal of #{service} service on #{ip} (exitcode #{exitcode})"
-      puts "Standard output: \n#{stdout}"
-      puts "Standard error: \n#{stderr}"
-      puts "End of journal of #{service} service on #{ip}"
-    rescue => e
-      puts "Cannot retrieve logs of service #{service} - failed to ssh exec on ip #{ip} with: #{e}"
-    end
   end
 end
--- a/tests/rspec/lib/cluster_support.rb
+++ b/tests/rspec/lib/cluster_support.rb
@@ -1,0 +1,58 @@
+# frozen_string_literal: true
+
+require 'fileutils'
+
+def save_docker_logs(ip, cluster_name)
+  # Get the Container IDs
+  command = "docker ps -a --format '{{.ID}} {{.Names}}'"
+  stdout, stderr, exitcode = ssh_exec(ip, command)
+  if exitcode != 0
+    puts "failed to execut docker ps on #{ip} (exitcode #{exitcode})"
+    puts1 "Standard error: \n#{stderr}"
+    return
+  end
+
+  containers_id = stdout.split("\n")
+
+  # For each container get the docker logs and save in a file
+  containers_id.each do |container|
+    id, image_name = container.split(' ')
+    command_log = "docker logs #{id}"
+    puts "Trying to get the docker logs: #{command_log} - image: #{image_name}"
+    stdout, stderr, exitcode = ssh_exec(ip, command_log)
+    output = ''
+    output << "Docker Logs of #{image_name} on #{ip} (exitcode #{exitcode})\n"
+    output << "Standard output: \n#{stdout}"
+    output << "\nStandard error: \n#{stderr}"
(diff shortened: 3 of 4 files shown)


Jenkinsfile
```diff
@@ -263,6 +263,7 @@ pipeline {
         TECTONIC_INSTALLER_ROLE = 'tectonic-installer'
         GRAFITI_DELETER_ROLE = 'grafiti-deleter'
         TF_VAR_tectonic_container_images = "${params.hyperkube_image}"
+        TF_VAR_tectonic_kubelet_debug_config = "--minimum-container-ttl-duration=8h --maximum-dead-containers-per-container=9999 --maximum-dead-containers=9999"
```
alexsomesan: Won't this conflict with @s-urbaniak PR?

cpanato: @alexsomesan the @s-urbaniak's PR add the functionality and here we are using that 😄

tests/rspec/lib/cluster_support.rb
```diff
@@ -0,0 +1,54 @@
...
+
+require 'fileutils'
+
+def save_docker_logs(ip, cluster_name)
+  docker_logs_path = "logs/docker_logs_#{cluster_name}"
+  FileUtils.mkdir_p(docker_logs_path)
+  begin
+    # Get the Container IDs
+    command = "docker ps -a --format '{{.ID}} {{.Names}}'"
+    stdout, stderr, exitcode = ssh_exec(ip, command)
```
mxinden: As far as I can tell `stderr` and the `exitcode` is unused. Can you check whether they are nil / zero before continuing. Otherwise we loose maybe helpful error messages.

Jenkinsfile
s-urbaniak: Teamwork! :-)
[https://media.giphy.com/media/TTfj0eequY6ru/giphy.gif](https://media1.giphy.com/media/26h0oQ3erWGYjdSi4/giphy.gif)

tests/rspec/lib/cluster_support.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+require 'fileutils'
+
+def save_docker_logs(ip, cluster_name)
+  docker_logs_path = "logs/docker_logs_#{cluster_name}"
```
mxinden: Shouldn't the `ip` be encoded in the file path as well? Otherwise we will overwrite the previous logs when we call `save_docker_logs` on two e.g. master machines.

tests/rspec/lib/cluster_support.rb
```diff
@@ -0,0 +1,54 @@
...
+      stdout, stderr, exitcode = ssh_exec(ip, command_log)
+      output = File.open("#{docker_logs_path}/#{ip}_#{image_name}.log", 'w+')
+      output << "Docker Logs of #{image_name} on #{ip} (exitcode #{exitcode})\n"
+      output << "Standard output: \n#{stdout}"
+      output << "\nStandard error: \n#{stderr}"
+      output << "\nEnd of journal of #{image_name} service on #{ip}"
+      output.close
+    end
+  rescue => e
+    puts "failed #{ip} with: #{e}"
```
mxinden: Can you add "what" failed here? Like e.g. `failed to retrieve docker logs on ip #{ip} with: #{e}`.

tests/rspec/lib/cluster_support.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  command = "journalctl --no-pager -u '#{service}'"
+  begin
+    stdout, stderr, exitcode = ssh_exec(ip, command)
+    puts "Journal of #{service} service on #{ip} (exitcode #{exitcode})"
+    puts "Standard output: \n#{stdout}"
+    puts "Standard error: \n#{stderr}"
+    puts "End of journal of #{service} service on #{ip}"
+
+    output = File.open("#{service_logs_path}/#{ip}_#{service}.log", 'w+')
```
mxinden: This seems to duplicate the logic in `save_docker_logs`. How about extracting it into a function instead?

tests/rspec/lib/cluster_support.rb
```diff
@@ -0,0 +1,54 @@
...
+      output << "\nEnd of journal of #{image_name} service on #{ip}"
+      output.close
+    end
+  rescue => e
+    puts "failed #{ip} with: #{e}"
+  end
+end
+
+def print_service_logs(ip, service, cluster_name)
+  service_logs_path = "logs/service_logs_#{cluster_name}"
```
mxinden: Am I right, that these logs will be saved in `tests/rspec/logs`? How about saving them in `build/#{CLUSTER}/logs` instead. So whenever someone runs RSpec twice in parallel the logs will not be overwritten for one of the two executions.

tests/rspec/lib/cluster_support.rb
```diff
@@ -0,0 +1,54 @@
...
+  FileUtils.mkdir_p(service_logs_path)
+
+  command = "journalctl --no-pager -u '#{service}'"
+  begin
+    stdout, stderr, exitcode = ssh_exec(ip, command)
+    puts "Journal of #{service} service on #{ip} (exitcode #{exitcode})"
```
mxinden: As far as I can tell you want to both log the outputs as well as save them to a file, right? How about first writing everything  into `output` and then both print and save it? This way you don't need to format the strings twice.

tests/rspec/lib/cluster_support.rb
cpanato: true, thanks!

tests/rspec/lib/cluster_support.rb
cpanato: ok

tests/rspec/lib/cluster_support.rb
cpanato: I did some tests locally and something in the docker logs I saw some strings in the stderr

tests/rspec/lib/cluster_support.rb
cpanato: ok sure, just making my life a little bit easy 😄

tests/rspec/lib/cluster_support.rb
cpanato: ok

tests/rspec/lib/cluster_support.rb
cpanato: actually not because the filename will contain the ip

tests/rspec/lib/cluster_support.rb
mxinden: Yes exactly, but as far as I can tell, this code never looks at `stderr` or `exitcode`, right? So when `docker ps -a --format '{{.ID}} {{.Names}}'` fails, we never know why it failed.

tests/rspec/lib/cluster_support.rb
cpanato: done

tests/rspec/lib/cluster_support.rb
cpanato: done

tests/rspec/lib/cluster_support.rb
cpanato: done

tests/rspec/lib/cluster_support.rb
cpanato: done

tests/rspec/lib/cluster_support.rb
cpanato: yes right, made confusing will update

Jenkinsfile
```diff
@@ -439,6 +440,7 @@ def runRSpecTest(testFilePath, dockerArgs) {
         throw error
       } finally {
         reportStatusToGithub((err == null) ? 'success' : 'failure', testFilePath)
+        archiveArtifacts allowEmptyArchive: true, artifacts: 'tests/rspec/logs/**/*'
```
mxinden: How about adding the `artifactDaysToKeep` and `artifactNumToKeep` option in the `logRotator`? Or is this not needed?
https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/tasks/LogRotator.java#L70

tests/rspec/lib/cluster.rb
```diff
@@ -192,8 +193,10 @@ def wait_for_bootstrapping
     puts 'HOORAY! The cluster is up'
   rescue
```
mxinden: This will swallow any error messages that were thrown earlier. Please rescue like: `rescue => e`.

This helped me understand the rescue system of Ruby: https://robots.thoughtbot.com/rescue-standarderror-not-exception

tests/rspec/lib/cluster.rb
```diff
@@ -192,8 +193,10 @@ def wait_for_bootstrapping
...
     ips.each do |master_ip|
-      ['bootkube', 'tectonic', 'kubelet', 'k8s-node-bootstrap'].each do |s|
-        print_service_logs(master_ip, s)
+      save_docker_logs(master_ip, @name)
+
+      ['bootkube', 'tectonic', 'kubelet', 'k8s-node-bootstrap'].each do |service|
+        print_service_logs(master_ip, service, @name)
       end
     end
     raise
```
mxinden: Here you can raise the exception `e` you caught in line 194.

tests/rspec/lib/cluster_support.rb
```diff
@@ -0,0 +1,53 @@
...
+  containers_id.each do |container|
+    id, image_name = container.split(' ')
+    command_log = "docker logs #{id}"
+    puts "Trying to get the docker logs: #{command_log} - image: #{image_name}"
+    stdout, stderr, exitcode = ssh_exec(ip, command_log)
+    output = ''
+    output << "Docker Logs of #{image_name} on #{ip} (exitcode #{exitcode})\n"
+    output << "Standard output: \n#{stdout}"
+    output << "\nStandard error: \n#{stderr}"
+    output << "\nEnd of journal of #{image_name} service on #{ip}"
```
mxinden: s/service/container

tests/rspec/lib/cluster.rb

tests/rspec/lib/cluster.rb

tests/rspec/lib/cluster_support.rb
cpanato: done

Jenkinsfile

cpanato:
ok to test

cpanato:
ok to test

PR #2358, opened 2017-11-10
module/tectonic: Asset cleanup after successful installation

kalmog:
Partially fixes #INST-475


s-urbaniak:
@kalmog CI fails with:
```
In ../modules/tectonic/resources/tectonic.sh line 107:
  rm -rf $ASSETS_PATH/*
         ^-- SC2115: Use "${var:?}" to ensure this never expands to /* .
         ^-- SC2086: Double quote to prevent globbing and word splitting.
```

s-urbaniak:
as discussed OOB: let's also tackle the cleanup of assets on S3 here.

kalmog:
test this please

kalmog:
test this please

modules/tectonic/resources/tectonic.sh
```diff
@@ -102,6 +102,11 @@ wait_for_pods() {
   set -e
 }
 
+asset_cleanup() {
+  echo "Cleaning up installation assets"
+  rm -rf "$${ASSETS_PATH:?}/"*
```
squat: Why `"$${ASSETS_PATH:?}/"*` and not just `"$ASSETS_PATH/*"`? All other parts of the file assume ASSETS_PATH exists and make that assumption before the asset_cleanup function is run. If we want to ensure the variable is set, then set the sh option `-u` at the top of the file e.g.:
```sh
#!/bin/sh
set -eu
```

modules/aws/master-asg/resources/rm-assets.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+set -e
+
+# shellcheck disable=SC2086,SC2154
+/usr/bin/docker run \
```
squat: This operation could fail due to transient network issues. Please consider adding a loop around this like we do in s3-puller.sh: https://github.com/coreos/tectonic-installer/blob/master/modules/ignition/resources/bin/s3-puller.sh#L33

modules/tectonic/resources/tectonic.sh
kalmog: this is due to this lint check: 

```In ../modules/tectonic/resources/tectonic.sh line 107:
  rm -rf "$ASSETS_PATH/*"
         ^-- SC2115: Use "${var:?}" to ensure this never expands to /* .```

https://github.com/koalaman/shellcheck/wiki/SC2115

Comment (squat) [reply modules/tectonic/resources/tectonic.sh]: OK, well, either way. Not a blocker from me.

**Comment by kalmog** [comment]
test this please

File: modules/ignition/resources/services/rm-assets.service [hunk 3]
```diff
@@ -0,0 +1,18 @@
+[Unit]
```
enxebre: Couldn't we call the s3 remove function from the puller script so we pull, move and remove in one go and we avoid to introduce a new systemd dependency?

squat:
@s-urbaniak @kalmog does anything ever set /run/metadata/master to `false` on the leader node? This script bails if /run/metadata/master is `true`, so it will never run on the leader. However, on non-leaders, this script will run /run/metadata/master is false. This logic seems funny to me. Can someone chime in?

s-urbaniak:
@squat good catch, I am looking into the logic right now and will submit a PR.

PR #2364, opened 2017-11-10
platform/gcp: add gcs support for upload assets

enxebre:
This moves from ssh model for assets to cloud storage pulling model
Remove unnecessary lb for sshing into masters
Allow the gcp tests retrieving the master ips by using gcloud


enxebre:
ok to test

cpanato:
ok to test

tests/rspec/lib/gcp_cluster.rb
```diff
@@ -24,19 +30,19 @@ def credentials_defined?
...
-    ip_addresses = []
-    Dir.chdir(@build_path) do
-      ip_address = `echo module.network.ssh_master_ip | terraform console ../../platforms/gcp`.chomp
-      if ip_address.empty?
-        raise 'should get the master_ip_address to use in the tests.'
-      end
-      ip_addresses.push(ip_address)
-      ip_addresses
-    end
+    gcloud_command = 'compute instances list \
```
cpanato: i dont have a strong opinion on this, but I guess we should use the ips we get from the terraform console.

enxebre: Thanks for feedback @cpanato 
We use a manage group https://cloud.google.com/compute/docs/instance-groups/ and instance templates for masters machines so there's no awareness from terraform about the specific machine ips.
We could use a ruby client library but all are quiet abandoned. Gcloud seems to have more energy focus and support from google and this is specific use with no potencial grown so it seems the simplest way to go to me here

tests/rspec/lib/gcloud_helper.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+# GcloudHelper contains helper functions to interact with the command line tool
+# gcloud
+class GcloudHelper
+  attr_accessor :authenticated
+
+  def initialize
+    @authenticated = true
+    auth_command = 'gcloud auth activate-service-account --key-file=${GOOGLE_APPLICATION_CREDENTIALS}'
```
cpanato: use `auth_command = "gcloud auth activate-service-account --key-file=\"${GOOGLE_APPLICATION_CREDENTIALS}\""`

enxebre:
@s-urbaniak @alexsomesan any thoughts?

modules/gcp/master-igm/resources/init-assets.sh
```diff
@@ -0,0 +1,24 @@
...
+    --network=host \
+    --entrypoint=/detect-master.sh \
+    ${gcloudsdk_image}
+
+MASTER=$(cat /run/metadata/master)
+if [ "$MASTER" != "true" ]; then
+    exit 0
+fi
+# Download the assets from GCS
+/usr/bin/bash /opt/gcs-puller.sh ${assets_gcs_location} /var/tmp/tectonic.zip
```
s-urbaniak: will `gcs-puller.sh` retry? If so, how often, and does it have back-off?

modules/gcp/master-igm/resources/detect-master.sh
```diff
@@ -0,0 +1,38 @@
...
+  INSTANCE_ID_LIST=$(echo "$INSTANCES_INFO" | jq -r  "sort_by(.id) | .[].id")
+  TARGET_SIZE=$(echo "$${INSTACE_GROUP_INFO}" | jq -r ".targetSize")
+  CURRENT_SIZE=$(echo "$${INSTACE_GROUP_INFO}" | jq -r ".currentActions.none")
+  if [ "$TARGET_SIZE" == "$CURRENT_SIZE" ]; then
+    break
+  fi
+
+  echo "Waiting for the IGM to be at desired capacity (Desired: $TARGET_SIZE, Current: $CURRENT_SIZE)"
+  sleep 15
+done
```
s-urbaniak: I suggest we move this to Go code eventually.

s-urbaniak: actually ... rather sooner than later ;-) Do you mind to open an issue for that?

s-urbaniak:
@enxebre another question: is `init-assets.service` not being used here? we already have one in `modules/aws` and `modules/ignition`. I believe we should unify them (the aws one includes the correct implementation).

modules/gcp/master-igm/ignition.tf
```diff
@@ -3,17 +3,21 @@ data "ignition_config" "main" {
...
+    "${data.ignition_file.init_assets.id}",
+    "${data.ignition_file.detect_master.id}",
   ]
 
   systemd = ["${compact(list(
     var.ign_docker_dropin_id,
     var.ign_locksmithd_service_id,
     var.ign_kubelet_service_id,
     var.ign_k8s_node_bootstrap_service_id,
+    var.ign_init_assets_service_id,
```
s-urbaniak: Ah it is used here, sorry ;-) Please update `init-assets.service` inside `modules/igntion` and use the implementation from aws.

enxebre: Done!

modules/gcp/master-igm/resources/init-assets.sh
enxebre: added basic retry now. Thanks!

modules/gcp/master-igm/resources/init-assets.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+# shellcheck disable=SC2086,SC2154
+set -e
+
+mkdir -p /run/metadata
+/usr/bin/docker run \
```
squat: This invocation has a ton of hidden network operations inside of `detect-master`, which could all fail. I suggest we add a retry around this as done in: https://github.com/coreos/tectonic-installer/pull/2384/files#diff-a8f67da9bff3df820b94ce8b1c1b3cd5R5

modules/ignition/resources/bin/gcs-puller.sh
```diff
@@ -0,0 +1,23 @@
...
+set -x
+
+if [ "$#" -ne "2" ]; then
+  echo "Usage: $0 location destination"
+  exit 1
+fi
+
+# Overriding /etc/profile.d/google-cloud-sdk.sh
+DOCKER_IMAGE=${gcloudsdk_image}
+alias gsutil="(docker images $DOCKER_IMAGE  || docker pull $DOCKER_IMAGE) > /dev/null; \
```
squat: Docker automatically downloads images if they are not present. I would prefer to rely on that build-in logic in docker rather than on this alias, which is a little over-engineered IMO.

modules/ignition/resources/bin/gcs-puller.sh
```diff
@@ -0,0 +1,23 @@
...
+docker run -i --net=host -v $HOME/.config:/.config -v /tmp:/gs $DOCKER_IMAGE gsutil"
+shopt -s expand_aliases
+
+assets=$(basename $${1})
+until gsutil cp gs://$${1} /gs/$${assets}; do
```
squat: The more appropriate construct is a POSIX function, rather than an ALIAS IMO. This way the operation does not have to be forced into a tricky one-liner.

Consider defining:
```sh
gsutil() {
    docker run --net=host -v...
}
```

modules/ignition/resources/bin/gcs-puller.sh
```diff
@@ -0,0 +1,23 @@
...
+
+if [ "$#" -ne "2" ]; then
+  echo "Usage: $0 location destination"
+  exit 1
+fi
+
+# Overriding /etc/profile.d/google-cloud-sdk.sh
+DOCKER_IMAGE=${gcloudsdk_image}
+alias gsutil="(docker images $DOCKER_IMAGE  || docker pull $DOCKER_IMAGE) > /dev/null; \
+docker run -i --net=host -v $HOME/.config:/.config -v /tmp:/gs $DOCKER_IMAGE gsutil"
```
squat: does this container really need to be interactive?

modules/ignition/resources/bin/gcs-puller.sh
```diff
@@ -0,0 +1,23 @@
...
+shopt -s expand_aliases
+
+assets=$(basename $${1})
+until gsutil cp gs://$${1} /gs/$${assets}; do
+    echo "Could not pull from GCS, retrying in 5 seconds"
+    sleep 5
+done
+
+gsutil rm gs://$${1}
```
squat: What if the remove operation fails due to some network issue?

enxebre:
@squat awesome feedback. Addressed the comments, thanks!

squat:
Thanks for cleaning that up so quickly!

PR #2365, opened 2017-11-11
frontend: Fix /terraform/apply error message display

kyoto:


robszumski:
What situation does this improve?

kyoto:
@robszumski If the installer binary was killed at a certain point in the install process, a JS error was thrown instead of gracefully displaying an error message.

PR #2372, opened 2017-11-13
test: fix typo / some error in frozen strings and save archiveArtifacts

cpanato:
and also fix the error: cant modify frozen strings and change the logs folder

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -344,7 +344,7 @@ pipeline {
                   throw error
                 } finally {
                   reportStatusToGithub((err == null) ? 'success' : 'failure', specFile)
-                  archiveArtifacts allowEmptyArchive: true, artifacts: 'build/**/logs/**/*'
+                  archiveArtifacts allowEmptyArchive: true, artifacts: 'tests/rspec/logs/**/*.log'
                   cleanWs notFailBuild: true
                 }
               }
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -344,7 +344,7 @@ pipeline {
                   throw error
                 } finally {
                   reportStatusToGithub((err == null) ? 'success' : 'failure', specFile)
-                  archiveArtifacts allowEmptyArchive: true, artifacts: 'build/**/logs/**/*'
+                  archiveArtifacts allowEmptyArchive: true, artifacts: 'tests/rspec/logs/**/*.log'
```
mxinden: Is it necessary to move this out of the `build` folder?

tests/rspec/lib/cluster_support.rb
```diff
@@ -21,10 +21,10 @@ def save_docker_logs(ip, cluster_name)
     puts "Trying to get the docker logs: #{command_log} - image: #{image_name}"
     stdout, stderr, exitcode = ssh_exec(ip, command_log)
     output = ''
-    output << "Docker Logs of #{image_name} on #{ip} (exitcode #{exitcode})\n"
```
mxinden: Whats is the needed difference between `<<` and `+=`?

Jenkinsfile
cpanato: we clean the build folder at the end of the tests. It was like that before, then I realize we do a cleanup then all logs are wiped. Then I move to the rspec, also because is part of the tests.

tests/rspec/lib/cluster_support.rb
cpanato: https://stackoverflow.com/questions/41375818/operator-appears-to-modify-frozen-string

PR #2374, opened 2017-11-13
tests: change reported spec names

squat:
cc @mxinden @cpanato


mxinden:
@squat Thanks for the improvement!

PR #2377, opened 2017-11-14
Fixes kubelet.env location per #FR-445

coresolve:
This moves the kubelet.env file back to /etc/kubernetes/kubelet.env
This affects bare metal and vmware installations.

### DIFF
--- a/modules/ignition/assets.tf
+++ b/modules/ignition/assets.tf
@@ -108,7 +108,7 @@ data "template_file" "installer_kubelet_env" {
 
 data "ignition_file" "installer_kubelet_env" {
   filesystem = "root"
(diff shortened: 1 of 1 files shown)


s-urbaniak:
As stated in https://jira.prod.coreos.systems/browse/FR-445, we have a different issue here. It seems that the torcx bootstrapper failed.

lucab:
Nack. The installer path is correct and the flow is documented:
 * https://github.com/coreos/tectonic-installer/blob/master/Documentation/dev/node-bootstrap-flow.md#k8s-node-bootstrapservice
 * https://github.com/coreos/tectonic-torcx/blob/master/Documentation/bootstrapper.md#sources-of-information

squat:
Closing this issue as discussed OOB.

PR #2380, opened 2017-11-14
frontend: Fix handling of errors when loading IAM roles from AWS API

kyoto:
If `/aws/iam-roles` returned an error, that error case was not being handled, resulting in a JS error.

screenshot-1


sudhaponnaganti:
@kyoto  @squat  - Can you merge this in to 1.7.9 Branch as well?

kyoto:
@sudhaponnaganti @squat Cherry pick PR: #2394

PR #2384, opened 2017-11-14
modules: retry init assets

squat:
This commit makees several modifications to ensure the init-assets
service will retry any network operations, e.g. wget, curl, aws etc.
Rather than adding individual retries for each of the dozens of
operations that could fail, this commit adds retries to the docker
commands that encompass those operations.

Fixes: INST-341, INST-574

cc @s-urbaniak

### DIFF
--- a/modules/aws/master-asg/resources/detect-master.sh
+++ b/modules/aws/master-asg/resources/detect-master.sh
@@ -1,4 +1,6 @@
 #!/bin/bash
+set -e
+set -o pipefail
 
 # Wait for the ASG to run at the expected scale.
 while true; do
--- a/modules/ignition/resources/bin/s3-puller.sh
+++ b/modules/ignition/resources/bin/s3-puller.sh
@@ -1,23 +1,38 @@
 #!/bin/bash
+set -e
+set -o pipefail
 
 if [ "$#" -ne "2" ]; then
     echo "Usage: $0 location destination"
     exit 1
 fi
 
-# shellcheck disable=SC2086,SC2154,SC2016
-/usr/bin/docker run \
-    --volume /tmp:/tmp \
-    --network=host \
-    --env LOCATION="$1" \
-    --entrypoint=/bin/bash \
-    ${awscli_image} \
-    -c '
-        REGION=$(wget -q -O - http://169.254.169.254/latest/meta-data/placement/availability-zone | sed s'/[a-zA-Z]$//')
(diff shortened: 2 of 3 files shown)


modules/aws/master-asg/resources/detect-master.sh
```diff
@@ -1,4 +1,6 @@
 #!/bin/bash
+set -e
```
squat: This script has dozes of operations that could fail, but the script does not exit on failure. We should exit on failure and simply retry the entire process if there was a failure.

modules/ignition/resources/bin/s3-puller.sh
```diff
@@ -1,23 +1,38 @@
...
+        --volume /tmp:/tmp \
+        --network=host \
+        --env LOCATION=$LOCATION \
+        --env LOCATION_ESCAPED=$LOCATION_ESCAPED \
+        --entrypoint=/bin/bash \
+        ${awscli_image} \
+        -c '
+            set -e
+            set -o pipefail
+            REGION=$(wget -q -O - http://169.254.169.254/latest/meta-data/placement/availability-zone | sed '"'"'s/[a-zA-Z]$//'"'"')
```
squat: We need this complicated escape sequence because in the script's previous form, `/[a-zA-Z]$//` was not surrounded in single quotes, meaning two problems:
 1. `/[a-zA-Z]$//` could have variables expanded by the shell running s3-puller; and
2. the sed command was actually executing `sed s/[a-zA-Z]$//`, so the `/[a-zA-Z]$//`  could also variables expanded by the subshell created by docker.

modules/aws/master-asg/resources/detect-master.sh
s-urbaniak: 👍

modules/ignition/resources/bin/s3-puller.sh
s-urbaniak: good catch 👍 this also amplifies the necessity to this in code rather than in a shell.

squat:
looks like one of the tests suffered the route53 DNS issue:
```
smoke_test.go:118: failed with error: failed to connect to API server: Get https://aws-pr-2384-218aee-api.tectonic.dev.coreos.systems:443/version: dial tcp: lookup aws-pr-2384-218aee-api.tectonic.dev.coreos.systems on 172.31.0.2:53: no such host
```
cc @cpanato @s-urbaniak

squat:
ok to test

squat:
AWS test dns fix landed. Retrying tests

squat:
ok to test

PR #2396, opened 2017-11-15
modules/bootkube: be explicit about runtime

squat:
We made the effort to disable alpha API groups in Tectonic clusters, but
we do not explicitly set the --runtime-config to "". "" is the default
setting for the --runtime-config flag, but setting it explicitly, as
@diegs explains, allows us to detect conflicts on updates.

fixes INST-321


squat:
ok to test

squat:
ok to test

PR #2403, opened 2017-11-16
frontend: Don't set custom subnets unless those fields have been edited

kyoto:
Don't include `tectonic_aws_master_custom_subnets` or `tectonic_aws_worker_custom_subnets` in the generated tfvars file unless one of those subnet fields was edited. This will also work when
restoring .progress files since those files include the `dirty` fields.

Note that currently, just opening the advanced VPC options is enough to mark the subnet fields as `dirty`.

Also removes the `progress` attribute that was being passed to `toAWS_TF()` / `toBaremetal_TF()` because they don't use it.


kyoto:
ok to test

PR #2404, opened 2017-11-16
frontend: Fix resetting the AWS existing subnet dropdown

kyoto:
This fixes a bug where resetting the subnet IDs in Redux (e.g. by selecting a different VPC), did not reset the dropdowns to the default "Select a subnet" option.

If there was only one subnet in the dropdown, this bug stopped you selecting that single subnet. Clicking / changing any other form element on the page reset the subnet dropdowns, fixing the problem.


squat:
Tests passed. Status is not being updated by jenkins. Merging.

PR #2415, opened 2017-11-17
frontend: Remove <option> "selected" attribute

kyoto:
Oops, shouldn't use `selected` with React. Fixes a JS warning.


kyoto:
@squat ``Warning: Use the `defaultValue` or `value` props on <select> instead of setting `selected` on <option>.`` 🙂

PR #2416, opened 2017-11-17
frontend: Remove ClusterInfo.canNavigateForward hack

kyoto:
Delay initial field validation until `facts` API call has completed.

The `facts` response contains initial values for the Tectonic license and pull secret, so making this change allows us to remove the `ClusterInfo.canNavigateForward()` hack.


kyoto:
ok to test

Issue #2419, opened 2017-11-17
AWS: Can't bootstrap cluster without private endpoints


aknuds1:
## Versions

* Tectonic version ([release](https://github.com/coreos/tectonic-installer/releases) or commit hash): 19a6e0b6b756a391d77f40d22a7684f7f700fa25
* Terraform version (`terraform version`): v0.10.8
* Platform (aws|azure|openstack|metal): aws

## What happened?
I try to create a cluster with Tectonic Installer on the command line (`CLUSTER=codersociety PLATFORM=aws make apply`) configured to disable private endpoints (`tectonic_aws_private_endpoints = false`). This causes an error in detect-master.sh during the earliest stages of cluster bootstrapping, on account of there being no load balancer called tectonic-test-int:
```
An error occurred (LoadBalancerNotFound) when calling the DescribeInstanceHealth operation: There is no ACTIVE Load Balancer named 'tectonic-test-int'
```

## How to reproduce it (as minimally and precisely as possible)?
Use a configuration similar to the following (in terraform.tfvars), disabling `tectonic_aws_private_endpoints`:

```
tectonic_admin_email = "me@example.com"
tectonic_admin_password = "password1234"
tectonic_aws_assets_s3_bucket_name = "tectonic.kopscluster.com"
tectonic_aws_extra_tags = {"cluster" = "tectonic-test"}
tectonic_aws_private_endpoints = false
tectonic_aws_public_endpoints = true
tectonic_aws_region = "eu-central-1"
tectonic_aws_ssh_key = "me"
tectonic_base_domain = "kopscluster.com"
tectonic_cluster_name = "tectonic-test"
tectonic_license_path = "tectonic-license.txt"
tectonic_pull_secret_path = "config.json"
tectonic_master_count = "1"
tectonic_worker_count = "2"
tectonic_aws_profile = "coder-society"
tectonic_self_hosted_etcd = "enabled"
```

squat:
We were aware of issues with disabling private endpoints and created https://github.com/coreos/tectonic-installer/pull/2392 for that reason. @s-urbaniak can you PTAL?

squat:
Indeed, we are hardcoding a loadbalancer in detect-master.sh:

https://github.com/coreos/tectonic-installer/blob/master/modules/aws/master-asg/resources/detect-master.sh#L28

squat:
[...]

PR #2420, opened 2017-11-17
modules/aws: detect-master assumes lb

squat:
This commit fixes issue #2419.

This issue occurs because we recently added -o pipefail to the
detect-master.sh script. This causes a previously undetected issue to
surface and break bootstrapping for clusters with no private-endpoints.

API_HEALTHY=$(aws elb describe-instance-health --region="$REGION"
--load-balancer-name "$CLUSTER_NAME-int" | jq -r '[ .InstanceStates[] |
select(.State | contains("InService")) ] | length > 1')
is hardcoded to use the internal loadbalancer. Without pipefail, this
aws command would 404 and the error would get swallowed by the pipe.
API_HEALTHY would never be true, even if the API was actually healthy,
so the rest of the script would run.

Now, since the pipefail option is set, that command correctly errors out
and the script execution is aborted.

We simply need to template the loadbalancer name based on whether
private endpoints are enabled or disabled.

cc @s-urbaniak

### DIFF
--- a/modules/aws/master-asg/ignition.tf
+++ b/modules/aws/master-asg/ignition.tf
@@ -23,13 +23,21 @@ data "ignition_config" "main" {
    ))}"]
 }
 
+data "template_file" "detect_master" {
+  template = "${file("${path.module}/resources/detect-master.sh")}"
+
+  vars {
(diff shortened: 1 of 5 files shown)


squat:
ok to test

squat:
ok to test

PR #2432, opened 2017-11-20
removed outdated ROADMAP file

sichvoge:
@sym3tri As per our discussion last week, the current ROADMAP file is significantly outdated. 

If no one updates it frequently, I suggest to remove it as it does introduce a lot of confusion on what we are working on or have finished. There are other ways to communicate working items. We can discuss that this week if you like.


sichvoge:
My PR has been approved 🎉  😄

PR #2437, opened 2017-11-21
test/ci: replace stash with env variable

cpanato:
we observed some issues in the initial steps of the Jenkins pipeline regarding the stash the original commit. since this is just a hash and not a file or other thing it is easier to create a variable and use that.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -34,7 +34,7 @@ creds = [
   ]
 ]
 
-quay_creds = [
+quayCreds = [
   usernamePassword(
     credentialsId: 'quay-robot',
     passwordVariable: 'QUAY_ROBOT_SECRET',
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -44,6 +44,7 @@ quay_creds = [
 
 default_builder_image = 'quay.io/coreos/tectonic-builder:v1.41'
 tectonic_smoke_test_env_image = 'quay.io/coreos/tectonic-smoke-test-env:v5.9'
+def original_commit_id = 'UNKNOWN'
```
mxinden: We should stick to either camelCase or snake_case. The [Groovy documentation](https://stackoverflow.com/a/14534685/4011134) follows the latter. Would you mind changing this and the two definitions above to camelCase?

cpanato: done!! 😄

PR #2440, opened 2017-11-21
tests: change logstash logic

erikkn:
Put all the information in the filename and let Logstash parse through the filename


erikkn:
ok to test

erikkn:
@mxinden

PR #2452, opened 2017-11-24
frontend: Fix Nightwatch test diff output and make it more readable

kyoto:


installer/frontend/ui-tests/utils/terraformTfvars.js
```diff
@@ -11,7 +11,7 @@ const diffTfvars = (client, assetsZip, expected) => {
       if (diff !== undefined) {
         client.assert.fail(
           'The following terraform.tfvars attributes differ from their expected value: ' +
-          diff.map(({attr, lhs, rhs}) => `${attr} (expected: ${rhs}, got: ${lhs})`).join(', ')
+          diff.map(({path: attr, lhs, rhs}) => `\n  ${attr.join('.')} (expected: ${rhs}, got: ${lhs})`)
```
squat: Looks like you have to change the `attr` to `path`

kyoto: The problem is that a variable `path` already exists in the outer scope, but using `path: attr` did look odd. I've changed it to use a loop variable instead so that it doesn't look strange.

kyoto:
ok to test

Issue #51, opened 2017-03-15
vmware: support vsphere cloud provider


alekssaul:
VMware has Kubernetes `--cloud-provider=vsphere` support as described in documentation[0]  . Cloud Provider plugins allow mainly integration with VMware's storage stack for block storage, similar to aws-ebs integration. Apart from the block storage integration it also supports self discovery such as nodes adding metadata about their placement in VMware infrastructure.

Created a short demo in; https://youtu.be/gfjwwkTYNRQ (needs CoreOS SSO)

Opening the issue to discuss the right way to implement. Per upstream documentation implementation requires: "Provide` the cloud config file to each instance of kubelet, apiserver and controller manager via --cloud-config=<path to file>"

Since kube-api and controller-manager are self-hosted it's possible to inject --cloud-config via Kubernetes secret. However kubelet will require the configuration file to be on disk. Since we would need to deploy this file onto disk, it's probably best to have kube-api and controller-manager use hostpath mounts as well. Since kubelet already mounts /etc/kubernetes I've added /etc/kubernetes/vsphere.conf via ignition_file and update kubelet.service to use 
```
--cloud-config=/etc/kubernetes/vsphere.conf \
--cloud-provider=vsphere \
```
Also have to modify the assets for kube-api and controller-manager pods with

```
        - --cloud-provider=vsphere
        - --cloud-config=/etc/kubernetes/vsphere.conf
```

```
        - mountPath: /etc/kubernetes/vsphere.conf
          name: vsphere
          readOnly: true
        - mountPath: /sys/devices/virtual/dmi/id/product_uuid
          name: vspheredmi
```
for hostpaths. `/sys/devices/virtual/dmi/id/product_uuid` [0] is needed for vmware code to find itself in the virtual machine infrastructure. This was merged in 1.5.3, and prior to it operator had to manually ad Node's VMware Virtual Machine UUID.

Master and Worker nodes also need `disk.enableUUID = "1"` as part of their Terraform custom_configuration_parameters.

Happy to submit a PR if above sounds acceptable.

[0] https://kubernetes.io/docs/getting-started-guides/vsphere/ 
[1] https://github.com/kubernetes/kubernetes/blob/v1.5.3/pkg/cloudprovider/providers/vsphere/vsphere.go#L207-L260

alekssaul:
related to https://github.com/coreos/tectonic-installer/issues/84

alekssaul:
[...]

PR #2454, opened 2017-11-27
module/tectonic: Add default storage classes

### LINKED ISSUE #711: platforms/*: install the default storage classes
Since we don't run the add-on manager we don't' install the default storage classes. We should do this!

- [AWS](https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/storage-class/aws/default.yaml)
- [OpenStack](https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/storage-class/openstack/default.yaml)
- [etc](https://github.com/kubernetes/kubernetes/tree/master/cluster/addons/storage-class)

cc @polvi

kalmog:
Fixes #INST-284
Fixes #711


kalmog:
test this please

kalmog:
test this please

Documentation/variables/aws.md
```diff
@@ -6,6 +6,10 @@ This document gives an overview of variables used in the AWS platform of the Tec
 
 | Name | Description | Type | Default |
 |------|-------------|:----:|:-----:|
+| extra_storage_params | Set to true, if extra parmeters are needed for the StorageClass definition | string | `true` |
+| extra_storage_params_list | A list of extra storage parameters to add to the storage class configuration (platform dependent), formatted as a list of yaml values.<br><br>Example: [ "type: gp2", "diskformat: thin" ] | list | `<list>` |
+| storage_class_provisioner | The provisioner to use for creating the storage class | string | `kubernetes.io/aws-ebs` |
+| storage_metadata_name | A metadata name tag for the storage class | string | `gp2` |
```
s-urbaniak: the convention is to prefix all top-level variables with `tectonic_`. let's rename `extra_storage_params` to `tectonic_extra_storage_params`, etc.

Documentation/variables/aws.md
```diff
@@ -6,6 +6,10 @@ This document gives an overview of variables used in the AWS platform of the Tec
 
 | Name | Description | Type | Default |
 |------|-------------|:----:|:-----:|
+| extra_storage_params | Set to true, if extra parmeters are needed for the StorageClass definition | string | `true` |
+| extra_storage_params_list | A list of extra storage parameters to add to the storage class configuration (platform dependent), formatted as a list of yaml values.<br><br>Example: [ "type: gp2", "diskformat: thin" ] | list | `<list>` |
```
s-urbaniak: naming question: is it necessary to prefix with `extra_`? or would `tectonic_storage_params` be more concise?

modules/tectonic/assets.tf
```diff
@@ -12,37 +12,45 @@ resource "template_dir" "tectonic" {
...
+    prometheus_operator_base_image      = "${var.container_base_images["prometheus_operator"]}"
+    tectonic_monitoring_auth_base_image = "${var.container_base_images["tectonic_monitoring_auth"]}"
+
+    storage_class_provisioner = "${var.storage_class_provisioner}"
+    storage_metadata_name     = "${var.storage_metadata_name}"
 
+    // We need this hack to generate a conditional yaml block
+    // for storage classes, since the extra parameters differ
+    // from platform to platform
+    storage_parameters = "${var.extra_storage_params ? "parameters:\n  ${indent(2, join("\n", var.extra_storage_params_list))}" : ""}"
```
enxebre: couldn't we just check if extra_storage_params_list is not empty so no need to coordinate both variables?

platforms/gcp/variables.tf
```diff
@@ -92,3 +92,32 @@ variable "tectonic_gcp_ssh_key" {
   type        = "string"
   description = "(required) Path to an SSH public key file to be provisioned as the SSH key for the 'core' user."
 }
+
+variable "storage_class_provisioner" {
+  description = "The provisioner to use for creating the storage class"
+  default     = "kubernetes.io/gce-pd"
```
enxebre: just so we are aware this most likely won't work as gcp is currently not using --cloud-provider flag

modules/tectonic/assets.tf
kalmog: a list type parameter cannot be parsed as bool, hence the dual parameters:
```
* data.template_file.init: 1 error(s) occurred:

* data.template_file.init: At column 3, line 1: condition must be type bool, not type list in:

${var.extra_params_list ? "parameters:\n  ${indent(2, join("\n", var.extra_params_list))}" : ""}
```

platforms/gcp/variables.tf
kalmog: Good catch. Similar to #51. I believe we should open a separate ticket for this.

modules/tectonic/assets.tf
squat: @kalmog I believe the suggestions is to use https://www.terraform.io/docs/configuration/interpolation.html#length-list- to determine if the list is empty or not and make decisions based on that

enxebre: yep something like ```"${length(var.extra_storage_params) > 0 ? "parameters:\n  ${indent(2, join("\n", var.extra_storage_params_list))}" : ""}"```

cpanato:
@squat  @kalmog  this is still valid? if yes, can you please rebase?

kalmog:
@cpanato the logic is likely going to be moved to the KAO. I'll close this one for now.

cpanato:
thanks @kalmog

PR #2455, opened 2017-11-27
tests: log upload feature

erikkn:
Added the feature that Terraform logs are uploaded to S3 as well.

All files are temporary stored in the 'templogfiles' directory.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -389,7 +389,7 @@ pipeline {
           withCredentials(creds) {
             unstash 'clean-repo'
             sh """#!/bin/bash -xe
-            ./tests/jenkins-jobs/scripts/log-analyzer-copy.sh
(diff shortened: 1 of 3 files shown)


erikkn:
ok to test

erikkn:
ok to test

mxinden:
@erikkn Can you add a description to the commit so we know in the future what happened here?

Jenkinsfile
```diff
@@ -432,6 +432,7 @@ def runRSpecTest(testFilePath, dockerArgs) {
                 sh """#!/bin/bash -ex
                   cd tests/rspec
                   bundler exec rspec ${testFilePath}
+                  cd ../..
```
mxinden: Can you check if this is needed?

erikkn:
ok to test

PR #2457, opened 2017-11-28
Change GUI AWS test to output tfvars for smoke test

kyoto:
Changes a GUI end-to-end test (output `/installer/frontend/ui-tests/output/aws.tfvars`) and a smoke test (input `/tests/smoke/aws/vars/aws.tfvars.json`) so that they match as closely as possible. Future PRs will make them actually use the same file and do the same for a bare metal test.


cpanato:
ok to test

installer/frontend/ui-tests/output/aws.tfvars
```diff
@@ -14,10 +14,10 @@
   "tectonic_aws_worker_ec2_type": "m4.large",
   "tectonic_aws_worker_root_volume_size": 15,
   "tectonic_aws_worker_root_volume_type": "gp2",
-  "tectonic_base_domain": "tectonic.dev.coreos.systems",
+  "tectonic_base_domain": "tectonic-ci.de",
   "tectonic_cluster_cidr": "10.2.0.0/16",
-  "tectonic_cluster_name": "test",
-  "tectonic_dns_name": "test",
+  "tectonic_cluster_name": "aws-basic",
```
cpanato: @kyoto  @squat  now I realize this will fail if we have more then 2 builds at the same time. We need somehow to overwrite this with random or I can change this in the rspec code.

PR #2460, opened 2017-11-28
*: fix cert and env binding for k8s-node-bootstrapper

coresolve:
This change adds a bindmount of /etc/ssl and /etc/profile.env to ensure
that we can operate in proxy environments.

@lucab @lander2k2

### DIFF
--- a/modules/ignition/resources/services/k8s-node-bootstrap.service
+++ b/modules/ignition/resources/services/k8s-node-bootstrap.service
@@ -12,10 +12,10 @@ TimeoutStartSec=1h
 ExecStartPre=/usr/bin/mkdir -p /etc/kubernetes
 ${kubeconfig_fetch_cmd}
 ExecStartPre=/usr/bin/docker run --rm \
+            --env-file /etc/profile.env \
             --tmpfs /tmp \
             -v /usr/share:/usr/share:ro \
             -v /usr/lib/os-release:/usr/lib/os-release:ro \
-            -v /usr/share/ca-certificates/ca-certificates.crt:/etc/ssl/certs/ca-certificates.crt:ro \
             -v /var/lib/torcx:/var/lib/torcx \
             -v /var/run/dbus:/var/run/dbus \
             -v /run/metadata:/run/metadata:ro \
@@ -24,6 +24,7 @@ ExecStartPre=/usr/bin/docker run --rm \
             -v /etc/coreos:/etc/coreos:ro \
             -v /etc/torcx:/etc/torcx \
             -v /etc/kubernetes:/etc/kubernetes \
+            -v /etc/ssl:/etc/ssl:ro \
             ${tectonic_torcx_image_url}:${tectonic_torcx_image_tag} \
             /tectonic-torcx-bootstrap \
             --upgrade-os=${bootstrap_upgrade_cl} \


modules/ignition/resources/services/k8s-node-bootstrap.service
```diff
@@ -24,6 +24,8 @@ ExecStartPre=/usr/bin/docker run --rm \
             -v /etc/coreos:/etc/coreos:ro \
             -v /etc/torcx:/etc/torcx \
             -v /etc/kubernetes:/etc/kubernetes \
+            -v /etc/ssl:/etc/ssl:ro \
```
squat: @coresolve the k8s-notebootstrapper service already bindmounts in /etc/ssl/certs/ca-certificates.crt [0]. Do you explicitly need all of /etc/ssl? If so, you should omit the first bind mount, since it's being stomped.

[0] https://github.com/coreos/tectonic-installer/pull/2460/files#diff-dea3ad9663a5b686a1d43a0d5d7b167eR18

modules/ignition/resources/services/k8s-node-bootstrap.service
```diff
@@ -24,6 +24,8 @@ ExecStartPre=/usr/bin/docker run --rm \
...
+            -v /etc/profile.env:/etc/profile.env:ro \
```
squat: JW, why is this needed?

modules/ignition/resources/services/k8s-node-bootstrap.service
coresolve: I've pulled the other mount. We need it to support things like egress proxy where all traffic is sent to an env set proxy server.

modules/ignition/resources/services/k8s-node-bootstrap.service
coresolve: Needed to makes sure we pull in settings like: 
```
HTTP_PROXY=thisisabadidea:3279
HTTPS_PROXY=stillbad:3279
```
etc

coresolve:
This  has been changed to use /etc/profile.env as a --env-file. The container has no shell and probably won't eval the /etc/profile.env file.

lucab:
@coresolve I haven't tested tectonic-torcx with any proxy, but I assume you ran this already. Do you want me to check any specific case/scenario (or just a random quick run)?

As a warning note, any proxy configurations applied here needs to be propagated to CLUO hook, otherwise cluster upgrades will get stuck into permanent network failures.

modules/ignition/resources/services/k8s-node-bootstrap.service
coresolve: I'm going to put this back

coresolve:
Yep. I need to close this and reopen against track-1

squat:
@coresolve there is no /etc/profile.env on CL machines by default. Good thing we did not merge without running tests. Please see #2757. All tests are failing because systemd fails to start kubelet.service due to:
```
Jan 18 11:49:08 ip-10-0-4-128 systemd[1]: kubelet.service: Failed to load environment files: No such file or directory
Jan 18 11:49:08 ip-10-0-4-128 systemd[1]: kubelet.service: Failed to run 'start-pre' task: No such file or directory
Jan 18 11:49:08 ip-10-0-4-128 systemd[1]: Failed to start Kubelet via Hyperkube ACI.
Jan 18 11:49:08 ip-10-0-4-128 systemd[1]: kubelet.service: Unit entered failed state.
Jan 18 11:49:08 ip-10-0-4-128 systemd[1]: kubelet.service: Failed with result 'resources'.
Jan 18 11:49:18 ip-10-0-4-128 systemd[1]: kubelet.service: Service hold-off time over, scheduling restart.
Jan 18 11:49:18 ip-10-0-4-128 systemd[1]: Stopped Kubelet via Hyperkube ACI.
```

PR #2463, opened 2017-11-29
tests: add rspec_tap for output

erikkn:
Added the rspec_tap gem, used for more detailed output.
The log files are streamed into ES and with this new more detailed output, we can actually search within Kibana for useful information.


erikkn:
@mxinden

erikkn:
ok to test

PR #2459, opened 2017-11-28
add ct into smoke test image for sandbox

enxebre:

PR #2469, opened 2017-11-30
tests: change Docker image to new version

erikkn:
Bump the quay.io/coreos/tectonic-smoke-test-env image(version 5.9) to a new version (v.5.10). This new version includes the rspec_tap gem, that is needed to get more detailed information in the log files and #2459


erikkn:
Followed up by max inden.

PR #2471, opened 2017-11-30
modules,platforms: variables cannot be "version"

squat:
Terraform minor version 0.11.0 contains improved module support,
with real support for pinning module versions using the `version`
directive, e.g.:

```hcl
module "foo" {
  source = "bar"
  version = "baz"
}
```

Unfortunately, this directive clashes with a variable in the
`container_linux` module by the same name; Terraform is unable to
handle variables and directives with the same name in a given module
block, so the `version` _variable_ is never passed to the module.

Fixes: https://github.com/coreos/terraform-aws-kubernetes/issues/10

cc @s-urbaniak @enxebre


enxebre:
@squat I reckon we need to update https://github.com/coreos/tectonic-installer/blob/master/tests/rspec/lib/shared_examples/k8s.rb#L87 otherwise lgtm

squat:
@enxebre no, that line is good. It's saying:

if version is `latest`, then set version to the output variable `version` of the container linux module.

The output variable name is still `version` so there should be no issues there I think.

enxebre:
oh right, the output remains the same 👍

PR #2474, opened 2017-12-01
tests: add git branches as attributes

erikkn:
This PR adds the source & target Git branch to the log file name, which is used by Logstash to add those values to the document as attributes.


squat:
@erikkn printing all environment variables like this could cause the leak of sensitive data. Are we sure there are no sensitive credentials saved as environment variables?

erikkn:
ok to test

PR #2477, opened 2017-12-01
platform/gcp: fix ectonic_etcd_count=0

enxebre:
tectonic_etcd_count has to be set explicitly to something different than "0" for deploying GCP. This fixes it


enxebre:
@cpanato @s-urbaniak @kalmog

PR #2481, opened 2017-12-04
Fix interpolation synatx for kube cert output id

thunderbird86:
Get error when try to use autogenerated cert's
```
Error: Error refreshing state: 1 error(s) occurred:

* module.kubernetes.module.kube_certs.output.id: At column 5, line 2: join: argument 1 should be type list, got type string in:

${sha1("
  ${join(" ",
    local_file.apiserver_key.id,
    local_file.apiserver_crt.id,
    local_file.kube_ca_key.id,
    local_file.kube_ca_crt.id,
    local_file.kubelet_key.id,
    local_file.kubelet_crt.id,
    )}
  ")}
```

### DIFF
--- a/modules/tls/kube/self-signed/outputs.tf
+++ b/modules/tls/kube/self-signed/outputs.tf
@@ -27,14 +27,5 @@ output "apiserver_key_pem" {
 }
 
 output "id" {
-  value = "${sha1("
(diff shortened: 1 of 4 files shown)


thunderbird86:
issue fixed

PR #2484, opened 2017-12-04
network operator install

rithujohn191:
This PR removes the network operator managed manifests and adds installation of the tectonic network operator.


dghubble:
I don't see how this can work. Control plane bootstrapping won't complete until pod-to-pod networking is functional (to schedule kube-dns, etc). tectonic-system components are created only after the control plane is bootstrapped so this won't get to a point where TNO can start to run. 

What plan did the original designs have in mind for this? @aaronlevy @diegs

diegs:
@dghubble I think this operator should be installed during the bootkube phase along with the KVO.

We were already creating the `tectonic-system` namespace in `modules/bootkube/resources/manifests`, so it should be possible to create the daemonset there too.

dghubble:
We discussed trying to add the tno manifest in kube-system so its one of the components initially started by `bootkube start`, I'll try to look at that soon.

aaronlevy:
It doesn't necessarily have to be in `kube-system` (although I don't see any issue with that -- probably makes more sense for network tools). We can create other namespaces during the bootstrap phase as well (bootkube just walks the files lexicographically, so just needs to be ordered before an object in that namespace)

cpanato:
@rithujohn191 can you please rebase the PR and resolve the conflicts? thanks!

aaronlevy:
Closing as this was superseded by https://github.com/coreos/tectonic-installer/pull/2618

PR #2486, opened 2017-12-04
Install addon operator. Remove addon installed components

thorfour:
removed self hosted etcd smoke tests (no longer being supported)


thorfour:
ok to test

cpanato:
@thorfour looks like all tests failed because those manisfests are missing.

```
cluster_test.go:320: MISSING : [Kind: tco.coreos.com/v1:AppVersion | Namespace: tectonic-system | Name: tectonic-etcd] - [tectonic/updater/app_versions/app-version-tectonic-etcd.yaml]
        	cluster_test.go:320: MISSING : [Kind: etcd.database.coreos.com/v1beta2:EtcdCluster | Namespace: kube-system | Name: kube-etcd] - [tectonic/updater/cluster-config.yaml]
        	cluster_test.go:320: MISSING : [Kind: apps/v1beta2:DeploymentOrDaemonSet | Namespace: tectonic-system | Name: tectonic-etcd-operator] - [tectonic/updater/operators/tectonic-etcd-operator.yaml]
```

if it is really removed, we need to update the tests as well to reflect that.

Also if you removed the self hosted thing we need to update jenkinsfile to remove the azure/self hosted test as well. lines https://github.com/coreos/tectonic-installer/blob/master/Jenkinsfile#L290 and 291

are you able to do that?

thanks so much!

thorfour:
ok to test

thorfour:
ok to test

thorfour:
@cpanato Need a little help debugging these failures. Not sure what's broken

cpanato:
@thorfour some errors are related to the console tests we do. This is already fixed. can you please rebase?

In meantime I will make it run again.

thorfour:
@cpanato all green, please review again

modules/tectonic/assets.tf
```diff
@@ -24,6 +24,7 @@ resource "template_dir" "tectonic" {
     tectonic_prometheus_operator_image = "${var.container_images["tectonic_prometheus_operator"]}"
     tectonic_etcd_operator_image       = "${var.container_images["tectonic_etcd_operator"]}"
     tectonic_cluo_operator_image       = "${var.container_images["tectonic_cluo_operator"]}"
+    kubernetes_addon_operator_image    = "${var.container_images["kubernetes_addon_operator"]}"
```
squat: Nit: everything else in this vars list is alphabetized. Please move up to like 19

tests/smoke/bare-metal/smoke.sh
```diff
@@ -279,7 +279,7 @@ test_cluster() {
   export SMOKE_NODE_COUNT=$(( MASTER_COUNT + WORKER_COUNT ))
   export SMOKE_MANIFEST_PATHS=${ROOT}/build/${CLUSTER}/generated/
   # shellcheck disable=SC2155
```
squat: nit: remove this shellcheck directive since it is no longer needed

tests/smoke/cluster_test.go
```diff
@@ -42,15 +42,11 @@ const (
 var (
 	// defaultIgnoredManifests represents the manifests that are ignored by
 	// testAllResourcesCreated by default.
-	defaultIgnoredManifests = []string{"bootstrap"}
-
-	// selfHostedEtcdManifests represents the manifests that are ignored by
-	// testAllResourcesCreated when manifestSelfHostedEtcdEnv isn't set to empty string.
-	selfHostedEtcdManifests = []string{
-		// Generated all the time but only deployed when experimental is enabled.
+	defaultIgnoredManifests = []string{
```
squat: I think this is a little backwards. Reading the comment:

> selfHostedEtcdManifests represents the manifests that are ignored by testAllResourcesCreated when manifestSelfHostedEtcdEnv isn't set to empty string

So, when `SMOKE_MANIFEST_SELF_HOSTED_ETCD` is set to `true` then `selfHostedEtcdManifests` is appended to the list of manifests to ignore. Otherwise the manifests to ignore is just `bootstrap`. So, if we are eliminating self-hosted etcd testing, then we should never have `SMOKE_MANIFEST_SELF_HOSTED_ETCD` set to true, so the ignored manifests should *always* be *just* `bootstrap`.

Please change so that
```go
defaultIgnoredManifests = []string{"bootstrap"}
```

tests/smoke/cluster_test.go
```diff
@@ -42,15 +42,11 @@ const (
...
 		"tectonic/updater/cluster-config.yaml",
```
squat: This entire slice can be removed per comment above

tests/smoke/cluster_test.go
```diff
@@ -352,9 +348,6 @@ func testAllResourcesCreated(t *testing.T) {
 	}
 
 	ignoredManifests := defaultIgnoredManifests
```
squat: since we are never appending anything to `defaultIgnoredManifests`, please consider cleaning this up by eliminating `ignoredManifests` and instead directly using `defaultIgnoredManifests` to avoid allocating another variable for no reason:
```go
err := retry(allResourcesCreated(manifestsPathsSp, defaultIgnoredManifests), t, 30*time.Second, max)
...
```

modules/tectonic/assets.tf
thorfour: Fixed

tests/smoke/bare-metal/smoke.sh
thorfour: removed

tests/smoke/cluster_test.go
thorfour: Done

tests/smoke/cluster_test.go
thorfour: fixed

thorfour:
@squat in hindsight I probably should have split the PR's into two separate ones. I just tacked it on out of necessity since the addon-operator changes broke self-hosted etcd

squat:
@thorfour ah i see this was necessary because of self-hosted etcd’s dependence on kube-dns

squat:
@thorfour I am realizing now that merging this will leave master in a semi-broken state where users can still enable self-hosted etcd but these clusters will never work. We need to at least eliminate that user-facing config option first so master is correct.

thorfour:
@squat care to point me to where I would disable that?

thorfour:
@squat removing the manifests from the "ignored" list causes some of these tests to fail. Am I missing something?

squat:
@thorfour looking over the manifest ignoring code again it appears to me that something is fishy. Clearly we _do_ need to add the self hosed etcd manifests to the ignores list as you originally did, however, the logic for determining when to ignore those files seems backwards. In any case, you were right when you originally igngored those manifests. Sorry for the runaround. Please ignore them still. In addition, we should eliminate the self hosted etcd manifest env const since it is no longer used.

squat:
I’ll post a link to the config option we need to eliminate as well

thorfour:
@squat do you have the link to remove the self-hosted option?

squat:
@thorfour I rebased your PR on master to eliminate the merge conflicts and added a new commit to the branch that eliminates the self-hosted etcd configuration option from the user-facing API. PTAL

thorfour:
@squat looks like just a single timeout. Something of concern or can I merge?

squat:
@thorfour looking good! I'm going to modify labels to only re-run the bare-metal tests. This is probably OK, but since we have touched code for every platform, I'd like to re-test just for safety.

squat:
ok to test

cpanato:
retest this please

squat:
All green! Merging.

Quentin-M:
This broke payload again.

PR #2490, opened 2017-12-05
tests: Enable teams to run component tests via provided containers

mxinden:
- Generalize conformance test container runner to run arbitrary images
- Add `RUN_COMPONENT_TESTS` Jenkins pipeline parameter and make RSpec
aware of option
- Add array to specify component test images
- Iterate over component test image array and generate rspec examples
(`it`)

Missing:
- Authentication to download component images from Quay

//CC @sudhaponnaganti @cpanato @Quentin-M (Please mind *work in progress* status, feel free to comment)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -80,6 +80,11 @@ pipeline {
       description: ''
     )
     booleanParam(
+      name: 'RUN_COMPONENT_TESTS',
+      defaultValue: true,
+      description: ''
+    )
+    booleanParam(
       name: 'RUN_GUI_TESTS',
       defaultValue: true,
       description: ''
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -262,7 +262,7 @@ pipeline {
     stage("Smoke Tests") {
       when {
         expression {
-          return params.RUN_SMOKE_TESTS || params.RUN_CONFORMANCE_TESTS
+          return params.RUN_SMOKE_TESTS || params.RUN_CONFORMANCE_TESTS || params.RUN_COMPONENT_TESTS
         }
       }
       environment {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -404,7 +404,7 @@ def runRSpecTest(testFilePath, dockerArgs) {
         timeout(time: 5, unit: 'HOURS') {
           forcefullyCleanWorkspace()
           ansiColor('xterm') {
-            withCredentials(creds) {
+            withCredentials(creds + quayCreds) {
               withDockerContainer(
                 image: tectonicSmokeTestEnvImage,
                 args: '-u root -v /var/run/docker.sock:/var/run/docker.sock ' + dockerArgs
@@ -414,8 +414,12 @@ def runRSpecTest(testFilePath, dockerArgs) {
                 sh """#!/bin/bash -ex
                   mkdir -p templogfiles && chmod 777 templogfiles
                   cd tests/rspec
+                  docker login -u="$QUAY_ROBOT_USERNAME" -p="$QUAY_ROBOT_SECRET" quay.io
+
                   # Directing test output both to stdout as well as a log file
                   rspec ${testFilePath} --format RspecTap::Formatter --format RspecTap::Formatter --out ../../templogfiles/format=tap.log
+
+                  docker logout quay.io
                 """
               }
             }
@@ -451,9 +455,10 @@ def runRSpecTestBareMetal(testFilePath) {
           ansiColor('xterm') {
             unstash 'clean-repo'
             unstash 'smoke-test-binary'
-            withCredentials(creds) {
+            withCredentials(creds + quayCreds) {
               sh """#!/bin/bash -ex
               cd tests/rspec
+              docker login -u="$QUAY_ROBOT_USERNAME" -p="$QUAY_ROBOT_SECRET" quay.io
               export RBENV_ROOT=/usr/local/rbenv
               export PATH="/usr/local/rbenv/bin:$PATH"
               eval \"\$(rbenv init -)\"
@@ -461,6 +466,7 @@ def runRSpecTestBareMetal(testFilePath) {
               gem install bundler
               bundler install
               bundler exec rspec ${testFilePath} --format RspecTap::Formatter
+              docker logout quay.io
               """
             }
           }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -80,14 +80,14 @@ pipeline {
       description: ''
     )
     booleanParam(
-      name: 'RUN_COMPONENT_TESTS',
-      defaultValue: true,
-      description: ''
-    )
-    booleanParam(
       name: 'RUN_GUI_TESTS',
       defaultValue: true,
       description: ''
+    )
+    string(
+      name: 'COMPONENT_TEST_IMAGES',
+      defaultValue: '',
+      description: 'List of container images for component tests to run (comma-separated)'
     )
     booleanParam(
       name: 'PLATFORM/AWS',
@@ -262,7 +262,7 @@ pipeline {
     stage("Smoke Tests") {
       when {
         expression {
-          return params.RUN_SMOKE_TESTS || params.RUN_CONFORMANCE_TESTS || params.RUN_COMPONENT_TESTS
+          return params.RUN_SMOKE_TESTS || params.RUN_CONFORMANCE_TESTS || params.COMPONENT_TEST_IMAGES != ''
         }
       }
       environment {
--- a/Makefile
+++ b/Makefile
@@ -197,9 +197,9 @@ tests/smoke: bin/smoke smoke-test-env-docker-image
 	-e TF_VAR_tectonic_admin_password \
 	-e TECTONIC_TESTS_DONT_CLEAN_UP \
 	-e RUN_SMOKE_TESTS \
-	-e RUN_COMPONENT_TESTS \
 	-e RUN_CONFORMANCE_TESTS \
 	-e KUBE_CONFORMANCE_IMAGE \
+	-e COMPONENT_TEST_IMAGES \
 	--cap-add NET_ADMIN \
 	--device /dev/net/tun \
 	quay.io/coreos/tectonic-smoke-test-env \
--- a/Makefile
+++ b/Makefile
@@ -197,6 +197,7 @@ tests/smoke: bin/smoke smoke-test-env-docker-image
 	-e TF_VAR_tectonic_admin_password \
 	-e TECTONIC_TESTS_DONT_CLEAN_UP \
 	-e RUN_SMOKE_TESTS \
+	-e RUN_COMPONENT_TESTS \
 	-e RUN_CONFORMANCE_TESTS \
 	-e KUBE_CONFORMANCE_IMAGE \
 	--cap-add NET_ADMIN \
--- a/tests/rspec/lib/shared_examples/k8s.rb
+++ b/tests/rspec/lib/shared_examples/k8s.rb
@@ -169,10 +169,10 @@ RSpec.shared_examples 'withRunningClusterExistingBuildFolder' do |vpn_tunnel = f
   end
 
   COMPONENT_TESTS.map do |test|
-    it "passes #{test.name} component tests", :component_tests do
+    it "passes #{test['name']} component tests", :component_tests do
       test_container = TestContainer.new(
-        test.image,
-        @cluster.kubeconfig,
+        test['image'],
+        @cluster,
         vpn_tunnel,
         @cluster.env_variables['PLATFORM']
       )
--- a/tests/rspec/lib/shared_examples/k8s.rb
+++ b/tests/rspec/lib/shared_examples/k8s.rb
@@ -161,9 +161,8 @@ RSpec.shared_examples 'withRunningClusterExistingBuildFolder' do |vpn_tunnel = f
   it 'passes the k8s conformance tests', :conformance_tests do
     conformance_test = TestContainer.new(
       ENV['KUBE_CONFORMANCE_IMAGE'],
-      @cluster.kubeconfig,
-      vpn_tunnel,
-      @cluster.env_variables['PLATFORM']
+      @cluster,
+      vpn_tunnel
     )
     expect { conformance_test.run }.to_not raise_error
   end
@@ -173,8 +172,7 @@ RSpec.shared_examples 'withRunningClusterExistingBuildFolder' do |vpn_tunnel = f
       test_container = TestContainer.new(
         test['image'],
         @cluster,
-        vpn_tunnel,
-        @cluster.env_variables['PLATFORM']
+        vpn_tunnel
       )
       expect { test_container.run }.to_not raise_error
     end
--- a/tests/rspec/lib/shared_examples/k8s.rb
+++ b/tests/rspec/lib/shared_examples/k8s.rb
@@ -13,10 +13,6 @@ require 'webdriver_helpers'
 require 'test_container'
 require 'with_retries'
 
-COMPONENT_TESTS = [
-  { 'name' => 'console', 'image' => 'quay.io/coreos/tectonic-console-tester:v2.7.1' }
-].freeze
-
 RSpec.shared_examples 'withRunningCluster' do |tf_vars_path, vpn_tunnel = false|
   include_examples('withBuildFolderSetup', tf_vars_path)
   include_examples('withRunningClusterExistingBuildFolder', vpn_tunnel)
@@ -167,10 +163,10 @@ RSpec.shared_examples 'withRunningClusterExistingBuildFolder' do |vpn_tunnel = f
     expect { conformance_test.run }.to_not raise_error
   end
 
-  COMPONENT_TESTS.map do |test|
-    it "passes #{test['name']} component tests", :component_tests do
+  (ENV['COMPONENT_TEST_IMAGES'] || '').split(',').each do |image|
+    it "passes component test '#{image}'", :component_tests do
       test_container = TestContainer.new(
-        test['image'],
+        image.chomp,
         @cluster,
         vpn_tunnel
       )
--- a/tests/rspec/lib/shared_examples/k8s.rb
+++ b/tests/rspec/lib/shared_examples/k8s.rb
@@ -10,8 +10,12 @@ require 'pages/login_page'
 require 'name_generator'
 require 'password_generator'
 require 'webdriver_helpers'
-require 'k8s_conformance_tests'
+require 'test_container'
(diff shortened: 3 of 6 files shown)


Quentin-M:
Thanks! This looks fantastic to begin with. I had a similar initial change, I'm happy😆

tests/rspec/lib/shared_examples/k8s.rb
```diff
@@ -9,9 +9,13 @@
 require 'name_generator'
 require 'password_generator'
 require 'webdriver_helpers'
-require 'k8s_conformance_tests'
+require 'test_container'
 require 'with_retries'
 
+COMPONENT_TESTS = [
+  { 'name' => 'console', 'image' => 'quay.io/repository/coreos/tectonic-console-tester:v11' }
```
kans: We ship a `tectonic-console-tester` for each version of console - the two versions must match when running tests.

mxinden: @kans Thanks for the note! Good to know.

What do you do, when you want to release a *test-only* fix?
- Bump both versions anyways
- Bump a subversion in the `tectonic-console-tester` image

kans: We build and push both images in the same pass, so we would bump both versions for the fix.  

I don't think this problem will come up too often past the initial merge because we run the same tests before we merge.

Quentin-M:
Added env variables injection,

However, the console image fails with: 
```
+ timeout 30 google-chrome --no-sandbox --headless --disable-gpu --dump-dom
libudev: udev_has_devtmpfs: name_to_handle_at on /dev: Operation not permitted
<html><head><script>start("/");</script>
<script>addRow("bin","bin",1,4096,"4.0 kB",1509809219,"11/4/17, 3:26:59 PM");</script>
<script>addRow("boot","boot",1,4096,"4.0 kB",1499950865,"7/13/17, 1:01:05 PM");</script>
<script>addRow("dev","dev",1,340,"340 B",1512662701,"12/7/17, 4:05:01 PM");</script>
<script>addRow("etc","etc",1,4096,"4.0 kB",1512662701,"12/7/17, 4:05:01 PM");</script>
[...]
+ set -e
+ echo qmachu-test-int.tectonic.dev.coreos.systems
+ sed s/https:\/\///
+ BRIDGE_HOST=qmachu-test-int.tectonic.dev.coreos.systems
+ echo GET /
+ timeout 30 openssl+ openssl x509 -outform PEM
 s_client -showcerts -connect qmachu-test-int.tectonic.dev.coreos.systems:443
gethostbyname failure
connect:errno=0
unable to load certificate
139940784416400:error:0906D06C:PEM routines:PEM_read_bio:no start line:pem_lib.c:696:Expecting: TRUSTED CERTIFICATE
```

/cc @ggreer

kans:
@Quentin-M 

The image is failing [here](https://github.com/coreos-inc/bridge/blob/master/frontend/test-e2e.sh#L11):

```echo 'GET /' | timeout 30 openssl s_client -showcerts -connect "$BRIDGE_HOST:443" | openssl x509 -outform PEM > bridge-e2e.pem```

qmachu-test-int.tectonic.dev.coreos.systems:443 gethostbyname failure

Does that name exist?

A bit of explanation for the test:  our tests visit console in a headless chrome, but we must click through the SSL warnings.  However, Chrome renders a white screen when it would otherwise warn a user and has purposefully disabled all command line switches for bypassing the warnings... when in headless mode!  As it turns out, Chrome also seems to ignore OS certs, but you can manually tell Chrome to trust w/e.  Chrome stores its trust stuff in a sql (lite?) db that needs to be initialized, which we do by running Chrome once and ignoring all errors.  Then, we visit the domain with openssl client, attempt to download the cert, and add it to googles trust store.

**Comment by Quentin-M** [comment]
Thanks! AFAIK the name existed. It's the address I was using to access the console from my computer. I'll try again tomorrow.

File: tests/rspec/lib/test_container.rb [hunk 2]
```diff
@@ -0,0 +1,55 @@
...
+    "--net=container:#{hostname}"
+  end
+
+  # Some tests require a few environment variables to run properly,
+  # build the environment parameters here.
+  def container_env(engine)
+    env = {
+      'BRIDGE_AUTH_USERNAME' => @cluster.tectonic_admin_email,
+      'BRIDGE_AUTH_PASSWORD' => @cluster.tectonic_admin_password,
+      'BRIDGE_BASE_ADDRESS' => @cluster.tectonic_console_url,
```
mxinden: `tectonic_console_url` is not an attribute of a `cluster` instance.

Would you mind adding a function to the `Cluster` class like e.g.:

``` ruby
  def tectonic_cluster_url
    "#{@name}.#{@tfvars_file.terraform_base_domain}"
  end
```

Quentin-M: Really? It seems working on AWS. I inspired this line from the following:

```
  describe 'Interact with tectonic console' do
    before(:all) do
      @driver = WebdriverHelpers.start_webdriver
      @login = Login.new(@driver)
      @console_url = @cluster.tectonic_console_url
    end
```

```
"Env": [
                "BRIDGE_AUTH_USERNAME=quentin.machu@coreos.com",
                "BRIDGE_AUTH_PASSWORD=*******",
                "BRIDGE_BASE_ADDRESS=qmachu-test-int.tectonic-ci.de",
                "BRIDGE_BASE_PATH=/",
                "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
                "NODE_VERSION=8.9.2",
                "YARN_VERSION=1.3.2"
            ],
```

cpanato: you can use the `tectonic_console_url`it is already defined for each platform

cpanato: just call `@cluster.tectonic_console_url`

mxinden: @cpanato @Quentin-M Sorry, I missed the inheriting platform specific cluster classes. Your approach is absolutely right.

Quentin-M:
Waiting now due to the bridge integration tests running but not passing.

cpanato:
where can I see those tests? @Quentin-M 
we need to set the labels in this PR as well to run the tests

Quentin-M:
I'm running these locally for now: https://hastebin.com/xixelofuce.xml
Geoff told me that https://github.com/coreos-inc/bridge/pull/1945 would help, let's see.

Makefile
```diff
@@ -197,6 +196,7 @@ tests/smoke: bin/smoke smoke-test-env-docker-image
 	-e TF_VAR_tectonic_admin_password \
 	-e TECTONIC_TESTS_DONT_CLEAN_UP \
 	-e RUN_SMOKE_TESTS \
+	-e RUN_COMPONENT_TESTS \
```
mxinden: Would you mind extending the `expose-environment-variables` section in the `tests/README.md` in the same PR?

https://github.com/coreos/tectonic-installer/blob/master/tests/README.md#1-expose-environment-variables

Quentin-M:
```
+ ./tests/jenkins-jobs/scripts/report-status-to-github.sh failure spec/aws/basic_spec.rb 48309e88e1a3e2d5289ba72bd15b628c9a4d5de3
/home/ubuntu/workspace/tectonic-installer_PR-2490-6XJVDYKHITSAZAPFKTDPKH3NXU6XQJ6FATBUNVOI5N2NXEJRKG7Q@tmp/durable-e00a0754/script.sh: line 2: ./tests/jenkins-jobs/scripts/report-status-to-github.sh: No such file or directory
```

For real? 😂

mxinden:
@Quentin-M We have seen this issue before. I will restart the test. I suspect Jenkins to have issues, syncing stashes between master and workers.

Ok to test

Makefile
Quentin-M: Done.

Quentin-M:
ok to test

Makefile
mxinden: @Quentin-M Thanks! :)

Jenkinsfile
```diff
@@ -400,7 +405,7 @@ def runRSpecTest(testFilePath, dockerArgs) {
         timeout(time: 4, unit: 'HOURS') {
           forcefullyCleanWorkspace()
           ansiColor('xterm') {
-            withCredentials(creds) {
+            withCredentials([creds, quayCreds]) {
```
mxinden: @Quentin-M The tests fail constantly with `./tests/jenkins-jobs/scripts/report-status-to-github.sh: No such file or directory`.  This means that Jenkins fails  somewhere before `unstash 'clean-repo'`, so that the `report-status-to-github.sh` is not unstashed when `reportStatusToGithub` is called.

I got a simple suspicion: `withCredentials` expects a simple array. Both `creds` and `quayCreds` return an array themselves, thereby `[creds, quayCreds]` returns an array containing two arrays instead of just an array of credentials. Can you try combining the two arrays instead like e.g. `withCredentials(creds + quayCreds)`?

Quentin-M: You are darn smart. Looking at how I do it there: https://github.com/coreos-inc/tectonic-release-automation/blob/d42f6e3/jobs/build-tectonic/Jenkinsfile#L101, I have `withCredentials([[1], [2], ...])`. Looks like here, `creds` is already `[[1], [2], ...]`, so indeed it cannot be wrapped more.

Quentin-M:
Still seeing:

```
[Crud] Test Suite
=====================
creating namespace qa-test-zpffibghqh

Running:  CRDs
 ✔ Passed [notEqual]: No unhandled JavaScript errors.
 ✔ Passed [equal]: No unhandled JavaScript errors.
navigated to aws-pr-2490-16e552b.tectonic-ci.de/



---- An unknown server-side error occurred while processing the command. ----



 ✖ Failed [equal]: (No Errors were thrown.)
    at err (/opt/bridge/frontend/integration-tests/tests/crud.js:89:18)

 ✔ Passed [notEqual]: No unhandled JavaScript errors.
 ✔ Passed [equal]: No unhandled JavaScript errors.

FAILED:  1 assertions failed and 4 passed (241ms)
```

Do you have any idea?

Also:
```
ok 8 - aws-custom-tls with a cluster passes console component tests 
```

This should not have happened if the CRUD test failed.

/cc @ggreer @kans

Quentin-M:
@ggreer @kans Looks like I've figured it out. `BRIDGE_BASE_ADDRESS` expects `https://`, now it can progress farther.. 

Every images I have tested return the same:
```
TEST FAILURE:  1 assertions failed, 244 passed. (1m 35s)

 ✖ crud

   - deployments (15.744s)
   Timed out while waiting for element <#save-changes> to be present for 15000 milliseconds.  - expected "found" but got: "not found"
       at async.series.cb (/opt/bridge/frontend/integration-tests/tests/crud.js:191:8)
       at createYAML (/opt/bridge/frontend/integration-tests/tests/crud.js:187:9)
   SKIPPED:
   - replicasets
   - replicationcontrollers
   - persistentvolumeclaims
   - statefulsets
   - resourcequotas
   - networkpolicies
   - roles
   - EditLabels
   - deleteNamespace
   - /clusterroles/view
   - /nodes
   - /settings/cluster
   - /all-namespaces/events
   - /crds
   - /
   - /k8s/all-namespaces/alertmanagers
   - /ns/tectonic-system/alertmanagers/main

error Command failed with exit code 1.
```

And the container exit code seems right.

sudhaponnaganti:
@mxinden - what does this container have? 
Why console tests are included here? Can that be a separate container
Are K8 conformance tests are also part of this. Would like to see that also a separate container
I am under the impression that this is smoke and basic tests container but seeing container tests as well.

Quentin-M:
@sudhaponnaganti 

```
Are K8 conformance tests are also part of this. Would like to see that also a separate container
I am under the impression that this is smoke and basic tests container but seeing container tests as well.
```

Console integration tests are failing, and nothing else: smoke tests, etc, are all passing. Smoke test is a pre-requisite for running integration tests. K8s conformance tests are not executed here: not the purpose of this PR and it takes too long.

```
Why console tests are included here?
```

Because this is the PR that introduces the framework to execute integration tests, and we have been told that console integration tests are the required POC to go with it.

Everything is in place and ready to go, only if the console integration tests were passing 100%.

sudhaponnaganti:
For POC it is ok but I would separate Console container from framework. 
Is there another PR for K8 Conformance test Container?

Quentin-M:
```
For POC it is ok but I would separate Console container from framework.
```

The console container is built by the console team. But we cannot merge the tests if they are not passing, as it will break master for everybody..

```
Is there another PR for K8 Conformance test Container?
```

Conformance tests are available since a fairly long time. Just need to add the label. RA triggers them too.

Quentin-M:
ok to test

Quentin-M:
ok to test

Quentin-M:
Still waiting for flakes in console integration tests to be resolved (and the inverted exit code issue), to merge this PR.

Quentin-M:
```
+geoff.greer@coreos.com +quentin.machu@coreos.com - Looks like one test scenario is failing. 
Would rather close on POD testing and not holding it back with agreement with Geoff and Matt that
this one test scenario will be addressed soon. Need to move on to other containers
```

/cc @sudhaponnaganti @kans @ggreer for LGTM then.

Quentin-M:
ok to test

kans:
I don't think it makes sense to merge failing tests - ours still failed in the last run, even though they didn't exit 1:

```23:27:38 [aws/network_canal_spec.rb] FAILED:  1 assertions failed and 7 passed (15.817s)```

We are in the process of a complete [rewrite](https://github.com/coreos-inc/bridge/pull/1918) of our integration tests using a new framework, which hopefully will increase stability.  If you want to drop our test and keep the framework, I'd be OK with that.

**Comment by ggreer** [comment]
1. Is the cluster still coming up when the console GUI tests are being run? If so, the API server might return 502s and other invalid response codes, causing our tests to fail.

2. I'm also confused that the tests are marked as passing despite the fact that they have a nonzero exit code. The latest tester image (v2.7.1) should correctly exit and return 1 if there are any errors. It's also compatible with versions of console as old as v2.6.3.

I think it's worthwhile to get the current console GUI tests working in the installer. We can push a new tester image once we're confident that the new tests are stable.

**Comment by Quentin-M** [comment]
1. No, the cluster is up and several smoke tests has already been executed against it.
2. It simply [uses](https://github.com/coreos/tectonic-installer/pull/2490/files#diff-76948777a6401ada2e06af39da6ff00fR24) `system()` to execute the container. Literally the next line will raise an exception and fail the test suite if the exit code is different than 0.

File: tests/rspec/lib/shared_examples/k8s.rb [hunk 5]
```diff
@@ -10,9 +10,13 @@
 require 'name_generator'
 require 'password_generator'
 require 'webdriver_helpers'
-require 'k8s_conformance_tests'
+require 'test_container'
 require 'with_retries'
 
+COMPONENT_TESTS = [
+  { 'name' => 'console', 'image' => 'quay.io/coreos/tectonic-console-tester:c703f7f764b6f28009bbd155e83c3e04c580d603' }
```
sym3tri: Why are we configuring these directly in the code? Shouldn't this be coming from config params to Jenkins or some other repository? I'd really like to get away from dumping _all the things_ into the installer repo. It's not scalable.

cpanato: we can add a input parameter in our Jenkins pipeline

Quentin-M: Having a parameter was definitely the idea behind using container images. That's why I pushed literally for *weeks* against the initial pod design that was proposed.

sym3tri: OK then... unless there's a good reason to ship this _inside_ the source code I'd suggest moving this to a Jenkins param before merging.

cpanato: agree

sudhaponnaganti:
@Quentin-M @kans @ggreer - can you remove the failing test and move on to the next container. We need to include this framework in Lithium and Beryllium runs. This is taking too long. Once fixed, that test can be added back.

Quentin-M:
@sudhaponnaganti I have already removed it before your comment. This is now a Jenkins Job parameter (which is why it was very extremely important to use containers, and not pods!). Teams will be able to test/run their container using the job, to verify that it works as expected (besides running their own clusters). Running to verify the framework is OK, and will merge once I have LGTM.

Quentin-M:
Framework is working as intended. See https://jenkins-tectonic-installer.prod.coreos.systems/blue/organizations/jenkins/tectonic-installer/detail/PR-2490/29/pipeline/81/, where I launched the job with the console-tester image twice (simulating running two different component test).

Can I get a rubber stamp on this?

tests/rspec/lib/shared_examples/k8s.rb
kans: The version of our test image is tightly coupled to the version of Console and it is not backwards compatible.  Eg, a PR to bump Console would require changing Jenkins and would break the tests in master.  It may not make sense to store the test version at this exact spot, but I think it should be in code somewhere.

| Shouldn't this be coming from config params to Jenkins or some other repository?

The original idea was that the tester would be a different repo but have the same exact version of Console, but we have been churning through test versions trying to get the tests to pass.

Quentin-M:
BTW, to clear confusion. Reported failures on the PR are with two console testers (triggered manually (console tests failed). https://jenkins-tectonic-installer.prod.coreos.systems/job/tectonic-installer/job/PR-2490/30/ here is a run without any failure whatsoever (no component image executed).

Quentin-M:
Everything passes. Only failure is the VPC failing to come up. Good to merge.
Component images come from a Jenkins job.

/cc @sudhaponnaganti @mxinden

cpanato:
retest this please

tests/rspec/lib/test_container.rb
```diff
@@ -0,0 +1,57 @@
...
+  end
+
+  def run
+    ::Timeout.timeout(3 * 60 * 60) do # 3 hours
+      command = if @cluster.env_variables['PLATFORM'].include?('metal')
+                  "sudo rkt run --volume kubecfg,kind=host,readOnly=false,source=#{@cluster.kubeconfig} \
+                  --mount volume=kubecfg,target=/kubeconfig #{network_config} --dns=host \
+                  #{container_env('rkt')} --insecure-options=image #{@image}"
+                else
+                  "docker run -v #{@cluster.kubeconfig}:/kubeconfig \
```
hongchaodeng: can we do a docker pull again here? or it might be running a out of date image when we force push an image to the same tag.

PR #2495, opened 2017-12-05
tests: add ssh proxy to be able to collect logs from workers and other instances

cpanato:
fixes: INST-700

### DIFF
--- a/tests/rspec/lib/aws_cluster.rb
+++ b/tests/rspec/lib/aws_cluster.rb
@@ -47,23 +47,21 @@ class AwsCluster < Cluster
   end
 
   def master_ip_addresses
-    ssh_master_ip_addresses = []
-    Dir.chdir(@build_path) do
-      terraform_state = `terraform state show module.masters.aws_autoscaling_group.masters`.chomp.split("\n")
-      terraform_state.each do |value|
-        attributes = value.split('=')
-        next unless attributes[0].strip.eql?('id')
-        instances_id = AwsSupport.sorted_auto_scaling_instances(attributes[1].strip.chomp, @aws_region)
-        instances_id.each do |instance_id|
-          ssh_master_ip_addresses.push AwsSupport.preferred_instance_ip_address(instance_id, @aws_region)
-        end
-      end
-    end
-    ssh_master_ip_addresses
+    aws_autoscaling_group_master = @tfstate_file.value('module.masters.aws_autoscaling_group.masters', 'id')
+    instances_id = AwsSupport.sorted_auto_scaling_instances(aws_autoscaling_group_master, @aws_region)
+
+    instances_id.map { |instance_id| AwsSupport.preferred_instance_ip_address(instance_id, @aws_region) }
   end
 
   def master_ip_address
     master_ip_addresses[0]
+  end
+
+  def worker_ip_addresses
+    aws_autoscaling_group_worker = @tfstate_file.value('module.workers.aws_autoscaling_group.workers', 'id')
+    instances_id = AwsSupport.sorted_auto_scaling_instances(aws_autoscaling_group_worker, @aws_region)
+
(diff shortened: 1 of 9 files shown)


tests/rspec/lib/aws_cluster.rb
```diff
@@ -66,6 +66,22 @@ def master_ip_address
     master_ip_addresses[0]
   end
 
+  def workers_ip_addresses
```
mxinden: To align with `master_ip_addresses`:

`s/workers_ip_addresses/worker_ip_addresses`

tests/rspec/lib/aws_cluster.rb
```diff
@@ -66,6 +66,22 @@ def master_ip_address
...
+    ssh_workers_ip_addresses = []
+    Dir.chdir(@build_path) do
+      terraform_state = `terraform state show module.workers.aws_autoscaling_group.workers`.chomp.split("\n")
```
mxinden: Why not use https://github.com/coreos/tectonic-installer/blob/master/tests/rspec/lib/tfstate_file.rb#L9 here?

tests/rspec/lib/aws_cluster.rb
cpanato: done

tests/rspec/lib/aws_cluster.rb
cpanato: done

mxinden:
@cpanato Let me know when I should have another look.

cpanato:
@mxinden can you please do another round?

tests/rspec/lib/aws_cluster.rb
```diff
@@ -48,24 +48,32 @@ def stop
...
-        instances_id = AwsSupport.sorted_auto_scaling_instances(attributes[1].strip.chomp, @aws_region)
-        instances_id.each do |instance_id|
-          ssh_master_ip_addresses.push AwsSupport.preferred_instance_ip_address(instance_id, @aws_region)
-        end
-      end
+
+    aws_autoscaling_group_master = @tfstate_file.value('module.masters.aws_autoscaling_group.masters', 'id')
+    instances_id = AwsSupport.sorted_auto_scaling_instances(aws_autoscaling_group_master, @aws_region)
+    instances_id.each do |instance_id|
+      ssh_master_ip_addresses.push AwsSupport.preferred_instance_ip_address(instance_id, @aws_region)
```
alexsomesan: Just as an observation, you can achieve this effect of constructing the `ssh_master_ip_addresses` array via the `.map` operation on `instances_id`, like this:

`instances_id.map { |instance_id| AwsSupport.preferred_instance_ip_address(instance_id, @aws_region) }

This ^ expression will evaluate to the same list of master IP addresses, but you would not have to use an explicit variable like `ssh_master_ip_addresses` for that. You could just use that expression as the return statement.

Not saying you should change anything, just wanted to let you about this Ruby feature.

cpanato: great! forgot about the map thing. thanks so much!

cpanato: done

tests/rspec/lib/aws_cluster.rb
```diff
@@ -47,25 +47,23 @@ def stop
...
-        attributes = value.split('=')
-        next unless attributes[0].strip.eql?('id')
-        instances_id = AwsSupport.sorted_auto_scaling_instances(attributes[1].strip.chomp, @aws_region)
-        instances_id.each do |instance_id|
-          ssh_master_ip_addresses.push AwsSupport.preferred_instance_ip_address(instance_id, @aws_region)
-        end
-      end
-    end
-    ssh_master_ip_addresses
+    aws_autoscaling_group_master = @tfstate_file.value('module.masters.aws_autoscaling_group.masters', 'id')
```
mxinden: @cpanato Thanks for updating it here as well.

cpanato:
retest this please

PR #2499, opened 2017-12-06
[WIP] Revert "Bump console to v2.5.3."

mxinden:
Reverts coreos/tectonic-installer#2498

PR #2501, opened 2017-12-06
tests: fix console tests

cpanato:


mxinden:
@cpanato Can we close https://github.com/coreos/tectonic-installer/pull/2499 then?

PR #2511, opened 2017-12-07
tests/frontend: Support multiple GUI end-to-end tests runs

kyoto:
This change is needed to allow the GUI tests and smoke tests to be unified.

Also changes a couple of GUI test inputs to bring them a little closer to the smoke tests.

### DIFF
--- a/installer/frontend/ui-tests/tests/metal.js
+++ b/installer/frontend/ui-tests/tests/metal.js
@@ -2,14 +2,40 @@ const log = require('../utils/log');
 const wizard = require('../utils/wizard');
 const tfvarsUtil = require('../utils/terraformTfvars');
 
-// Test input .progress file
-const input = tfvarsUtil.loadJson('metal.progress').clusterConfig;
-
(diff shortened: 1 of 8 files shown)


kyoto:
ok to test

installer/frontend/ui-tests/tests/metal.js
```diff
@@ -2,14 +2,40 @@ const log = require('../utils/log');
...
-// Test input .progress file
-const input = tfvarsUtil.loadJson('metal.progress').clusterConfig;
-
-const testPage = (page, nextInitiallyDisabled) => wizard.testPage(page, 'metal', input, nextInitiallyDisabled);
-
 const REQUIRED_ENV_VARS = ['TF_VAR_tectonic_license_path', 'TF_VAR_tectonic_pull_secret_path'];
 
-module.exports = {
+// Expects an input file <prefix>.progress and an expected output file <prefix>.tfvars to both exist
+const steps = prefix => {
```
squat: In the future we should consider re-factoring and making a stepsFactory func to avoid some of the duplication here now.

PR #2515, opened 2017-12-07
tests: collect etcd logs

cpanato:
fixes: INST-624

### DIFF
--- a/tests/rspec/lib/aws_cluster.rb
+++ b/tests/rspec/lib/aws_cluster.rb
@@ -64,6 +64,10 @@ class AwsCluster < Cluster
     instances_id.map { |instance_id| AwsSupport.preferred_instance_ip_address(instance_id, @aws_region) }
   end
 
+  def etcd_ip_addresses
+    @tfstate_file.output('etcd', 'ip_addresses')
+  end
+
   def check_prerequisites
     raise 'AWS credentials not defined' unless credentials_defined?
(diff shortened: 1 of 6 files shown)


cpanato:
@mxinden can you take a look when you have some time? thanks

tests/rspec/lib/cluster.rb
```diff
@@ -133,6 +133,12 @@ def forensic
         print_service_logs(worker_ip, service, @name, master_ip_address)
       end
     end
+
+    etcd_ip_addresses.each do |etcd_ip|
+      ['etcd-member'].each do |service|
```
mxinden: Why do this map over an array with a single entry? Why not something like:
```ruby
print_service_logs(etcd_ip, 'etcd-member', @name, master_ip_address)
```

cpanato: because in some cases we have more then one etcd server

tests/rspec/lib/aws_cluster.rb
```diff
@@ -64,6 +64,10 @@ def worker_ip_addresses
     instances_id.map { |instance_id| AwsSupport.preferred_instance_ip_address(instance_id, @aws_region) }
   end
 
+  def etcd_ip_addresses
```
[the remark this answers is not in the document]
cpanato: the way to get the etcd hosts differ from platforms

PR #2520, opened 2017-12-08
aws: add etcd iam custom role

cpanato:
fixes: INST-713

### DIFF
--- a/modules/aws/etcd/nodes.tf
+++ b/modules/aws/etcd/nodes.tf
@@ -20,7 +20,22 @@ data "aws_ami" "coreos_ami" {
   }
 }
 
-resource "aws_iam_role" "etcd" {
+resource "aws_iam_instance_profile" "etcd" {
+  count = "${length(var.external_endpoints) == 0 ? 1 : 0}"
+  name  = "${var.cluster_name}-etcd-profile"
+
+  role = "${var.etcd_iam_role == "" ?
+    join("|", aws_iam_role.etcd_role.*.name) :
+    join("|", data.aws_iam_role.etcd_role.*.name)
+  }"
+}
+
+data "aws_iam_role" "etcd_role" {
+  count = "${var.etcd_iam_role == "" ? 0 : 1}"
+  name  = "${var.etcd_iam_role}"
+}
+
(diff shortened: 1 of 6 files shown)


hongchaodeng:
So I'm wondering why the CI didn't break before due to lack of permissions?
Checking [here](https://github.com/coreos/tectonic-installer/blob/c1089828696b3a23b9f3ae1acdbf4d57bd775c3e/Jenkinsfile#L123:78):
```
. ${WORKSPACE}/tests/smoke/aws/smoke.sh assume-role "$TECTONIC_INSTALLER_ROLE"
```

What does TECTONIC_INSTALLER_ROLE give? And we should probably converge that with the roles given in https://github.com/coreos-inc/installer-infra?

hongchaodeng:
cc @pbx0 on this.

cpanato:
@hongchaodeng we are not using the smoke.sh anymore, all tests we moved to rspec

modules/aws/etcd/nodes.tf
```diff
@@ -20,7 +20,22 @@ data "aws_ami" "coreos_ami" {
...
+  count = "${length(var.external_endpoints) == 0 ? 1 : 0}"
+  name  = "${var.cluster_name}-etcd-profile"
+
+  role = "${var.etcd_iam_role == "" ?
+    join("|", aws_iam_role.etcd_role.*.name) :
+    join("|", data.aws_iam_role.etcd_role.*.name)
+  }"
+}
+
+data "aws_iam_role" "etcd_role" {
```
squat: Discussed OOB. It is not totally obvious why this data source is needed. If we are just using the iam role name, then we could directly use the etcd_role variable in the instance profile rather than this data source. Spoke with @cpanato and it seems like this exists here for consistency with other role customizations. We should follow up and check if this data source is indeed needed, i.e. if the output `data.aws_iam_role.etcd_role.*.name` is something different than just the role name, e.g. a full ARN.

hongchaodeng:
@cpanato 
> all tests we moved to rspec

Got it. Thanks for explaining.
Why it didn't break the CI before if IAM role was missing?

cpanato:
answered in pvt

PR #2528, opened 2017-12-11
frontend: Fix max value validation for instance count fields

kyoto:
Also change validation error wording from e.g. "Provide a value less than 11" to "Cannot be greater than 10".

cc: @robszumski 

screenshot-1


squat:
Awesome work

PR #2531, opened 2017-12-11
tests: add idempotency tests

cpanato:
We have a test that checks the idempotency but I will disable the test because today we are not idempotency 😢. we can enable that when we start to work on INST-287

### DIFF
--- a/tests/rspec/lib/shared_examples/k8s.rb
+++ b/tests/rspec/lib/shared_examples/k8s.rb
@@ -84,6 +84,15 @@ RSpec.shared_examples 'withRunningClusterExistingBuildFolder' do |vpn_tunnel = f
     end
   end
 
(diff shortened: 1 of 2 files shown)


tests/rspec/lib/shared_examples/k8s.rb
```diff
@@ -84,6 +84,14 @@
     end
   end
 
+  it 'is idempotency when doing a terraform plan after an terraform apply' do
```
mxinden: %s/is idempotency when doing a terraform plan after an terraform apply/terraform plan after a terraform apply is an idempotent operation (does not suggest further changes)/

cpanato: I will update that in the next commit which I will disable the test as well.

PR #2536, opened 2017-12-12
config.tf: Bump KVO to v1.8.4-kvo.3

yifan-gu:
This is to make upgrade test pass as it's trying to upgrade from
179-3, which the current kvo in the config.tf doesn't know about.


yifan-gu:
ok to test

PR #2552, opened 2017-12-14
tests: Remove unused tectonic_aws_az_count var from smoke test input

kyoto:
This var doesn't appear to do anything.


kyoto:
ok to test

PR #2553, opened 2017-12-14
test: enable test but if it fails not fail the whole test suit.

cpanato:
Enabling the idempotency tests however if it fails it will not fail the entire test suit.


mxinden:
Closing here, discussed offline. `pending` fails the test if no error was raised.

https://relishapp.com/rspec/rspec-core/v/2-0/docs/pending/pending-examples

PR #2558, opened 2017-12-14
test: add option to set expiration date to the tls certs

cpanato:


sym3tri:
is there a jira issue for this?

cpanato:
@sym3tri  @mxinden  there is no jira because I did that in my spare time in the weekend.

This custom cert expiration date I did because Antoine asked me to help him to test a scenario when the certification is about to expire and to check if prometheus report the correct status. Then was more easy to make this small change to add an option.

PR #2562, opened 2017-12-14
platform/vmware: Revert vsphere provider version to 0.2.2

lander2k2:
The latest version 0.4.2 is causing mysterious errors - like preventing
the CA cert from being generated on self-signed installs.  Reverting
until the new version can be properly tested.


sudhaponnaganti:
@sym3tri  - can you get this reviewed and merge in to 1.8.4

PR #2568, opened 2017-12-15
Increase fs.inotify.max_user_watches limit

wallies:
I was getting a lot of "failed to create fsnotify watcher: too many open files" errors on one of my pods, so naturally I went hunting around the internet and found https://github.com/kubernetes/kops/pull/2913. Tried the fix and by magic everything is working perfectly again. Thought I would throw this open to see if it helps anyone.

https://github.com/coreos/tectonic-installer/pull/2568 could probably be closed as well.


enxebre:
ok to test

modules/ignition/outputs.import
```diff
@@ -4,6 +4,10 @@ variable "ign_max_user_watches_id" {
   type = "string"
 }
 
+variable "ign_max_user_instances_id" {
```
enxebre: test are broken because this has no value. Default values could be set here for watches and instances or just set  opinionated values in the template.

modules/ignition/outputs.tf
```diff
@@ -6,6 +6,14 @@ output "max_user_watches_rendered" {
...
 }
 
+output "max_user_instances_id" {
+  value = "${data.ignition_file.max_user_instances.id}"
+}
+
+output "max_user_instances_rendered" {
+  value = "${data.template_file.max_user_instances.rendered}"
+}
+
```
enxebre: Output would need to be injected for all platforms as per https://github.com/coreos/tectonic-installer/blob/master/platforms/aws/main.tf#L148

enxebre:
Thanks @wallies, dropped a few comments

modules/ignition/resources/sysctl.d/max-user-watches.conf
```diff
@@ -1 +1 @@
-fs.inotify.max_user_watches=16184
+fs.inotify.max_user_watches=524288
```
sym3tri: I see that these were copied from the referenced PR, but I didn't see any reasoning about these values. Are these just arbitrary?

wallies: Same value came from release notes for docker for mac https://docs.docker.com/docker-for-mac/release-notes/#beta-181-release-notes-2016-07-07-1120-rc3-beta181, also is referenced in some other moby issue (https://github.com/moby/moby/issues/2259), as well as some other third party issues (http://confluence.jetbrains.com/display/IDEADEV/Inotify%2BWatches%2BLimit)

sym3tri:
@wallies Thanks for this contribution. We're working on a more general way to provide customizations to the ignition configs without requiring installer code-changes like this. I've made a note to consider this as the default going forward.

PR #2572, opened 2017-12-15
frontend: Fix display of error message when AWS credentials are invalid

kyoto:
This error was already rendered as a field validation error on the
regions dropdown, but that meant it was only shown if the dropdown was
marked as "dirty" in Redux, which doesn't happen until its value is
first set.

Explicitly rendering the AWS API error solves this problem, but it does
mean that the error message can now possibly be displayed twice. Solving
that will require a larger code change, so this is an initial quick fix.

screenshot-1


kyoto:
For reference, the duplicated error message introduced by this PR occurs if you:
1. Enter valid credentials
2. Select a region
3. Change the credentials so that they are not valid, but do pass field validation

screenshot-1

PR #2578, opened 2017-12-15
platform/vmware: Add missing ignition vars on masters

lander2k2:


sym3tri:
merging since only affects vmware

PR #2579, opened 2017-12-15
platform/vmware: Revert vsphere provider version to 0.2.2

lander2k2:


sym3tri:
merging b/c only affects vmware

PR #2581, opened 2017-12-18
tests/frontend: Diff AWS GUI test output against smoke test input

kyoto:
Delete the expected AWS GUI test output file and diff the actual GUI test output directly against the smoke test input file instead.


cpanato:
ok to test

PR #2582, opened 2017-12-18
tests/frontend: Ignore fewer TF vars

kyoto:
- Some tests don't output TF vars for smoke tests, so they don't need to ignore any TF vars.
- Some TF vars are specific to a single platform, so only the tests for that platform need to ignore them.


installer/frontend/ui-tests/tests/aws.js
```diff
@@ -47,8 +47,21 @@ const toExport = {
   },
 };
 
+const skipKeys = {
+  actual: [
```
squat: I don't think we need to make this distinction at all. We should just have a list of keys to skip and remove those keys, if they exist, from both the expected list and the actual list. This will simplify our lives and remove some complexity IMO

kyoto: 👍 I changed it to not distinguish between `actual` and `expected` vars.

PR #2588, opened 2017-12-18
config.tf: add "none" networking option to config

spangenberg:


modules/tectonic/assets.tf
```diff
@@ -101,9 +108,7 @@ resource "template_dir" "tectonic" {
     image_re            = "${var.image_re}"
     kube_dns_service_ip = "${cidrhost(var.service_cidr, 10)}"
 
-    tectonic_networking = "${var.tectonic_networking}"
-    calico_mtu          = "${var.calico_mtu}"
-    cluster_cidr        = "${var.cluster_cidr}"
+    network_config = "${indent(6, local.network_config_yaml)}"
```
alexsomesan: Very fancy 👍

modules/tectonic/assets.tf
```diff
@@ -1,3 +1,10 @@
+locals {
+  calico_config_yaml   = "${var.tectonic_networking == "calico" ? "calicoConfig:\n    mtu: ${var.calico_mtu}" : ""}"
+  network_config_yaml  = "${join("\n", compact(list(local.network_profile_yaml, local.pod_cidr_yaml, local.calico_config_yaml)))}"
```
alexsomesan: I presume the `compact()` ensures no empty lines when some attributes are missing, right?

spangenberg: Exactly

modules/tectonic/assets.tf
spangenberg: :smirk:

PR #2595, opened 2017-12-19
frontend: Fix TF action title when action is neither apply nor destroy

kyoto:
Normally we expect to only have an `apply` or `destroy` action in progress, that is not actually true 100% of the time.

screenshot-1

### DIFF
--- a/installer/frontend/components/tf-poweron.jsx
+++ b/installer/frontend/components/tf-poweron.jsx
@@ -265,7 +265,7 @@ export const TF_PowerOn = connect(stateToProps, dispatchToProps)(
         );
       }
 
-      const tfTitle = `${isApply ? 'Applying' : 'Destroying'} Terraform`;
+      const tfTitle = `Terraform ${_.startCase(action)}`;
 
       return <div>
         {!isBareMetal &&


installer/frontend/components/tf-poweron.jsx
```diff
@@ -265,7 +265,7 @@ export const TF_PowerOn = connect(stateToProps, dispatchToProps)(
         );
       }
 
-      const tfTitle = `${isApply ? 'Applying' : 'Destroying'} Terraform`;
+      const tfTitle = isApply ? 'Applying Terraform' : `Terraform ${_.startCase(action)}`;
```
squat: This seems a little funny. Now it is possible for titles to read “Terraform Destory” and other to read “Applying Terraform”. Should be made consistent like before? Also, what other Terraform actions do we recognize?

kyoto: I changed it to “Terraform <ACTION>“ for all actions.

I've seen a case where the action is `show` when the app is launched and detects an install in progress, but I think that case will go away with planned changes to the restore logic. Still, this change makes it more general / tolerant.

PR #2597, opened 2017-12-19
frontend: Remove unneeded exception when getting AWS hosted zone domain

kyoto:
Because of the way we validate all field when the page is reloaded, it's
possible to end up calling `getZoneDomain()` when the platform is bare
metal, but this is harmless so let's just remove the exception.

Also rename to `getAwsZoneDomain()` to show it is AWS specific.


kyoto:
ok to test

squat:
All tests have passed. For some reason the final result is not being displayed in GitHub. Merging.

PR #2602, opened 2017-12-20
frontend: Make autoFocus use more consistent

kyoto:
Generally, if the first field on a page is a `text` or `textarea` input, it should be focused.
- Add `autoFocus` to first input on Console Login page.
- Remove `autoFocus` from AWS VPC field that is actually half way down the page.
- Consistently use `autoFocus={true}` syntax rather than `autoFocus="true"` or just `autoFocus`.

cc @tlwu2013


squat:
All tests are green; status is simply not updated in GitHub. Merging.

Issue #356, opened 2017-04-25
GUI: provide a way to get a kubeconfig for troubleshooting installation


sym3tri:
If you go thru the GUI flow and everything is good, you'll get to the step where it says: "Starting Tectonic console", which could end up never working for various reasons. If this step were to silently fail there is now way for the user to debug what has happened. We don't generate the kubeconfig or document how to get it. You can only download the assets once the install has completed successfully.

tlwu2013:
Hi @kyoto ,

Please see attached latest mocks for "Console startup failed case".

There are three of them:
1. **2.4_Starting_Tectonic_failed__ver3.3.png**:
--> look-n-feel

2. **2.4_Starting_Tectonic_failed__ver3.3__redline.png** 
--> see notes for more info

3. **2.4_Starting_Tectonic_failed__ver3.3__spacing.png**
--> see spacing info

Please ping me for any question you have, thanks!
Tony

**2.4_Starting_Tectonic_failed__ver3.3.png**:
![2 4_starting_tectonic_failed__ver3 3](https://user-images.githubusercontent.com/5903705/28588797-91f7d6ce-7130-11e7-8c89-ad1b8e365a36.png)

**2.4_Starting_Tectonic_failed__ver3.3__redline.png** 
![2 4_starting_tectonic_failed__ver3 3__redline](https://user-images.githubusercontent.com/5903705/28588805-962fe98e-7130-11e7-8fb4-fc0176ec690c.png)

**2.4_Starting_Tectonic_failed__ver3.3__spacing.png**
![2 4_starting_tectonic_failed__ver3 3__spacing](https://user-images.githubusercontent.com/5903705/28588789-8d7224b0-7130-11e7-9d47-a90473bed9c2.png)

tlwu2013:
And here's the mock for showing the flow:
![installer__starting-tectonic-screens__ after_terraform_success](https://user-images.githubusercontent.com/5903705/28588883-d59b1896-7130-11e7-831e-02ac154b4c28.png)

kyoto:
Blocked on #1192

sym3tri:
@kyoto unblocked now that #1192 has merged?

kyoto:
@sym3tri Yes, unblocked now. I have started working on this.

PR #2606, opened 2017-12-20
frontend: Add help text for Download Assets button

kyoto:
Also capitalize "Assets" in button text.

This text was in the mocks (see https://github.com/coreos/tectonic-installer/issues/356#issuecomment-317838399), but seems to have been overlooked.

screenshot-1

### DIFF
--- a/installer/frontend/components/tf-poweron.jsx
+++ b/installer/frontend/components/tf-poweron.jsx
@@ -376,9 +376,14 @@ export const TF_PowerOn = connect(stateToProps, dispatchToProps)(
             <div className="col-xs-12">
               <a href="/terraform/assets" download>
                 <button className={classNames('btn btn-primary wiz-giant-button')}>
-                  <i className="fa fa-download"></i>&nbsp;&nbsp;Download assets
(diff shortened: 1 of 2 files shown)


installer/frontend/components/tf-poweron.jsx
```diff
@@ -376,10 +376,15 @@ export const TF_PowerOn = connect(stateToProps, dispatchToProps)(
...
               <a href="/terraform/assets" download>
                 <button className={classNames('btn btn-primary wiz-giant-button')}>
-                  <i className="fa fa-download"></i>&nbsp;&nbsp;Download assets
+                  <i className="fa fa-download"></i>&nbsp;&nbsp;Download Assets
                 </button>
               </a>
             </div>
+            <div className="col-xs-12">
+              <div className="wiz-pending-fg">
+                <p>The assets contain TLS contents used by Kubelets and manifests for all Tectonic components.</p>
```
squat: We should override the wording of the mock and change the text to `TLS assets` or even `TLS certificates` since `contents` doesn't actually make sense in this context.

kyoto: I changed it to "TLS certificates".

PR #2607, opened 2017-12-20
tests: call forensic to collect the logs when timeout happen during start

cpanato:
When a timeout in RSpec happen during the bootstrap we now collect the logs to analyze later

### DIFF
--- a/tests/rspec/lib/cluster.rb
+++ b/tests/rspec/lib/cluster.rb
@@ -163,6 +163,8 @@ class Cluster
         return true if system(env, "make -C ../.. apply | tee ../../build/#{@name}/terraform-apply.log")
       end
     end
+  rescue Timeout::Error
+    forensic
     raise 'Applying cluster failed'
   end
 


tests/rspec/lib/cluster.rb
```diff
@@ -154,7 +154,7 @@ def license_and_pull_secret_defined?
   end
 
   def apply
-    ::Timeout.timeout(30 * 60) do # 30 minutes
```
enxebre: Why removing :: here? if so, we should do the same for destroy method for consistency

cpanato: for destroy, we don't care if timeout. but i will update

enxebre: afaik it sticks to the styling guide both ways with :: and without :: https://github.com/bbatsov/ruby-style-guide#double-colons so lets be consistent and choose the same approach for all the cases

PR #2609, opened 2017-12-20
modules/ignition,vmware: Enable install behind HTTP proxy

lander2k2:
This addition will allow a cluster to be installed behind an HTTP proxy
on VMware.  The changes have been implemented in the ignition module to
allow it to be readily added to the other platforms.

Note: this only addresses the env vars needed on the CL nodes.  There
are tectonic components - like tectonic-channel-operator - that also
need these env vars to properly function.  That will be addressed
separately.


enxebre:
Hey @lander2k2 need to run `terraform fmt` to get it green

Documentation/variables/config.md
```diff
@@ -33,11 +33,15 @@ This document gives an overview of variables used in all platforms of the Tecton
...
 | tectonic_etcd_servers | (optional) List of external etcd v3 servers to connect with (hostnames/IPs only). Needs to be set if using an external etcd cluster. Note: If this variable is defined, the installer will not create self-signed certs. To provide a CA certificate to trust the etcd servers, set "tectonic_etcd_ca_cert_path".<br><br>Example: `["etcd1", "etcd2", "etcd3"]` | list | `<list>` |
 | tectonic_etcd_tls_enabled | (optional) If set to `true`, all etcd endpoints will be configured to use the "https" scheme.<br><br>Note: If `tectonic_experimental` is set to `true` this variable has no effect, because the experimental self-hosted etcd always uses TLS. | string | `true` |
+| tectonic_http_proxy | (optional) HTTP proxy address.<br><br>Note: `tectonic_http_proxy_enabled` must be set to `true` to take effect.<br><br>Example: `http://myproxy.example.com` | string | `` |
+| tectonic_http_proxy_enabled | (optional) Enable the use of an HTTP proxy to egress the local network.<br><br>Note: Only available for VMware modules at this time. | string | `false` |
```
sym3tri: I'd prefer we not add any new top-level config options unless we also add support for other platforms. 

Can you keep these changes localized to VMWare only for now? Otherwise please add & test similar changes for the other main platforms (aws, azure, baremetal)?

Documentation/variables/config.md
```diff
@@ -33,11 +33,15 @@ This document gives an overview of variables used in all platforms of the Tecton
...
 | tectonic_master_count | The number of master nodes to be created. This applies only to cloud platforms. | string | `1` |
 | tectonic_networking | (optional) Configures the network to be used in Tectonic. One of the following values can be used:<br><br>- "flannel": enables overlay networking only. This is implemented by flannel using VXLAN.<br><br>- "canal": [ALPHA] enables overlay networking including network policy. Overlay is implemented by flannel using VXLAN. Network policy is implemented by Calico.<br><br>- "calico": [ALPHA] enables BGP based networking. Routing and network policy is implemented by Calico. Note this has been tested on baremetal installations only. | string | `flannel` |
+| tectonic_no_proxy | (optional) Comma delimited list of local endpoints that will not use HTTP proxy.<br><br>Note: `tectonic_http_proxy_enabled` must be set to `true` to take effect.<br><br>Example: `127.0.0.1,localhost,.example.com,10.3.0.1` | string | `` |
```
sym3tri: If this is a list, why not make it a list type instead of a string?

Documentation/variables/config.md
```diff
@@ -33,11 +33,15 @@ This document gives an overview of variables used in all platforms of the Tecton
 | tectonic_etcd_count | The number of etcd nodes to be created. If set to zero, the count of etcd nodes will be determined automatically.<br><br>Note: This is not supported on bare metal. | string | `0` |
 | tectonic_etcd_servers | (optional) List of external etcd v3 servers to connect with (hostnames/IPs only). Needs to be set if using an external etcd cluster. Note: If this variable is defined, the installer will not create self-signed certs. To provide a CA certificate to trust the etcd servers, set "tectonic_etcd_ca_cert_path".<br><br>Example: `["etcd1", "etcd2", "etcd3"]` | list | `<list>` |
 | tectonic_etcd_tls_enabled | (optional) If set to `true`, all etcd endpoints will be configured to use the "https" scheme.<br><br>Note: If `tectonic_experimental` is set to `true` this variable has no effect, because the experimental self-hosted etcd always uses TLS. | string | `true` |
+| tectonic_http_proxy | (optional) HTTP proxy address.<br><br>Note: `tectonic_http_proxy_enabled` must be set to `true` to take effect.<br><br>Example: `http://myproxy.example.com` | string | `` |
```
sym3tri: nit: `tectonic_http_proxy_address` would be a clear name.

Documentation/variables/config.md
sym3tri: I realize this makes the code uglier, but I'd rather infer proxy is enabled by the presence of the above 2 vars and removing this one entirely. We've already got way too many top-level config options, I'd like to avoid adding any that aren't absolutely necessary. 

Also it causes confusing behavior if this is true but the others are both empty & there's no way to validate that in this PRs current form.

Documentation/variables/config.md
```diff
@@ -33,11 +33,15 @@ This document gives an overview of variables used in all platforms of the Tecton
...
+| tectonic_http_proxy_enabled | (optional) Enable the use of an HTTP proxy to egress the local network.<br><br>Note: Only available for VMware modules at this time. | string | `false` |
+| tectonic_https_proxy | (optional) HTTPS proxy address.<br><br>Note: `tectonic_http_proxy_enabled` must be set to `true` to take effect.<br><br>Example: `http://myproxy.example.com` | string | `` |
```
sym3tri: nit: `tectonic_https_proxy_address` would be a clear name.

sym3tri:
Hey, can you also make VMware PRs against master first then cherry-pick into 'track-1'?

Documentation/variables/config.md
lander2k2: @sym3tri Agreed.  That would be clearer.  I'll change that.

Documentation/variables/config.md
lander2k2: I gave this a lot of consideration.  Normally I would not add top-level configs when changes are only effective on one platform.  But localizing to VMware will make the change to add other platforms quite a bit more work.  Whereas this change will allow other platforms to be readily added.  I've written the code for AWS and SE's just completed testing it this morning for a POC they're working on for Pearson.  T-Mobile will need it for BM very soon (next month).  I don't have the means to test on Azure or BM but it's possible I could add an AWS PR later today.

So I _could_ shuffle this around to localize it to VMware but I that will burn a lot more precious cycles in the long-run.

lander2k2: @sym3tri Are you sure you want me to do that?  I totally appreciate the effort to limit new config options but the trade-off is make the code uglier to save one config option.  Of course, I'll do it if you feel strongly.  I would just argue that cleaner, more maintainable code would be a higher priority in this case.

Documentation/variables/config.md
lander2k2: @sym3tri  Because it's not really a list.  It's just a string that gets set as an environment variable.  To make it a terraform list type would just mean I'd have to iterate over the list and create a string to set as an env var.

sym3tri: Isn't it as simple as calling `join(",", var.mylist)`?

https://www.terraform.io/docs/configuration/interpolation.html#join-delim-list-

I'll leave it up to you, I don't feel too strongly about it.

Documentation/variables/config.md
sym3tri: How about you add the code for the other platforms as well (best-effort) in a separate PR, then we'll take over the PR and make sure they're all tested and working? If so pls create a Jira issue in our backlog with a link so we can track it.

sym3tri: I'll defer to @alexsomesan on what he thinks is best here.

sym3tri: Spoke to him on slack...
I think it's not much of an issue given we'll be refactoring quite a bit soon. No need to change.

lander2k2: Deal!  Will do.

alexsomesan: I'm with @sym3tri here. I think the `enabled` flag is unnecessary in this context as that state can be inferred from `tectonic_http_proxy_address` having a non-empty value. The code is not necessarily uglier if you use a local Terraform variable to proxy the `enabled` value to the other resources. Here's an example of how I'd express that:

```
locals {
  tectonic_http_proxy_enabled = "${length(var.tectonic_http_proxy_address) > 0}"
}
```
Then to use it you would do something like:
```
http_proxy_enabled      = "${local.tectonic_http_proxy_enabled}"
```

Documentation/variables/config.md
alexsomesan: I'm in favor of a list variable + `join()`.
We plan to add validation to user inputs in the near future and it would help to know from the variable type what kind of data to expect in it, etc.

lander2k2: I guess you're right.  TIL.  I'll make that change.

Documentation/variables/config.md
lander2k2: Yeah, that is a clean solution.  I'll add that change.

lander2k2:
@sym3tri @alexsomesan 
I think I've addressed all the changes requested.  LMK if there are any other issues.  I'll add the other platforms next.  Then submit the same changes against master.  I'll submit further VMware/on-prem changes against master first.

sym3tri:
Basic tests are failing with
```
outdated examples (run 'make examples' to fix)
19:59:57 Makefile:139: recipe for target 'structure-check' failed
```

lander2k2:
@sym3tri It's failing on examples because I removed the examples for platforms besides VMware after running `make examples`.  Those platforms will all be coming in a subsequent PR.  LMK if you want me to add the other examples so the test passes.

sym3tri:
@lander2k2 we need to have the examples, otherwise tests will always fail.

sym3tri:
Going to kick off smoke tests since this is modifying common code.

sym3tri:
ok to test

lander2k2:
@sym3tri @alexsomesan I've added all the other platforms in addition to VMware.  There are just a couple of thins I want to call on BM.

* I added a new "proxy" directory in `modules/ignition/resources/` LMK if there's a better way to organize those files.
* [1] will deposit empty files on the nodes when an HTTP proxy is not being used.  While I think that will be harmless, it's a little messy.  If you guys have a suggestion on how to improve, I'm all ears.

[1] https://github.com/coreos/tectonic-installer/pull/2609/commits/0906c4daeffb8c9b25ec536f5083cc8d0ab33fea#diff-f3dbed0a854bcf42775243e54dda3256

sym3tri:
Hey @lander2k2 there are conflicts now, can you rebase this?

aaronlevy:
> There are tectonic components - like tectonic-channel-operator - that also
need these env vars to properly function. That will be addressed
separately.

If this isn't addressed that means the cluster will not be able to upgrade - probably should make a very clear note around the option if that's the case.

kalmog:
Conflicting files in branch. Opened new PR (#2714)

PR #2610, opened 2017-12-20
BUILD => BUILD.bazel

diegs:
Resolves the conflict for case-insensitive filesystems (macs).
BUILD.bazel is the currently recommended BUILD filename anyway.


coresolve:
nice! Thanks @diegs

alexsomesan:
@diegs Thanks for the fix.

The reason for having both names around is we are trying to make a distinction between human editable and generated (e.g. with Gazelle) build files.

This change is alright for now, we'll figure out the convention again after we deprecate the make logic.

PR #2620, opened 2017-12-21
frontend: Fix AWS VPC subnets resetting to defaults after page refresh

kyoto:
If we already have subnet values, they should not be overwritten by default values.


squat:
We should ensure this is generally the case for all fields, no? If we refresh we’d like to maintain state? We should discuss with tony

kyoto:
@squat Currently all fields do maintain their state when you refresh the page. This was a bug that only affected the AWS VPC subnet fields because the default subnet values were being loaded and applied after the session state was restored.

PR #2625, opened 2017-12-21
config.tf: Remove images that are already managed by KAO.

yifan-gu:


enxebre:
ok to test

yifan-gu:
@enxebre Fixed, tests passed now.

PR #2628, opened 2017-12-22
frontend: Prevent unnecessary "validate" API calls

kyoto:


kyoto:
@squat FYI, I went ahead and merged this small fix.

PR #2632, opened 2017-12-22
rspec/lib: pulls test image before running it

hongchaodeng:
TestContainer'd better pull the component test image before running it.
Otherwise it could run a local out of date cache.

ref: https://github.com/coreos/tectonic-installer/pull/2490/files#r158543036


hongchaodeng:
cc @Quentin-M

Quentin-M:
@hongchaodeng Thank you! Do you know if we need to do the same for rkt, just above?

hongchaodeng:
That's actually good catch. Will add the same for rkt too

hongchaodeng:
fixed.

Quentin-M:
ok to test

PR #2637, opened 2017-12-30
frontend: Handle case where cluster state data is not populated

kyoto:
This happens if the app detects that an installation is already in progress and jumps to the Start Installation page.

Hide the Terraform actions menu and the Tectonic domain since they will not work without the cluster state data. The UI will still allow you to download assets to destroy the cluster manually.

### DIFF
--- a/installer/frontend/components/tf-poweron.jsx
+++ b/installer/frontend/components/tf-poweron.jsx
@@ -222,7 +222,9 @@ export const TF_PowerOn = connect(stateToProps, dispatchToProps)(
         const dnsReady = tectonic.console.success || ((tectonic.console.message || '').search('no such host') === -1);
         consoleSubsteps.push(
           <Step pending={isTFRunning} done={dnsReady && !isTFRunning} cancel={tfError} key="dnsReady">
-            Resolving <a href={`https://${tectonicDomain}`} target="_blank">{tectonicDomain}</a>
+            Resolving {tectonicDomain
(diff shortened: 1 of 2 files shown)


installer/frontend/components/tf-poweron.jsx
```diff
@@ -316,7 +321,7 @@ export const TF_PowerOn = connect(stateToProps, dispatchToProps)(
             </Step>
             <div style={{marginLeft: 22}}>
               {isAWS && isApply && statusMsg !== 'success' && <ProgressBar progress={state.terraformProgress} isActive={isTFRunning} />}
-              {showLogs && output && !isApplySuccess &&
+              {showLogs && output && !showTfActions &&
```
squat: I don't think showing logs should be conditioned on `clusterName`. As it stands, the logs would be shown if the terraform apply has succeeded and the cluster state is not populated, which is not desired. IMO we do not want to show logs ever if apply has succeeded.
 
This should probably remain as `isApplySuccess`

installer/frontend/components/tf-poweron.jsx
```diff
@@ -300,7 +305,7 @@ export const TF_PowerOn = connect(stateToProps, dispatchToProps)(
               {tfError
                 ? <span><span className="wiz-running-fg">{tfTitle}:</span> [Failure] {_.startCase(action)} failed</span>
                 : tfTitle}
-              {output && !isApplySuccess &&
+              {output && !showTfActions &&
```
squat: Same as below

PR #2640, opened 2018-01-02
frontend: Merge etcd.jsx and aws-define-nodes.jsx

kyoto:
These 2 files were closely related and depended on each other, so it simplifies things to make them a single file.

Also add `etcdForm` as a field of `awsNodesForm` to be consistent with how most other pages work and also to simplify `AWS_Nodes.canNavigateForward`.


squat:
All tests have passed.

PR #2642, opened 2018-01-02
frontend: Remove "blurry" property

kyoto:
This seems to no longer do anything useful.


sym3tri:
@kyoto just wondering what this originally was intended to do?

kyoto:
@sym3tri Actually, I've realized it does in fact slightly alter the behavior by causing a `Both fields are required` validation error if the field is focussed then unfocussed without entering a value. Other fields only show that error if you set a value and then delete it again.

It seems to me that all fields should have the same behavior, so we should remove this `blurry` option. In other words, I don't see why these fields should be a special case.

PR #2645, opened 2018-01-02
platforms/vmware: Add ability to map VMs to datastores

lander2k2:
Currently, a single datastore must be used by all VMs of a particular
role (etcd, master, worker).  In an environment where the cluster spans
physical datacenters VMs will need to use datastores located at their
datacenter.  This change allows datastores to be mapped to each VM to
accommodate this.


lander2k2:
Jira issue: https://jira.coreos.com/browse/FR-366

PR #2650, opened 2018-01-03
tests: set bash as shell and set -o pipefail to return the correct error code

cpanato:
when using "command 1 | tee bla.log" if the command 1 fails it will return always 0 if tee did not fail. 

using -o pipefail will return the correct error code.

I observed some errors in the tests when the terraform fail but the execution continues. using this it will not continue and fail properly.


enxebre:
retest this please

PR #2652, opened 2018-01-03
modules/ignition,platforms/*: Optionally allow users to set NTP servers for the cluster

lander2k2:
NTP servers on Container Linux default to CoreOS time servers,
which may not be accessible on private cluster deployments.
Allow users to set a list of primary NTP servers with the CoreOS
servers as the default fallback.

This is cherry-picked onto track-1 branch from @lblackstone PR: https://github.com/coreos/tectonic-installer/pull/2213


lander2k2:
@kalmog So it was.  I saw the examples (for all platforms) and didn't look deeper.  I'll add a commit for the rest of the platforms...

kalmog:
rebased at #2747

PR #2661, opened 2018-01-04
tests: bump conformance docker image

cpanato:


enxebre:
You probably need to rebase or may be is github being being drunk, but the diff for examples/terraform.tfvars.govcloud is incorrect as the vanilla stuff is not in master anymore

cpanato:
both tests passed in different runs. merging

PR #2663, opened 2018-01-04
modules/ignition,platforms/*: Enable install behind HTTP proxy

lander2k2:
This addition will allow a cluster to be installed behind an HTTP proxy
on all platforms.


platforms/azure/main.tf
```diff
@@ -184,6 +194,8 @@ module "ignition_workers" {
   kubelet_debug_config    = "${var.tectonic_kubelet_debug_config}"
   kubelet_node_label      = "node-role.kubernetes.io/node"
   kubelet_node_taints     = ""
+  no_proxy                = "${var.tectonic_no_proxy}"
+  tectonic_vanilla_k8s    = "${var.tectonic_vanilla_k8s}"
```
enxebre: tectonic_vanilla_k8s has been removed, need rebase

lander2k2: Ugh.  Sorry 'bout that.  I did see it had been removed.  Mistakenly left in during merge conflict resolution.

cpanato:
@lander2k2  @enxebre  lets wait for this https://github.com/coreos/tectonic-installer/pull/2662 to be merged and then we rebase this PR as well

brianredbeard:
Does this actually need changes to 42 files?  If that's the case what's the argument for avoiding some type of inheritance to keep it [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)?

lander2k2:
@brianredbeard In order to implement across all platforms, it does indeed need changes to 42 files.  8 of those files are just docs and examples, but nonetheless, there is considerable repetition.  The ignition module has improved the DRYness of these kinds of changes but deeper changes to the installer are probably a little beyond the scope of this PR.  We're definitely stretching the capabilities of terraform here.

sym3tri:
DRYness issues are planned to be resolved in forthcoming PRs.

brianredbeard:
@lander2k2 As this is referencing an internal resource which is not managed by the installing administrator, this should also allow for the specification of an internal certificate authority (common with the use of BlueCoat and similar products).

lander2k2:
@brianredbeard I haven't encountered the internal CA requirement related to the proxy at Ford or T-Mobile.  I bet you're right about needing to do it in the future.  Hopefully, we'll be on track 2 by that time and will have a cleaner mechanism to meet that requirement.

kalmog:
ok to test

kalmog:
ok to test

PR #2666, opened 2018-01-05
frontend: Combine some navigation logic

kyoto:
Combine some navigation logic that was split between `Trail`, `app.jsx` and `Base`.
- Refactor away `Trail.canNavigateForward()` method.
- Merge `fixLocation()` and `fixPath()` and move to `Base`.
- Remove unused `Trail.cmp()` method.
- Remove unnecessary `Trail.navigable()` method. This was only used by the sidebar and the sidebar should be able to assume that all pages in the current `Trail` section are navigable.

### DIFF
--- a/installer/frontend/components/base.jsx
+++ b/installer/frontend/components/base.jsx
@@ -4,9 +4,8 @@ import React from 'react';
 import { connect } from 'react-redux';
 import { Route, Switch, withRouter } from 'react-router-dom';
 
-import { fixLocation } from '../app';
 import { withNav } from '../nav';
-import { sections as trailSections, trail } from '../trail';
+import { sections, trail } from '../trail';
 
 import { Loader } from './loader';
 import { PLATFORM_TYPE } from '../cluster-config';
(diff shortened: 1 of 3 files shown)


installer/frontend/components/base.jsx
```diff
@@ -4,44 +4,35 @@ import React from 'react';
...
+    {_.map(section, page => {
+      const classes = classNames('wiz-wizard__nav__step', {'wiz-wizard__nav__step--active': page === currentPage});
+      const link = <li className={classes} key={page.path}>
+        <button
+          className="wiz-wizard__nav__link btn btn-link btn-link-ordinary"
+          disabled={disabled}
+          onClick={() => navigate(page)}
+        >{page.title}</button>
+      </li>;
+      disabled = disabled || !canNavigateForward(page);
```
squat: Shouldn't this value be updated before line 25?

installer/frontend/components/base.jsx
```diff
@@ -90,33 +83,55 @@ const Wizard = withNav(withRouter(connect(stateToProps)(
...
       document.title = `Tectonic - ${this.props.title}`;
     }
 
     componentWillReceiveProps (nextProps) {
-      if (!nextProps.currentPage) {
-        fixLocation();
-      }
-      if (nextProps.title === this.props.title) {
-        return;
+      this.fixLocation();
```
squat: Why we need to fix location on both willMount and on willReceiveProps?

installer/frontend/components/base.jsx
```diff
@@ -4,44 +4,35 @@ import React from 'react';
...
-                >{page.title}</button>
-              </li>
-            );
-          })
-        }
-      </ul>
-    );
-  }
-);
+const NavSection = ({canNavigateForward, currentPage, navigate, navSections, navTrail, title}) => {
```
squat: Purely stylistic, but I prefer `sections` to `navSections` since it is more legible and we already understand that it is related to `nav` from context. Not a blocker.

kyoto: Yeah, I just I'm just trying to avoid clashing with the `sections` defined in `trails.js`, but I guess both should be renamed.

installer/frontend/components/base.jsx
kyoto: It's because the link should only be disabled if the *next* page's `canNavigateForward()` returns `false`.

installer/frontend/components/base.jsx
kyoto: When the component first mounts, `willMount()` is called, but not `willReceiveProps()`. We might need to fix the location when the component first mounts as well as on subsequent property updates, so we need both.

kyoto:
Rebased to handle `canReset` introduced by #2635.

installer/frontend/components/base.jsx
kyoto: I've now added a commit to this PR that renames them.

PR #2681, opened 2018-01-09
config.tf: fix monitoring-grafana quay repo.

diegs:
Also add it to the tectonic-monitoring-config since it was previously
(and mysteriously) unused.

### DIFF
--- a/config.tf
+++ b/config.tf
@@ -101,7 +101,7 @@ variable "tectonic_container_base_images" {
     config_reload            = "quay.io/coreos/configmap-reload"
     addon_resizer            = "quay.io/coreos/addon-resizer"
     kube_state_metrics       = "quay.io/coreos/kube-state-metrics"
-    grafana                  = "quay.io/coreos/grafana-monitoring"
+    grafana                  = "quay.io/coreos/monitoring-grafana"
     grafana_watcher          = "quay.io/coreos/grafana-watcher"
     prometheus_operator      = "quay.io/coreos/prometheus-operator"
     prometheus_config_reload = "quay.io/coreos/prometheus-config-reloader"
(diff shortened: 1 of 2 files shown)


diegs:
cc @coresolve

diegs:
Backport is #2682

diegs:
@cpanato this one failed with:

```
End of journal of etcd-member service on 10.0.102.221
not ok 1 - aws-vpc with a cluster generates operator manifests 
#
#  aws-vpc with a cluster generates operator manifests
#  Failure/Error: raise "timeout waiting for #{service} service to bootstrap on any of: #{ips}"
#
#  RuntimeError:
#    timeout waiting for bootkube service to bootstrap on any of: ["10.0.101.120"]
```

enxebre:
hey @diegs this seems to be unused as you pointed, so we should probably just remove it, I reckon this is now handle by Prometheus operator? cc @brancz

brancz:
> The value doesn't appear to be used anywhere, but we should ensure that it is correct.

You're right, that is the correct value, and it's curious that it's not being used, it should be used here: https://github.com/coreos/tectonic-installer/blob/master/modules/tectonic/resources/manifests/updater/tectonic-monitoring-config.yaml

The default used by the TPO is here, so I wonder what problem this caused (we should still fix it, but I wonder how we got here)?

enxebre:
As far as I can see, it's never been used, I'd suggest to add it
```
grafana:
  baseImage: ${grafana_base_image}
  grafanaWatcherBaseImage: ${grafana_watcher_base_image}
```
run the tests and see what it reveals

brancz:
That sounds good to me.

diegs:
Added the entry to the configmap as well. We'll need to write a migration for that in track 1 if we go forward with it...

brancz:
We have yet to figure out whether this actually caused the problem, as we could not reproduce the problem from master or track 1. Once we figure out what actually caused the problem we’ll make sure we have a migration.

brancz:
This can be closed in favor of #2783

PR #2686, opened 2018-01-09
tests: Set platform and spec for bare-metal

erikkn:
Set the platform and spec value for the bare-metal tests.


erikkn:
@cpanato

mxinden:
@erikkn Why do you introduce the new `action` `baremetal-smoke-test-logs` here?

erikkn:
@mxinden I agree, using a different argument (in this case baremetal-smoke-test-logs) is useless, because nothing is changing in the logic of the loganalyzer-script anyway. I'll create a new PR (will comment the PR number here later) to remove this argument and give it the same name as the 'runRSpecTest' function: smoke-test-logs .

PR #2687, opened 2018-01-09
test: add new argument to logstash

cpanato:


cpanato:
retest this please. on aws

mxinden:
@erikkn Is this needed for the *amount-of-started-clusters* graph? Why not just count the number of ['HOORAY! The cluster is up'](https://github.com/coreos/tectonic-installer/blob/master/tests/rspec/lib/cluster.rb#L242) in a given timeframe?

erikkn:
@mxinden , you are right, we can use this line to create a visualization. I discussed this with Carlos as well, and I think that we could still use this attribute to create something useful. I'll update the Jira ticket and let's discuss this when you are back.
Status: Let's keep the PR open until Max is back

cpanato:
@mxinden @erikkn  should we merge this? or this makes no sense?

erikkn:
Discussed this with Max and we actually don't need this PR. Closing PR.

PR #2690, opened 2018-01-10
frontend: Set the default assets download filename to "assets.zip"

kyoto:
Chrome was already defaulting to `assets.zip`, but Firefox was defaulting to just `assets`.


cpanato:
retest this please

PR #2693, opened 2018-01-10
tests: optimize logstash upload script

erikkn:
Remove the 'baremetal-smoke-test-logs' argument, because this makes no sense. The logic of the script stays the same, so we can just use the smoke-test-logs argument.


erikkn:
@mxinden

cpanato:
retest this please

PR #2694, opened 2018-01-10
Remove region specification in aws basic smoke test

mxinden:
@arithx is using the Tectonic installer to test nightly Container Linux
AMIs. They are only present in `us-west-2`.

Terraform lets values in the tfvars file take precedence over
environment variables. To easily configure the regoin for the aws basic
tests, this patch removes the specification. The other smoke test tfvars
files don't have a region specified.

@cpanato @alexsomesan @squat Do you know of a particular reason, why the region is specified here?

### DIFF
--- a/tests/smoke/aws/vars/aws.tfvars.json
+++ b/tests/smoke/aws/vars/aws.tfvars.json
@@ -6,7 +6,6 @@
   "tectonic_aws_master_root_volume_size": 15,
   "tectonic_aws_master_root_volume_type": "gp2",
   "tectonic_aws_private_endpoints": false,
-  "tectonic_aws_region": "us-west-1",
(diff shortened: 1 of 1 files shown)


alexsomesan:
It may be a leftover from when we didn't have large enough resource limits in most regions. 
The installer does require a region to be specified nonetheless, so we need to make sure the corresponding environment variable is injected by RSpec.

We can try removing this and see what happens.

mxinden:
@alexsomesan Yes, the corresponding env var is injected here: https://github.com/coreos/tectonic-installer/blob/master/tests/rspec/lib/aws_cluster.rb#L17

cpanato:
now I realize we use this for the UI tests

mxinden:
@cpanato could you go into more details?

PR #2698, opened 2018-01-10
Fixed calico bug

thorfour:
The location of calico manifest had changed in x-operator and the operator uses the `tectonic_networking` variable to lookup that path. Changed the setting to reflect the new directory structure. 

The calico manifests also had the same bug that plagued the canal manifests in that the ClusterRole had a namespace in it's metadata which caused x-operator to fail in creating it.


thorfour:
https://github.com/coreos-inc/tectonic-operators/pull/181 reflects the manifest fix

diegs:
ok to test

diegs:
@thorfour i think you need to run `make docs`

PR #2700, opened 2018-01-11
modules/ignition,platforms/*: Allow NFS configuration on CL nodes

lander2k2:
This will allow for the configuration of NFS to, for example, specify
versions of NFS protocol that may be used.  It is flexible enough to
allow any configuration desired by allowing the contents of the config
file to be passed in as a tfvar.


lander2k2:
Closing as this functionality will be implemented differently on track 2.

PR #2707, opened 2018-01-11
tests: ignore was region in the UI tests

cpanato:
We remove the harcoded aws region from the tfvars used in one of our tests. however the UI tests check the aws region, and now it is not there. This PR remove the check against aws region


cpanato:
thanks @sym3tri

PR #2709, opened 2018-01-11
Enable kubelet token authN and RBAC authZ

brancz:
This enables token authN and RBAC authZ on the kubelet (let me know if I missed any files or if this is the only place I have to change something). It goes towards disabling the insecure ports all together. The monitoring stack needs to gather metrics from the kubelets and it would be preferable to use ServiceAccount tokens for that rather than client certificates.

This is only needed in track 2, this does not need to be merged back into track 1.

@alexsomesan @sym3tri 

/cc @ericchiang

### DIFF
--- a/modules/ignition/resources/services/kubelet.service
+++ b/modules/ignition/resources/services/kubelet.service
@@ -42,6 +42,8 @@ ExecStart=/usr/lib/coreos/kubelet-wrapper \
   --pod-manifest-path=/etc/kubernetes/manifests \
   --require-kubeconfig \
   --rotate-certificates \
+  --authentication-token-webhook \
+  --authorization-mode=Webhook \
   ${cloud_provider_config} \
   ${debug_config} \
   ${node_taints_param}


ericchiang:
I remember this having some interesting interactions with the checkpointer. Would be good to know there was a reboot test before this merged.

ericchiang:
The change looks correct to me.

brancz:
I've been running the kubelet with those flags for a while on my private cluster, and it seems to be working fine. Maybe @diegs or @thorfour can comment how this might affect the checkpointer. My understanding is that since the checkpointer switched to requesting the CRI, the kubelet's API is not used anymore.

diegs:
Correct, checkpointer does not use the kubelet internal (`/runningpods`) API, and the other API doesn't do much either. I think it would be prudent to test this change in bootkube first if you haven't already.

brancz:
I'm happy to test any scenario you'd like. What exactly with bootkube would you like me to look at? Given that the smoke tests worked fine and I run my private cluster (which is also using bootkube) with these settings I'm not sure what else to test.

In any case I'll be going through the kubelet systemd units in bootkube to make sure we test there with the same settings.

diegs:
Smoke tests did not run on this PR.

brancz:
My mistake, of course we need to run those and ensure they pass reliably.

enxebre:
ok to test

alexsomesan:
@CI failing with this error `outdated yarn.lock (commit it to fix)`
Maybe a rebase is needed? Haven't seen this one yet...

alexsomesan:
Is this change effectively disabling TLS auth in the Kubelet or is it going to coexist with token based auth?
Is the kubeconfig required to contain a valid token for this to work?

alexsomesan:
/cc: @aaronlevy and @crawford as this may affect the node configuration operator and the TLS bootstrapping story.

brancz:
TLS client certs continue to work, this is just that you can use ServiceAccount tokens in addition to client certs for authN. This is what the `--authentication-token-webhook` flag does. Then the `--authorization-mode=Webhook` flag allows performing authZ for the obtained identity (obtained either through the client cert or ServiceAccount token).

Come to think about it, actually I needed the apiserver to have additional permissions, otherwise reading logs did not work. This is what I had to add:

```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: kube-apiserver
rules:
- apiGroups:
  - ""
  resources:
  - nodes/proxy
  verbs:
  - get
  - create
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: kube-apiserver
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: kube-apiserver
subjects:
- apiGroup: rbac.authorization.k8s.io
  kind: User
  name: kube-apiserver
```

Which component would be the appropriate one for us to add these with?

alexsomesan:
@brancz Since yesterday, the control plane manifests are no longer managed by the installer.
Instead, they are the concern of the new kube-core-operator, which renders them in a similar manner to bootkube.

You can find the templates here:
https://github.com/coreos-inc/tectonic-operators/tree/master/operator/kube-core/spec/manifests
and here for the bootstrap control plane:
https://github.com/coreos-inc/tectonic-operators/tree/master/operator/kube-core/spec/bootstrap-manifests

@diegs is the point of contact for that code.

cpanato:
removing conformance tests because it fail in the delete pods/namespaces.

ericchiang:
Should this be closed?

brancz:
I would still like this to land for Beryllium, I'm planning to get back to this soon. My initial hope was that it would just take this PR :)

diegs:
@brancz should be a quick PR in the tectonic operators repo to move the manifest changes there. LMK if you need any help.

brancz:
Are kubelets already handled by the operators now?

diegs:
No, kubelets will be handled by the NCG/TNCO stuff.

sym3tri:
@brancz closing b/c stale & this should be handled in the TNC now. Let's discuss if you still need this.

PR #2720, opened 2018-01-13
removed version from operators that are unversioned

thorfour:
some manifests don't have associated versions. Removing the placeholder version and updating their appversions to facilitate the install


yifan-gu:
ok to test

PR #2722, opened 2018-01-13
frontend: Improve Matchbox error message

kyoto:


installer/frontend/components/bm-matchbox.jsx
```diff
@@ -61,7 +61,7 @@ const dispatchToProps = dispatch => ({
       .then(value => {
         const available = value.coreos;
         if (!available || available.length === 0) {
-          return Promise.reject(`could not find any coreos images at ${endpointURL}`);
+          return Promise.reject(`Could not find any coreos images at ${endpointURL}`);
```
squat: Can we also fix the `coreos` here? It should really say `Container Linux images`. Or at the very least `CoreOS`.

sym3tri: Please use `Container Linux`

kyoto: Thanks. I changed it to `Container Linux images`.

PR #2723, opened 2018-01-15
frontend: Clean up FieldList and fix AWS tags validation

kyoto:
Add `FieldRowList` component, which replaces `ConnectedFieldList` and `InnerFieldList_` and also handles the add and remove buttons.

Greatly simplify `FieldList` class.

This makes the `autoFocus` behavior consistent for the 3 field lists (AWS Tags, BM master nodes, BM worker nodes).

Make `appendField` and `removeField` only available to field lists instead of to any component wrapped by `Connect` because they are only applicable to field lists.

Merge `TagsWithPlaceholder` functionality into `AWS_Tags`.

Also fixes the AWS tags validation:
  - Fixes a bug where the `Tag keys must be unique` error was only displayed if the first tag row contained one of the duplicate values.
  - Allow the `Both fields are required` error to be displayed on either the key input or the value input, rather than just on the value input. This fixes an issue where no error was displayed if the key input was cleared, but the value input was not marked as "dirty".

Add more Nightwatch tests for AWS tags validation.

### DIFF
--- a/installer/frontend/components/bm-nodeforms.jsx
+++ b/installer/frontend/components/bm-nodeforms.jsx
@@ -11,13 +11,8 @@ import { MAX_MASTERS, MAX_WORKERS } from './nodes';
 import { validate } from '../validate';
 import { readFile } from '../readfile';
 import { FieldList, Form } from '../form';
-
-import {
-  BM_MASTERS,
-  BM_WORKERS,
-} from '../cluster-config';
-
-import { Input, Connect } from './ui';
+import { BM_MASTERS, BM_WORKERS } from '../cluster-config';
+import { Input, Connect, FieldRowList } from './ui';
 
 const BulkUpload = connect(null, dispatch => ({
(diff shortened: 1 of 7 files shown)


installer/frontend/components/bm-nodeforms.jsx
```diff
@@ -266,59 +254,43 @@ class NodeForm extends React.Component {
...
         </div>
+
+        <FieldRowList id={id} Row={NodeRow} rowFields={rowFields} />
       </div>
     </div>;
   }
 }
 
-const mastersFields = generateField(BM_MASTERS, 'Master', 9);
+const mastersFields = makeFieldList(BM_MASTERS, 'Master', 9);
```
squat: ah, needs a rebase here

squat:
Needs rebase, then let's merge it! :shipit:

kyoto:
@squat Rebased

squat:
Great work with all this clean up!!!

PR #2731, opened 2018-01-16
frontend: Merge file upload handlers

kyoto:
Add a new component `FileInput` to extract out the file upload handler logic from `FileArea` and `cluster-info.jsx`.

Also clean up `FileArea` a bit.


kyoto:
ok to test

PR #2735, opened 2018-01-16
tests: run tests on a running cluster

cpanato:
fixes: https://jira.coreos.com/browse/INST-520

### DIFF
--- a/tests/rspec/lib/cluster.rb
+++ b/tests/rspec/lib/cluster.rb
@@ -27,6 +27,7 @@ class Cluster
     @name = ENV['CLUSTER'] || NameGenerator.generate(@tfvars_file.prefix)
     @tectonic_admin_email = ENV['TF_VAR_tectonic_admin_email'] || NameGenerator.generate_fake_email
     @tectonic_admin_password = ENV['TF_VAR_tectonic_admin_password'] || PasswordGenerator.generate_password
+    save_console_creds(@name, @tectonic_admin_email, @tectonic_admin_password)
 
     @build_path = File.join(File.realpath('../../'), "build/#{@name}")
     @manifest_path = File.join(@build_path, 'generated')
@@ -50,6 +51,10 @@ class Cluster
 
   def update_cluster
     start
+  end
+
+  def init
+    terraform_init
   end
 
   def stop
@@ -188,6 +193,17 @@ class Cluster
     raise e
   end
 
+  def terraform_init
+    ::Timeout.timeout(30 * 60) do # 30 minutes
+      env = env_variables
+      env['TF_INIT_OPTIONS'] = '-no-color'
+      return true if system(env, 'make -C ../.. terraform-init')
+    end
+  rescue Timeout::Error
+    forensic(false)
+    raise 'Terraform init failed'
+  end
+
   def recover_from_failed_destroy() end
 
   def wait_til_ready
@@ -237,6 +253,7 @@ class Cluster
 
   def wait_for_bootstrapping
     ips = master_ip_addresses
+    raise 'Empty master ips. Aborting...' if ips.empty?
     wait_for_service('bootkube', ips)
     wait_for_service('tectonic', ips)
     puts 'HOORAY! The cluster is up'
--- a/tests/rspec/lib/shared_examples/k8s.rb
+++ b/tests/rspec/lib/shared_examples/k8s.rb
@@ -12,6 +12,7 @@ require 'password_generator'
 require 'webdriver_helpers'
 require 'test_container'
 require 'with_retries'
+require 'jenkins'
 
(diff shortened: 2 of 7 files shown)


tests/smoke/azure/README.md
```diff
@@ -33,3 +33,10 @@ To run the entire test suite for Azure, use this command:
 export TEST_VARS=vars/Azure-exp.tfvars
 make tests/smoke TEST='azure_*'
 ```
+
+
+## Running the rspec tests in a existing cluster
```
squat: This should either go in all the READMEs or in a general one

tests/rspec/lib/shared_examples/k8s.rb
```diff
@@ -24,20 +25,29 @@
   include_examples('withRunningClusterExistingBuildFolder', vpn_tunnel)
 end
 
-RSpec.shared_examples 'withRunningClusterExistingBuildFolder' do |vpn_tunnel = false|
+RSpec.shared_examples 'withRunningClusterExistingBuildFolder' do |vpn_tunnel = false, exist_tf = nil, exist_plat = nil|
```
squat: can we make the order of `exist_tf` and `exist_plat` the same here and for `ClusterFactory.from_variable`

tests/smoke/azure/README.md
cpanato: yes, wrong place. my bad

tests/rspec/lib/cluster.rb
```diff
@@ -188,6 +193,17 @@ def destroy
...
 
+  def terraform_init
+    ::Timeout.timeout(30 * 60) do # 30 minutes
+      env = env_variables
+      env['TF_INIT_OPTIONS'] = '-no-color'
+      return true if system(env, 'make -C ../.. terraform-init')
+    end
+  rescue Timeout::Error
+    forensic(false)
+    raise 'Applying cluster failed'
```
mxinden: s/Applying cluster failed/terraform init failed/

tests/rspec/lib/shared_examples/k8s.rb
```diff
@@ -145,6 +155,7 @@
       # remove platform AZURE when the JIRA https://jira.prod.coreos.systems/browse/INST-619 is fixed
       skip_platform = %w[metal azure gcp]
       skip "This test is not ready to run in #{platform}" if skip_platform.include?(platform)
+      skip 'Skipping this tests. running locally' unless Jenkins.environment?
```
mxinden: Why are you skipping this test, when it is run locally?

tests/rspec/spec/existing_cluster_spec.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+require 'shared_examples/k8s'
+require 'env_var'
+
+RSpec.describe 'existing_cluster' do
+  before(:all) do
+    raise 'Missing existing platform. Please set PLATFORM environment variable' unless EnvVar.set?(%w[PLATFORM])
```
mxinden: Great, I like the "fail fast" approach!
Can you do the same with the `CLUSTER` variable?

tests/rspec/lib/shared_examples/k8s.rb

tests/rspec/lib/cluster.rb
cpanato: done, thanks!

tests/rspec/lib/shared_examples/k8s.rb
cpanato: because this is the scale up, and if you run for example 3-4 times or more it will scale up your cluster a lot. also takes some time and the developer maybe need just a fast feedback loop.

the scale-up tests are used for a single test or in CI. not for local tests and debug.
if the dev really need this he can comment out this part

tests/rspec/spec/existing_cluster_spec.rb
cpanato: done

mxinden:
@cpanato One general idea: The console password and username are both saved inside a Kubernetes secret in the `tectonic-system` namespace. How about reading it from there instead of saving it into a `.json` file locally and reading it from there.

This would make the process more robust, because users don't need to worry about the environment variables. What do you think?

cpanato:
@mxinden this definitely is a good idea, however, the password is a hash and I cannot decode that
for example, this is the configmap looks like:

```
<....>
 staticPasswords:
    - email: "zora_fisher@zieme.biz"
      hash: "$2a$12$msOJPEYv.agnG6Yf6FYmYeKS7zIy93ft0oqqxiG5bX6JfKXzTvNxK"
      username: "admin"
      userID: "Yqo-0IiDywGQJzkghXaGKw"
<....>
```

tests/rspec/lib/shared_examples/k8s.rb
mxinden: 👍

PR #2736, opened 2018-01-16
Bazel: fix Golang compilation and introduce X-compilation

squat:
This PR adds a `gazelle` rule so that all the Golang Bazel.build
files can be updated by running:

```sh
bazel run gazelle
```

This commit also fixes the WORKSPACE and Bazel.build files in several
directories so that:

1. the installer binary can be successfully built; and
2. the binary can be cross-compiled from Linux to Darwin.

To build the binary, run:

```sh
bazel build backend
```

To cross-compile the binary for Darwin, run:

```sh
bazel build backend --experimental_platforms=@io_bazel_rules_go//go/toolchain:darwin_amd64
```

*Note:* currently, the cross-compiled binary for Darwin cannot be
compiled statically because GCC on Linux does is missing some
Darwin-specific options.

*Note*: the second commit is all BUILD file that were regenerated by gazelle. Please focus on the first and third commits.


WORKSPACE
```diff
@@ -11,8 +11,8 @@ supported_platforms = [
 
 http_archive(
     name = "io_bazel_rules_go",
-    sha256 = "91fca9cf860a1476abdc185a5f675b641b60d3acf0596679a27b580af60bf19c",
-    url = "https://github.com/bazelbuild/rules_go/releases/download/0.7.0/rules_go-0.7.0.tar.gz",
+    url = "https://github.com/bazelbuild/rules_go/releases/download/0.8.1/rules_go-0.8.1.tar.gz",
```
squat: We can't yet use 0.9 because that version breaks cross compilation https://github.com/bazelbuild/rules_go/issues/1240

PR #2740, opened 2018-01-16
tests: all pods should be in running and none in crashloopbackoff

cpanato:
This is blocked by: https://jira.coreos.com/browse/ALM-423

### DIFF
--- a/tests/smoke/cluster_test.go
+++ b/tests/smoke/cluster_test.go
@@ -83,6 +83,21 @@ func testAllPodsRunning(t *testing.T) {
 }
 
 func allPodsRunning(t *testing.T) error {
+	err := checkPodsRunning(t)
+	if err != nil {
+		return err
+	}
+
+	// check one more time, because in case of crashloop we might get the transition
+	time.Sleep(5 * time.Second)
+	err = checkPodsRunning(t)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func checkPodsRunning(t *testing.T) error {
 	c, _ := newClient(t)
 	pods, err := c.Core().Pods("").List(meta_v1.ListOptions{})
 	if err != nil {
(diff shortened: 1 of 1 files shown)


tests/smoke/cluster_test.go
```diff
@@ -94,14 +94,27 @@ func allPodsRunning(t *testing.T) error {
 
 	allReady := len(pods.Items) != 0
```
enxebre: Seems like a good case for a retry function? e.g. https://github.com/coreos/tectonic-torcx/blob/6a9620014cce2cf580ad9b91e6b82f202d8eb392/internal/kube.go#L198

cpanato: this test already use retry: https://github.com/cpanato/tectonic-installer/blob/dcee8e0d356b53ed163981c0b2bdf4d382ab8e5c/tests/smoke/cluster_test.go#L81

squat: All test points are already wrapped in retry logic :+1:

cpanato:
@squat @enxebre can you PTAL ?

tests/smoke/cluster_test.go
```diff
@@ -91,14 +91,27 @@ func allPodsRunning(t *testing.T) error {
...
+	// check one more time, because in case of crashloop we might get the transition
+	time.Sleep(5 * time.Second)
+	pods, err = c.Core().Pods("").List(meta_v1.ListOptions{})
+	if err != nil {
+		return fmt.Errorf("could not list pods: %v", err)
+	}
+
+	for _, p := range pods.Items {
+		if p.Status.Phase != v1.PodRunning || p.Status.ContainerStatuses[0].State.Running == nil {
+			return errors.New("pods are not all ready")
```
squat: duplicating all this code is a little dirty. can we make this whole section identical to the one above, extrapolate it into a function, and just call the function twice with a timer in the middle?

cpanato: sure, will do that

cpanato:
@squat  PTAL

PR #2742, opened 2018-01-17
Allow manually setting resolved domains to get around AWS DHCP search path issues

bshelton229:
Hi, we're running into an issue where we use multiple space separated domain names in our AWS dhcp option set for the VPC we've been installing tectonic into. One of the domain names in the search path is the AWS default, but we're running into this "bug" https://github.com/coreos/bugs/issues/1934. This causes the required default DNS resolution needed by tectonic in AWS to not work.

Hopefully this PR can start a conversation. I'm not completely sure what the best way to generalize the ability to add additional domains to /etc/resolv.conf would be. This approach is working in my testing. It seems to behave as I would expect when the domains string is empty. We would set the region appropriate "us-west-2.compute.internal" domain in "resolved_domains" to make tectonic work in our VPC where dhcp isn't setting the right dns search path. The patch we're currently using in our installer is here - https://github.com/roverdotcom/tectonic-installer/pull/1/files

Does it seem reasonable to include something like this in the installer? If so, I'd be glad to help with a better PR if there are other ideas for generalizing the ability to set custom domain names in the search path.

Thanks!

### DIFF
--- a/modules/aws/master-asg/ignition.tf
+++ b/modules/aws/master-asg/ignition.tf
@@ -8,6 +8,7 @@ data "ignition_config" "main" {
     "${var.ign_max_user_watches_id}",
     "${var.ign_s3_puller_id}",
     "${var.ign_ca_cert_id_list}",
+    "${var.ign_resolved_domains_id}",
   ]
 
   systemd = ["${compact(list(
--- a/modules/aws/master-asg/variables.tf
+++ b/modules/aws/master-asg/variables.tf
@@ -147,3 +147,7 @@ variable "dns_server_ip" {
   type    = "string"
   default = ""
 }
+
+variable "ign_resolved_domains_id" {
+  type = "string"
+}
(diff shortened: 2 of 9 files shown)


robszumski:
Hi there @bshelton229, we are currently refactoring how the installer bootstraps machines, and a large part of that is how the Ignition profiles work under the hood. The ideal solution for this (and is planned) if for each user to be able to specify an Ignition profile that can contain a number of things, such as files to place on disk (like your drop in), set up bounded networking, mount storage in a particular way, etc. This will cover the bulk of the customizations folks have been making to the installer. This new profile will be appended to the base profile before the machines are booted.

Note that this functionality is not in master but we are rapidly working towards it. Using this code is a good option in the short term but will be a base feature with the refactor.

bshelton229:
Thanks @robszumski that solution sounds great.

enxebre:
Hey @bshelton229 many thanks for reporting this. I'm closing it for now as @robszumski mentioned this feature should be covered with the new approach. Thanks!

PR #2748, opened 2018-01-17
platforms/metal: fix iscsi variable in Matchbox templates

squeed:
Looks like this variable was typo'd in the iscsi PR.


cpanato:
i will follow this, looks like this is the issue with our bare metal tests

cpanato:
will cherry pick this to track-1

PR #2749, opened 2018-01-17
Fix example generator rules for openstand + vendored deps

alexsomesan:
This hand-tweaks the paths to dependencies for the Gazelle generated build files.
What we should really do is unify the vendored packages under a single vendor folder at the top of the repo.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -41,7 +41,7 @@ genrule(
     cmd = "cp $(<) $(@)",
 )
 
-PLATFORMS = ["aws", "azure", "gcp", "govcloud", "metal", "openstack", "vmware"]
+PLATFORMS = ["aws", "azure", "gcp", "govcloud", "metal", "openstack/neutron", "vmware"]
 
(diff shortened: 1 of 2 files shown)


BUILD.bazel
```diff
@@ -41,7 +41,7 @@ genrule(
     cmd = "cp $(<) $(@)",
 )
 
-PLATFORMS = ["aws", "azure", "gcp", "govcloud", "metal", "openstack", "vmware"]
+PLATFORMS = ["aws", "azure", "gcp", "govcloud", "metal", "openstack/neutron", "vmware"]
```
squat: Unfortunately, this breaks the example tfvars generation since the slash causes the path to be interpreted as a directory with a leaf called `neutron`

alexsomesan: I think we can drop example var generation. We'll soon need to replace that with example cluster config generation, anyway.

mxinden:
Closing here in favour of #2762

PR #2750, opened 2018-01-17
bazel: create tarball target

squat:
This commit adds tarball targets for Bazel.

We can now create tarballs using:

```sh
bazel build tarball
```

This will create a gzip compressed tarball named tectonic.tar.gz. The
contents of the tarball will be nested in a directory named `tectonic`.

To create a tarball with a version string in the directory name, we can
build the tarball using:

```sh
export VERSION=<some version>
bazel build tarball --action_env=VERSION
```

In this case, the file structure will be nested in a directory called
`tectonic_<some version>`.

This commit also makes the installer binaries pure go so that they are
statically linked, which is nicer for containers.

cc @alexsomesan

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -102,3 +102,83 @@ alias(
     name = "backend",
     actual = "//installer/cmd/installer:installer",
 )
+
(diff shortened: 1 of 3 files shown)


squat:
ok to test

PR #2752, opened 2018-01-17
platforms/metal: fix iscsi variable in Matchbox templates

kalmog:
fix var typo in matchbox templates (see #2748)


squat:
ok to test

PR #2775, opened 2018-01-19
 Add terraform fmt test rule to bazel

mxinden:
Before adding all other format checkers I would like to propose this patch for early feedback. I will follow up with the others once this is merged.

Let me know what you think? Happy for any feedback.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,3 +1,5 @@
+load("//bazel-rules:gen_test.bzl", "gen_test")
+
 package(
     default_visibility = ["//visibility:public"],
 )
@@ -28,6 +30,7 @@ genrule(
     }),
     outs = ["bin/terraform"],
     cmd = "cp $(<) $(@)",
+    executable = True,
 )
 
 genrule(
@@ -181,4 +184,19 @@ pkg_tar(
     mode = "0755",
     package_dir = "tectonic-installer/linux",
     srcs = ["@terraform_provider_matchbox_linux//:terraform-provider-matchbox"],
+)
+
+filegroup(
+    name = "terraform_files",
+    srcs = glob(
+        ["**/*.tf"],
+        exclude=["bazel-bin/**/*", "bazel-out/**/*", "bazel-genfiles/**/*", "build/**/*"],
+        exclude_directories=1,
+    ),
+)
+
+gen_test(
+    name = "terraform_fmt",
+    command = "./bin/terraform fmt -list -check -write=false",
+    deps = [":terraform_runtime", ":terraform_files"],
 )
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,6 +1,6 @@
 workspace(name = "tectonic_installer")
 
-terrafom_version = "0.10.8"
+terrafom_version = "0.11.2"
 
 provider_matchbox_version = "0.2.2"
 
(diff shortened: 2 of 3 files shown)


WORKSPACE
```diff
@@ -1,6 +1,6 @@
 workspace(name = "tectonic_installer")
 
-terrafom_version = "0.10.8"
+terrafom_version = "0.11.0"
```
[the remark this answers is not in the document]
alexsomesan: Yeah, definitely. Upstream is already at 0.11.2.
We can try going all the way there. There should be no breaking changes, according to changelog.

BUILD.bazel
```diff
@@ -102,3 +105,18 @@ alias(
...
+    name = "terraform_files",
+    srcs = glob(
+        ["**/*.tf"],
+        exclude=["bazel-bin/**/*", "bazel-out/**/*", "bazel-genfiles/**/*", "build/**/*"],
+        exclude_directories=1,
+    ),
+)
+
+gen_test(
+    command = "./bin/terraform fmt -list -check -write=false",
```
squat: nit: it's convention to put the name first

squat: rather than use `./bin/terraform` this should be `$(location :terraform_runtime)`

squat: Also, this command should operate on a set of file, otherwise, how do we know what files are being checked? The whole command should probably be something like:
```
$(location :terraform_runtime) -list -check -write=false $(SRCS)
```
where the `:terraform_files` filegroup is specified as a source.

mxinden: Using `$(location :terraform_runtime)` would be my preferred method as well. I am still facing the following issues when trying to achieve this:
- `$(location :terraform_runtime)` via the [`ctx.expand_location`](https://docs.bazel.build/versions/master/skylark/lib/ctx.html#expand_location) function expands to `bazel-out/k8-fastbuild/bin/bin/terraform`.
- The `bazel-out` folder is not symlinked into the working directory of a test.
- There are no enviroment variables injected that point to the `bazel-out` directory.

According to the bazel [test encyclopedia](https://docs.bazel.build/versions/master/test-encyclopedia.html#runfiles) one should leverage the [`runfiles`](https://docs.bazel.build/versions/master/test-encyclopedia.html#runfiles) concept. I am doing this via:
``` python
runfiles = ctx.runfiles(files=ctx.files.deps)
return [DefaultInfo(runfiles=runfiles)]
```
This makes all directly referenced files available in the current directory itself and all binary targets available in the `bin/` folder. Downside to this approach, is, that the files are not referenced directly in the `command =` string.

I have looked into how other `github.com/bazelbuild/rules_***` projects solve this. One example would be [`_jsonnet_to_json_test_impl`](https://github.com/bazelbuild/rules_jsonnet/blob/745566196e26107042a5f787f34ded8eaeb908fb/jsonnet/jsonnet.bzl#L177), which heavily depends on setup functions, which I don't feel comfortable just importing into this project.

Let me know if you have more ideas. As of now, I would suggest this solution as a first iteration.

WORKSPACE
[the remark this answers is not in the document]
mxinden: 👍

mxinden:
@squat Linking [comment](https://github.com/coreos/tectonic-installer/pull/2775#discussion_r162862200) here, so it woun't get lost.

mxinden:
Import path is currently broken. Fixing it right now.

mxinden:
Should be fixed now.

PR #2777, opened 2018-01-19
Explicitly set `kubelet-preferred-address-types` flag for apiserver

justaugustus:
Explicitly sets `--kubelet-preferred-address-types=InternalDNS,InternalIP,Hostname,ExternalDNS,ExternalIP` on the apiserver.

This allows API --> kubelet communication to succeed in instances where a DNS search domain is not present in `/etc/resolv.conf`

Fixes: https://jira.coreos.com/browse/TECREL-150

### DIFF
--- a/modules/bootkube/resources/manifests/kube-apiserver.yaml
+++ b/modules/bootkube/resources/manifests/kube-apiserver.yaml
@@ -51,7 +51,7 @@ spec:
         - --tls-private-key-file=/etc/kubernetes/secrets/apiserver.key
         - --kubelet-client-certificate=/etc/kubernetes/secrets/apiserver.crt
         - --kubelet-client-key=/etc/kubernetes/secrets/apiserver.key
-        - --kubelet-preferred-address-types=InternalIP
+        - --kubelet-preferred-address-types=InternalIP,ExternalIP,Hostname
         - --service-account-key-file=/etc/kubernetes/secrets/service-account.pub
         - --client-ca-file=/etc/kubernetes/secrets/ca.crt
         - --authorization-mode=RBAC
--- a/modules/bootkube/resources/manifests/kube-apiserver.yaml
+++ b/modules/bootkube/resources/manifests/kube-apiserver.yaml
@@ -51,6 +51,7 @@ spec:
         - --tls-private-key-file=/etc/kubernetes/secrets/apiserver.key
         - --kubelet-client-certificate=/etc/kubernetes/secrets/apiserver.crt
         - --kubelet-client-key=/etc/kubernetes/secrets/apiserver.key
+        - --kubelet-preferred-address-types=InternalIP
         - --service-account-key-file=/etc/kubernetes/secrets/service-account.pub
         - --client-ca-file=/etc/kubernetes/secrets/ca.crt
         - --authorization-mode=RBAC
--- a/modules/bootkube/resources/bootstrap-manifests/bootstrap-apiserver.yaml
+++ b/modules/bootkube/resources/bootstrap-manifests/bootstrap-apiserver.yaml
@@ -28,6 +28,7 @@ spec:
     - --advertise-address=${advertise_address}
     - --kubelet-client-certificate=/etc/kubernetes/secrets/apiserver.crt
     - --kubelet-client-key=/etc/kubernetes/secrets/apiserver.key
+    - --kubelet-preferred-address-types=InternalIP,ExternalIP,Hostname
     - --runtime-config=""
     - --secure-port=443
     - --service-account-key-file=/etc/kubernetes/secrets/service-account.pub


diegs:
cc @thorfour in case this ends up being cherry-picked into lithium and requires a migration in the KVO.

coresolve:
I think the order should be:
`InternalIP,Hostname,InternalDNS,ExternalDNS,ExternalIP`
cause in the AWS case the InternalDNS is only resolvable by AWS domain name servers. One of the assumptions that lead to this PR is that the cluster admin has change the resolvers and presumably the searchpath. 

If we use `InternalIP` then that also saves us a dns call. 

One interesting note. The kubelet generates a cert and places this cert in `/var/lib/kubelet/pki` It's a self signed cert and it has a SAN matching the hostname. 
In Azure this cert looks like: 

```
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number: 1 (0x1)
    Signature Algorithm: sha256WithRSAEncryption
        Issuer: CN=kube-lego-master-0@1516654753
        Validity
            Not Before: Jan 22 20:59:13 2018 GMT
            Not After : Jan 22 20:59:13 2019 GMT
        Subject: CN=kube-lego-master-0@1516654753
        Subject Public Key Info:
            Public Key Algorithm: rsaEncryption
                Public-Key: (2048 bit)
                Modulus:
                      Exponent: 65537 (0x10001)
        X509v3 extensions:
            X509v3 Key Usage: critical
                Digital Signature, Key Encipherment, Certificate Sign
            X509v3 Extended Key Usage:
                TLS Web Server Authentication
            X509v3 Basic Constraints: critical
                CA:TRUE
            X509v3 Subject Alternative Name:
                DNS:kube-lego-master-0
    Signature Algorithm: sha256WithRSAEncryption
```
So I wanted to ensure that we can connect to it without a problem. Turns out the apiserver doesn't validate the cert when connecting to the node for calls like `exec logs attach cp`

Looks like InternalIP may be the winner here. Until we get to bootstrap tls and then we need to ensure that the SAN for the kubelet includes the InternalIP.

justaugustus:
@sym3tri / @alexsomesan / @squat--
Can we get eyes on this tomorrow? @coresolve and I have done some validation, but we'd like to be sure this is a reasonable change for all platforms.

justaugustus:
To Duffie's point (https://github.com/coreos/tectonic-installer/pull/2777#issuecomment-359594193), if we use `InternalIP,Hostname,InternalDNS,ExternalDNS,ExternalIP`, theoretically, we should be able to _ALWAYS_ return `InternalIP`, right?

In that case, why not just have the stringSlice as `--kubelet-preferred-address-types=InternalIP`?

If `InternalIP` is broken, we're in a bad state anyway, right?

cc: @ericchiang / @diegs / @rphillips

justaugustus:
Setting `--kubelet-preferred-address-types=InternalIP` and updated the discussion in https://jira.coreos.com/browse/TECREL-150

justaugustus:
@sym3tri / @mxinden / @cpanato --
Can we have someone apply all of the platform labels and run smoke tests, so that we can get some initial signal on this?

diegs:
ok to test

diegs:
retest this please

modules/bootkube/resources/manifests/kube-apiserver.yaml
```diff
@@ -51,6 +51,7 @@ spec:
         - --tls-private-key-file=/etc/kubernetes/secrets/apiserver.key
         - --kubelet-client-certificate=/etc/kubernetes/secrets/apiserver.crt
         - --kubelet-client-key=/etc/kubernetes/secrets/apiserver.key
+        - --kubelet-preferred-address-types=InternalIP
```
diegs: FWIW this is what kubeadm sets:

https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/phases/controlplane/manifests.go#L169

cc @aaronlevy

justaugustus:
retest this please

modules/bootkube/resources/manifests/kube-apiserver.yaml
diegs: @justaugustus I would feel slightly better copying what kubeadm does...

justaugustus: @diegs I'm down for that too. 
Honestly, any method that prefers `InternalIP` over `Hostname` should address the issue.

I'll push that change. Thanks for the feedback!!

cpanato:
retest this please. metal only

cpanato:
retest this please. govcloud only

modules/bootkube/resources/manifests/kube-apiserver.yaml
justaugustus: Updated!

justaugustus:
retest this please. azure only

sudhaponnaganti:
@justaugustus @robszumski -a temp branch has been created for you to submit this PR 1.8.7-cust. We will make a build once your PR gets merged. 
/cc @coverprice

justaugustus:
@sudhaponnaganti / @robszumski -- here's the PR to merge to `1.8.7-cust`: #2849

sudhaponnaganti:
@justaugustus - can you merge it - I do not see any option to merge it by myself. Also custom_tls check is failing but you can delete it from checks as this is only temp branch. This suite has been passed on 1.8.7 branch.

cpanato:
@justaugustus do we need to merge this pr? if yes can you rebase? otherwise please close

diegs:
@cpanato I believe this is on hold until further testing is done in the custom branch (#2849). Adding a do not merge label.

justaugustus:
@cpanato Yup, what Diego said. Testing in progress on the custom branch.
Thanks @diegs for updating labels on this!!

justaugustus:
I took a look at using `--kubelet-preferred-address-types=InternalIP,ExternalIP,Hostname` on the following platforms:
- AWS
- AWS GovCloud
- Azure

Confirming that QA checks out on each platform.

cpanato:
retest this please

cpanato:
retest this please. aws/govcloud/metal

squat:
@diegs @thorfour just clarifying one of Diego’s comment and towards the top of this issue; 1.8.7 _is_ Lithium so we will need a migration once this merges even if there is no cherry-pick

diegs:
Yes, @thorfour will have to make a migration for existing clusters if this goes in with Lithium. Thanks @cpanato !

PR #2789, opened 2018-01-22
frontend: Remove old AsyncSelect component

kyoto:
Switch single use of the old `AsyncSelect` to use `Selector` instead. Then rename `Selector` to `AsyncSelect`.


squat:
looks good, we just need to rebase

kyoto:
@squat Rebased. Thanks!

PR #2791, opened 2018-01-22
tests: add max log file size (1GB) feature

erikkn:
It happens sometimes that log files are extremely large, which causes Logstash to crash. This could happen due Jenkins failure, infinite loop or something else. A typical log files is less than a GB, so this PR adds this check. If the file is bigger, just remove it and continue.


erikkn:
ok to test

erikkn:
@mxinden

PR #2795, opened 2018-01-23
frontend: Output custom subnet tfvars if Advanced Networking is opened

kyoto:
It used to be that `AWS_CONTROLLER_SUBNETS` and `AWS_WORKER_SUBNETS` were
marked as dirty as soon as the Advanced Networking section was opened,
but that stopped being the case due to AWS VPC and forms changes. We
should therefore track the dirtiness of `AWS_ADVANCED_NETWORKING` itself
and use that to decide whether to output master and worker custom
subnet tfvars.


kyoto:
For reference, if one of the Advanced Networking fields was changed, then those advanced options were correctly passed to the backend. However, if you opened the Advanced Networking section and saw the default values, but didn't edit them. Then the actual subnets created would not match those seen in the Advanced Networking options.

PR #2802, opened 2018-01-23
Multistep bootstrap - WIP

enxebre:
This decouples the main bootstrapping process into 5 steps:
To deploy: `mkdir cluster-aws; cd cluster-aws` and create your  `terraform.tfvars`
- Generate tls content
`terraform apply -state tls.tfstate ../steps/tls/`
- Generate assets e.g operator manifests (and NCG ignition config manifests, this will need to be consolidated with NCG/chain model generation)
`terraform apply -state assets.tfstate ../steps/assets/`
- Generate Ignition config for bootstrap node
`terraform apply -state ignition.tfstate ../steps/ignition/`
- Deploy the bootstrap node which runs the ncg
`terraform apply -state bootstrap.tfstate ../steps/bootstrap/`
wait for ncg to run, then:
`terraform destroy -force -state=bootstrap.tfstate -target aws_route53_record.tectonic_ncg ../steps/bootstrap/`
- Scale the cluster to the target size by deploying the joining nodes consuming from the NCG
`terraform import -state=joining.tfstate -config=../steps/joining/ aws_autoscaling_group.masters cluster-aws-masters`
`terraform apply -state joining.tfstate ../steps/joining`

This also removes all leader election logic.

TODO:
- To avoid skew between the bootstrap node and ncg nodes, currently the bootstrap node gets its s3 ignition config (identical to NCG config) through an http endpoint. This needs to be addressed before merging.
- To stop NCG config manifests to be deployed to other platforms (doing it for convenience at the minute)
- Each step before 4 bootstrap and 5 joining are almost cloud agnostic and should be moved from terraform to best suited specific tools
- Once https://github.com/coreos/tectonic-installer/pull/2806 lands it will need to be added support for this multistep workflow

PR #2806, opened 2018-01-23
Cli tool

alexsomesan:
Same as original, but correctly rebased.


BUILD.bazel
```diff
@@ -44,7 +44,29 @@ genrule(
     cmd = "cp $(<) $(@)",
 )
 
-exports_files(["config.tf"])
+PLATFORMS = ["aws", "azure", "gcp", "govcloud", "metal", "openstack/neutron", "vmware"]
```
squat: I think something funny happened with the rebase: these definitions are out of date with ut2-integration. We no longer define `PLATFORMS` in BUILD.bazel

BUILD.bazel
```diff
@@ -44,7 +44,29 @@ genrule(
...
+
+[genrule(
```
squat: Same here, these rules are no defined in `examples/BUILD.bazel` and `Documentation/BUILD.bazel`

installer/cmd/tectonic/BUILD.bazel
```diff
@@ -0,0 +1,22 @@
...
+    importpath = "github.com/coreos/tectonic-installer/installer/cmd/tectonic",
+    visibility = ["//visibility:private"],
+    deps = [
+        "//installer/pkg/workflow:go_default_library",
+        "//installer/pkg/tectonic:go_default_library",
+        "//installer/vendor/gopkg.in/alecthomas/kingpin.v2:go_default_library",
+    ],
+)
+
+go_binary(
```
squat: we should add explicit rules for darwin and linux. can be done in a followup but it is a requirement before the task is done.

installer/cmd/tectonic/main.go
```diff
@@ -0,0 +1,52 @@
...
+				},
+			)
+			w.Execute()
+		}
+	case clusterDeleteCommand.FullCommand():
+		{
+			buildPath := *deleteClusterDir
+			pathStat, err := os.Stat(buildPath)
+			// TODO: add deeper checking of the path for having cluster state
+			if os.IsNotExist(err) || !pathStat.IsDir() {
```
squat: what about any other syscall-related errors? is this error actually safe to ignore?

installer/pkg/tectonic/buildstate.go
```diff
@@ -0,0 +1,52 @@
...
+	vf, err := os.Open(varfile)
+	if err != nil {
+		log.Print(err)
+		return "", err
+	}
+	defer vf.Close()
+	scanner := bufio.NewScanner(vf)
+	for scanner.Scan() {
+		line := scanner.Text()
+		if strings.HasPrefix(line, "tectonic_cluster_name") {
```
squat: `tectonic_cluster_name` should be made into a constant

installer/pkg/tectonic/buildstate.go
```diff
@@ -0,0 +1,52 @@
...
+			value := strings.Split(line, "=")[1]
```
squat: currently this method only works with .tfvars and not .tfvars.json, right? maybe add a note saying that

installer/pkg/tectonic/buildstate.go
```diff
@@ -0,0 +1,52 @@
...
+			value = strings.TrimSpace(value)
+			value = strings.Trim(value, "\"")
```
squat: This is a little bit cleaner and easier to read if you use a rune, e.g. 
```go
value = strings.Trim(value, `"`)
```

installer/pkg/tectonic/buildstate.go
```diff
@@ -0,0 +1,52 @@
...
+			return value, nil
+		}
+	}
+	return "", errors.New("not found")
```
squat: This case only occurs if the given file does not include a `tectonic_cluster_name`, right? I think the error should reflect that, e.g. 
```go
fmt.Errorf("the given file does not specify a %q", TectonicClusterNameKey)
```

installer/pkg/tectonic/buildstate.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func NewBuildLocation(clusterName string) string {
+	pwd, _ := os.Getwd()
```
squat: why is this error suppressed?

installer/pkg/tectonic/buildstate.go
```diff
@@ -0,0 +1,52 @@
...
+	err := os.MkdirAll(buildPath, os.ModeDir|0755)
+	if err != nil {
+		log.Fatalf("Failed to create build folder at %s", buildPath)
+	}
+	return buildPath
+}
+
+// implement actual detection of templates
+func FindTemplatesForType(buildType string) string {
+	pwd, _ := os.Getwd()
```
squat: same here

installer/pkg/workflow/install.go
```diff
@@ -0,0 +1,86 @@
...
+			terraformPrepareStep{},
+			terraformInitStep{},
+			terraformApplyStep{},
+		},
+	}
+}
+
+type terraformPrepareStep struct{}
+
+func (s terraformPrepareStep) Execute(m *Metadata) error {
```
squat: An unexported type should not have exported methods

installer/pkg/workflow/destroy.go
```diff
@@ -0,0 +1,40 @@
...
+			terraformInitStep{},
+			terraformDestroyStep{},
+		},
+	}
+}
+
+type terraformDestroyStep struct{}
+
+func (s terraformDestroyStep) Execute(m *Metadata) error {
+	bp := (m.GetValue("build_path")).(string)
```
squat: `build_path` should probably be a const as well; maybe a type imported from another pkg

installer/pkg/workflow/workflow.go
```diff
@@ -0,0 +1,42 @@
...
+
+import "log"
+
+// Workflow is a high-level representation
+// of a set of actions performed in a predictable order
+type Workflow interface {
+	Execute() error
+}
+
+type Metadata map[string]interface{}
```
squat: likewise, these types are implementation details of the `Workflowtype` type and should not be exported.

installer/pkg/workflow/workflow.go
```diff
@@ -0,0 +1,42 @@
...
+
+type Step interface {
```
squat: Rather than model `Step` as an interface, why not model it as a function type? This way defining the list of steps in install.go would be easier and you wouldn't need to define throwaway structs just to define methods.

installer/pkg/workflow/workflow.go
```diff
@@ -0,0 +1,42 @@
...
+	Execute(*Metadata) error
+}
+
+// Context represents the state of a workflow
+type WorkflowType struct {
```
squat: This type should strictly be unexported, otherwise there is no real point to exporting the interface. If the `Workflow` interface is the API for all other packages, then this type should be internal, i.e. an implementation that is of no concern to users.

squat: Please rename this type; it is odd in Go to suffix any type with `Type` just as we do not suffix interfaces with `IFace` or the like. If this is truly going to be one of several implementations then I suggest that this type be named after that implementation. On the other hand, if this type is going to be the only implementation, then there should be no interface at all and this should be the only exported type from this package. In this case this type should be named simply `Workflow`.

installer/pkg/workflow/workflow.go
```diff
@@ -0,0 +1,42 @@
...
+	metadata Metadata
+	steps    []Step
+}
+
+func (w WorkflowType) Execute() error {
+	var stepError error
```
squat: why not just name this `err`?

installer/pkg/workflow/workflow.go
```diff
@@ -0,0 +1,42 @@
...
+	for _, oneStep := range w.steps {
```
squat: it's convention in Go to name the range value either  the singular of the splice, i.e. `step`, or the first letter, i.e. `s`.

installer/pkg/tectonic/buildstate.go
alexsomesan: This whole method is only meant to be around until we merge Daniel's config object parser. Then we scrap this and get the name from the config object.
I agree though, technically "tectonic_cluster_name" should be a constant. I also expect the compiler would optimize towards that, no?

installer/pkg/tectonic/buildstate.go
alexsomesan: Same as above. This is just mock-up code - Daniel's config parser will be the long term solution.

installer/pkg/tectonic/buildstate.go
alexsomesan: Good point. Forgot about that form. Will fix

installer/pkg/tectonic/buildstate.go
squat: Im suggesting it from code maintainability perspective.

installer/pkg/tectonic/buildstate.go
alexsomesan: Crappy mock-up code. Meant to be dropped entirely when config object parsing gets in.

installer/pkg/tectonic/buildstate.go
alexsomesan: This is not the actual implementation we want to have here.
Only meant as a minimally working surrogate until we implement a proper way to find the templates within the release bundle.

installer/pkg/tectonic/buildstate.go
alexsomesan: Out of haste. Should be handled. I'll make the change.

installer/pkg/workflow/workflow.go
alexsomesan: Old habit from other more talkative languages. I'll make it 'err'.

installer/pkg/workflow/workflow.go
alexsomesan: I had the same dilema myself.
The reality is we (or I) don't know for sure if this is going to suffice as a sole implementation or not in the long run. That's why the interface is there and it's being referenced via the interface. I think this is the safer design choice at this point.

How about we call this implementation SimpleWorkflow for now? I'm not great with picking good names..

squat: sure call it something like `basicWorkflow`/`simpleWorkflow` but the type should really be unexported

installer/pkg/workflow/workflow.go
alexsomesan: Simple steps appear as overkill with this approach I know, but I can imagine more complex ones would be split across multiple functions that would want to share some state. Even then we could explicitly pass around all necessary state as arguments (functional style) and still use a function type as entrypoint.
I tend to think this form is easier to reason with for most people so I'm tempted to keep it. If you feel strongly the other way, we can switch though.
Either way, empty structs should get optimised away by the compiler (unless they are being very "original") so they should not bloat in any way.

installer/pkg/workflow/workflow.go
alexsomesan: Agree. Making it unexported.

installer/pkg/workflow/install.go
alexsomesan: That's right. Will lower-case it.

installer/pkg/workflow/destroy.go
alexsomesan: I'm actually thinking to make build path a first-class attribute of metadata.
It's always going to be set and will be accessed from pretty much everywhere.
Same applies for other "core" attributes we'll run across as we develop this.
WDYT?

installer/pkg/workflow/workflow.go
alexsomesan: Will change them.

installer/pkg/workflow/workflow.go
squat: Agreed on the compiler angle. However, in order to avoid premature optimization for a potential future complexity, I would have gone with the function style first. It's possible we may need to make the change in the future, though it's also possible that if we go down the current route, we will never need the structs to carry any state and we'll have struct steps simply because we thought we'd need the complexity. Not a big problem either way. Again, not concerned about the binary performance but more about the legibility and maintainability

alexsomesan:
@squat Can you have a second look?
Made the changes you suggested (minus the ones to parts soon to be made obsolete by @spangenberg ).

installer/pkg/workflow/destroy.go
squat: My instinct is that `metadata` should be a config struct inside the Workflow. The struct should only define the fields we know it needs now. If we need arbitrary keys and values in the future then we can move to `map[string]interface{}` but it's a little much right now, especially for type that only has one known field ATM. Plus it would be good to avoid having to do type assertions down the road

squat:
Looks like there may be some unnecessary commit in there. Can we rebase and drop dee61c50d4faa3ac09d8a3b59b0b0d647aac57b3? that was never merged into any branch.

installer/cmd/tectonic/main.go
```diff
@@ -0,0 +1,29 @@
+package main
+
+import (
+	"github.com/coreos/tectonic-installer/installer/pkg/workflow"
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+var (
+	dryRunFlag            = kingpin.Flag("dry-run", "Just pretend, but don't do anything.").Bool()
+	clusterInstallCommand = kingpin.Command("install", "Create a new Tectonic cluster.")
```
squat: nit: the descriptions should all end without a period or all end with a period. I prefer no period but either way is ok as long as it is consistent

installer/cmd/tectonic/main.go
```diff
@@ -0,0 +1,29 @@
...
+		{
+			w := workflow.NewInstallWorkflow(*clusterConfigFlag)
+			w.Execute()
+		}
+	case clusterDeleteCommand.FullCommand():
+		{
+			w := workflow.NewDestroyWorkflow(*deleteClusterDir)
+			w.Execute()
+		}
+	}
```
squat: can we have a default case that shows the help?

installer/pkg/workflow/workflow.go
```diff
@@ -0,0 +1,42 @@
...
+	for _, step := range w.steps {
+		err = step.execute(&w.metadata)
+		if err != nil {
+			log.Fatal(err) // TODO: actually do proper error handling
+		}
+	}
+	return nil
+}
+
+func (m *metadata) SetValue(key string, value interface{}) error {
```
squat: These methods should be unexported as well. Also, why are we wrapping the map with setters and getters?

installer/cmd/tectonic/main.go
alexsomesan: good point, I'll drop the periods.

installer/cmd/tectonic/main.go
alexsomesan: Not sure I understand this right, but the Kingpin library actually produces automatic help based on those descriptions strings of each command. Here's how it looks today:
```
bazel-bin/installer/cmd/tectonic/darwin_amd64_stripped/tectonic --help
usage: tectonic [<flags>] <command> [<args> ...]

Flags:
  --help     Show context-sensitive help (also try --help-long and --help-man).
  --dry-run  Just pretend, but don't do anything.

Commands:
  help [<command>...]
    Show help.

  install --config=CONFIG
    Create a new Tectonic cluster.

  delete [<dir>]
    Delete an existing Tectonic cluster.
```

We could obviously add additional information to it, but I presume this is enough to get started.

installer/pkg/workflow/workflow.go
alexsomesan: Two reasons I had in mind when I did this.

The practical one is avoid the ugly syntax of dereferencing the m pointer every time since this is going to be done in a lot of places. May be a personal bias though.

The other one is a bit more on the hypothetical side: make sure we don't have to sweep through the whole code if we want to swap the map out for something else.

You think we should switch to accessing the map directly?

installer/pkg/tectonic/buildstate.go
alexsomesan: You've got a valid point. But this method is disposable and should only be in here until Daniel gets to merge his stuff in. I'll add the constant thought.

installer/cmd/tectonic/main.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/coreos/tectonic-installer/installer/pkg/workflow"
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+var (
+	dryRunFlag            = kingpin.Flag("dry-run", "Just pretend, but don't do anything.").Bool()
+	clusterInstallCommand = kingpin.Command("install", "Create a new Tectonic cluster.")
+	clusterDeleteCommand  = kingpin.Command("delete", "Delete an existing Tectonic cluster.")
+	deleteClusterDir      = clusterDeleteCommand.Arg("dir", "The name of the cluster to delete").String()
+	clusterConfigFlag     = clusterInstallCommand.Flag("config", "Cluster specification file").Required().ExistingFile()
```
cpanato: add a `.` at the end of the description. like the other 😄

installer/cmd/tectonic/main.go
cpanato: what happen if just execute `tectonic`? it will show the help as well?

cpanato: 👍

enxebre:
Just for having an starting point to discuss once this gets in, it will need to be integrated with https://github.com/coreos/tectonic-installer/pull/2802 so new organic abstractions will arise and in particular for Cloud specific step logic.
Also an option to consider once this gets mature would be rather than encapsulate into hidden step struct functions, could use nested user facing commands, so if required the user could `tectonic-cli create-cluster igntion`, this should be pretty straight forward if we do the right abstractions

installer/pkg/tectonic/buildstate.go
```diff
@@ -0,0 +1,40 @@
...
+func ClusterNameFromConfig(varfile string) (string, error) {
+	// TODO @spangenberg: implement this based on parsed config object.
+	return "", errors.New("not found")
+}
+
+func NewBuildLocation(clusterName string) string {
+	var err error
+	pwd, err := os.Getwd()
+	if err != nil {
+		log.Fatalf("Failed to current directory because: %v", err)
```
sym3tri: missing a word in this sentence "failed to ____ current directory..."

installer/pkg/workflow/workflow.go
```diff
@@ -0,0 +1,34 @@
...
+
+type metadata struct {
+	statePath   string
+	clusterName string
+	configFile  string
+}
+
+type Step func(*metadata) error
+
+// Context represents the state of a workflow
```
squat: What “Context”?

installer/pkg/workflow/workflow.go
```diff
@@ -0,0 +1,34 @@
...
+type SimpleWorkflow struct {
```
squat: This type should be unexported. The exports type is the interface. No one outside of the workflow package should care about or depend on this type at all

installer/pkg/workflow/install.go
```diff
@@ -0,0 +1,83 @@
...
+		steps: []Step{
+			terraformPrepareStep,
+			terraformInitStep,
+			terraformApplyStep,
+		},
+	}
+}
+
+func terraformPrepareStep(m *metadata) error {
+	if m.statePath == "" {
```
squat: It seems odd to me that the steps mutate the metadata. This should happen in the install workflow constructor IMO

installer/pkg/workflow/install.go
```diff
@@ -0,0 +1,83 @@
...
+	err := tfInit.Run()
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func terraformApplyStep(m *metadata) error {
+	log.Printf("Installation is running...")
+	tfInit := exec.Command("terraform", "apply", tectonic.FindTemplatesForType("aws")) // TODO: get from cluster config
```
enxebre: tfInit sounds like a carry over name from previews ```terraformInitStep```
in a follow up PR we can probably
```func runTfCommand(buildPath string, args ...string) error {
	tfCommand := exec.Command("terraform", args...)
	tfCommand.Dir = buildPath
	tfCommand.Stdin = os.Stdin
	tfCommand.Stdout = os.Stdout
	tfCommand.Stderr = os.Stderr
	err := tfCommand.Run()
	if err != nil {
		return err
	}
	return nil
}
```

installer/pkg/workflow/install.go
alexsomesan: In my mind that was the whole point of passing the metadata object around.
It was supposed to provide a vehicle for passing data between steps, without coupling the steps to each other.

installer/pkg/workflow/workflow.go
alexsomesan: leftover. needs fixing.

installer/pkg/workflow/install.go
squat: If you want to decouple the steps then the metadata must be an invariant that is completely filled during construction. This implementation is the opposite. IMO. All steps that require the statePath metadata are now dependent on this step.

installer/pkg/workflow/install.go
alexsomesan: I deliberately chose to stay away from abstracting this just yet, because I wanted to get a feel of what the interface to the TF runner abstraction would need to look like once we have a few workflows in place.

I'll incorporate your suggestion and we'll watch how it maps to some real workflows and evolve it from there if needed.

squat:
Go Lint is failing due to improper commenting on a few funcs:
NewDestroyWorkflow
NewInstallWorkflow
ClusterNameFromConfig
FindTemplatesForType

othewise looks good for now

PR #2810, opened 2018-01-24
tests: Add bazel build tarball to Jenkinsfile

mxinden:


mxinden:
@cpanato @squat @alexsomesan This PR adds the `bazel build tarball` step to the Jenkinsfile. For now the purpose of this is not to use the output in any way, but instead make sure we are not hitting any regressions on master. I will build on top of this in the next couple of days.

This comes with the cost of 4 minutes added to the overall time of the test pipeline. I think this sounds reasonable. What do you think?

images/tectonic-builder/Dockerfile
```diff
@@ -1,111 +1,13 @@
...
-# docker build \
-# --build-arg TERRAFORM_URL=<upstream terraform download url> \
-# -t quay.io/coreos/tectonic-builder:<next-semver>-upstream-terraform \
-# -f images/builder/Dockerfile .
-# docker push quay.io/coreos/tectonic-builder:<next-semver>-upstream-terraform
-###
+FROM openjdk:8
 
-FROM golang:1.9.1-stretch
+RUN echo "deb [arch=amd64] http://storage.googleapis.com/bazel-apt stable jdk1.8" | \
```
squat: We should really pin the version of Bazel. Specifically, 0.9.x is known to have issues when crossconpiling go binaries.

images/tectonic-builder/Dockerfile.bazel
```diff
@@ -0,0 +1,13 @@
+FROM openjdk:8
+
+RUN echo "deb [arch=amd64] http://storage.googleapis.com/bazel-apt testing jdk1.8" | \
+    tee /etc/apt/sources.list.d/bazel.list && \
+    curl https://bazel.build/bazel-release.pub.gpg | apt-key add -
+
+RUN apt-get update \
+    && apt-get -y install bazel=0.10.0~rc6 \
```
mxinden: @squat as discussed this pins the Bazel version to `0.10.0~rc6` from the deb repository. Good hint, thanks! Let me know if there is anything else.

installer/frontend/BUILD.bazel
```diff
@@ -34,10 +37,13 @@ genrule(
...
-    cmd = "cp -RL installer tmp && cp -R $(GENDIR)/installer tmp && OUT=$$(pwd)/$@ $(location @yarn//:yarn) --cwd $$(pwd)/tmp/installer/frontend run build",
-    tools = ["@yarn//:yarn"],
+    cmd = "PATH=$$(pwd)/$$(dirname $(location @nodejs//:node)):$$PATH"
+          + " && cp -RL installer tmp"
+          + " && cp -R $(GENDIR)/installer tmp"
+          + " && OUT=$$(pwd)/$@ $(location @yarn//:yarn) --cwd $$(pwd)/tmp/installer/frontend run build",
+    tools = ["@yarn//:yarn", "@nodejs//:node"],
 )
 
 genrule(
```
squat: @mxinden could you add nodejs to PATH for these two frontend tests as well? that way all of the frontend building is consistent and hermetic?

mxinden: @squat Done. In addition I added the `frontend_test` target to the Jenkinsfile to run after every build.

mxinden:
ok to test

PR #2813, opened 2018-01-24
modules: fix ca-certificates mount src

squat:
This PR changes the ca-certificates mounts for:

* k8s-node-bootstrap.service; and
* tectonic-channel-operator.yaml

Rather than rely on /usr/share/ca-certificates/ca-certificates.crt they
should be using /etc/ssl/certs/ca-certificates.crt since this will
actually include admin-installed certs.

Fixes: INST-886

cc @brianredbeard @coresolve @alexsomesan


brianredbeard:
The paths look fine. Since we're always adding the Tectonic CA certificates
this should always be the correct path, correct? @squat

PR #2814, opened 2018-01-24
modules: fix ca-certificates mount src

squat:
This PR changes the ca-certificates mounts for:

* k8s-node-bootstrap.service; and
* tectonic-channel-operator.yaml

Rather than rely on /usr/share/ca-certificates/ca-certificates.crt they
should be using /etc/ssl/certs/ca-certificates.crt since this will
actually include admin-installed certs.

Fixes: INST-886

cc @brianredbeard @coresolve @alexsomesan

### DIFF
--- a/modules/bootkube/resources/bootstrap-manifests/bootstrap-apiserver.yaml
+++ b/modules/bootkube/resources/bootstrap-manifests/bootstrap-apiserver.yaml
@@ -59,7 +59,7 @@ spec:
       path: /etc/kubernetes/cloud
   - name: ssl-certs-host
     hostPath:
-      path: /usr/share/ca-certificates
+      path: /etc/ssl/certs
   - name: var-lock
     hostPath:
       path: /var/lock
--- a/modules/bootkube/resources/bootstrap-manifests/bootstrap-controller-manager.yaml
+++ b/modules/bootkube/resources/bootstrap-manifests/bootstrap-controller-manager.yaml
@@ -33,4 +33,4 @@ spec:
       path: /etc/kubernetes
   - name: ssl-host
     hostPath:
-      path: /usr/share/ca-certificates
+      path: /etc/ssl/certs
--- a/modules/bootkube/resources/manifests/kube-apiserver.yaml
+++ b/modules/bootkube/resources/manifests/kube-apiserver.yaml
@@ -92,7 +92,7 @@ spec:
       volumes:
       - name: ssl-certs-host
         hostPath:
-          path: /usr/share/ca-certificates
+          path: /etc/ssl/certs
       - name: secrets
         secret:
           secretName: kube-apiserver
--- a/modules/bootkube/resources/manifests/kube-controller-manager.yaml
+++ b/modules/bootkube/resources/manifests/kube-controller-manager.yaml
@@ -83,5 +83,5 @@ spec:
           secretName: kube-cloud-cfg
       - name: ssl-host
         hostPath:
-          path: /usr/share/ca-certificates
+          path: /etc/ssl/certs
       dnsPolicy: Default # Don't use cluster DNS.
--- a/modules/bootkube/resources/manifests/kube-proxy.yaml
+++ b/modules/bootkube/resources/manifests/kube-proxy.yaml
@@ -53,7 +53,7 @@ spec:
         effect: "NoSchedule"
       volumes:
       - hostPath:
-          path: /usr/share/ca-certificates
+          path: /etc/ssl/certs
         name: ssl-certs-host
       - name: etc-kubernetes
         hostPath:
--- a/modules/ignition/resources/services/k8s-node-bootstrap.service
+++ b/modules/ignition/resources/services/k8s-node-bootstrap.service
@@ -1,6 +1,8 @@
 [Unit]
 Description=Bootstrap Kubernetes Node Components
 ConditionPathExists=!/etc/kubernetes/kubelet.env
+Wants=update-ca-certificates.service
+After=update-ca-certificates.service
 Before=kubelet.service
 
 [Service]
@@ -15,7 +17,7 @@ ExecStartPre=/usr/bin/docker run --rm \
             --tmpfs /tmp \
             -v /usr/share:/usr/share:ro \
             -v /usr/lib/os-release:/usr/lib/os-release:ro \
-            -v /usr/share/ca-certificates/ca-certificates.crt:/etc/ssl/certs/ca-certificates.crt:ro \
+            -v /etc/ssl/certs/ca-certificates.crt:/etc/ssl/certs/ca-certificates.crt:ro \
             -v /var/lib/torcx:/var/lib/torcx \
             -v /var/run/dbus:/var/run/dbus \
             -v /run/metadata:/run/metadata:ro \
--- a/modules/tectonic/resources/manifests/console/deployment.yaml
+++ b/modules/tectonic/resources/manifests/console/deployment.yaml
@@ -140,9 +140,6 @@ spec:
         - mountPath: /etc/ssl/certs
           name: ssl-certs-host
           readOnly: true
-        - mountPath: /usr/share/ca-certificates
-          name: ca-certs-host
-          readOnly: true
         - mountPath: /etc/tectonic/licenses
           name: tectonic-license-secret
           readOnly: true
@@ -164,9 +161,6 @@ spec:
       - hostPath:
           path: /etc/ssl/certs
         name: ssl-certs-host
-      - hostPath:
-          path: /usr/share/ca-certificates
-        name: ca-certs-host
       - name: tectonic-license-secret
         secret:
           secretName: tectonic-license-secret
--- a/modules/tectonic/resources/manifests/updater/operators/tectonic-channel-operator.yaml
+++ b/modules/tectonic/resources/manifests/updater/operators/tectonic-channel-operator.yaml
@@ -51,4 +51,4 @@ spec:
       volumes:
       - name: certs
         hostPath:
-          path: /usr/share/ca-certificates
+          path: /etc/ssl/certs


brianredbeard:
Looking at
https://github.com/squat/tectonic-installer/blob/b2bc1fc33a0ab549f43af0d5a9a67f0095f9ac9c/modules/ignition/resources/services/k8s-node-bootstrap.service#L5

Is there anything else in place to ensure the rebuild of the cert chain is
happening early enough (E.g before docker.service)?

squat:
@brianredbeard good question. I don't think update-ca-certs needs to occur any earlier than kubelet bootrstrapping since this is gating k8s and all docker workloads (besides those run by some seated operater). Am I missing anything? We could modify the update-ca-certs to always finish before docker.service just in case

squat:
ok to test

brianredbeard:
FYI, @squat / @robszumski / @mike-saparov this change set failed in an attempt with the nightly build system:

https://jenkins-tectonic-installer.prod.coreos.systems/job/tectonic-release-automation/job/build-tectonic/job/master/120/console

squat:
@brianredbeard the change is unrelated to this change set but point to some go/vendoring issue. Cc @cpanato

brianredbeard:
Keep in mind, that in my testing I was doing this from first principals with a fresh CoreOS image without the use of Terraform.  This info is provided as a reference for what had to happen from a fresh CoreOS image... I can't speak to the workflow of what's happening under the hood with the Tectonic installer.

To begin with, I started with the need to affect the storage of the host in the following ways:
  - Add a CA certificate to the host, which I desired to be at the path `/etc/ssl/certs/coreos-ca.pem`
  - Create a symbolic link from the known hash of that CA certificate (`296964d6`) to the new CA, performing the OpenSSL anchoring process

These two steps began simply with adding the following stanza to a [Container Linux Config](https://coreos.com/os/docs/latest/configuration.html) (CLC) inside of a [Matchbox](https://github.com/coreos/matchbox) template:

```
storage:
  files:
    - path: /etc/ssl/certs/coreos-proxy.crt
      filesystem: root
      mode: 0644
      contents:
        inline: |
          -----BEGIN CERTIFICATE-----
          MIIFNjCCAx6gAwIBAgIEeDrWpTANBgkqhkiG9w0BAQsFADA0MTIwMAYDVQQDDClD
          b3JlT1MgUHJveHkgU2VydmVyIENlcnRpZmljYXRlIEF1dGhvcml0eTAeFw0xNzEy
          MTkyMjM1NThaFw0xOTEyMTkyMjM1NThaMDQxMjAwBgNVBAMMKUNvcmVPUyBQcm94
          eSBTZXJ2ZXIgQ2VydGlmaWNhdGUgQXV0aG9yaXR5MIICIjANBgkqhkiG9w0BAQEF
          AAOCAg8AMIICCgKCAgEAvcSghpzJlI6RNGKJnFPt0MAgzKtTWp+xItvQaZMMiDSZ
          OSRl+RqdBEXARQVW7Y2ydPEOUDCkMUSbgqPWm1TwJ1JsjnOK1Bnow3ezlgp9cAh1
          Ov4CC/4uDiSuTSg5QBy3DFbIYai7wMMXMAT4AnVLKJr57u9Z2R7H2GRxXVqX5UC/
          4/fAvOf2DgJUUqpVjZThj9uueEYfAhe30SgXWQfgvhx61GAaRijX4BTfOkUj3zQ5
          ZozyRx3HO0ZwitBjcEmqNfHrUTUXKAyARmFjbowjbRYHfaBAZjFzi0I4j+sb19lJ
          c/4tPZbWHRjucu2rBowZn4WlAz21QquhkJHJqt/y522WT0nonltHHQGwJ6PJzZIn
          uTnlYNGlO/eognNnvXhpbJQst37cxEjx5rwfzDRjAzP82B1/N7xfAnJOk3s8W0rA
          Rx6IsDh0njUsZOD/stul1hFQXVB+GCS77m9qS4FnpxDjzePIcz2rqR57/FfWYZRr
          dSP7AYXp7IPcohTXw1G9RgWW4HPul+lNxy4+FnjDRdhCeIRxdSdfZ27WcA3GjBW8
          4OPisL0kiwuDSdCjJgXSEhrfu8hJe2Y/ATQ7Nv91Q9qd63YeUQnBhoMcnWhjPtho
          NtlXCbhqydj13IrTSjWYk7NsC99nNTmQkkyAJ2Lknmg/Y7/tupFuwh22WjRGJcsC
          AwEAAaNQME4wHQYDVR0OBBYEFLOBvT98kNAweaC+1b2VhqHw2PI3MB8GA1UdIwQY
          MBaAFLOBvT98kNAweaC+1b2VhqHw2PI3MAwGA1UdEwQFMAMBAf8wDQYJKoZIhvcN
          AQELBQADggIBABHutaXamhEWI9WEALTd5uLL1IRj3pmckiubuFLfclJocRL4zKdH
          BYad8vf7E0lRH7OSRU87e4qjw8S4Mp0kBXDeQW295rnPl4xqTn9cfD66yPhJ2qMd
          oT7LmX+eVCeXELDeuw+lzkTjeWgbvAUBb0BVk/jF4qgxuAet7f4ayY87BABr/LVe
          DB+EJzn0KUuaCauriL7O+3DZxYi+T+4kyOwDvePdfxureG9dz31uPye0kc4P6VAT
          Au9O9Gz3yUOKR0mMFcCyCxcBpIkCLrVVT+nfdftpJX56dWbkrJsCTEu48HGaQZuv
          f+xgqTBJY6SS0pgJMnnngNAtHKXXOOMmzWhjhRj/br5T9Y8E33dMkirD9uaalqPL
          4TBufQSTze1trG8efPSRV0QQoFF2JD/4JqJrxc3PSBInDA4rKb+az/GR4674vx3O
          xz3M8OtNcPy3Tn833HL/w3BXuCi83vBxAczK7N5shMuD55uQR0lEMr9deftzvY8S
          2EuYIHBtnEifsw21+r682T9kOiqanx8NVt8mtvXgm28W28jtiqTlqZDKmkYwMvsp
          SSA3pTsZeKDY62sXSLrSTADXa+UWN324qsNO9rgmKZu/TqqR9CpSBv/7we77hzop
          v9VnekZLcfxG7j6qqTu5XBr5rsTI7Mo4d8+8mAQKmRYOSibaCkdzYVSo
          -----END CERTIFICATE-----

  links:
    - path: /etc/ssl/certs/296964d6.0
      filesystem: root
      target: /etc/ssl/certs/coreos-proxy.crt
```

This successfully laid down the certificate on the host but did not affect the running state of any of the Go based binaries.  At first, I thought that this was due to the ordering of the CA certificate generation services on the host, as Go binaries only read the [canonical cert location](https://golang.org/src/crypto/x509/root_linux.go) on the host at application startup.

After digging in further, the first issue that I found was with the naming of the certificate.  

The script [coreos/coreos-overlay@app-misc/ca-certificates/files/update-ca-certificates](https://github.com/coreos/coreos-overlay/blob/2a31d771a485afb52fb6019b1a4e7dd5b98bfe29/app-misc/ca-certificates/files/update-ca-certificates#L26) will attempt to perform the symbolic linking step on behalf of the user, but only if the flag `--skip-rehash` is *NOT* passed (this is not the default behavior though of the _systemd_ service `update-ca-certificates.service`).  Exploring further, the script `[c_rehash](https://gitweb.gentoo.org/repo/gentoo.git/tree/app-misc/c_rehash)` _only_ works on files ending in `.pem`, thus even if the systemd service had run, the symbolic links would not be created.

OpenSSL will happily reference the files as long as the filename matches the form `${HASH}.${INDEX}` (where `${HASH}` conforms to the value output from `openssl x509 -noout -hash` for the given certificate and `${INDEX}` is an integer (for use when there are multiple files which hash to the same value). 

I worked around the above notes by changing the above snippet to use the `.pem` extension mentioned and once again ensured that the symbolic link was created.

This, alas, did not resolve the issue.

The next challenge found was related to the fact that the service will also _only_ run if the path `/etc/ssl/certs/ca-certificates.crt` is *NOT* a symbolic link (as is the default behavior) due to these stanzas in the systemd unit:

```
ConditionPathIsReadWrite=/etc/ssl/certs
# Do nothing if update-ca-certificates has never been run before
ConditionPathIsSymbolicLink=!/etc/ssl/certs/ca-certificates.crt
```

Thus, I needed a mechanism to break the symbolic link (or force running the unit).  I did not want to run this script every time the host booted (as an administrator, if I haven't changed the certs through some expected process, I don't want more roots of trust magically supported).

At this point, I added the following CLC snippet:

```
storage:
  files:
    - path: /etc/ssl/certs/ca-certificates.crt
      filesystem: root
      mode: 0755
      contents:

```

This replaces the symbolic link with a zero byte file.

Alas,  this _still_ did not resolve the issue as now my directory was older than the file `/etc/ssl/certs/ca-certificates.crt` which fails the conditional expression [located here](https://github.com/coreos/coreos-overlay/blob/2a31d771a485afb52fb6019b1a4e7dd5b98bfe29/app-misc/ca-certificates/files/update-ca-certificates#L29):

```
if [[ ! -e "${CERTBUNDLE}" || "${CERTSDIR}" -nt "${CERTBUNDLE}" ]]; then
```
As the file `${CERTBUNDLE}` exists (albeit with a zero byte size) and the timestamp is newer than that of the directory, the `update-ca-certificates` script fails to see the need for taking any action.

Ultimately this was solved by adding my _own_ execution of the systemd unit to the CLC as follows:

```
systemd:
  units:
    - name: ca-firstboot.service
      enable: true
      contents: |
        [Unit]
        DefaultDependencies=no
        Wants=systemd-tmpfiles-setup.service clean-ca-certificates.service
        After=systemd-tmpfiles-setup.service clean-ca-certificates.service
        Before=sysinit.target
        ConditionPathIsReadWrite=/etc/ssl/certs
        ConditionFirstBoot=True

        [Service]
        Type=oneshot
        ExecStart=/usr/sbin/update-ca-certificates
        [Install]
        WantedBy=multi-user.target
```

At this point because it will execute `c_rehash` on the first boot (but after ignition has laid out the files) it will perform the additional steps as needed.

cc @squat / @sym3tri / @coresolve

cpanato:
retest this please

cpanato:
checking the issue in the RA build system I saw one error in the makefile. will fix and build the repo @brianredbeard

brianredbeard:
FYI, after an initial check it does not seem that this worked (run against `tectonic_b2bc1fc3.zip`):

```
core@node1 ~ $ find /etc/ssl/certs/ -type f
/etc/ssl/certs/custom_ca_certs.pem
/etc/ssl/certs/ca-certificates.crt
```

Looking into this to see what went wrong.

brianredbeard:
@cpanato / @alexsomesan Just so I can spot check my work to ensure I didn't do anything wrong can one of you send me a `tfvars` file used in testing to validate this?  I'm paranoid that there's something goofy I did and want to make sure.

brianredbeard:
Strike that.  I'm confusing #2814 and #2825 .  Let me check again.

brianredbeard:
Confirmed:

```
# /etc/systemd/system/k8s-node-bootstrap.service
[Unit]
Description=Bootstrap Kubernetes Node Components
ConditionPathExists=!/etc/kubernetes/kubelet.env
Wants=update-ca-certificates.service
After=update-ca-certificates.service
Before=kubelet.service

[Service]
Type=simple
RemainAfterExit=true
Restart=on-failure
RestartSec=10
TimeoutStartSec=1h
ExecStartPre=/usr/bin/mkdir -p /etc/kubernetes

ExecStartPre=/usr/bin/docker run --rm \
            --tmpfs /tmp \
            -v /usr/share:/usr/share:ro \
            -v /usr/lib/os-release:/usr/lib/os-release:ro \
            -v /etc/ssl/certs/ca-certificates.crt:/etc/ssl/certs/ca-certificates.crt:ro \
            -v /var/lib/torcx:/var/lib/torcx \
            -v /var/run/dbus:/var/run/dbus \
            -v /run/metadata:/run/metadata:ro \
            -v /run/torcx:/run/torcx:ro \
            -v /run/systemd:/run/systemd \
            -v /etc/torcx:/etc/torcx \
            quay.io/coreos/tectonic-torcx:v0.2.0 \
            /tectonic-torcx-bootstrap \
            --upgrade-os=true \
            --torcx-manifest-url="" \
            --torcx-skip-setup=false \
            --verbose=debug
ExecStart=/usr/bin/echo "node components bootstrapped"

[Install]
WantedBy=multi-user.target kubelet.service
```

PR #2816, opened 2018-01-24
tests: not fail the post operations and send slack notifications

cpanato:
fixes: INST-808

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -63,7 +63,6 @@ pipeline {
   agent none
   environment {
     KUBE_CONFORMANCE_IMAGE = 'quay.io/coreos/kube-conformance:v1.8.4_coreos.0'
-    LOGSTASH_BUCKET = 'log-analyzer-tectonic-installer'
   }
   options {
     // Individual steps have stricter timeouts. 360 minutes should be never reached.
@@ -127,6 +126,11 @@ pipeline {
     booleanParam(
       name: 'PLATFORM/BARE_METAL',
       defaultValue: true,
+      description: ''
+    )
+    booleanParam(
+      name: 'NOTIFY_SLACK',
+      defaultValue: false,
       description: ''
     )
   }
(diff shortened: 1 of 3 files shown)


cpanato:
ok to test

Jenkinsfile
```diff
@@ -522,3 +530,16 @@ def reportStatusToGithub(status, context, commitId) {
...
 }
+
+def notifyBuildSlack(String message) {
+  if(message.contains("master") || message.contains("track-1") || message.contains("log-es-error")) {
+    def colorCode = '#FF0000'
+    def subject = "Job '${env.JOB_NAME} [${env.BUILD_NUMBER}]': ${message}"
+    def summary = "${subject} (${env.BUILD_URL})"
+
+    // Send notifications
+    slackSend(color: colorCode, message: summary, channel: "#team-installer")
```
enxebre: This is probably ok to start, but experience tells me that it's going to end up being to noisy and we'll move it to a dedicated channel

cpanato: this will send just when master/track-1 fail as we have today

cpanato:
@enxebre @mxinden can you please take a look? thanks

Jenkinsfile
```diff
@@ -396,19 +395,31 @@ pipeline {
   post {
     always {
       node('worker && ec2') {
-        forcefullyCleanWorkspace()
```
mxinden: What is the reason for removing this line?

Jenkinsfile
```diff
@@ -522,3 +533,16 @@ def reportStatusToGithub(status, context, commitId) {
     """
   }
 }
+
+def notifyBuildSlack(String message) {
+  if(message.contains("master") || message.contains("track-1") || message.contains("log-es-error")) {
```
mxinden: If an `log-es-error` happens on a PR, we still want to be notified, right?

Jenkinsfile
cpanato: forgot to readd, will put it back

Jenkinsfile
cpanato: yes

cpanato:
@enxebre @mxinden can you please take a look? thanks

PR #2819, opened 2018-01-24
ignition: flag to disable proxy env vars in global profile

alexsomesan:
@brianredbeard this is an interim solution to unblock you.
I'm working on getting you an actual white-list type setting that would take an explicit list of sys-d units to enable the proxy for. That will most likely happen tomorrow.

In the mean time, using this pach, if you set `tectonic_proxy_exclusive_units = true` in tfvars it will not generate `/etc/profile.env` with the proxy settings. That should limit TOP access to just systemd units that still receive the DefaultEnv dropin.


squat:
I think we need this to work on bare metal, no? @brianredbeard

brianredbeard:
Correct @squat / @alex.somesan, though for certificates and http_proxy
there's nothing platform specific about those settings.

cpanato:
@alexsomesan `outdated docs (run 'make docs' to fix)`

alexsomesan:
@brianredbeard Correct. They use the same ignition module. I just omitted to pass the parameter into the module in the bare-metal case.
I'm also testing locally for BM now. Will update shortly.

alexsomesan:
retest this please

alexsomesan:
retest this please

alexsomesan:
random failures on smoke test.
re-running

alexsomesan:
retest this please

alexsomesan:
Different random failures. This time we're hitting AWS limits:
```
* aws_vpn_gateway.vpg: 1 error(s) occurred:

* aws_vpn_gateway.vpg: Error creating VPN gateway: VpnGatewayLimitExceeded: The maximum number of virtual private gateways has been reached.
	status code: 400, request id: 378c9881-7517-46da-b073-bd65c2fc326a
```

I expect this change to be safe to merge since each test case has passed at least in one of the test runs, before being affected by flakiness.

alexsomesan:
I've asked @cpanato to dig a little deeper into the failures.

cpanato:
retest this please. aws only

config.tf
```diff
@@ -577,3 +577,13 @@ variable "tectonic_ntp_servers" {
...
 EOF
 }
+
+variable "tectonic_proxy_exclusive_units" {
+  default = false
+
+  description = <<EOF
+(optional) When set to true, it restricts the application of proxy settings to processes started through systemd units.
+Global proxy settings will not be affected in this case.
+When set to false, the proxy settings will apply globally, including to all processes lauched by users.
```
squat: So, when set to `false`, the default, we want `profile.env` to include proxy settings

modules/ignition/assets.tf
```diff
@@ -238,6 +238,7 @@ data "template_file" "profile_env" {
 }
 
 data "ignition_file" "profile_env" {
+  count      = "${var.proxy_exclusive_units}"
```
squat: This does the opposite of the documented behavior: when set to `false`, we DO want profile.env to be created so that ALL processes are proxied, however here, if `proxy_exclusive_units` is `false`, no profile.env will be created.

platforms/metal/matchers.tf
```diff
@@ -113,6 +114,7 @@ module "ignition_workers" {
   no_proxy                = "${var.tectonic_no_proxy}"
   ntp_servers             = "${var.tectonic_ntp_servers}"
   tectonic_vanilla_k8s    = "${var.tectonic_vanilla_k8s}"
+  proxy_exclusive_units   = "${var.tectonic_proxy_exclusive_units}"
```
squat: This variable has no effect on the ignition generated for Matchbox since it does metal does not use the ignition provider, so profile.env will always be created when http_proxy is defined.

squat:
I made some comments to document the current status of this PR. It is broken in a couple places at the moment, namely because the current implementation has no effect on bare-metal since it does not use the same ignition provider as the other platforms but rather directly leverages Terraform templates.

squat:
I will make follow-up commits to fix the issues I identified.

cpanato:
retest this please. azure

platforms/metal/matchers.tf
squat: Fixed in the latest commit

squat:
cc @coresolve @brianredbeard PTAL

squat:
ok to test

squat:
ok to test. Govcloud only

squat:
Govcloud failing due to account limit errors. We need to manually delete some VPCs. Will test tcp now for conpleteness

enxebre:
retest this please

squat:
ok to test. Gcp only

squat:
ok to test

PR #2825, opened 2018-01-25
platforms/metal: unbundle custom certs

squat:
This commit un-bundles the custom CA certificates that are passed to the
Tectonic Installer. This is needed for OpenSSL certificate hashing to
work. Without this, CAs cannot be anchored. This commit adds an
intermediate templating step to the Matchbox profile generation where
custom CA certs are inserted. Because of the fact that the Tectonic
Installer modifies the `update-ca-certificates.service` unit via a
drop-in, certificates are rehashed on every reboot so the custom CA
certs are rehashed automatically and are avaiable on the first boot to
programs compiled against OpenSSL.

Fixes: INST-885

cc @brianredbeard @coresolve @alexsomesan @sym3tri 

Related: https://github.com/coreos/tectonic-installer/pull/2813, https://github.com/coreos/tectonic-installer/pull/2814

### DIFF
--- a/platforms/metal/cl/bootkube-worker.yaml.tmpl
+++ b/platforms/metal/cl/bootkube-worker.yaml.tmpl
@@ -37,16 +37,6 @@ systemd:
 
 storage:
   files:
-    - path: /etc/kubernetes/installer/kubelet.env
-      filesystem: root
-      mode: 0644
-      contents:
-        inline: {{.ign_installer_kubelet_env_json}}
-    - path: /etc/kubernetes/installer/runtime-mappings.yaml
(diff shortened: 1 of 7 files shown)


platforms/metal/cl/bootkube-worker.yaml.tmpl
```diff
@@ -37,16 +37,6 @@ systemd:
 
 storage:
   files:
-    - path: /etc/kubernetes/installer/kubelet.env
```
squat: This file has just been re-arranged so that this entire section matches the corresponding section in the master profile

PR #2833, opened 2018-01-26
Tests: option to run tests for only one config for each platform

cpanato:


cpanato:
retest this please. now with azure and metal :)

cpanato:
did not report back the results. but PR is green

Jenkinsfile
```diff
@@ -92,6 +92,11 @@ pipeline {
       defaultValue: true,
       description: ''
     )
+    booleanParam(
+      name: 'BASIC_CLUSTER_CONFIG_ONLY',
+      defaultValue: true,
```
mxinden: Default set to true would mean, that we have to edit all trigger jobs, right?

cpanato: will set to false

cpanato:
aws failed in the destroy phase because the credentials expired. I will create a PR to fix this issue

PR #2834, opened 2018-01-26
move dns api key to a variable for govcloud

enxebre:
move dns api key to a variable for govcloud https://github.com/coreos/tectonic-installer/pull/2836

PR #2836, opened 2018-01-26
move dns api key to a variable

enxebre:
Move dns api key to a variable for govcloud https://github.com/coreos/tectonic-installer/pull/2834
Remove unnecessary ternaries https://github.com/coreos/tectonic-installer/pull/2838

### DIFF
--- a/platforms/govcloud/main.tf
+++ b/platforms/govcloud/main.tf
@@ -6,11 +6,6 @@ provider "aws" {
 
 data "aws_availability_zones" "azs" {}
 
-locals {
-  tectonic_govcloud_private_endpoints = true
-  tectonic_govcloud_public_endpoints  = false
-}
-
 module "container_linux" {
   source = "../../modules/container_linux"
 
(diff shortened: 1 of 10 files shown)


platforms/govcloud/variables.tf
```diff
@@ -321,5 +303,16 @@ EOF
 }
 
 variable "tectonic_govcloud_dns_server_ip" {
-  type = "string"
+  description = "The resolver api of the dns server."
```
squat: Should this be "The resolver _IP_..."?

platforms/govcloud/main.tf
```diff
@@ -6,6 +6,11 @@ provider "aws" {
 
 data "aws_availability_zones" "azs" {}
 
+locals {
+  tectonic_govcloud_private_endpoints = true
```
squat: @enxebre here you are hardcoding `private_endpoints` to `true` and `public_endpoints` to `false` but you are keeping all of the ternary expressions. Why not just simplify this and eliminate all of the ternarys? It makes it much easier to read. These changes are in VCS so we don't need to keep this around. If you really need to keep it around even if the ternarys do nothing, then please add a comment

PR #2846, opened 2018-01-29
azure: add option to enable boot diagnostics for VMs in azure

cpanato:
in order to collect the boot logs for debug investigation. we need to enable the boot diagnostics on Azure.

fixes: INST-904

next step is to start collecting the boot logs in the test framework. This PR enable this feature in the tests area.


platforms/azure/variables.tf
```diff
@@ -294,3 +294,9 @@ EOF
     usdodeast          = 2
   }
 }
+
+variable "tectonic_azure_enable_boot_diagnostics" {
```
sym3tri: If there are similar settings for other cloud providers it's probably better to have a single top-level variable that each of them is aware of. You can indicate which platforms are supported in the var description.

cpanato: I will make top level. thanks for the review

cpanato: done @sym3tri

cpanato:
retest this please

cpanato:
@squat  / @alexsomesan  please review when you guys have time. thanks

sym3tri:
thanks for making the changes. seems good to me, but i'll let @alexsomesan give final review

justaugustus:
+1

modules/azure/resource-group/resource-group.tf
```diff
@@ -35,10 +39,30 @@ resource "azurerm_resource_group" "tectonic_cluster" {
...
+
 output "name" {
   value = "${var.external_rsg_id == "" ? element(concat(azurerm_resource_group.tectonic_cluster.*.name, list("")), 0) : element(split("/", var.external_rsg_id), 4)}"
 }
 
 output "storage_id" {
   value = "${random_id.storage_id.hex}"
 }
+
+output "storage_blob_endpoint" {
```
squat: we should break apart this module into variables.tf and outputs.tf to match the other modules

cpanato: done

PR #2847, opened 2018-01-29
tests/frontend: fix ui tests. to use the session role in aws

cpanato:
We are using the role of Jenkins and the UI tests start failing. this fix the issue.


cpanato:
ran several times and all passed

PR #2853, opened 2018-01-30
frontend: Fix style of nodes form error

kyoto:
Fix nodes form error (currently just the `IOPS can't be larger than ### (50 IOPS/GiB)`) to use the form error style (`Alert`) instead of the field error style.

Also rename `toKey()` to `toId()` because we use "id" to refer to field IDs everywhere else and rename a couple of variables too.


squat:
Grrr looks good but needs a rebase

kyoto:
ok to test

PR #2854, opened 2018-01-30
tests: add option to use Role in the aws provider and rspec refactor in the IAM part

cpanato:


Documentation/variables/aws.md
```diff
@@ -21,6 +21,7 @@ This document gives an overview of variables used in the AWS platform of the Tec
 | tectonic_aws_external_vpc_id | (optional) ID of an existing VPC to launch nodes into. If unset a new VPC is created.<br><br>Example: `vpc-123456` | string | `` |
 | tectonic_aws_external_worker_subnet_ids | (optional) List of subnet IDs within an existing VPC to deploy worker nodes into. Required to use an existing VPC and the list must match the AZ count.<br><br>Example: `["subnet-111111", "subnet-222222", "subnet-333333"]` | list | `<list>` |
 | tectonic_aws_extra_tags | (optional) Extra AWS tags to be applied to created resources.<br><br>Example: `{ "key" = "value", "foo" = "bar" }` | map | `<map>` |
+| tectonic_aws_iam_role | (optional) Name of IAM role to use to access AWS in order to deploy the Tectonic Cluster. The name is also the full role's ARN.<br><br>Example:  * Role ARN  = arn:aws:iam::123456789012:role/tectonic-installer | string | `` |
```
alexsomesan: Let's call it something that will make it clear it's used during installation and not for any on-cluster stuff.

tests/rspec/lib/aws_cluster.rb
```diff
@@ -14,11 +14,12 @@
...
   def initialize(tfvars_file)
     export_random_region_if_not_defined if Jenkins.environment?
-    AWSIAM.assume_role if ENV.key?('TECTONIC_INSTALLER_ROLE')
     @aws_region = tfvars_file.tectonic_aws_region
+    @role_credentials = nil
+    @role_credentials = AWSIAM.assume_role(@aws_region) if ENV.key?('TECTONIC_INSTALLER_ROLE')
 
     unless ssh_key_defined?
-      ENV['TF_VAR_tectonic_aws_ssh_key'] = AwsSupport.create_aws_key_pairs(@aws_region)
+      ENV['TF_VAR_tectonic_aws_ssh_key'] = AwsSupport.create_aws_key_pairs(@aws_region, @role_credentials)
```
alexsomesan: Why do we now need to pass credentials into these functions?

cpanato: this function will create an aws client and then will create the keys

Documentation/variables/aws.md
cpanato: ok, any ideias? 😄

tests/rspec/lib/aws_cluster.rb
cpanato: this is same we do for other functions that interacts with AWS

cpanato:
@alexsomesan @mxinden please can you guys take a look?

Jenkinsfile
```diff
@@ -30,12 +31,20 @@ commonCreds = [
   ]
 ]
 
+credsUI = commonCreds.collect()
```
mxinden: As far as I can tell, `credsUI` and `creds` are always used together, right? Why not merge them into one?

tests/rspec/spec/aws/vpc_internal_spec.rb
```diff
@@ -11,7 +11,7 @@
 
   before(:all) do
     export_random_region_if_not_defined
-    AWSIAM.assume_role if ENV.key?('TECTONIC_INSTALLER_ROLE')
+    # AWSIAM.assume_role(ENV['TF_VAR_tectonic_aws_region']) if ENV.key?('TECTONIC_INSTALLER_ROLE')
```
mxinden: Can you add a comment here why you commented this line out?

Jenkinsfile
cpanato: no. I can explain to you in person. 
But the credsUI have the aws cred that Jenkins provides the session token for us. then we don't need to assume the role explicitly.

the creds have the aws credentials with no session token and we need to assume that. this is useful in the terraform and in our rpec.

I also don't like this approach to have two creds, but I would like to see this running for a while and see if works, then I can do a followup to we assume tole in all parts of the build.

tests/rspec/spec/aws/vpc_internal_spec.rb
cpanato: we are not using this in this part of the code. I will remove that in a followup.

PR #2855, opened 2018-01-30
tests: add logic to not fail when trying to send the logs to s3

cpanato:
followup of: https://github.com/coreos/tectonic-installer/pull/2816

master will fail now because the env variable is not defined. was my fault. I thought we send logs only in the end. this fixes the issue.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -63,6 +63,7 @@ pipeline {
   agent none
   environment {
     KUBE_CONFORMANCE_IMAGE = 'quay.io/coreos/kube-conformance:v1.8.4_coreos.0'
+    LOGSTASH_BUCKET= "log-analyzer-tectonic-installer"
   }
   options {
     // Individual steps have stricter timeouts. 360 minutes should be never reached.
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -302,6 +302,7 @@ pipeline {
         TF_VAR_tectonic_container_images = "${params.hyperkube_image}"
         TF_VAR_tectonic_kubelet_debug_config = "--minimum-container-ttl-duration=8h --maximum-dead-containers-per-container=9999 --maximum-dead-containers=9999"
         GOOGLE_PROJECT = "tectonic-installer"
+        LOGSTASH_BUCKET= "log-analyzer-tectonic-installer"
```
mxinden: This makes `LOGSTASH_BUCKET` defined twice, right?

cpanato: forgot to remove, doing that

cpanato: done @mxinden

cpanato: already pushed the changes. remove the variable from the post operation and keep that in high level

cpanato:
retest this please. aws only

PR #2857, opened 2018-01-30
frontend: Remove unused clusterKind option

kyoto:
`clusterKind` was passed from the frontend to the backend, but the backend doesn't actually use it.


kyoto:
ok to test

kyoto:
ok to test

PR #2863, opened 2018-01-30
Support multiple clusters per Matchbox

ethernetdan:
The installer currently overrides the Matchbox profile and groups from previously installed clusters. This makes it difficult to use for multiple clusters and simultaneous installs impossible with the same Matchbox instance.

This change prefixes Matchbox profile and group names with the cluster name to support installing multiple clusters with the same Matchbox instance.

/cc @dghubble


ethernetdan:
ok to test

PR #2867, opened 2018-01-31
tests: update console tests

cpanato:


tests/rspec/lib/pages/base_page.rb
```diff
@@ -98,4 +98,12 @@ def enabled?(locator)
   rescue
     retry if (retries += 1) < 10
   end
+
+  def wait_for_load
```
mxinden: Why not use https://github.com/coreos/tectonic-installer/blob/master/tests/rspec/lib/with_retries.rb here?

cpanato: done

cpanato:
@mxinden please review and merge on green if you agree. thanks

PR #2873, opened 2018-01-31
Govcloud

enxebre:
cherry-picks https://github.com/coreos/tectonic-installer/commit/f3aec294a3c5bfac2784d85023b67b68908a1de9 and https://github.com/coreos/tectonic-installer/commit/4b252440737b642fbe78f3673b0c627937a4358e


contrib/govcloud/vars.tf
```diff
@@ -2,48 +2,72 @@
...
 }
 
-variable base_domain {
+variable "instance_type" {
+  description = "The type of the ec2 machine."
+  default     = "t2.micro"
+}
+
+variable "base_domain" {
   default     = "tectonic-ci.de"
```
cpanato: this is correct to keep hardcoded?

enxebre: there's no sensible information there so it's fine, we could remove it in a follow up to keep it less opinionated

cpanato: sound good

PR #2875, opened 2018-02-01
frontend: Change minimum AWS node storage size to 30 GiB

kyoto:
We should enforce a sensible minimum node volume size. I've used 30 GiB because https://coreos.com/tectonic/docs/latest/install/aws/requirements.html states "minimum of 30GB of storage for each node".

Also adds Nightwatch tests for these fields.

cc @sym3tri @robszumski


kyoto:
ok to test

PR #2878, opened 2018-02-01
tests: using CL beta in master branch

cpanato:
fixes: https://jira.coreos.com/browse/INST-911

### DIFF
--- a/tests/rspec/lib/metal_support.rb
+++ b/tests/rspec/lib/metal_support.rb
@@ -46,9 +46,7 @@ module MetalSupport
       FileUtils.cp(cert, dest_folder)
     end
 
-    # Download CoreOS images
-    cl_version = varfile.tectonic_container_linux_version
-    system(env_variables_setup, "#{root}/matchbox/scripts/get-coreos stable #{cl_version} ${ASSETS_DIR}")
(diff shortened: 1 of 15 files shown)


tests/rspec/lib/metal_support.rb
```diff
@@ -89,6 +86,17 @@ def self.start_matchbox(varfile)
     end
   end
 
+  def self.download_cl(varfile)
+    # Download CoreOS images
+    cl_channel = varfile.tectonic_container_linux_channel
+    cl_version = varfile.tectonic_container_linux_version
+    if cl_version.include?('latest')
```
squat: this is tricky. what if there is no version specified in the tf_vars? since the cl_version defaults to `latest` maybe we should account for that here.

cpanato: for metal tests we need to define the version, otherwise, the setup part that downloads the CL will not work.
previous this change we always have the version set in the tfvars.

let me know if you want to change this then I will rethink the logic

@squat

squat: sounds good. I don't think we need to over-engineer this :sunglasses:

PR #2879, opened 2018-02-01
Build smoke

squat:
This PR removes the QA test suite from the smoke test, since all of that was stats-specific and has been moved to https://github.com/coreos/container-images/tree/master/tectonic-stats-extender/test.

This PR also adds a lot of work enabling bazel to build the smoke test binary using:

```sh
bazel build tests/smoke
```

cc @alexsomesan @mxinden


squat:
ok to test

mxinden:
@squat Would you mind adding `bazel build tests/smoke` in the [Jenkinsfile](https://github.com/coreos/tectonic-installer/blob/master/Jenkinsfile#L168)?Just to make sure, we don't hit any regressions.

squat:
@mxinden good call. I just added it to the Jenkinsfile. Please take another look

PR #2884, opened 2018-02-02
[WIP] frontend: Generate YAML cluster config instead of tfvars file

kyoto:
Change `cluster-config.js` to output the new Tectonic cluster config values and change the backend to write out a YAML config file instead of a tfvars file.

Update GUI unit tests and end-to-end tests to test the YAML config. Adds `js-yaml` as a dev dependency for the end-to-end tests.

cc @alexsomesan @spangenberg


kyoto:
ok to test

installer/api/handlers_terraform.go
```diff
@@ -19,14 +20,26 @@ import (
...
-	PullSecret    string                 `json:"pullSecret"`
-	DryRun        bool                   `json:"dryRun"`
-	Retry         bool                   `json:"retry"`
+	Credentials terraform.Credentials `json:"credentials"`
+	DryRun      bool                  `json:"dryRun"`
+	License     string                `json:"license"`
+	PullSecret  string                `json:"pullSecret"`
+	Retry       bool                  `json:"retry"`
+	Variables   struct {
+		AWS              map[string]interface{} `json:"AWS,omitempty"`
```
squat: by convention, JSON is encoded in camel-case. Why is this using title casing?

installer/pkg/terraform/executor_test.go
```diff
@@ -93,28 +93,29 @@ func TestExecutorSimple(t *testing.T) {
 		assert.FailNow(t, "TerraForm apply timed out")
 	}
 
+	// TODO: Update tests for new CLI
```
squat: rather that comment these out, can you skip the tests so that their lack of implementation is noted in the test output?

squat:
Overall LGTM. needs rebase. I'm a little confused about the marshaling of json but if you're just keeping up with @spangenberg 's changes, then that shouldn't block this PR. We will need to fix json marshalling in his code generation

installer/pkg/terraform/executor_test.go
kyoto: Done

installer/api/handlers_terraform.go
kyoto: Like you said in your other comment, I'm just matching the existing JSON keys. I agree that they should be changed to camelCase though.

squat: Ok thanks for confirming. We’ll need to follow up and fix this on the config-generator side of things

kyoto:
ok to test

squat:
ok to test

kyoto:
We will wait to merge this until after `ut2-integration` has merged into `master`.

squat:
we'll need to re-open this against master now that ut2 has merged.

PR #2895, opened 2018-02-05
frontend: Add IP range start and end addresses to CIDR tooltip

kyoto:
screenshot-1

### DIFF
--- a/installer/frontend/components/cidr.jsx
+++ b/installer/frontend/components/cidr.jsx
@@ -3,27 +3,19 @@ import React from 'react';
 import { connect } from 'react-redux';
 
 import { CIDR, Connect, Deselect } from './ui';
-import { validate } from '../validate';
-import { DESELECTED_FIELDS } from '../cluster-config.js';
-
-export const cidrSize = cidr => {
-  if (validate.CIDR(cidr)) {
-    return undefined;
-  }
-  const [, bits] = cidr.split('/');
-
-  // JavaScript's bit shifting only works on signed 32bit ints so <<31 would be negative :(
-  return Math.pow(2, 32 - parseInt(bits, 10));
-};
+import { cidrEnd, cidrSize, cidrStart } from '../cidr';
+import { DESELECTED_FIELDS } from '../cluster-config';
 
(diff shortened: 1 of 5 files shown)


installer/frontend/components/cidr.jsx
```diff
@@ -16,14 +17,35 @@ export const cidrSize = cidr => {
   return Math.pow(2, 32 - parseInt(bits, 10));
 };
 
+// Convert an IPv4 string to an integer
+const ip2int = ip => {
+  const [a, b, c, d] = ip.split('.').map(i => parseInt(i, 10));
+  return a * (256 ** 3) + b * (256 ** 2) + c * 256 + d;
```
squat: another, faster way to do this, I think, would be to bitshift the pieces and `|` them together.

installer/frontend/components/cidr.jsx
```diff
@@ -16,14 +17,35 @@ export const cidrSize = cidr => {
...
+};
+
+// Convert an integer to an IPv4 string
+const int2ip = int => [0, 1, 2, 3].map(i => Math.floor(int / (256 ** i)) % 256).reverse().join('.');
+
+// Get the first IP address in a CIDR range
+const cidrStart = cidr => {
+  const [ip, bits] = cidr.split('/');
```
squat: should you `parseInt` here?

kyoto: Good catch, but I discovered another bug in `cidrStart()` and ended up changing it anyway.

installer/frontend/components/cidr.jsx
kyoto: I checked this again, but unfortunately we do run into the signed 32bit int issue, so I it this as is.

kyoto:
@squat I ended up reworking this by
- Rewriting `cidrStart()` to avoid another problem with bit shifting signed ints
- Moving all the CIDR helpers to new file `cidr.js`
- Adding unit tests

Could you please look again.

kyoto:
cc @tlwu2013 @sym3tri

PR #2901, opened 2018-02-06
add multistep support for the cli

enxebre:
Introduce multistep workflow for the cli, based on https://github.com/coreos/tectonic-installer/pull/2802
The steps defined here could be exposed a user facing subcommands.
Unit test coverage should be increased once we agree on the approach and shape this model.


installer/pkg/workflow/destroy.go
```diff
@@ -19,29 +18,80 @@ func NewDestroyWorkflow(buildPath string) Workflow {
...
 	}
+
+	// TODO: Discrimitate by config provider. if platform is aws:
 	return simpleWorkflow{
 		metadata: metadata{
 			statePath: buildPath,
 		},
 		steps: []Step{
 			terraformPrepareStep,
-			terraformInitStep,
```
alexsomesan: The init step should be included in the destroy workflow as well.
If everything is in order, it will just no-op and move on.
But it does help the user experience when modules or plugins may be missing, in which case it repairs rather than throwing an error.

installer/pkg/workflow/destroy.go
```diff
@@ -19,29 +18,80 @@ func NewDestroyWorkflow(buildPath string) Workflow {
...
-			terraformDestroyStep,
+			joiningDestroyStep,
+			bootstrapDestroyStep,
+			ignitionDestroyStep,
+			assetsDestroyStep,
+			tlsDestroyStep,
 		},
 	}
+
+	//return simpleWorkflow{
```
alexsomesan: Why are the commented parts still in there?

installer/pkg/workflow/workflow_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+func test2Step(m *metadata) error {
+	return nil
+}
+
+func test3Step(m *metadata) error {
+	return errors.New("Step failed!")
+}
+
+func TestWorkflowTypeExecute(t *testing.T) {
```
alexsomesan: Awesome! Thanks for adding units for this!

enxebre:
let's wait for https://github.com/coreos/tectonic-installer/pull/2905 and the subsequent PRs which complete https://github.com/coreos/tectonic-installer/pull/2802 before getting this one in, so once those are merged we ensure the cli match the underlying steps

installer/pkg/workflow/destroy.go
enxebre: this is commented out base on the TODO above, so once there's a config parser, initially we would run multistep if platform is aws or single step otherwise

enxebre:
closing in favour of https://github.com/coreos/tectonic-installer/pull/2949

PR #2903, opened 2018-02-06
Bazel docs

squat:
Notable changes:

- bazel: add terraform examples to tarball
- installer: refactor so binaries do not have platform suffix
- tarball: add helper to build tarball with darwin bin
- Documentation: add docs for building with bazel

fixes: INST-747

cc @alexsomesan @mxinden

### DIFF
--- a/Documentation/dev/build.md
+++ b/Documentation/dev/build.md
@@ -1,0 +1,142 @@
+# Building Tectonic Installer
+
+The Tectonic Installer leverages the [Bazel build system](https://bazel.build/) to build all artifacts, binaries, and documentation.
+
+## Getting Started
+
+Install Bazel > 0.10.x using the instructions [provided online](https://docs.bazel.build/versions/master/install.html).
+*Note*: compiling Bazel from source requires Bazel.
+*Note*: cross-compilation of Golang binaries, which the Tectonic Installer leverages, is [broken in Bazel 0.9.0](https://github.com/bazelbuild/rules_go/issues/1161).
+
+Clone the Tectonic Installer git repository locally:
+
+```sh
+git clone git@github.com:coreos/tectonic-installer.git && cd tectonic-installer
+```
+
+## Quickstart
+
+To build Tectonic for development or testing, run the `tarball` script located in the base directory:
+
+```sh
+./tarball
+```
+
+This will produce an archive named `tectonic.tar.gz` containing all the assets necessary to bring up a Tectonic cluster, namely the:
+
+* Tectonic Installer binary;
+* Terraform modules;
+* Terraform binary;
+* Terraform provider binaries; and
+* examples
+
+To build a versioned Tectonic release tarball, set a `VERSION` environment variable and run the `tarball` script:
+
+```sh
+export VERSION=1.2.3-beta
+./tarball
+```
+
+This will create an archive named `tectonic_1.2.3-beta.tar.gz` in the base directory.
+
+For more details on building a Tectonic release or other Tectonic assets as well as workarounds to some known issues, read on.
+
+## Building A Release Tarball
+
+To build a release tarball for the Tectonic Installer, issue the following command from the `tectonic-installer` root directory:
+
+```sh
+bazel build tarball
+```
+
+*Note*: Bazel < 0.9.0 is known to [fail to build tarballs when using Python 3](https://github.com/bazelbuild/bazel/issues/3816); to avoid this issue, force Python 2 by using:
+
+```sh
+bazel build --force_python=py2 --python_path=/usr/bin/python2 tarball
+```
+
+This will create a tarball named `tectonic.tar.gz` in the `bazel-bin` directory with the following directory structure:
+
+```
+tectonic
+├── config.tf
+├── examples
+├── modules
+├── platforms
+└── tectonic-installer
+    ├── darwin
+    │   ├── terraform
+    │   └── terraform-provider-matchbox
+    └── linux
+        ├── installer
+        ├── terraform
+        └── terraform-provider-matchbox
+```
+
+*Note*: declarative cross-compilation of Golang binaries from Linux to other platforms is [currently broken in Bazel](https://github.com/bazelbuild/rules_go/issues/1148) so the tarball will not include a Darwin binary. 
+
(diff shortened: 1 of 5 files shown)


Documentation/dev/build.md
```diff
@@ -0,0 +1,109 @@
...
+bazel build backend
+```
+
+To build a cross-compiled binary for another platform, e.g. a Darwin binary on a Linux machine, specify the target platform explicitly:
+
+```sh
+bazel build backend --experimental_platforms=@io_bazel_rules_go//go/toolchain:darwin_amd64
+```
+
+## Building A Release Tarball
```
mxinden: How about switching `## Building A Release Tarball` with `## Building the Installer Binary`? Building the tarball should be the goal of the majority of the readers.

Documentation/dev/build.md
```diff
@@ -0,0 +1,109 @@
...
+    ├── darwin
+    │   ├── terraform
+    │   └── terraform-provider-matchbox
+    └── linux
+        ├── installer
+        ├── terraform
+        └── terraform-provider-matchbox
+```
+
+*Note*: declarative cross-compilation of Golang binaries from Linux to other platforms is [currently broken in Bazel](https://github.com/bazelbuild/rules_go/issues/1148) so the tarball will not include a Darwin binary.
```
mxinden: If I understand this correctly, running `./tarball` should work for everyone? How about moving this section to the top of `## Building A Release Tarball` then, so users know what to do and can dive deeper if needed.

Documentation/dev/build.md
```diff
@@ -0,0 +1,109 @@
...
+
+Clone the Tectonic Installer git repository locally:
+
+```sh
+git clone git@github.com:coreos/tectonic-installer.git && cd tectonic-installer
+```
+
+## Building the Installer Binary
+
+To build the Tectonic Installer binary, issue the following command from the `tectonic-installer` root directory:
```
sym3tri: Maybe callout why I would choose this option over `bazel build tarball` and vice vesra. It's not really clear from the docs.

Documentation/dev/build.md
```diff
@@ -0,0 +1,109 @@
...
+├── config.tf
+├── examples
+├── modules
+├── platforms
+└── tectonic-installer
+```
+
+*Note*: the generated tarball will not include the version string in its own name since output names must be known ahead of time in Bazel. To include the version in the tarball name copy or move the archive with the desired name in the destination.
+
+Similarly, do build a versioned tarball with the Darwin binary included, export the `VERSION` environment variable and run the `tarball` shell script:
```
sym3tri: `s/do build/to build/`

Documentation/dev/build.md
```diff
@@ -0,0 +1,109 @@
...
+To build a tarball with the Darwin binary included, run the temporary `tarball` script:
+
+```sh
+./tarball
+```
+
+In order to build a release tarball with the version string in the directory name within the tarball, export a `VERSION` environment variable and then build the tarball while passing the variable to the build:
+
+```sh
+export VARIABLE=1.2.3-beta
```
sym3tri: Shouldn't this be `VERSION` instead of `VARIABLE`?

Documentation/dev/build.md
```diff
@@ -0,0 +1,109 @@
...
+├── platforms
+└── tectonic-installer
+```
+
+*Note*: the generated tarball will not include the version string in its own name since output names must be known ahead of time in Bazel. To include the version in the tarball name copy or move the archive with the desired name in the destination.
+
+Similarly, do build a versioned tarball with the Darwin binary included, export the `VERSION` environment variable and run the `tarball` shell script:
+
+```sh
+export VARIABLE=1.2.3-beta
```
sym3tri: same?

Documentation/dev/build.md
```diff
@@ -0,0 +1,109 @@
...
+./tarball
+```
+
+This will create a tarball named `tectonic_1.2.3-beta.tar.gz` in the base directory.
+
+
+## Building the Smoke Test Binary
```
sym3tri: Can you link to the docs on running smoke tests?

Documentation/dev/build.md
sym3tri: Maybe a brief sentence or two about what's included in the tarball too.

Documentation/dev/build.md
sym3tri: consider just having a "Quickstart" section at the top with the most common/basic flow.

squat: Done. I think Max's comment about changing the order helps disambiguate the importance of the different targets. I also added a small blurb about why you might want only the binary. I added a Quickstart documenting the two most important use-cases:

1. build a tarball for development/testing
2. build a release tarball

Documentation/dev/build.md
squat: done

Documentation/dev/build.md
squat: thanks!

Documentation/dev/build.md
squat: yup thanks!

Documentation/dev/build.md
squat: got it

Documentation/dev/build.md
squat: sure! I thought the file trees would be helpful for this. I added this info explicitly in the quickstart now

squat:
@sym3tri I prefer not to modify the root README.md until we actually pull the trigger on modifying the workflow and eliminate the Makefile. This PR is really intended to document using Bazel. We should follow up with a different PR that entirely eliminates the old system and replaces the documentation.

squat:
thanks for the comments, @mxinden and @sym3tri. PTAL again for some updates

alexsomesan:
Looks good to me too.

As a second iteration on this, I would prefer to decouple the development workflow from the release one, to make the average developer's experience as simple as it can get.

I mostly say that because the additional complexity and workarounds needed for cross-compilation are only interesting in a release setting, where a developer would not have a need to cross-compile as part of his normal workflow and so his experience could be simpler.
Ideally, at the very least we have separate chapters for these two topics but I would go as far as separate documents next to each other.

PR #2904, opened 2018-02-06
monitoring: Fix migration and insecure image org

brancz:
This closes TECREL-160 (release blocker for Lithium).

@alexsomesan @cpanato @sym3tri 

@pst @fabxc @ant31 

@coresolve @justaugustus could you test to make sure the reported issue on your providers disappears? I tried on a new AWS cluster, there everything seems fine and in general the bug should be provider independant, but let's make double sure.


coresolve:
Verified on metal and aws vanilla clusters. Now works.

coresolve:
LGTM

PR #2905, opened 2018-02-06
platform/aws: move assets generation into own tf step

enxebre:
Breaking down https://github.com/coreos/tectonic-installer/pull/2802 into smaller PRs.
Move assets and tls generation into its own step for aws.


enxebre:
thanks a lot for reviewing! next follow up https://github.com/coreos/tectonic-installer/pull/2924

PR #2909, opened 2018-02-07
tests/frontend: Ignore tectonic_container_linux_* tfvars in GUI tests

kyoto:


squat:
@kyoto what is the motivation behind this? can you elaborate in the git commit

cpanato:
The UI tests check the fields generated in the UI against the existing tfvars we have in the smoke tests. We added the CL_channel in our tfvars to test the CL beta. and this change broke the UI tests. This PR ignore the fields during the validation.

squat:
ok sounds fine. can you add that to the commit message so that this is documented? otherwise it is unclear why this change occurred.

kyoto:
@squat Sorry, I've added the commit message now.

PR #2912, opened 2018-02-08
frontend: Remove CSS rule that actually makes vertical alignment worse

kyoto:
Before | After
--- | ---
screenshot-1 | screenshot-2


squat:
ok to test

PR #2914, opened 2018-02-08
frontend: Add additional info for AWS VPCs and subnets

kyoto:
For VPCs, add CIDR to dropdown labels and add CIDR tooltip with number of IP addresses and address range.

For subnets, add Name (from AWS tag) to dropdown labels and add CIDR tooltip with number IP addresses, address range and number of available addresses.

Change label format of both dropdowns to `<CIDR> (<NAME> | <ID>)`, which matches the format in AWS Console.

Change `Select` to accept an `options` array instead of an `availableValues` object.

Change `Deselect` to also output the field label, which makes things easier because the deselect checkbox is expected to be rendered alongside the label and it's easier to ensure they use matching `id` / `htmlFor` attributes if they are rendered together.

Rename `fieldName` to `deselectId` and change `CIDRRow` to take `deselectId` instead of a `selectable` boolean.

cc @robszumski 

Before | After
--- | ---
screenshot-2 | screenshot-1

### DIFF
--- a/installer/api/handlers_aws.go
+++ b/installer/api/handlers_aws.go
@@ -22,6 +22,12 @@ import (
 	paws "github.com/coreos/tectonic-installer/installer/pkg/aws"
 )
 
+type awsVPC struct {
+	CIDR string `json:"instanceCIDR"`
+	ID   string `json:"id"`
+	Name string `json:"name"`
+}
+
 type labelValue struct {
 	Label string `json:"label"`
 	Value string `json:"value"`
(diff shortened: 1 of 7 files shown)


squat:
This is failing due to Golint issues:

```
/go/src/github.com/coreos/tectonic-installer/installer/api/handlers_aws.go:25:6: exported type VPC should have comment or be unexported
```

installer/api/handlers_aws.go
```diff
@@ -22,6 +22,12 @@ import (
 	paws "github.com/coreos/tectonic-installer/installer/pkg/aws"
 )
 
+type VPC struct {
```
squat: no need to export this type

kyoto: Thanks @squat. I renamed it to `awsVPC`.

kyoto:
ok to test

installer/api/handlers_aws.go
```diff
@@ -145,26 +151,26 @@ func awsGetVPCsHandler(w http.ResponseWriter, req *http.Request, _ *Context) err
...
-	response := []labelValue{}
-
+	response := make([]awsVPC, 0, len(vpcs.Vpcs))
 	for _, vpc := range vpcs.Vpcs {
 		if vpc.VpcId == nil {
 			continue
 		}
 
-		label := aws.StringValue(vpc.VpcId)
+		name := ""
```
squat: in Go types all have `zero-values`. the zero-value for a string is `""`; so the conventional way to declare a string and have memory allocated for its value on the stack is just:

```go
var name string
```

installer/pkg/aws/getters.go
```diff
@@ -261,11 +263,22 @@ func getVPCSubnetsByIDs(sess *session.Session, subnetIDs []*string) ([]VPCSubnet
 
 	subnets := make([]VPCSubnet, 0, len(output.Subnets))
 	for _, subnet := range output.Subnets {
+		name := ""
```
squat: same here with the declaration of the string.

installer/api/handlers_aws.go
kyoto: @squat Fixed. Thanks!

installer/pkg/aws/getters.go
kyoto: Fixed

kyoto:
@squat Could you please review again?

PR #2915, opened 2018-02-08
Installer API: Add buildTime and tectonicVersion

kalmog:
this will add a new endpoint `/releases/current` to the API
which will fetch TECTONIC_VERSION from the build environment
and displays the build time in a JSON format.

### DIFF
--- a/buildvars.sh
+++ b/buildvars.sh
@@ -1,0 +1,5 @@
+#!/bin/bash -ex
+
+# Vars exported to the build info
+echo TECTONIC_VERSION ${TECTONIC_VERSION}
+echo BUILD_TIME $(date -u '+%d_%b_%Y_%I:%M:%S%p')
--- a/installer/api/api.go
+++ b/installer/api/api.go
@@ -103,6 +103,7 @@ func New(config *Config) (http.Handler, error) {
 
 	//handlers_latest_release.go
 	mux.Handle("/releases/latest", logRequests(httpHandler("GET", ctx, latestReleaseHandler)))
+	mux.Handle("/releases/current", logRequests(httpHandler("GET", ctx, currentReleaseHandler)))
 	return mux, nil
 }
 
--- a/installer/api/handlers_latest_release.go
+++ b/installer/api/handlers_latest_release.go
@@ -3,6 +3,8 @@ package api
 import (
 	"io"
 	"net/http"
+
+	"github.com/coreos/tectonic-installer/installer/version"
 )
 
 // latestReleaseHandler gets the tectonic's latest release from coreos.com.
@@ -14,4 +16,17 @@ func latestReleaseHandler(w http.ResponseWriter, req *http.Request, _ *Context) 
 	io.Copy(w, res.Body)
 	res.Body.Close()
 	return nil
+}
+
+// Fetch tectonic's Build version and return in JSON format
+func currentReleaseHandler(w http.ResponseWriter, req *http.Request, _ *Context) error {
+	version := struct {
+		TectonicVersion string `json:"tectonicVersion"`
+		BuildTime       string `json:"buildTime"`
+	}{
+		TectonicVersion: version.TectonicVersion,
+		BuildTime:       version.BuildTime,
+	}
+
+	return writeJSONResponse(w, req, http.StatusOK, version)
 }
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,0 +1,1 @@
+build --color=yes --workspace_status_command=./buildvars.sh
--- a/installer/api/BUILD.bazel
+++ b/installer/api/BUILD.bazel
@@ -16,6 +16,7 @@ go_library(
     visibility = ["//visibility:public"],
     deps = [
         "//installer/assets:go_default_library",
+        "//installer/version:go_default_library",
         "//installer/pkg/aws:go_default_library",
         "//installer/pkg/containerlinux:go_default_library",
         "//installer/pkg/terraform:go_default_library",
--- a/installer/version/BUILD.bazel
+++ b/installer/version/BUILD.bazel
@@ -1,0 +1,12 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library")
+
+go_library(
+    name = "go_default_library",
+    srcs = ["version.go"],
+    importpath = "github.com/coreos/tectonic-installer/installer/version",
+    visibility = ["//visibility:public"],
+    x_defs = {
+        "TectonicVersion": "{TECTONIC_VERSION}",
+        "BuildTime": "{BUILD_TIME}"
+    },
+)
--- a/installer/version/version.go
+++ b/installer/version/version.go
@@ -1,0 +1,7 @@
+package version
+
+// TectonicVersion - defaults to 'dev-build' if TECTONIC_VERSION was not set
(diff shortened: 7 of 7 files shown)


installer/api/api.go
```diff
@@ -103,6 +103,7 @@ func New(config *Config) (http.Handler, error) {
 
 	//handlers_latest_release.go
 	mux.Handle("/releases/latest", logRequests(httpHandler("GET", ctx, latestReleaseHandler)))
+	mux.Handle("/releases/build", logRequests(httpHandler("GET", ctx, currentReleaseHandler)))
```
cpanato: need to use the name `build`? because the other is using `latest` and this new endpoint is to get the current. then maybe use `current` instead of `build`

kalmog: I was deliberating between the two (I actually created the endpoint with "current" to start with). I thought that "current" sounded weird. can be changed, if you think it's better.

kalmog:
Output example:
image

kalmog:
changed endpoint from `/releases/build` to `/releases/current` per @cpanato suggestion.

installer/stamp/stamp.go
```diff
@@ -0,0 +1,7 @@
+package stamp
```
squat: What does `stamp` mean? Is it `timestamp`? It is convention to call this package `version`, see for example https://github.com/coreos/matchbox/blob/master/matchbox/version/version.go or https://github.com/coreos-inc/bridge/blob/master/version/version.go. Please rename for legibility and consistency

installer/stamp/stamp.go
```diff
@@ -0,0 +1,7 @@
...
+
+// TectonicVersion - defaults to 'dev-build' if TECTONIC_VERSION was not set
+var TectonicVersion = "dev-build"
```
squat: For simplicity can we just call this `Version`?

installer/api/handlers_latest_release.go
```diff
@@ -15,3 +17,13 @@ func latestReleaseHandler(w http.ResponseWriter, req *http.Request, _ *Context)
 	res.Body.Close()
 	return nil
 }
+
+// Fetch tectonic's Build version and return in JSON format
+func currentReleaseHandler(w http.ResponseWriter, req *http.Request, _ *Context) error {
+
+	verMap := map[string]string{
+		"TectonicVersion": stamp.TectonicVersion,
```
squat: This returns a json object with title-cased keys, which is a little odd. By convention, json keys are normally camel-cased, or sometimes snake-cased. Can you fix the casing?

installer/api/handlers_latest_release.go
```diff
@@ -15,3 +17,13 @@ func latestReleaseHandler(w http.ResponseWriter, req *http.Request, _ *Context)
 	res.Body.Close()
 	return nil
 }
+
+// Fetch tectonic's Build version and return in JSON format
+func currentReleaseHandler(w http.ResponseWriter, req *http.Request, _ *Context) error {
+
+	verMap := map[string]string{
```
squat: For cases such as this where all the fields are known, we typically use an anonymous struct rather than a map. This should also be more memory-efficient

squat: There is no need to specify the type of this variable in the name. Please rename to just “version”

installer/api/handlers_latest_release.go
```diff
@@ -15,3 +17,13 @@ func latestReleaseHandler(w http.ResponseWriter, req *http.Request, _ *Context)
...
+		"TectonicVersion": stamp.TectonicVersion,
+		"BuildTime":       stamp.BuildTime,
+	}
+	return writeJSONResponse(w, req, http.StatusOK, verMap)
```
squat: Where is this API endpoint consumed in the front end? Without it, this pull request does not fully solve our issue of the reported build version being wrong

buildvars.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+# Vars exported to the build info
+echo TECTONIC_VERSION ${TECTONIC_VERSION}
```
squat: Who sets this environment variable? Please add this detail to the build documentation. For consistency, with the rest of the build process, I think that this variable should be simply `VERSION`

buildvars.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
squat: please add `-e` here

buildvars.sh
alexsomesan: I'd actually prefer to keep these environment inputs namespaced like this, with TECTONIC_*

Two reasons for this:
1. VERSION is much too common and the probability of it affecting other tools or just being used for something else in the system seems rather high.
2. The prefixing helps to visualize the current state of the environment much easier by piping `env` to `sort`.

squat: that's fine; the main issues here are twofold:

1. Who sets this variable? We need this documented somewhere.
2. We need to make this consistent with the rest of the build process; if we want to go the namespaced route, then that means changing the variables used by the tarball bazel build rule so that we do not need to set two different variables.

installer/stamp/stamp.go
kalmog: The name was taken from a bazel example of using x_defs to add variables to compiled binaries, and they named it "stamped bin". But your suggestion is actually clearer.

installer/api/handlers_latest_release.go
kalmog: I'm planning to add that change in a separate commit

squat: Sounds good. Separate commit in this same PR or a different PR?

buildvars.sh
kalmog: I agree with @alexsomesan in the using just `VERSION` would be too generic. We could use this api endpoint in the future, to output any kind of build information that would help us debug issues, such as go / bazel versions, etc, etc. So I'd opt for modifying the documentation and existing vars.

squat: Sounds good. Please make those changes in this PR then

alexsomesan: @squat Both are valid questions.
I think for 1) it's whatever RA ends up running in - most likely their Jenkins pipeline.

For 2), if there're no hard blockers, I'd prefer we align everything that's an input to the build process under TECTONIC_ or TECTONIC_BUILD_. We have to keep in mind that the CLI will also support env var overrides. We might want to make a distinction between the runtime and the buildtime inputs (namespaces). But that's mostly off-topic for this change.

kyoto:
@kalmog Looks like the commit comment needs updating with the new API endpoint and JSON keys.

Let me know if you need any help with the frontend changes.

installer/api/handlers_latest_release.go
```diff
@@ -15,3 +17,17 @@ func latestReleaseHandler(w http.ResponseWriter, req *http.Request, _ *Context)
 	res.Body.Close()
 	return nil
 }
+
+// Fetch tectonic's Build version and return in JSON format
+func currentReleaseHandler(w http.ResponseWriter, req *http.Request, _ *Context) error {
+
```
squat: nit: can you eliminate this extra newline?

kalmog:
ok to retest

kalmog:
retest this please

PR #2918, opened 2018-02-08
networking: Update Calico and Canal

tmjd:
The changes include:
- Update calico/node to v2.6.7
- Update calico/cni to v1.11.2
- Add terminationGracePeriodSeconds to Canal and Calico manifests
- Fixed env vars in Calico manifest

Before creating your PR, please make sure to add the appropriate GitHub labels
like e.g. `run-smoke-tests` + `platform/<xxx>`. For more details see
[tests/README.md](../tests/README.md).

(In case you don't have permissions to add labels, please ask a
[Maintainer](../MAINTAINERS).)


cpanato:
ok to test

modules/net/calico/resources/manifests/kube-calico.yaml
```diff
@@ -151,10 +154,10 @@ spec:
               value: "warning"
             # Don't enable BGP.
             - name: CALICO_NETWORKING_BACKEND
-              value: "k8s,bgp"
+              value: "bgp"
```
robszumski: what does this mean for cloud environments or anywhere you don't want to use BGP?

tmjd: I'm glad you pointed this out.  `k8s,bgp` is not a valid value for that field and what I updated it to was not any better.  It should be 'bird'.
That field specifies how Calico will distribute the routes and for this it uses BGP and defaults to using bird for that.

As for your question.  Calico uses BGP if it is setting up pod networking, by default it creates a BGP mesh between the nodes.  If one doesn't want to use BGP then they must use a different method to configure networking between pods (for example Canal).

That field should be set to `none`, `bird`, or `gobgp`.  If it is set to none then calico operates in a policy only mode.

I'll push up changes in a minute switching it to `bird`.

robszumski: @alexsomesan How would we need to template this for canal users?

tmjd: @robszumski Canal currently exists in the tectonic-installer as a separate manifest, are you thinking about combining those manifests into one and templating the differences? You can see that Canal has `CALICO_NETWORKING_BACKEND` set to `none`.

sym3tri:
Thanks for the contribution and sorry about the delay. I'm going to close this b/c the 1.8 release already shipped.

tmjd:
@sym3tri Is there a branch where I can submit a PR so the next 1.8.x series release can get these changes?

PR #2922, opened 2018-02-09
Support additive node labels

hhoover:

Solves #667 and partially impacts #1439, from #1468


cpanato:
ok to test

cpanato:
don't we need to push this changes to master as well?

cc @sym3tri

sym3tri:
We're solving this in a different way and in master.

PR #2924, opened 2018-02-09
platform/aws: add bootstrap node and step for joining it

enxebre:
Breaking down #2802 into smaller PRs.
Fix INST-823 and INST-824
At this point this would require:
`terraform import aws_autoscaling_group.masters cluster-aws-masters`
`terraform import aws_autoscaling_group.workers cluster-aws-workers` for the joining step. Just a note that in the future the tradeoffs of simplifying this workflow by using single instances could be re-evaluated


alexsomesan:
@enxebre What's the reasoning in favor of using import `terraform import aws_autoscaling_group.masters cluster-aws-masters` versus referencing the autoscaling group via remote state datasource?

enxebre:
@alexsomesan good question! to my knowledge data source are read only so we can take advantage of it to take action based on data from a previews step but I can't see how to modify the state of the resource without importing it. The way to avoid it as far as I can see would be to use independent resources, so single instances as stated in the desc above

PR #2926, opened 2018-02-09
tests/frontend: Better tests for file upload fields and clean up temporary files when tests complete

kyoto:
Add `setFileField()` and `testFileTextCombo()` custom commands for testing file upload inputs.

Use the same temporary file location for all file upload tests and delete that file at the end of the test run. This should prevent temporary files being left on the file system when the tests exit.

Also improves testing of upload / textarea combos:
  - Test that the textarea is populated with the file contents
  - Test textarea field validation
  - Stop using xpath selectors because all combos now have id attributes that we can use instead.


installer/frontend/ui-tests/commands/setFileField.js
```diff
@@ -0,0 +1,14 @@
+const fs = require('fs');
+
+exports.command = function (selector, text, textSelector = undefined) {
```
squat: why are we explicitly setting a default to `undefined`? this is the default value of any parameter that was not set. moreover, we don't do this anywhere else, I believe

kyoto: This was just meant to indicate that the parameter is optional, but after looking at it again, I think it's better to move the textarea tests to `testFileTextCombo` and remove the `textSelector` parameter completely.

kyoto:
@squat Could you take another look please?

installer/frontend/ui-tests/tests/aws.js
```diff
@@ -1,3 +1,5 @@
+const fs = require('fs');
+
```
squat: is this extra newline wanted?

kyoto: We're generally putting a newline between package imports and imports from the same repo.

PR #2927, opened 2018-02-09
installer: support declarative x-compile

squat:
This PR simplifies tarball generation so that only a single command is needed to build a complete tarball including any cross-compiled binaries:

```sh
bazel build tarball
```

Details:
This commit eliminates platform-specific `select` statements from Golang
rules in several BUILD.bazel files. This allows us to workaround Bazel being
unable to modify the platform toolchain. Sourcing all of the files for
all platforms still results in a correctly compiled binary because
rules_go performs file filtering
https://github.com/bazelbuild/rules_go/issues/1148#issuecomment-353136033.

TL;DR we no longer need the `tarball` shell script to build a complete
Tectonic tarball with binaries for all supported platforms.

cc @mxinden @alexsomesan @sym3tri


squat:
@mxinden I just fixed compilation darwin compilation issues. To verify that a complete tarball builds on both Linux and Mac, both commands below should exit 0:

```sh
bazel build tarball --experimental_platforms=@io_bazel_rules_go//go/toolchain:darwin_amd64
bazel build tarball --experimental_platforms=@io_bazel_rules_go//go/toolchain:linux_amd64
```

PTAL again

PR #2928, opened 2018-02-09
Support additive node labels

hhoover:

Solves #667 and partially impacts #1439, from #1468 (Dupe of #2922 on 1.8.7 branch)


squat:
@hhoover, thanks for this, however, we are going in a different direction for supporting node labels for arbitrary node pools.

PR #2936, opened 2018-02-12
installer/pkg: Integrate config generator into CLI

spangenberg:


installer/pkg/workflow/destroy.go
```diff
@@ -35,15 +32,8 @@ func terraformDestroyStep(m *metadata) error {
...
-	tfDestroy.Stdin = os.Stdin
-	tfDestroy.Stdout = os.Stdout
-	tfDestroy.Stderr = os.Stderr
-	err := tfDestroy.Run()
-	if err != nil {
-		return err
-	}
-	return nil
+
+	return terraformExec(m, "destroy", "-force")
```
alexsomesan: Talk to @enxebre about his solution from #2901 
We need a common ground :)

enxebre: This seems a reasonable abstraction according to the code changes introduced by this PR, we can always reshape organically once https://github.com/coreos/tectonic-installer/pull/2901 gets in and the underlying steps introduce more tf hassle

PR #2939, opened 2018-02-13
Remove obsolete git-version script

kalmog:


mxinden:
Looks good to me. I think @squat already implemented this in Bazel, right?

kalmog:
yes, this is a leftover, I suppose, from the old build method.

PR #2940, opened 2018-02-13
Ignition assets

enxebre:
Move ignition config creation into assets step.
Remove leader election code.
Add support for ncg.
This will require `terraform destroy -target aws_route53_record.tectonic_ncg -var-file terraform.tfvars ../../platforms/aws/` before running the joining nodes step


enxebre:
cc @alexsomesan

modules/aws/master-asg/ignition_s3.tf
```diff
@@ -1,21 +1,17 @@
-resource "aws_s3_bucket_object" "ignition_master" {
-  bucket  = "${var.s3_bucket}"
-  key     = "ignition_master.json"
-  content = "${data.ignition_config.main.rendered}"
-  acl     = "private"
-
-  server_side_encryption = "AES256"
+data "ignition_config" "s3" {
```
alexsomesan: It think these should be called something  'bootstrap_*'  rather then "s3"

enxebre: that will be the ignition config for all masters not only bootstrap, so may be ignition_masters

PR #2947, opened 2018-02-14
Installer: Move tectonicVersion and buildTime to /tectonic/facts

kalmog:
This commit moves the `tectonicVersion and` `buildTime` build facts to `/tectonic/facts`.

### DIFF
--- a/installer/api/handlers_latest_release.go
+++ b/installer/api/handlers_latest_release.go
@@ -3,8 +3,6 @@ package api
 import (
 	"io"
 	"net/http"
-
-	"github.com/coreos/tectonic-installer/installer/version"
 )
 
 // latestReleaseHandler gets the tectonic's latest release from coreos.com.
@@ -16,17 +14,4 @@ func latestReleaseHandler(w http.ResponseWriter, req *http.Request, _ *Context) 
 	io.Copy(w, res.Body)
 	res.Body.Close()
 	return nil
-}
-
-// Fetch tectonic's Build version and return in JSON format
-func currentReleaseHandler(w http.ResponseWriter, req *http.Request, _ *Context) error {
-	version := struct {
-		TectonicVersion string `json:"tectonicVersion"`
-		BuildTime       string `json:"buildTime"`
-	}{
-		TectonicVersion: version.TectonicVersion,
-		BuildTime:       version.BuildTime,
-	}
-
-	return writeJSONResponse(w, req, http.StatusOK, version)
 }
--- a/installer/api/api.go
+++ b/installer/api/api.go
@@ -103,7 +103,6 @@ func New(config *Config) (http.Handler, error) {
 
 	//handlers_latest_release.go
 	mux.Handle("/releases/latest", logRequests(httpHandler("GET", ctx, latestReleaseHandler)))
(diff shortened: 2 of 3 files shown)


installer/api/handlers_latest_release.go
```diff
@@ -1,31 +1,55 @@
...
 func currentReleaseHandler(w http.ResponseWriter, req *http.Request, _ *Context) error {
+	LatestRelease := fetchLatestRelease()
+	BuildTime := version.BuildTime
+	TectonicVersion := version.TectonicVersion
+
 	version := struct {
-		TectonicVersion string `json:"tectonicVersion"`
 		BuildTime       string `json:"buildTime"`
+		LatestRelease   string `json:"latestRelease"`
+		TectonicVersion string `json:"tectonicVersion"`
```
squat: Can we unify these names? I.e. why is one “latestRelease” and the other “tectonicVersion”? They are both versions of Tectonic. And it should already be assumed that they are releases since they are in the /release/ path. Could this be renamed `latest` and `current` to make it more intelligible?

installer/api/handlers_latest_release.go
```diff
@@ -1,31 +1,55 @@
...
 
 // latestReleaseHandler gets the tectonic's latest release from coreos.com.
-func latestReleaseHandler(w http.ResponseWriter, req *http.Request, _ *Context) error {
-	res, err := http.Get("https://coreos.com/tectonic/api/releases/latest")
+func fetchLatestRelease() string {
+	var rel string
+	response, err := http.Get("https://coreos.com/tectonic/api/releases/latest/")
 	if err != nil {
-		return newBadRequestError("Failed to get a response from coreos.com: %s", err)
+		fmt.Printf("Failed to get a response from coreos.com: %s", err)
```
squat: Just logging an error with “fmt.Printf” is not really sufficient. We should return an error here.

installer/api/handlers_latest_release.go
```diff
@@ -1,31 +1,55 @@
...
+	"strings"
 
 	"github.com/coreos/tectonic-installer/installer/version"
+	"golang.org/x/net/html"
 )
 
 // latestReleaseHandler gets the tectonic's latest release from coreos.com.
-func latestReleaseHandler(w http.ResponseWriter, req *http.Request, _ *Context) error {
-	res, err := http.Get("https://coreos.com/tectonic/api/releases/latest")
+func fetchLatestRelease() string {
```
squat: This function can fail, so let’s restore the function signature to also return an error.

installer/api/handlers_latest_release.go
```diff
@@ -1,31 +1,55 @@
...
+	var rel string
+	response, err := http.Get("https://coreos.com/tectonic/api/releases/latest/")
 	if err != nil {
-		return newBadRequestError("Failed to get a response from coreos.com: %s", err)
+		fmt.Printf("Failed to get a response from coreos.com: %s", err)
 	}
-	io.Copy(w, res.Body)
-	res.Body.Close()
-	return nil
+	defer response.Body.Close()
```
squat: None of the code below this point is valid if we got an error earlier. We need to skip all of this by returning an error early.

kalmog:
retest this please

squat:
Looks good!

PR #2949, opened 2018-02-14
add multistep support for the cli - WIP - DO NOT MERGE

enxebre:
Add support for deploying using a cluster the cli:
`cli install --config=terraform.tfvariables`
cluster-name and platform are currently hardcoded

### DIFF
--- a/installer/cmd/tectonic/main.go
+++ b/installer/cmd/tectonic/main.go
@@ -1,29 +1,61 @@
 package main
 
 import (
+	"log"
+
 	"github.com/coreos/tectonic-installer/installer/pkg/workflow"
 	"gopkg.in/alecthomas/kingpin.v2"
 )
 
 var (
-	dryRunFlag            = kingpin.Flag("dry-run", "Just pretend, but don't do anything").Bool()
-	clusterInstallCommand = kingpin.Command("install", "Create a new Tectonic cluster")
-	clusterDeleteCommand  = kingpin.Command("delete", "Delete an existing Tectonic cluster")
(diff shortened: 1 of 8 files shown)


installer/cmd/tectonic/main.go
```diff
@@ -1,29 +1,61 @@
...
 
 var (
-	dryRunFlag            = kingpin.Flag("dry-run", "Just pretend, but don't do anything").Bool()
-	clusterInstallCommand = kingpin.Command("install", "Create a new Tectonic cluster")
-	clusterDeleteCommand  = kingpin.Command("delete", "Delete an existing Tectonic cluster")
-	deleteClusterDir      = clusterDeleteCommand.Arg("dir", "The name of the cluster to delete").String()
-	clusterConfigFlag     = clusterInstallCommand.Flag("config", "Cluster specification file").Required().ExistingFile()
+	dryRunFlag                = kingpin.Flag("dry-run", "Just pretend, but don't do anything").Bool()
+	clusterInstallCommand     = kingpin.Command("install", "Create a new Tectonic cluster")
+	clusterFullInstallCommand = clusterInstallCommand.Command("full", "Create a new Tectonic cluster").Default()
```
alexsomesan: This should be the default path. No need to explicitly have a sub-command.
#lessismore #minimalism

enxebre: By adding full as default the user can just run `cli install --config` otherwise the same command complains about it expecting a subcommand. Not sure yet how to tell kingpin that the subcommand is not mandatory

PR #2950, opened 2018-02-14
Upgrade fix for CA certs

thorfour:
Proposed fix for https://jira.coreos.com/browse/TECREL-165

Regression introduced by
https://github.com/coreos/tectonic-installer/pull/2814


squat:
@coverprice @thorfour I dug a little more and it turns out that the systemd drop-in that enables update-ca-certificates to run on every boot, and thus create a real (read: non-symlink) file at /etc/ssl/certs/ca-certificates.crt, was a new feature in 1.8.7. So this feature not exist in 1.8.4 and earlier, meaning that /etc/ssl/certs/ca-certificates.crt will be a symlink on old clusters and as a result anything that only bind mounts /etc/ssl/certs will not be able to traverse the symlink to the actual file as described in the issue.

This PR LGTM

PR #2954, opened 2018-02-15
modules/tectonic: fix alertmanager ingress config.

sym3tri:
The alert manager ingress config expects both the host and the port.
Previously we only provided the host. This can break on some platforms
where the port is required to access Tectonic ingress.


cpanato:
ok to test

cpanato:
tested in `azure` using private vps and was able to ge the source link from the alertmanager

PR #2955, opened 2018-02-15
tests: Add delay to ssh and update cl channel to beta

cpanato:
Adding this sleep to wait for some time before we start ssh into the server
if we ssh during the reboot from torcx this might put the shutdown in a weird state
and that's might be the reason why we saw several connection timeouts in tests while spinning up a cluster

Also, removing the cl channel stable leftover form another PR. we now are using channel beta for our tests in master.

ps. this is an experiment. if we continue seeing the issue we will remove this.


cpanato:
failed during terraform apply and the other failure was in the bootkube bootstrap.

tests/rspec/lib/cluster.rb
```diff
@@ -283,6 +284,20 @@ def wait_for_bootstrapping
     puts 'HOORAY! The cluster is up'
   end
 
+  # Adding this sleep to wait for some time before we start ssh into the server
+  # if we ssh during the reboot from torcx this might put the shutdown in a weird state
+  # and that's might be the reason why we saw several connection timeouts in tests while spinning up a cluster
```
squat: i've seen this exact issue and root cause testing on bare-metal. I think this hack should work for now

PR #2958, opened 2018-02-15
remove bootstrap ignition as part of rm-assets

enxebre:
overwrite bootstrap ignition as part of rm-assets service

### DIFF
--- a/steps/assets/resources/rm-assets.sh
+++ b/steps/assets/resources/rm-assets.sh
@@ -11,14 +11,15 @@ s3_clean() {
     /usr/bin/docker run \
         --volume /tmp:/tmp \
         --network=host \
-        --env LOCATION="${assets_s3_location}" \
+        --env LOCATION="${bucket_s3_location}" \
         --entrypoint=/bin/bash \
         ${awscli_image} \
         -c '
             set -e
             set -o pipefail
             REGION=$(wget -q -O - http://169.254.169.254/latest/meta-data/placement/availability-zone | sed '"'"'s/[a-zA-Z]$//'"'"')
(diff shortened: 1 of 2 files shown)


steps/assets/resources/rm-assets.sh
```diff
@@ -11,14 +11,15 @@ s3_clean() {
...
-        --env LOCATION="${assets_s3_location}" \
+        --env LOCATION="${bucket_s3_location}" \
         --entrypoint=/bin/bash \
         ${awscli_image} \
         -c '
             set -e
             set -o pipefail
             REGION=$(wget -q -O - http://169.254.169.254/latest/meta-data/placement/availability-zone | sed '"'"'s/[a-zA-Z]$//'"'"')
-            /usr/bin/aws --region="$REGION" s3 cp /tmp/assets.zip s3://"$LOCATION"
+            /usr/bin/aws --region="$REGION" s3 cp /tmp/assets.zip s3://"$LOCATION/assets.zip"
```
alexsomesan: I think we actually want to invalidate both `assets.zip` and `vögeln`.
Would something like `echo > ignition` and `echo > assets.zip` before each `cp` operation work for that?

enxebre: isn't what we are doing by `touch /tmp/assets.zip`? I can rename if it's not clear

alexsomesan: Unless /tmp/assets.zip doesn't already exist.
In that case it just gets it's timestamps updated and ends up being uploaded to S3.
Not the right time and place to obsess about edge cases tough, so this might be nitpicking.

PR #2960, opened 2018-02-15
installer: Integrate multistep cli with configuration

spangenberg:


installer/pkg/workflow/terraform.go
```diff
@@ -1,31 +1,42 @@
...
-	tfCommand.Dir = buildPath
-	tfCommand.Stdin = os.Stdin
-	tfCommand.Stdout = os.Stdout
-	tfCommand.Stderr = os.Stderr
-	err := tfCommand.Run()
-	if err != nil {
-		return err
-	}
-	return nil
+const (
```
enxebre: I'd rather move this constants and findTemplatesForStep to a different package may be utils or step, and keep terraform workflow agnostic, with which is purely terraform

enxebre:
just had a quick look, is looking great to me.
- Can you share an example config to test this?
- As there's assets that rely on terraform yet, e.g ignition, bootkube.sh, tectonic.sh, ncg ncg-ignition-config, for now we'll need to consolidate the output path of the kube config generated here with the generated by terraform, and make sure it keep satisfying the bootstrap step assumptions to upload and deploy all  of them

installer/pkg/workflow/install.go
```diff
@@ -1,190 +1,87 @@
...
-			clusterName: clusterName,
-			configFile:  configFile,
-		},
+func NewInstallAssetsWorkflow(clusterDir string) Workflow {
+	return Workflow{
+		metadata: metadata{clusterDir: clusterDir},
 		steps: []Step{
-			terraformPrepareStep,
-			assetsStep,
+			readClusterConfigStep,
```
enxebre: why is the readClusterConfigStep step needed?

spangenberg: Metadata always just gets the cluster dir:
```go
metadata: metadata{clusterDir: clusterDir}
```
The `readClusterConfigStep` takes care of ensuring that the config gets parsed, e.g. for cluster `Name` or `Platform`.

installer/pkg/workflow/terraform.go
spangenberg: Makes sense, will change that.

spangenberg:
@enxebre can you point me towards the output path where you want the operator yml's to be. I'll include a `config.yml` in the contrib for the moment.

enxebre:
@spangenberg the casuistic is as follows:
"init" step will generate: `foo/kube-system.yaml` and `foo/kube-system.yaml`
Then "assets" will generate what is needed (https://github.com/coreos/tectonic-installer/tree/master/modules/bootkube/resources and https://github.com/coreos/tectonic-installer/tree/master/modules/tectonic/resources/manifests) inside foo/generated
The expectation for the system the way it's right now are:
- Your kube-system.yml ends up in generated/manifests/cluster-config.yaml
- Your tectonic-system.yml ends up in generated/tectonic/cluster-config.yaml

enxebre:
@spangenberg  thanks a lot, this is pretty awesome work!
I'm currently deploying from this branch doing smoke functional testing
Some gotchas:
- EtcdCount is being ignored, after a quick look, probably because is expecting a string https://github.com/coreos/tectonic-installer/blob/ut2-integration/installer/pkg/terraform-generator/tectonic.go#L27?
- Kube yamls are generated with multiple blanks
- The way this is at the moment, the kube yaml files will be overridden by the "assets" step, as terraform will override the whole "generated" folder. As a transition we should probably
`run init
run assets
mv init yaml into expected folder`
Would you mind to add TODO note to spot that out?
- In follow ups we need to add high coverage for unit test the templates generation from different config files, which will give a huge benefit at the time of making changes

spangenberg:
@enxebre thanks, this sounds fantastic.

I changed the workflow order that the config yamls will be generated after assets step was run.
This should circumvent the issues you encountered.
I will try to target the config issues in another PR.

PR #2961, opened 2018-02-16
frontend: Use monospace font for Terraform log output

kyoto:
Terraform log output area now uses the same font-family rule as the textarea inputs.

cc @robszumski, @tlwu2013 

Before | After
--- | ---
screenshot-2 | screenshot-1


coverprice:
ok to test

PR #2962, opened 2018-02-16
frontend: Add progress bar for Terraform destroy actions

kyoto:
The implementation of `estimateTerraformProgress()` will need to change for the new CLI, but the other logic and the UI should remain valid.

screenshot-1


coverprice:
ok to test

PR #2968, opened 2018-02-16
tests: Archive tarball for release automation to consume

mxinden:
This patch archives the Tectonic installer build tarball from Bazel as a
Jenkins artifact to be consumed by e.g. the release automation pipeline.

@coverprice Would this be a possible solution for the release pipeline to consume build artefacts from the Tectonic installer pipeline? I would imagine a workflow like:
1. Release automation triggers Tectonic installer pipeline
2. Once Tectonic installer pipeline finishes and build result is green
    - Retrieve tarball artefact from Tectonic installer build
    - Release tarball artefact e.g. as a nightly

@cpanato This might result in a high number of tarballs, that we persist. The [`buildDiscarder`](https://github.com/coreos/tectonic-installer/blob/master/Jenkinsfile#L81) should take care of that, right?

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -175,6 +175,7 @@ pipeline {
 
                   stash name: 'tectonic.tar.gz', includes: 'tectonic.tar.gz'
                   stash name: 'smoke-tests', includes: 'smoke'
+                  archiveArtifacts allowEmptyArchive: true, artifacts: 'tectonic.tar.gz'
                 }
 
                 withDockerContainer(params.builder_image) {


coverprice:
@mxinden If we want to share the build steps between RA + Installer, then this situation looks great. The `bazel build tarball` step above does all the heavy lifting, and it's mostly wrapped by a vanilla "clean / checkout / archive results" Jenkins stage. It should be trivial to update the RA pipeline to call out to the `bazel build` step. Ideally I'd like the RA pipeline to share the Jenkins stage definition too, but that's tricky across repos without us resorting to Jenkins Shared Libraries, so as a stop-gap we can just copy the Jenkins stage to the RA pipeline.

cpanato:
One solution is we save the tarball in S3 and then RA can consume form there

coverprice:
@mxinden You suggested that RA could trigger Installer and then retrieve the artifact bundle after the build stage. That's one way to do it, but I'd like to detangle the 2 pipelines if possible. I feel that having one job rely on another increases the complexity. However I'm fine with continuing as-is for now, until I can suggest a cleaner way of separating them.

coverprice:
@cpanato An S3 intermediary would definitely work. I'm a little wary of using that though, just because it feels like adding another step in complex Rube Goldberg machine. :) I don't have enough Jenkins-fu to suggest anything better though.

cpanato:
@coverprice @mxinden  To use Jenkins and get the tarballs from master nightly builds is easy as well, you just need to access this URL:

<jenkins_URL>/job/tectonic-installer/job/master/lastSuccessfulBuild/artifacts

we can talk during the week because we will need to set up an user in order to have access only for the artifacts

PR #2976, opened 2018-02-19
frontend: Get current version info from /tectonic/facts

kyoto:
Fetch current version info from `/tectonic/facts` API endpoint instead of
using `babel-plugin-git-version`.

`isReleaseVersion()` (used to determine whether to send Google Analytics
events) will now return `true` if the value of `tectonicVersion` is a valid
semver string that doesn't contain the substring `-rc.`.

If `buildTime` is set, use it as the version info's `title` attribute.

Simplify `hasNewVersion()` by using the fact that our version strings are
now defined such that determining which is greater should just be a case
of comparing them lexicographically.

Slightly simplify version info CSS.

cc: @kalmog, @robszumski, @tlwu2013 

screenshot-1

screenshot-2

### DIFF
--- a/installer/frontend/components/header.jsx
+++ b/installer/frontend/components/header.jsx
@@ -1,5 +1,6 @@
 import _ from 'lodash';
 import React from 'react';
+import { connect } from 'react-redux';
 import semver from 'semver';
 
 import { A, DocsA, DropdownMixin } from './ui';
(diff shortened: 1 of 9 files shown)


kyoto:
ok to test

installer/frontend/components/header.jsx
```diff
@@ -23,45 +24,48 @@ const parseLatestVersion = (html) => {
...
   }
 
-  const latestReleasePr = semver.prerelease(latestRelease);
-  const gitTagPr = semver.prerelease(GIT_TAG);
+  const latestPreRelease = semver.prerelease(latestRelease);
+  const currentPreRelease = semver.prerelease(currentRelease);
 
   //No rc string in the latest release
-  if (latestReleasePr.length < gitTagPr.length) {
+  if (latestPreRelease.length < currentPreRelease.length) {
```
squat: Why don’t we return the comparison expression here like in the other cases?

squat:
Looks like this is failing until #2980 merges

installer/frontend/components/header.jsx
kyoto: Yes, that logic looks like it was incorrect to me.

Actually, I think we can greatly simplify this because our version strings can now be ordered alphabetically. The current logic dates from June, before our release naming became consistent, so I've changed it.

kyoto:
ok to test

kyoto:
ok to test

PR #2977, opened 2018-02-19
tests/frontend: Add AWS GUI test for external etcd and custom CA

kyoto:
Adds some AWS GUI checks not included in our other tests:
  - External etcd input fields
  - Custom CA input fields
  - Variety of different valid input values for other fields


kyoto:
ok to test

PR #2979, opened 2018-02-19
frontend: Add message when in progress install is detected

kyoto:
The case where the app detects that an install is already in progress when the app loads is potentially confusing, so add a message to explain why it skipped to the Start Installation page.

screenshot-1


kyoto:
cc: @robszumski, @tlwu2013

kyoto:
ok to test

PR #2981, opened 2018-02-19
tests: Stash installer binary for gui tests

mxinden:
In the long run the GUI tests should use the tarball produced by Bazel
as well. As a short fix, this patch stashes the installer binary again,
so the GUI tests can use it.


squat:
@mxinden this isn’t not needed since we have a bazel test for guitests

PR #2982, opened 2018-02-19
cli: populate core config

enxebre:
This fills unpopulated core config fields


enxebre:
@spangenberg I'm not sure there's a better way to set this "default" non user input values. If this is a valid approach we could also move them to constants

PR #2984, opened 2018-02-19
cli: add missing tectonic config

enxebre:
Add missing tectonic config

### DIFF
--- a/installer/pkg/config-generator/generator.go
+++ b/installer/pkg/config-generator/generator.go
@@ -161,19 +161,19 @@ func (c ConfigGenerator) utilityConfig() (*tectonicutility.OperatorConfig, error
 		return nil, err
 	}
 	hashedAdminPassword := string(bytes)
-	adminUserID, err := generateRandomId(16)
+	adminUserID, err := generateRandomID(16)
 	if err != nil {
 		return nil, err
 	}
-	consoleSecret, err := generateRandomId(16)
+	consoleSecret, err := generateRandomID(16)
 	if err != nil {
 		return nil, err
 	}
-	KubectlSecret, err := generateRandomId(16)
+	KubectlSecret, err := generateRandomID(16)
 	if err != nil {
 		return nil, err
 	}
-	clusterId, err := generateClusterId(16)
+	clusterID, err := generateClusterID(16)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


installer/pkg/config-generator/generator.go
```diff
@@ -45,26 +52,35 @@ func New(cluster config.Cluster) ConfigGenerator {
...
-	return &kubeaddon.OperatorConfig{
+	addonConfig := kubeaddon.OperatorConfig{
 		TypeMeta: metav1.TypeMeta{
 			APIVersion: kubeaddon.APIVersion,
-			Kind:       kubeaddon.Kind,
+			// TODO: get Kind from kubeaddon.Kind
+			Kind: "AddonConfig",
 		},
 	}
+	cidrhost, _ := cidrhost(c.Cluster.Networking.ServiceCIDR, 10)
```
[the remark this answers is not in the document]
alexsomesan: Cool, thanks for the clarification. The fact that it's not obvious (at least to me) where that comes from is a reason enough to try to improve such cases.

Would you mind also dropping in a `//TODO: re-evaluate solution` in such cases which are not meant to be the long term solution?

This would allow us to have a simple auditing command to collect all these when we plan future work.

squat:
@alexsomesan agreed. We have made a lot of concessions over several PRs and features for the sake of getting this overall work merged quickly. We need to dedicate time as soon as possible for this cleanup otherwise we will have proven ourselves right and created permanent debt

installer/pkg/config-generator/generator.go
```diff
@@ -162,3 +242,58 @@ func marshalYAML(obj interface{}) (string, error) {
...
+		hexStr[16:20],
+		hexStr[20:32]), nil
+}
+
+// cidrhost takes an IP address range in CIDR notation
+// and creates an IP address with the given host number.
+// If given host number is negative, the count starts from the end of the range
+// Reproduces tf behaviour.
+// TODO: re-evaluate solution
+func cidrhost(iprange string, hostNum int) (string, error) {
```
sym3tri: nit & not a blocker: i'd suggest later returning an IP instead of a string.

installer/pkg/config-generator/generator.go
```diff
@@ -109,19 +141,67 @@ func (c ConfigGenerator) networkConfig() *tectonicnetwork.OperatorConfig {
...
+	clusterID, err := generateClusterID(16)
+	if err != nil {
+		return nil, err
+	}
+	utilityConfig.IdentityConfig.AdminEmail = c.Console.AdminEmail
+	utilityConfig.IdentityConfig.AdminPasswordHash = hashedAdminPassword
+	utilityConfig.IdentityConfig.AdminUserID = adminUserID
+	utilityConfig.IdentityConfig.ConsoleClientID = "tectonic-console"
+	utilityConfig.IdentityConfig.ConsoleSecret = consoleSecret
+	utilityConfig.IdentityConfig.KubectlClientID = "tectonic-kubectl"
```
sym3tri: TODO for later: all these should be isolated as constants somewhere.

enxebre:
Thanks for reviews, will merge this and some follow ups can hopefully be addressed as part of INST-845

alexsomesan:
I'm merging to incorporate into CLI for demo.

enxebre:
follow up https://github.com/coreos/tectonic-installer/pull/2994

PR #2917, opened 2018-02-08
tectonic: pull in default network policy (utility), enable ingress controller operator

squeed:
This moves the ingress controller to its own namespace.

PR #2985, opened 2018-02-19
modules/tectonic: set correct utility config kind

squeed:


squat:
ok to test

squat:
ok to test

enxebre:
Not sure how the logic behind it fully works yet but something else probably need to be changed as the operators seems to be only happy with kind AddonConfig and kind UtilityConfig rather than TectonicUtilityOperatorConfig and TectonicUtilityOperatorConfig

squeed:
I think you're right. The Kind needs to match exactly the kind defined in tectonic-config, e.g. https://github.com/coreos/tectonic-config/blob/master/config/tectonic-utility/config.go
I'll update this PR and check.

squeed:
I think I'll actually fix this in #2917

squat:
Closing in favor of #2917

PR #2986, opened 2018-02-20
Some CLI fixes

alexsomesan:
This change fixes destroy steps on incomplete builds. 

In cases where a user does a `tectonic install assets...` and then attempts a `tectonic destroy ...` the tool will fail with an error. The error is caused by the destroy operations that don't have a preceding initialized state. In this case, all destroy steps after `assets` don't have their modules and state initialized and will fail.

It also enables the CLI executable to determine its location and extrapolate the locations of other resources as per expected tarball layout.

### DIFF
--- a/installer/pkg/workflow/destroy.go
+++ b/installer/pkg/workflow/destroy.go
@@ -16,13 +16,37 @@ func NewDestroyWorkflow(clusterDir string) Workflow {
 }
 
 func destroyAssetsStep(m *metadata) error {
-	return tfDestroy(m.clusterDir, assetsStep, findTemplatesForStep(assetsStep))
+	if !hasStateFile(m.clusterDir, assetsStep) {
+		// there is no statefile, therefore nothing to destroy for this step
+		return nil
+	}
+	templateDir, err := findTemplatesForStep(assetsStep)
+	if err != nil {
+		return err
+	}
+	return tfDestroy(m.clusterDir, assetsStep, templateDir)
 }
 
 func destroyBootstrapStep(m *metadata) error {
-	return tfDestroy(m.clusterDir, bootstrapStep, findTemplatesForStep(bootstrapStep))
+	if !hasStateFile(m.clusterDir, bootstrapStep) {
+		// there is no statefile, therefore nothing to destroy for this step
+		return nil
+	}
+	templateDir, err := findTemplatesForStep(bootstrapStep)
+	if err != nil {
+		return err
+	}
+	return tfDestroy(m.clusterDir, bootstrapStep, templateDir)
 }
 
 func destroyJoinStep(m *metadata) error {
-	return tfDestroy(m.clusterDir, joinStep, findTemplatesForStep(joinStep))
+	if !hasStateFile(m.clusterDir, joinStep) {
+		// there is no statefile, therefore nothing to destroy for this step
+		return nil
+	}
+	templateDir, err := findTemplatesForStep(joinStep)
+	if err != nil {
+		return err
+	}
+	return tfDestroy(m.clusterDir, joinStep, templateDir)
 }
--- a/installer/pkg/workflow/destroy.go
+++ b/installer/pkg/workflow/destroy.go
@@ -16,37 +16,26 @@ func NewDestroyWorkflow(clusterDir string) Workflow {
 }
 
 func destroyAssetsStep(m *metadata) error {
-	if !hasStateFile(m.clusterDir, assetsStep) {
-		// there is no statefile, therefore nothing to destroy for this step
-		return nil
-	}
-	templateDir, err := findTemplatesForStep(assetsStep)
-	if err != nil {
-		return err
-	}
-	return tfDestroy(m.clusterDir, assetsStep, templateDir)
+	return runDestroyStep(m.clusterDir, assetsStep)
 }
 
 func destroyBootstrapStep(m *metadata) error {
-	if !hasStateFile(m.clusterDir, bootstrapStep) {
-		// there is no statefile, therefore nothing to destroy for this step
-		return nil
-	}
-	templateDir, err := findTemplatesForStep(bootstrapStep)
-	if err != nil {
-		return err
-	}
-	return tfDestroy(m.clusterDir, bootstrapStep, templateDir)
+	return runDestroyStep(m.clusterDir, bootstrapStep)
 }
 
 func destroyJoinStep(m *metadata) error {
-	if !hasStateFile(m.clusterDir, joinStep) {
+	return runDestroyStep(m.clusterDir, joinStep)
+}
+
+func runDestroyStep(clusterDir, step string) error {
+	if !hasStateFile(clusterDir, step) {
 		// there is no statefile, therefore nothing to destroy for this step
 		return nil
 	}
-	templateDir, err := findTemplatesForStep(joinStep)
+	templateDir, err := findTemplates(step)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


installer/pkg/workflow/utils.go
```diff
@@ -42,13 +45,35 @@ func copyFile(fromFilePath, toFilePath string) error {
...
+	if err != nil {
+		return "", err
+	}
+	base = filepath.Join(base, "steps")
+	stat, err := os.Stat(base)
+	if err != nil {
+		return "", err
+	}
+	if !stat.IsDir() {
+		fmt.Errorf("invlid templates path: %s", base)
```
squat: Return here?

installer/pkg/workflow/utils.go
```diff
@@ -42,13 +45,35 @@ func copyFile(fromFilePath, toFilePath string) error {
...
+	}
+	base = filepath.Join(base, step)
+	stat, err = os.Stat(base)
+	if err != nil {
+		return "", err
+	}
+	if !stat.IsDir() {
+		fmt.Errorf("invlid templates path: %s", base)
```
squat: And here?

installer/pkg/workflow/utils.go
enxebre: typo

installer/pkg/workflow/utils.go
enxebre: typo

installer/pkg/workflow/utils.go
```diff
@@ -42,13 +45,35 @@ func copyFile(fromFilePath, toFilePath string) error {
...
+	}
+	return terraformExec(clusterDir, "destroy", "-force", fmt.Sprintf("-state=%s.tfstate", bootstrapStep), "-target=aws_route53_record.tectonic_ncg", templatesPath)
 }
 
-// TODO: Handle errors for Getwd
-func findTemplatesForStep(step string) string {
-	dir, _ := os.Getwd()
-	return filepath.Join(dir, "steps", step)
+func findTemplatesForStep(step string) (string, error) {
+	base, err := baseLocation()
```
squat: We should either wrap this bare error or wrap the first error in baselocation, which is also currently bare (ideally both). Otherwise, we will see an error initializing an executable and wonder where it’s coming from

installer/pkg/workflow/destroy.go
```diff
@@ -16,13 +16,37 @@ func NewDestroyWorkflow(clusterDir string) Workflow {
...
+	templateDir, err := findTemplatesForStep(assetsStep)
+	if err != nil {
+		return err
+	}
+	return tfDestroy(m.clusterDir, assetsStep, templateDir)
 }
 
 func destroyBootstrapStep(m *metadata) error {
-	return tfDestroy(m.clusterDir, bootstrapStep, findTemplatesForStep(bootstrapStep))
+	if !hasStateFile(m.clusterDir, bootstrapStep) {
```
enxebre: As a follow up we should create `runDestroyStep` analogous `runInstallStep`

installer/pkg/workflow/utils.go
alexsomesan: there should be one indeed! :)

installer/pkg/workflow/utils.go
alexsomesan: 👍

installer/pkg/workflow/utils.go
alexsomesan: That's a good point!
Would you rather have a concrete type for each or just more meaningful messages via fmt.Errorf(..)?

squat: For now wrapping the error via fmt.Errorf is plenty. Unless we needed to do some programmatic error recovery and need the types we can skip creating the error types right now IMO

squat: Again, this is needed here in particular since we are doing a few levels of error bubbling without wrapping the messages making the source more difficult to identify

installer/pkg/workflow/destroy.go
alexsomesan: I think this is similar to what @squat suggested above by the tryDestroy method, no?
I'll see what I can do there.

installer/pkg/workflow/utils.go
alexsomesan: 👍

installer/pkg/workflow/utils.go
alexsomesan: 👍

installer/pkg/workflow/utils.go
alexsomesan: OK, fixed. Log line for such errors now looks like this:
```
2018/02/20 12:50:44 invalid path for '/Users/alex/Desktop/tectonic/steps/assets' templates: stat /Users/alex/Desktop/tectonic/steps/assets: no such file or directory
```

installer/pkg/workflow/destroy.go
alexsomesan: I implemented `runDestroyStep` like you suggested.

installer/pkg/workflow/destroy.go
```diff
@@ -16,13 +16,37 @@ func NewDestroyWorkflow(clusterDir string) Workflow {
...
+	if err != nil {
+		return err
+	}
+	return tfDestroy(m.clusterDir, bootstrapStep, templateDir)
 }
 
 func destroyJoinStep(m *metadata) error {
-	return tfDestroy(m.clusterDir, joinStep, findTemplatesForStep(joinStep))
+	if !hasStateFile(m.clusterDir, joinStep) {
+		// there is no statefile, therefore nothing to destroy for this step
```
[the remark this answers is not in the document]
alexsomesan: I implemented a `runDestroyStep` to mach the already present `runInstallStep` (per suggestion from @enxebre).
Looks cleaner now.

installer/pkg/workflow/utils.go
```diff
@@ -162,3 +196,19 @@ func writeFile(path, content string) error {
 
 	return nil
 }
+
+func baseLocation() (string, error) {
```
enxebre: just to clarify if I'm reading this correctly this assumes the output of `bazel build tarball` and it won't work in a different context e.g dev running bazel build //installer/cmd/tectonic:darwing May be worth a comment clarifying that?

enxebre:
lgtm

installer/pkg/workflow/utils.go
alexsomesan: You're right, this assumes running from the built tarball.
This *is on purpose* because all testing should be done against the release layout, not the source repository one.

The main reason why we adopted Bazel is to have a fast, reproducible build that we could use during development to test against the same artefacts as the one we release. 
We've failed multiple times in the past because of this usage skew between development and released items.

If you're testing the repository itself, please switch to testing from the build output.
Whatever rough edges there are in this process we should identify and iron out so that we make it as smooth as possible to follow this flow.

alexsomesan: Here's the steps I'm taking to test a new change:
1. `bazel build tarball && tar -C ~/Desktop/ -xf bazel-bin/tectonic.tar.gz`
2. `cd ~/Desktop`
3. `./tectonic/tectonic-installer/darwin/tectonic init --config test-cluster.xml`
4. `./tectonic/tectonic-installer/darwin/tectonic install assets --dir TestCluster`
5.  `...`

PR #2995, opened 2018-02-20
bazel/build/ci: add terraform fmt check

cpanato:
fixes: https://jira.coreos.com/browse/INST-879

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -165,9 +165,10 @@ pipeline {
                 echo "originalCommitId: ${originalCommitId}"
 
                 withDockerContainer(tectonicBazelImage) {
+                  sh "bazel test terraform_fmt --test_output=all"
+                  sh "bazel test installer/frontend:unit --test_output=all"
                   sh"""#!/bin/bash -ex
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -166,8 +166,9 @@ pipeline {
 
                 withDockerContainer(tectonicBazelImage) {
                   sh"""#!/bin/bash -ex
+                    bazel test terraform_fmt --test_output=all
```
mxinden: I think you can combine the two test targets into one, like the `bazel build tarball tests/smoke` below, right?

squat: We can but if we separate them then we can view them as separated steps in blue ocean I think

mxinden: @squat Ah, sounds like a good idea. As far as I know, you need to use a separate `sh` step for each one then.

cpanato: I will split in separate `sh`

squat:
hmm not sure about the need for different sh blocks. I’ve seen that each of the cp calls result in separate blocks in blue ocean

PR #3001, opened 2018-02-21
Remove makefile / examples folder and documentation/variables folder

cpanato:
Examples and Documentation we can build using bazel now. Don't need to keep those files in the repo.


Makefile
```diff
@@ -170,7 +16,7 @@ smoke-test-env-docker-image:
 	docker build -t quay.io/coreos/tectonic-smoke-test-env -f images/tectonic-smoke-test-env/Dockerfile .
 
 .PHONY: tests/smoke
-tests/smoke: bin/smoke smoke-test-env-docker-image
```
squat: the docker build needs to ADD the smoke test binary to the container, so the binary needs to be built. Either we use the output from Bazel in the Dockerfile, we re-build the binary using Make, or we build the container entirely using Bazel. I vote for the latter, so we can entirely eliminate this Makefile

PR #3012, opened 2018-02-22
digitalocean: Upgrade do-spaces-tool

aknuds1:
Use the latest version of aknudsen/do-spaces-tool as it fixes a [critical bug](https://github.com/aknuds1/do-spaces-tool/issues/1), which makes cluster bootstrapping fail due to not being able to upload to the DigitalOcean bucket.


sym3tri:
ok to test

PR #3013, opened 2018-02-22
Ip

enxebre:
https://github.com/coreos/tectonic-installer/commit/c9468fcd24874d8c54734db371c11b455291cadd is needed until INST-953 gets fixed
Create independent bootstrap node. Lock down s3 permissions for ignition config by ip. Fix INST-944


modules/aws/master-asg/bootstrap.tf
```diff
@@ -0,0 +1,36 @@
...
+    create_before_destroy = true
+  }
+}
+
+resource "aws_eip_association" "eip_assoc" {
+  instance_id   = "${aws_instance.bootstrap_node.id}"
+  allocation_id = "${var.eip_bootstrap_id}"
+}
+
+// Value of count cannot be computed https://github.com/hashicorp/terraform/issues/15172
```
enxebre: Hoping for better ideas here

enxebre:
retest this please

modules/aws/master-asg/bootstrap.tf
sym3tri: What is `count` used for here?

steps/joining/masters.tf
```diff
@@ -11,9 +11,9 @@ provider "aws" {
 
 resource "aws_autoscaling_group" "masters" {
   name                 = "${var.tectonic_cluster_name}-masters"
-  desired_capacity     = "${var.tectonic_master_count}"
-  max_size             = "${var.tectonic_master_count * 3}"
-  min_size             = "${var.tectonic_master_count}"
+  desired_capacity     = "${local.master_count}"
+  max_size             = "${local.master_count * 3}"
+  min_size             = "${local.master_count}"
```
sym3tri: What is the purpose of the changes here?

enxebre: https://github.com/coreos/tectonic-installer/pull/3013/files#diff-ab1cdfcf8c59247cffcabbac2f236b96R21
If the user desired capacity `tectonic_master_count` is 3 masters, as the bootstrap node is not part of the scaling group,  the scaling group expected capacity would be the user input  - the bootsrapnode

enxebre:
Ideally we could eventually attach the bootstrap node to the autoscaling group rather than to the load balancers https://docs.aws.amazon.com/autoscaling/ec2/userguide/attach-instance-asg.html
we'll need to add support in terraform as it seems it only support attaching elb to autoscaling groups  https://www.terraform.io/docs/providers/aws/r/autoscaling_attachment.html cc @sym3tri

squat:
closing since ut2 has merged. we should re-open this against master if needed.

PR #3014, opened 2018-02-22
modules/aws: enable cluster recovery after cleanup

squat:
If the assets zip is 0 bytes, then this means that the cluster already
bootstrapped once and the assets bundle was cleaned. In this case, exit
successfully so the kubelet will start and the cluster can be recovered
via bootkube.

Fixes: TECREL-187

cc @coresolve @coverprice @sym3tri

### DIFF
--- a/modules/aws/master-asg/resources/init-assets.sh
+++ b/modules/aws/master-asg/resources/init-assets.sh
@@ -27,6 +27,13 @@ fi
 # Download the assets from S3.
 # shellcheck disable=SC2154
 /usr/bin/bash /opt/s3-puller.sh "${assets_s3_location}" /var/tmp/tectonic.zip
+
+# If the file is 0 bytes, then this means that the cluster already bootstrapped once and the assets bundle was cleaned. In this case, exit successfully so the kubelet will start and this master will join the existing cluster or support bootkube recovery.
+if [ ! -s /var/tmp/tectonic.zip ]; then
+    echo "zip files is 0 bytes; assuming cluster is being recovered"
+    exit 0
+fi
+
 unzip -o -d /var/tmp/tectonic/ /var/tmp/tectonic.zip
 rm /var/tmp/tectonic.zip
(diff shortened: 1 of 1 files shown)


modules/aws/master-asg/resources/init-assets.sh
```diff
@@ -27,6 +27,10 @@ fi
 # Download the assets from S3.
 # shellcheck disable=SC2154
 /usr/bin/bash /opt/s3-puller.sh "${assets_s3_location}" /var/tmp/tectonic.zip
+
+# If the file is 0 bytes, then this means that the cluster already bootsrapped once and the assets bundle was cleaned. In this case, exit successfully so the kubelet will start and the cluster can be recovered via bootkube.
```
coresolve: kubelet will start and this new master will join the existing cluster or support bootkube recovery.

modules/aws/master-asg/resources/init-assets.sh
```diff
@@ -27,6 +27,10 @@ fi
...
+[ ! -s /var/tmp/tectonic.zip ] && exit 0
```
coresolve: maybe put something in the log that says you are calling it a success?

modules/aws/master-asg/resources/init-assets.sh
```diff
@@ -27,6 +27,13 @@ fi
 # Download the assets from S3.
 # shellcheck disable=SC2154
 /usr/bin/bash /opt/s3-puller.sh "${assets_s3_location}" /var/tmp/tectonic.zip
+
+# If the file is 0 bytes, then this means that the cluster already bootsrapped once and the assets bundle was cleaned. In this case, exit successfully so the kubelet will start and this master will join the existing cluster or support bootkube recovery.
```
coverprice: nit typo: bootsrapped -> bootstrapped

squat:
@coverprice thanks, updated

coverprice:
retest this please

PR #3015, opened 2018-02-22
added initial bootkube rendering

thorfour:
@enxebre @alexsomesan Looking for some feedback on the TNC bootstrapping changes I have so far. I was hoping to get some guidance on how to point booting nodes to use the tnc ignition endpoint.


enxebre:
Hey @thorfour currently the process is as follows:
All nodes have identical user-data ignition config https://github.com/coreos/tectonic-installer/blob/ut2-integration/modules/aws/master-asg/master.tf#L64
https://github.com/coreos/tectonic-installer/blob/ut2-integration/modules/aws/worker-asg/worker.tf#L28 We need to move from query string to rest style. (does TNC support this?)

For the bootstrap node a cname translate the NCG url into an s3 endpoint that contains a boot ignition config which is currently generated by the installer (can we use the TNC libraries to generate this config from the cli?)

The NCG is statically deployed by bootkube https://github.com/coreos/tectonic-installer/blob/ut2-integration/modules/bootkube-ut2/resources/manifests/ncg.yaml
with some ignition config generated by the installer https://github.com/coreos/tectonic-installer/blob/ut2-integration/modules/bootkube-ut2/resources/manifests/ncg-config.yaml
(All this needs to be removed and generated by the TNC)

Once the bootstrap node is up (so the NCG running), the cname is removed and the same NCG url points now to a load balancer which sits on top of the NCG pods

The rest of the masters and workers get their config form the NCG.

thorfour:
@enxebre Thanks for the info.

 TNC has been updated to serve configs over a REST path. ex: `server:port/ign/v1/worker` 

TNC at this time doesn't have a way to render a boot ignition config. It's bootstrap mode is just the server. However it might be a good idea to move that rendering into TNC as well to keep all ignition knowledge in a single place. 

The first step to integrating TNC is to get node bootstrapping off of it, and once that is happening, to have it replaced by the long-running TNC that serves configs from cluster objects. 

I currently have the `bootkube.sh` script launching TNC, is my understanding correct that I would just need to point the cname to the bootstrap node TNC endpoint? I'm not familiar with modifying cname objects in aws, can you explain how I might do that?

enxebre:
NCG manifests (https://github.com/coreos/tectonic-installer/blob/ut2-integration/modules/bootkube-ut2/resources/manifests/ncg.yaml https://github.com/coreos/tectonic-installer/blob/ut2-integration/modules/bootkube-ut2/resources/manifests/ncg-config.yaml) should be removed in favour of TNC ones

The NCG port should match the TNC config now, so listening and healthcheking TNC port https://github.com/coreos/tectonic-installer/blob/ut2-integration/modules/aws/vpc/master-elb.tf#L13

Masters, workers and initial s3 config must match new TNC path:
Masters - https://github.com/coreos/tectonic-installer/blob/ut2-integration/modules/aws/master-asg/master.tf#L64
Workers - https://github.com/coreos/tectonic-installer/blob/ut2-integration/modules/aws/worker-asg/worker.tf#L30
Bootstrap node - https://github.com/coreos/tectonic-installer/blob/ut2-integration/steps/bootstrap/s3-assets.tf#L62

modules/bootkube-ut2/resources/bootkube.sh
```diff
@@ -11,6 +11,17 @@ set -e
 
 mkdir -p /etc/kubernetes/manifests/
 
+# shellcheck disable=SC2154
```
enxebre: shouldn't the TNC be rendered by kube-render? if not, it should be deployed as a static pod, so etcd can get its config from it, so the cluster can get state and continue with the bootstrap process

enxebre: cc @alexsomesan @squat

modules/bootkube-ut2/resources/bootkube.sh
```diff
@@ -11,6 +11,17 @@ set -e
...
+/usr/bin/docker run \
+    --detatch \
+    --volume "$(pwd)":/assets \
+    --publish 45900:45900
+    "${tnc_bootstrap_image}" \
```
squat: @enxebre I think that is what @thorfour  is trying to accomplish here. He is manually running TNC on docker so that it can be used to bootstrap etcd.

Although this works, I would prefer to see this run as a static pod as well so that the kubelet can take care of retries. Currently, `kube-core render` and `bootkube start` are one-shot operations with short lifetimes, while TNC is a long running process that does not exit. Also, as a static pod, all of its configuration could be defined as a declarative manifest. Finally, if we make it a static pod then we can avoid overloading this script with operations that are not directly related to bootkube.

steps/assets/tectonic.tf
```diff
@@ -75,6 +75,14 @@ module "bootkube" {
   cloud_config_path   = ""
   tectonic_networking = "${var.tectonic_networking}"
   calico_mtu          = "1480"
```
enxebre: Why do we need this for tectonic?

modules/bootkube/resources/tnc-config.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: v1
```
enxebre: this should be bootkube-ut2

modules/bootkube-ut2/resources/bootkube.sh
thorfour: @squat you're mostly correct. However tnc is split into two separate binaries. This is running the bootstrap binary which isn't the long-running tnc, and needs to be torn down along with bootkube. The follow on TNC will be the long-running one that needs to run in-cluster.

steps/assets/tectonic.tf
thorfour: No idea? I didn't add that

modules/bootkube-ut2/resources/bootkube.sh
thorfour: Got it. Yea static-pod might be the way to go. We just need a story on how to ensure it gets torn down after bootstrapping. I only globbed it onto the bootkube.sh as a path of least resistance in an attempt to get a POC working.

thorfour: FWIW original design was to have tnc-boostrap be a systemd unit, and have a follow-on service have a conflict with that so it gets torn down.

squat: Yeah I see. Ultimately, the bootstrap node is supposed to be recycled, so any static pods will be properly removed and will no longer be run when the node comes back up as a master

enxebre: @thorfour @squat while there's no pivoting mechanism in place we could modify this PR to just run it as a long running static pod for now

squat: The problem with that is that this bootstrap containeris actually a different image. So we would have a daemonset running on other nodes exposing one api and a static pod running here exposing a different API so if you happen to get load balanced to the bootstrap node you may not be able to boot

enxebre: @thorfour @squat so the goal is to get the bootstrap TNC serving static config running as pod, and a pivot mechanism to tear it down and switch to a long running TNC (different binary) serving "dynamic" config from CRD. To get there while there's no such mechanism we need to define a transition, choices IMO are:
- Swap the current NCG daemonSet by a TNC daemonSet serving static config (etcd nodes can't be bootstrapped from it this way as the daemonset requires state)
- Remove the current NCG and only run the TNC as a static pod (cluster might be unstable as if the node running the pod gets rebooted other nodes will fail to join)
- Run both TNC as a pod and as a DaemonSet simultaneously with the problem stated in the comment above and probably other conflicts

enxebre:
@thorfour ut2-integration branch has been merge into master, can you please change the base branch of this PR to master? thanks

squat:
closing this PR for now, since ut2 has been merged. We need to re-open against master.

PR #2911, opened 2018-02-07
platforms/vmware: Use NodePort for tectonic-lb service

lander2k2:
Currently, the VMware platform uses a HostPort service for the tectonic
ingress controller which exposes port 443 on all worker nodes.  This
change will route https traffic to port 32000 for tectonic ingress.

PR #3016, opened 2018-02-22
platforms/vmware: Use NodePort for tectonic-lb service

lander2k2:
Currently, the VMware platform uses a HostPort service for the tectonic
ingress controller which exposes port 443 on all worker nodes. This
change will route https traffic to port 32000 for tectonic ingress.


lander2k2:
Corresponding PR on 1.8.7 branch: https://github.com/coreos/tectonic-installer/pull/2911

coverprice:
ok to test

PR #3018, opened 2018-02-23
Parameterize Slack integration

coverprice:
To support Jenkinsfile development:
- Parameterize the slack channel that gets notified on failure.
- Check the is-enabled flag on every call to notifyBuildSlack, not just one of them.
- Remove notifyBuildSlack callsite param that was never used.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -26,7 +26,7 @@ commonCreds = [
   ],
   [
     $class: 'StringBinding',
-    credentialsId: 'jrussell-github',
+    credentialsId: 'github-coreosbot',
     variable: 'GITHUB_CREDENTIALS'
   ]
 ]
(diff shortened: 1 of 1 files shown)


coverprice:
ok to test

Jenkinsfile
```diff
@@ -142,7 +142,12 @@ pipeline {
     booleanParam(
       name: 'NOTIFY_SLACK',
       defaultValue: false,
-      description: ''
+      description: 'Notify slack channel on failure'
+    )
+    string(
+      name: 'slack_channel',
```
mxinden: All of the Parameters are in upper case style. Would you mind doing the same here?

coverprice: I thought uppercase = for users, and lowercase = system config (i.e. less likely to need modification). But if there's no pattern, I'll be happy to make them consistently uppercase.

PR #3019, opened 2018-02-23
frontend: Use Redux action creators for all action types

kyoto:
Benefits
- Makes action dispatch consistent (was a mix of action creators, action dispatching "helpers" and using the action types directly)
- Moves complexity from reducers (which should be kept simple) to action creators
- Reduces boilerplate

Remove `configActionTypes.APPEND` and `configActionTypes.ADD_IN` reducers (logic moved to action creators).

Convert `configActionTypes.REMOVE_FIELD_LIST_ROW` action to a more generic
`configActionTypes.DELETE_IN` action and convert `dirtyActionTypes.CLEAN` to `dirtyActionTypes.DELETE_IN` to match.

Greatly simplify the `dirtyActionTypes.ADD` reducer by using Immutable.js.

Separate form actions out into `formActions`.

Make `extraData`, `inFly` and `refreshExtraData()` only available to components that need them (currently just `AsyncSelect`).

### DIFF
--- a/installer/frontend/actions.js
+++ b/installer/frontend/actions.js
@@ -64,7 +64,7 @@ export const commitPhases = {
   FAILED: 'COMMIT_FAILED',
 };
 export const FIELDS = {};
-const FIELD_TO_DEPS = {};
+export const FIELD_TO_DEPS = {};
 export const FORMS = {};
 
 const getField = name => {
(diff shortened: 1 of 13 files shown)


installer/frontend/actions.js
```diff
@@ -5,66 +5,98 @@ import { DEFAULT_CLUSTER_CONFIG } from './cluster-config';
...
-  RESET: 'CONFIG_ACTION_RESET',
+  BATCH_SET_IN: 'CONFIG_BATCH_SET_IN',
+  DELETE_IN: 'CONFIG_DELETE_IN',
+  RESET: 'CONFIG_RESET',
+  SET: 'CONFIG_SET',
+  SET_IN: 'CONFIG_SET_IN',
+};
+export const configActions = {
+  // Adds a value at a given path or no-op if a value already exists for the path
+  addIn: (path, value) => (dispatch, getState) => {
```
squat: Is this a valid definition for the action creator? Now that all actions have been re-written as action creators rather than helpers, calling `dispatch(configActions.addIn(...,...))` will be dispatching a function rather than an action object with a payload and type. Does dispatch support this?

kyoto: Sorry, I should have said in the PR description that as well as standard action creators, some are Redux Thunk action creators.

From https://github.com/gaearon/redux-thunk:
> Redux Thunk middleware allows you to write action creators that return a function instead of an action. The thunk can be used to delay the dispatch of an action, or to dispatch only if a certain condition is met. The inner function receives the store methods dispatch and getState as parameters.

PR #3020, opened 2018-02-23
frontend: Fix confusing Cluster Name validation error message

kyoto:
We don't allow lower case characters in the name, so change the validation error message to indicate that.

Before this change, the name `1a` would result in the error `Value must be a valid AWS Stack Name: [a-zA-Z][-a-zA-Z0-9]*`, which was confusing because we don't actually allow upper case characters.


kyoto:
ok to test

PR #3023, opened 2018-02-23
platforms/metal: add fail and retry flags to curl

lucab:
This adds additional flags to curl, so that it can tolerate transient
network failures but hard-fail on permanent errors too.
This way we block a node earlier while still in a debuggable environment,
instead of rebooting into an initramfs emergency console.


lucab:
/cc @squat @pst. This follows our offline discussion.

I don't have permissions to add labels in here, please handle them as required.

squat:
ok to test

squat:
Closing this since we have removed bare-metal from master :(

PR #3028, opened 2018-02-26
Add tectonic tarball instead of repo to tectonic installer docker image

mxinden:
Previously the quay.io/coreos/tectonic-installer:master image contained
the entire Tectonic installer repository. Instead with this patch, only
the Tectonic tarball build by Bazel is added to the image on /tectonic.

### DIFF
--- a/images/tectonic-installer/Dockerfile
+++ b/images/tectonic-installer/Dockerfile
@@ -1,14 +1,7 @@
-FROM golang:1.9.2-stretch
+FROM alpine:3.7
 
-ENV TERRAFORM_VERSION="0.11.1"
-
-RUN apt-get update \
-    && apt-get install --no-install-recommends -y -q \
-    unzip
-
-# Install Terraform
-RUN curl https://releases.hashicorp.com/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_linux_amd64.zip | funzip > /usr/local/bin/terraform && chmod +x /usr/local/bin/terraform
-
-ENV PROJECT_DIR /go/src/github.com/coreos/tectonic-installer
-
-ADD . $PROJECT_DIR/
+# Docker build does not follow symlinks (see
+# https://github.com/moby/moby/issues/1676) Make sure to first copy the tarball
+# from bazel-bin/tectonic.tar.gz to the docker build context folder of your
+# choice (e.g. the root of the repository).
+ADD tectonic.tar.gz /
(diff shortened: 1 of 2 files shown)


images/tectonic-installer/Dockerfile
```diff
@@ -1,14 +1,7 @@
...
-RUN curl https://releases.hashicorp.com/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_linux_amd64.zip | funzip > /usr/local/bin/terraform && chmod +x /usr/local/bin/terraform
-
-ENV PROJECT_DIR /go/src/github.com/coreos/tectonic-installer
-
-ADD . $PROJECT_DIR/
+# Docker build does not follow symlinks (see
+# https://github.com/moby/moby/issues/1676) Make sure to first copy the tarball
+# from bazel-bin/tectonic.tar.gz to the docker build context folder of your
+# choice (e.g. the root of the repository).
+ADD tectonic.tar.gz /
```
sym3tri: why not untar it for user convenience?

mxinden: @sym3tri `docker build` already does that for us: https://docs.docker.com/engine/reference/builder/#add

images/tectonic-installer/Dockerfile
```diff
@@ -1,14 +1,7 @@
-FROM golang:1.9.2-stretch
+FROM debian:stretch
```
squat: Why do we rely on such a large base image? Is there a reason we can't use scratch or even alpine?

mxinden: I wanted other teams to easily install whatever tools they need in addition via `apt`. But if you want I am happy to change.

squat: I think this depends on how other teams leverage this container. If they want to install custom packages, then maybe `alpine` is sufficient since that distro has its own package manager. If they want to install very custom things nicely using a more advanced package manager and larger eco-system then `debian` will be better. If they only need to rely on it as a base image and COPY/ADD in other deps then `scratch` would be fine.

images/tectonic-installer/Dockerfile
sym3tri: fancy

images/tectonic-installer/Dockerfile
sym3tri: +1 for smaller a base image. I don't think we need to try to support every use-case, if so we'd be publishing images based on every linux distro. If other users want to do more customizations they can create their own images & copy the files out. All we really need for now is a place to put the bits.

mxinden: @squat Switched to `alpine` so users still have bash to look around. Let's see what the requirements for this are in the future. Long term I would prefer nightly signed releases with just the tarball on e.g. an S3 bucket.

PR #3029, opened 2018-02-26
Ut2 integration

spangenberg:


squat:
how is the example generated?

installer/pkg/config/types.go
```diff
@@ -10,63 +10,85 @@ import (
...
-
 type containerLinux struct {
 	Channel string `yaml:"Channel"`
 	Version string `yaml:"Version"`
 }
 
-type dns struct {
-	BaseDomain string `yaml:"BaseDomain"`
+type ddns struct {
+	Key    ddnsKey `yaml:"Secret"`
```
squat: both these fields have the same key

installer/pkg/config/types.go
```diff
@@ -10,63 +10,85 @@ import (
...
-	NodeCount       int      `yaml:"NodeCount"`
-	MachineType     string   `yaml:"MachineType"`
-	ExternalServers []string `yaml:"ExternalServers"`
+	Count    int          `yaml:"Count"`
+	External etcdExternal `yaml:"External"`
+}
+
+type etcdExternal struct {
+	CACertPath     string   `yaml:"CACertPath"`
+	ClientCertPath string   `yam:"ClientCertPath"`
```
squat: this struct tag should say `yaml`

mxinden:
@squat As far as I understood, this is manually written. We should invest into generating these soon. Maintaining them manually would be impossible in the long run.

mxinden:
@spangenberg Kubernetes uses camelCasing, not PascalCasing for any yaml file. Any thoughts on this?

installer/pkg/config/types.go
spangenberg: will fix

spangenberg:
@mxinden we already spoke about renaming everything. As soon as the structure is fixed I'm going to rename everything.

squat:
@mxinden I raised the same issue with casing last week. It is very tedious to write all the caps and it is also inconsistent with all other yaml we see: k8s, container linux config etc

mxinden:
@squat are you ok with proceeding here and following up with the casing before merging `ut2-integration` into master?

mxinden:
@squat Thanks.

PR #3033, opened 2018-02-27
frontend: Async fixes and improvements

kyoto:
Several related fixes for handling async API calls.

Also prevents some unnecessary API calls.


ggreer:
If you go through the default wizard, you get stuck on the network config page: screenshot from 2018-03-06 13-47-04

The GUI tests don't test creating a VPC, so tests passed despite it not working.

PR #3036, opened 2018-02-27
tests: update govcloud to use roles and the new aws govcloud account

cpanato:
fixes: INST-921

### DIFF
--- a/tests/rspec/spec/govcloud/vpc_internal_spec.rb
+++ b/tests/rspec/spec/govcloud/vpc_internal_spec.rb
@@ -13,7 +13,10 @@ RSpec.describe 'govcloud-vpc' do
   )
 
   before(:all) do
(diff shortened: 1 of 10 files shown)


tests/rspec/spec/govcloud/vpc_internal_spec.rb
```diff
@@ -13,7 +13,10 @@
   )
 
   before(:all) do
-    @ssh_key = ENV['TF_VAR_tectonic_govcloud_ssh_key'] || AwsSupport.create_aws_key_pairs('us-gov-west-1')
+    @role_credentials = nil
```
enxebre: curious why are we initialising to nil

cpanato: because if we don't use that, it will pass nil to the functions that call aws

cpanato: then we can support both cases: with roles and without role

cpanato:
adding other platforms just to test the changes in the Jenkins file.
retest this please.

PR #3043, opened 2018-02-28
tests: Adjust framework to config file changes

mxinden:
The Tectonic config format changed with
f678d42dbb5ea6b61553a47c98336aeb800fa664

This patch adjusts the test framework accordingly.


tests/rspec/lib/config_file.rb
```diff
@@ -26,20 +26,20 @@ def node_count
   end
 
   def master_count
-    data['Clusters'][0]['Masters']['NodeCount']
```
enxebre: need to be:
`
data['Clusters'][0]['Master']['Count']
data['Clusters'][0]['Worker']['Count']
`

PR #3045, opened 2018-02-28
modules/container_linux: avoid shell

squat:
Due to the way Terraform ternary expressions are evaluated, even if you
specify a Container Linux version to install, Terraform will invoke the
container_linux module external data source and try to run `curl` and
make an HTTP request. If the host is offline or does not have `curl`
installed, then this module will fail. This commit enables this module
to be circumvented by explicitly defining the Container Linux version to
install. Furthermore, this enables non-unix hosts to use the module.

Due to Terraform's limited support for maps, there is no way to interpolate
a list of maps using `join`, `element`, `merge`, etc. This approach serializes
the maps and extracts the version using the known format of the map.

cc @mxinden @enxebre

### DIFF
--- a/modules/container_linux/main.tf
+++ b/modules/container_linux/main.tf
@@ -1,3 +1,4 @@
 data "external" "version" {
+  count   = "${var.release_version == "latest" ? 1 : 0}"
   program = ["sh", "-c", "curl https://${var.release_channel}.release.core-os.net/amd64-usr/current/version.txt | sed -n 's/COREOS_VERSION=\\(.*\\)$/{\"version\": \"\\1\"}/p'"]
 }
--- a/modules/container_linux/outputs.tf
+++ b/modules/container_linux/outputs.tf
@@ -1,3 +1,13 @@
+locals {
+  // Create a map that matches the structure of the output of the external data source
+  // so we can avoid running the shell script and still parse the output consistently.
+  // Here, we jsonencode because ternaries can only operate on flat data types and
+  // Terraform `merge` and `element` do not play nicely with maps.
+  json = "${var.release_version == "latest" ? jsonencode(data.external.version.*.result) : jsonencode(map("version", var.release_version))}"
+}
+
 output "version" {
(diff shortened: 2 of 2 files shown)


modules/container_linux/main.tf
```diff
@@ -1,3 +1,4 @@
 data "external" "version" {
```
mxinden: Would you mind adding a comment with parts of your PR description for the future us?

squat: The description is already in the commit message. Would you like to see it in the code as well?

enxebre:
terraform_fmt failing, lgtm

squat:
@enxebre yeah just pushed a fix. PTAL again

modules/container_linux/main.tf
mxinden: @squat I would have done so, but I don't know what your convention is here. This is not a strong opinion. Go ahead with whatever suits best.

squat: OK @mxinden I just added some comments. PTAL and leave a review

squat:
ok to test

modules/container_linux/outputs.tf
```diff
@@ -1,3 +1,13 @@
+locals {
+  // Create a map that matches the structure of the output of the external data source
+  // so we can avoid running the shell script and still parse the output consistently.
+  // Here, we jsonencode because ternaries can only operate on flat data types and
+  // Terraform `merge` and `element` do not play nicely with maps.
+  json = "${var.release_version == "latest" ? jsonencode(data.external.version.*.result) : jsonencode(map("version", var.release_version))}"
```
enxebre: would may be lookup help here?
 https://www.terraform.io/docs/configuration/interpolation.html#lookup-map-key-default-

squat: this does not help because the splat gives us a list of maps and we first of all need to use a trick to allow evaluation of the 0th element of the list. This breaks because join and element do not work of lists of maps. Using lookup directly does not work in the case that count = 0

PR #3049, opened 2018-02-28
add bazel test suite for cli units

enxebre:
Enable cli go unit tests to be run by bazel and the cli

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -215,3 +215,11 @@ gen_test(
     deps = [":terraform_runtime", ":terraform_files"],
     size = "small",
 )
+
+test_suite(
+    name = "cli_units",
+    tests = [
+        "//installer/pkg/workflow:go_default_test",
+        "//installer/pkg/config-generator:go_default_test",
+    ],
+)
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -215,11 +215,3 @@ gen_test(
     deps = [":terraform_runtime", ":terraform_files"],
     size = "small",
 )
-
-test_suite(
-    name = "cli_units",
-    tests = [
-        "//installer/pkg/workflow:go_default_test",
-        "//installer/pkg/config-generator:go_default_test",
-    ],
-)
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -181,7 +181,7 @@ pipeline {
                 withDockerContainer(tectonicBazelImage) {
                   sh "bazel test terraform_fmt --test_output=all"
                   sh "bazel test installer/frontend:unit --test_output=all"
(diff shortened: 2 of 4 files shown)


enxebre:
retest this please

installer/pkg/config-generator/BUILD.bazel
```diff
@@ -19,3 +19,20 @@ go_library(
         "//installer/vendor/golang.org/x/crypto/bcrypt:go_default_library",
     ],
 )
+
+filegroup(
+    name = "config_generator_fixtures",
+    srcs = [
+        "fixtures/kube-system.yaml",
```
squat: Why not use a splat so that this does not have to be updated in the future when more fixtures are added? It’s also fewer LOC. Also, no one else depends on these fixtures, so do we really need a file group or can we just put a the expression `glob([“fixtures/**”])` in the `data` field of the test rule

Jenkinsfile
```diff
@@ -181,6 +181,7 @@ pipeline {
                 withDockerContainer(tectonicBazelImage) {
                   sh "bazel test terraform_fmt --test_output=all"
                   sh "bazel test installer/frontend:unit --test_output=all"
+                  sh "bazel test cli_units --test_output=all"
```
mxinden: @cpanato @squat @enxebre Do we have any tests that we don't want to run on Jenkins? If not, we could replace the `bazel test xxx` lines by `bazel test ...`. What do you think?

squat: There are tests that we do not want to run together e.g. GUI tests. We only want to run those some times. It would be better to tag the different tests and then run `bazel test` to either ignore those suites or run all PRs in a particular group, e.g. `unit`

installer/pkg/config-generator/BUILD.bazel
enxebre: Thanks a lot! still getting familiar with bazel semantics

Jenkinsfile
enxebre: Mm I like how this shows as individual drop-downs on blue ocean. Also don't have strong knowledge   but bazel test ... might run undesirable tests e.g in go vendors?

enxebre:
@mxinden @squat @cpanato happy to get this in now?

BUILD.bazel
```diff
@@ -215,3 +215,11 @@ gen_test(
     deps = [":terraform_runtime", ":terraform_files"],
     size = "small",
 )
+
+test_suite(
```
squat: the one thing I wonder is if this is the correct place for this test suite. shouldn't it go in installer? or installer/pkg?

enxebre: make sense, moved to installer package

PR #3052, opened 2018-02-28
Tectonic version bump

coverprice:
Bump the tectonic version to match the 1.8.7 branch.


coverprice:
ok to test

diegs:
odd that you have to ok to test your own PR...

config.tf
```diff
@@ -113,7 +113,7 @@ variable "tectonic_versions" {
 
   default = {
     monitoring = "1.9.3"
-    tectonic   = "1.8.4-tectonic.2"
+    tectonic   = "1.8.7-tectonic.2"
```
diegs: Really we should bump this to something that is more appropriate for the master branch. As of right now we're using Kubernetes 1.9.3, so perhaps `1.9.3-tectonic.1`?

coverprice: Done

cpanato: the version of `hyperkube` is `v1.9.1_coreos.0` does that means we are using 1.9.1 instead of 1.9.3 or am I missing something? this is always confusing to me

cpanato: @coverprice  @diegs  ⬆️

coverprice: Waiting for @diegs or someone else to weigh in on the comment above before this gets merged.

diegs: The hyperkube in the installer currently only affects the kubelets IIUC. Once the TNC is merged that will no longer be true. We should update that one to 1.9.3 as well though to match what's used by the control plane, as installed by the kube-core operator.

coverprice:
I bumped the hyperkube version to match the Tectonic version, i.e. v1.9.3_coreos.0

coverprice:
ok to test

coverprice:
retest this please

coverprice:
retest this please

coverprice:
Closing this, we can bump when the team decides to upgrade the hyperkube version.

PR #3054, opened 2018-03-01
Cluster id fix

enxebre:
Fix INST-953
Introduce buildInternalStep
- Generates a source of truth file in `cluster-name/internal.yaml` containing internal config i.e cluster_id
- readClusterConfigStep now fills `cluster.Internal` struct with the data from `cluster-name/internal.yaml`
- This data is exposed in the auto-generated tfvars file so it is made available via tf variable to the different tf steps


enxebre:
ok to test

enxebre:
retest this please

enxebre:
retest this please

installer/pkg/workflow/init.go
```diff
@@ -1,10 +1,14 @@
...
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
 
+	yaml "gopkg.in/yaml.v2"
+
+	configgenerator "github.com/coreos/tectonic-installer/installer/pkg/config-generator"
```
cpanato: why not `configGenerator` ?

installer/pkg/config/parser.go
```diff
@@ -32,12 +32,33 @@ func Parse(data []byte) (*Config, error) {
...
+
+	if err := yaml.Unmarshal(data, internal); err != nil {
+		return nil, err
+	}
+
+	return internal, nil
+}
+
+// ParseInternalFile parses a yaml file and returns, if successful, an internal.
+func ParseInternalFile(path string) (*internal, error) {
```
squat: I think the internal file should not be YAML. YAML is intended to be human readable-writable, but this is internal-only, right? No humans should be touching this ever, so I think we should serialize to some other format, e.g. JSON

installer/pkg/workflow/init_test.go
```diff
@@ -38,22 +39,94 @@ func TestGenerateTerraformVariablesStep(t *testing.T) {
...
+
+	for _, tc := range testCases {
+		match, _ := regexp.MatchString(tc.expected, tc.got)
+		if !match {
+			t.Errorf("test case %s: expected: %s, got: %s", tc.test, tc.expected, tc.got)
+		}
+	}
+
+	// clean up
+	if err := os.Remove(internalFilePath); err != nil {
```
mxinden: I would add this as a `defer` to the top, just to prevent future issues were someone adds a `t.Fatal()` before. Not very important though.

steps/variables.tf
```diff
@@ -364,3 +364,9 @@ Example:
  * `["ingress-nginx"]`
 EOF
 }
+
+// This variable is generated by tectonic internally. Do not modify
+variable "tectonic_cluster_id" {
+  type        = "string"
+  description = "The tectonic cluster id."
```
mxinden: Small thing: We have been writing `Tectonic` instead of `tectonic` in previous descriptions in this file.

installer/pkg/config/parser.go
mxinden: JSON sounds good. Argument for YAML would be, that we can add a comment at the top like `Do not touch, auto-generated`.

squat: hmm yes that's a good point. the question would be: what discourages modification more?

installer/pkg/config/types.go
```diff
@@ -49,6 +49,7 @@ type Cluster struct {
 	TLSValidityPeriod int              `yaml:"tlsValidityPeriod"`
 	VMware            vmware.Config    `yaml:"vmware,omitempty"`
 	Worker            worker           `yaml:"worker"`
+	Internal          internal         `yaml:"internal,omitempty"`
```
spangenberg: Can you make that:

```golang
  Internal `json:",inline" yaml:"internal,omitempty"`
```

I changed that in another PR which is not merged yet and this simplifies mine.

Highly appreciated 😃

installer/pkg/config/types.go
```diff
@@ -139,3 +140,7 @@ func (n nodePools) Map() map[string]int {
 	}
 	return m
 }
+
+type internal struct {
```
spangenberg: Same here:

```golang
type Internal struct {
```

installer/pkg/config/types.go
```diff
@@ -139,3 +140,7 @@ func (n nodePools) Map() map[string]int {
...
+	ClusterID string `yaml:"clusterId"`
```
spangenberg: And here:

```golang
  ClusterID string `json:"tectonic_cluster_id,omitempty" yaml:"clusterId"`
```

installer/pkg/workflow/init.go
enxebre: that's how the package is called in generator.go What's the golang convention here? cc @spangenberg

installer/pkg/config/parser.go
enxebre: That's a good point, however while I can't see any drawbacks from using yaml here, I can see the benefits of sharing and reusing the approach

installer/pkg/workflow/init.go
mxinden: @enxebre Go imports should be [all lower case](https://blog.golang.org/package-names).

mxinden: > Package names
> 
> Good package names are short and clear. They are lower case, with no under_scores or mixedCaps. They are often simple nouns, such as:
> 
>   -  time (provides functionality for measuring and displaying time)
>   -  list (implements a doubly linked list)
>   -  http (provides HTTP client and server implementations)
> 
> The style of names typical of another language might not be idiomatic in a Go program. Here are two examples of names that might be good style in other languages but do not fit well in Go:
> 
>   - computeServiceClient
>   - priority_queue
> 
> A Go package may export several types and functions. For example, a compute package could export a Client type with methods for using the service as well as functions for partitioning a compute task across several clients.
> 
> Abbreviate judiciously. Package names may be abbreviated when the abbreviation is familiar to the programmer. Widely-used packages often have compressed names:
> 
>   - strconv (string conversion)
>   - syscall (system call)
>   - fmt (formatted I/O)
> 
> On the other hand, if abbreviating a package name makes it ambiguous or unclear, don't do it.
> 
> Don't steal good names from the user. Avoid giving a package a name that is commonly used in client code. For example, the buffered I/O package is called bufio, not buf, since buf is a good variable name for a buffer.

installer/pkg/config/parser.go
enxebre: @mxinden @squat added the "Do not touch, auto-generated" message but happy to discuss more about this and concerns about state

enxebre:
rebased PTAL cc @mxinden @squat @spangenberg

enxebre:
retest this please

installer/pkg/config/cluster.go
```diff
@@ -40,6 +40,7 @@ type Cluster struct {
 	metal.Metal         `json:",inline" yaml:"metal,omitempty"`
 	openstack.OpenStack `json:",inline" yaml:"openstack,omitempty"`
 	vmware.VMware       `json:",inline" yaml:"vmware,omitempty"`
+	Internal            `json:",inline" yaml:"internal,omitempty"`
```
mxinden: As this should never appear in the yaml config, how about adding `json:"-"` and `yaml:"-"` here? Just in case we ever read and write a config yaml in the future.

installer/pkg/config/types.go
```diff
@@ -96,3 +96,7 @@ type Worker struct {
 	Count     int      `json:"tectonic_worker_count,omitempty" yaml:"-"`
 	NodePools []string `json:"-" yaml:"nodePools"`
 }
+
+type Internal struct {
+	ClusterID string `json:"tectonic_cluster_id,omitempty" yaml:"clusterId"`
```
spangenberg: can you omitempty clusterID
Otherwise this will make converting from tfvars weird.

installer/pkg/config/cluster.go
enxebre: I put back `json:",inline" as is needed by the tf vars generation

PR #3061, opened 2018-03-02
tests: Optionally trigger pipeline with specific git ref

mxinden:
In order for the release automation pipeline to consume the results of
the Tectonic installer pipeline, we have to ensure, that both are
configured with the same Tectonic installer git ref (commit).

This patch adds a Jenkins pipeline parameter to the Tectonic installer
pipeline to trigger a build with a specific git ref.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -154,9 +154,14 @@ pipeline {
       description: 'Slack channel to notify on failure.'
     )
     string(
-      name : 'GITHUB_REPO',
+      name: 'GITHUB_REPO',
       defaultValue: 'coreos/tectonic-installer',
       description: 'Github repository'
+    )
+    string(
+      name: 'SPECIFIC_GIT_COMMIT',
+      description: 'Checkout a specific git ref (e.g. sha or tag). If not set, Jenkins uses the most recent commit of the triggered branch.',
+      defaultValue: '',
     )
   }
 
@@ -173,10 +178,27 @@ pipeline {
             try {
               timeout(time: 20, unit: 'MINUTES') {
                 forcefullyCleanWorkspace()
-                checkout scm
-                stash name: 'clean-repo', excludes: 'installer/vendor/**,tests/smoke/vendor/**'
-                originalCommitId = sh(returnStdout: true, script: 'git rev-parse origin/"\${BRANCH_NAME}"').trim()
+
+                /*
+                  This supports users who require builds at a specific git ref
+                  instead of the branch tip.
+                */
+                if (params.SPECIFIC_GIT_COMMIT == '') {
+                  checkout scm
+                  originalCommitId = sh(returnStdout: true, script: 'git rev-parse origin/"\${BRANCH_NAME}"').trim()
+                } else {
+                  checkout([
+                    $class: 'GitSCM',
+                    branches: [[name: params.SPECIFIC_GIT_COMMIT]],
+                    userRemoteConfigs: [[url: "https://github.com/${params.GITHUB_REPO}.git"]]
+                  ])
+                  // In case params.SPECIFIC_GIT_COMMIT is a mutable tag instead
+                  // of a sha
+                  originalCommitId = sh(returnStdout: true, script: "git rev-parse ${params.SPECIFIC_GIT_COMMIT}").trim()
+                }
+
                 echo "originalCommitId: ${originalCommitId}"
+                stash name: 'clean-repo', excludes: 'installer/vendor/**,tests/smoke/vendor/**'
 
                 withDockerContainer(tectonicBazelImage) {
                   sh "bazel test terraform_fmt --test_output=all"


Jenkinsfile
```diff
@@ -154,10 +154,15 @@ pipeline {
...
     )
     string(
-      name : 'GITHUB_REPO',
+      name: 'GITHUB_REPO',
       defaultValue: 'coreos/tectonic-installer',
       description: 'Github repository'
     )
+    string(
+      name: 'SPECIFIC_GIT_COMMIT',
+      description: 'Checkout a specific git commit. If not set, Jenkins uses the most recent commit of the triggered branch.',
```
coverprice: Nit: "Checkout a specific git ref **(e.g. sha or tag)**. If not set, ..."

Jenkinsfile
```diff
@@ -173,10 +178,28 @@ pipeline {
...
+                } else {
+                  checkout([
+                    $class: 'GitSCM',
+                    branches: [[name: params.SPECIFIC_GIT_COMMIT]],
+                    userRemoteConfigs: [[url: "https://github.com/${params.GITHUB_REPO}.git"]]
+                  ])
+                  originalCommitId = params.SPECIFIC_GIT_COMMIT
+                }
+
                 echo "originalCommitId: ${originalCommitId}"
```
coverprice: If it doesn't exist, can Jenkins add a "build_metadata.txt" file to the artifact that contains the following?
```
- Git repo: github.com/${params.GITHUB_REPO}
- Git SHA: ${originalCommitId}
[If built from branch tip] - Branch: ${BRANCH_NAME}
[otherwise] - Git ref: ${params.SPECIFIC_GIT_COMMIT}
```

I think this would be useful when determining the provenance of an artifact bundle. E.g. when the RA pipeline pulls the bundle, it can check that the artifact's build SHA is the SHA it asked for.

Jenkinsfile
```diff
@@ -173,10 +178,28 @@ pipeline {
...
+                if (params.SPECIFIC_GIT_COMMIT == '') {
+                  checkout scm
+                  originalCommitId = sh(returnStdout: true, script: 'git rev-parse origin/"\${BRANCH_NAME}"').trim()
+                } else {
+                  checkout([
+                    $class: 'GitSCM',
+                    branches: [[name: params.SPECIFIC_GIT_COMMIT]],
+                    userRemoteConfigs: [[url: "https://github.com/${params.GITHUB_REPO}.git"]]
+                  ])
+                  originalCommitId = params.SPECIFIC_GIT_COMMIT
```
coverprice: It's possible for a user to enter a tag, and unfortunately the SHA that a tag points to is mutable. So I would prefer if originaCommitId was actually the SHA. i.e.
`originalCommitId = sh(returnStdout: true, script: "git rev-parse ${params.SPECIFIC_GIT_COMMIT}").trim()`

However the above code is not ideal because it uses unparsed user input directly in a shell command. I'd much prefer it if it was escaped, but that apparently is not easily done. A workaround is described a the bottom of this issue: https://issues.jenkins-ci.org/browse/JENKINS-44231, so perhaps something like:

```
def shellEscape(s) {
    '\'' + s.replace('\'', '\'\\\'\'') + '\''
}
originalCommitId = sh(returnStdout: true, script: "git rev-parse " + shellEscape(params.SPECIFIC_GIT_COMMIT)).trim()
```

Jenkinsfile
mxinden: I guess embedding the commit sha could be a general feature of our Bazel build process. I will suggest it in the installer team.

Jenkinsfile
```diff
@@ -173,10 +178,28 @@ pipeline {
...
-                originalCommitId = sh(returnStdout: true, script: 'git rev-parse origin/"\${BRANCH_NAME}"').trim()
+
+                /*
+                  The Tectonic release automation pipeline consumes the build
+                  result (tarball) of the Tectonic installer pipeline. Triggering
+                  the Tectonic installer pipeline by branch name can result in
+                  race conditions. In this case triggering by specific git commit is
+                  necessary.
+                */
+                if (params.SPECIFIC_GIT_COMMIT == '') {
```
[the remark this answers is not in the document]
mxinden: I will follow up on this with https://github.com/coreos/tectonic-installer/pull/3069

Good point.

Jenkinsfile
```diff
@@ -173,10 +178,28 @@ pipeline {
             try {
               timeout(time: 20, unit: 'MINUTES') {
                 forcefullyCleanWorkspace()
-                checkout scm
-                stash name: 'clean-repo', excludes: 'installer/vendor/**,tests/smoke/vendor/**'
-                originalCommitId = sh(returnStdout: true, script: 'git rev-parse origin/"\${BRANCH_NAME}"').trim()
+
+                /*
+                  The Tectonic release automation pipeline consumes the build
```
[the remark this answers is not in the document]
mxinden: :+1:

Jenkinsfile
mxinden: :+1:

Jenkinsfile
mxinden: Tag to sha conversion is a great idea. I have added that.

I think untrusted user input validation and escaping should be considered as a general improvement to the pipeline, not particularly scoped to this PR. If you don't mind I will create a follow up task for this.

mxinden:
@coverprice Thanks for the constructive feedback! Would you mind taking another look?

Jenkinsfile
coverprice: @mxinden We'll agree to disagree there; My feeling is that style + security concerns are fundamentally part of any PR, unless they overwhelm its original purpose. Leaving security concerns as techdebt doesn't guarantee they'll ever be addressed. However I think the risk is fairly low in this case so I won't press the point.

mxinden: Fair point. See the direct follow up PR here: https://github.com/coreos/tectonic-installer/pull/3076

PR #3062, opened 2018-03-02
tests: Wrap entire first stage with ansiColor tag

mxinden:
Bazel also prints out color codes. This patch extends the scope of the
ansiColor function to the entire first stage.


mxinden:
@squat Thanks for the review.

Let's wait for @cpanato to review as well. I remember he spotted some issues with the `ansiColor` feature in the past.

PR #3064, opened 2018-03-02
tests: Adjust log-analyzer-copy to missing build folder

mxinden:
With the new Tectonic cli flow, the cluster build folder does not
exist anymore. Instead the cli creates a folder on the top level with
the name of the cluster, containing all relevant files.

For now we need to support both flows. For AWS the Tectonic CLI is used.
For Azure and all other platforms we still use plain terraform.

Similar to #3063

### DIFF
--- a/tests/jenkins-jobs/scripts/log-analyzer-copy.sh
+++ b/tests/jenkins-jobs/scripts/log-analyzer-copy.sh
@@ -102,14 +102,30 @@ case "${action}" in
     fields[platform]=$(echo -n "${testFilePath}" | cut -d'/' -f2)
     fields[specName]=$(echo -n "${testFilePath}" | cut -d'/' -f3 | sed -e 's/_spec\.rb//')
 
-    # Checks whether there are any log files
-    if ! compgen -G ../build/*/*.log >/dev/null ; then
-      # Nothing to do
-      exit 0
+    logfiles_found="false"
+
+    # Old _terraform_ workflow copies logs to `build/<cluster_name>/*.log`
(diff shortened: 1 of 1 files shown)


tests/jenkins-jobs/scripts/log-analyzer-copy.sh
```diff
@@ -102,14 +102,27 @@ case "${action}" in
     fields[platform]=$(echo -n "${testFilePath}" | cut -d'/' -f2)
     fields[specName]=$(echo -n "${testFilePath}" | cut -d'/' -f3 | sed -e 's/_spec\.rb//')
 
+    # Old _terraform_ workflow copies logs to `build/<cluster_name>/*.log`
     # Checks whether there are any log files
     if ! compgen -G ../build/*/*.log >/dev/null ; then
```
coverprice: It seems that if the new workflow is used, then this check for the old workflow will fail and the script will prematurely exit on the next line, so the new workflow's files won't be uploaded. Is that correct? The script should only exit if files from neither location were found.

coverprice: If the new workflow is used but the old one is not, then wouldn't this check fail to find any old workflow build logs and then exit, so the new workflow files would not be uploaded? It seems to me that the script should only exit if no files are found from either old/new workflows.

mxinden:
@coverprice You are absolutely right, sorry about that. Can you have another look?

coverprice:
@mxinden Looks like the errors are transient, but I'm not too familiar with those tests. I'm ok with merging if you are.

PR #3068, opened 2018-03-05
installer: Remove installer/Makefile

kyoto:
All necessary functionality from the Makefile should now have been ported to Bazel.


kyoto:
ok to test

PR #3069, opened 2018-03-05
tests: Trim all Jenkins string parameters in pipeline

mxinden:
Once we run update to a Jenkins version with https://github.com/jenkinsci/jenkins/pull/3106 we can use the trim option in all our string Jenkins parameters.


sym3tri:
closing b/c stale

PR #3072, opened 2018-03-05
aws: added `worker_iam_role_policy_attachments`

Dominik-K:
The deletion of ECR pull permissions in #2745 will be a big breaking change for running clusters. To avoid that I added the new, generic variable `tectonic_aws_worker_iam_role_policy_attachments` to highlight that change and offer a migration. 
I added it only to the workers as (perhaps) others and we will only run own images on workers.

@cpanato Can you review and add the necessary labels, please?


sym3tri:
Thanks for the contribution & apologies for the delay. Since we are converging Tectonic with Open Shift we will not be adding any other features to this repo.

See our blog for more details: https://coreos.com/blog/coreos-tech-to-combine-with-red-hat-openshift

krish7919:
I don't know if you are still tracking this, but the current way to do it would be to create your own role and attach it to the instances using `tectonic_aws_master_iam_role_name` and `tectonic_aws_worker_iam_role_name`.

PR #3073, opened 2018-03-05
Retrieve Jenkins console logs without verifying SSL cert

coverprice:
To support development of this Jenkinsfile on other instances that may not have a valid SSL cert, we retrieve the consoleText with the `--insecure` flag.

### DIFF
--- a/tests/jenkins-jobs/scripts/log-analyzer-copy.sh
+++ b/tests/jenkins-jobs/scripts/log-analyzer-copy.sh
@@ -113,7 +113,8 @@ case "${action}" in
     ;;
 
   "jenkins-logs")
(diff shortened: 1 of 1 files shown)


coverprice:
ok to test

coverprice:
@mxinden I agree. If there is an easy way to detect prod vs test, then the `--insecure` flag could be conditional. I'd check the hostname but that seems a little fragile.

cpanato:
@coverprice @mxinden you can have (another 😄 ) Jenkins parameters that set that, like a boolean. but it is one more parameter to take care :/ .

I dont have a strong opinion about this, and also our instances are acessible only if you loginm then I guess we are good

sym3tri:
@coverprice  status on this?

sym3tri:
@coverprice closing b/c stale. If still needed please resubmit (probably against track-1).

PR #3078, opened 2018-03-07
frontend: Fix AWS custom subnet validation bug when AWS region changes

kyoto:
Change the action to always overwrite the AWS subnets so that the subnet keys (which include the AWS AZs) are updated when the AWS region is changed.

The bug was introduced by #2620, so this PR partially reverts #2620. Unfortunately, this brings back the original bug where any user entered subnet CIDRs are overwritten if the page is refreshed. Fixing that will require a larger code change.


kyoto:
ok to test

PR #3079, opened 2018-03-07
Etcd tnc

enxebre:
Bootstrap etcd nodes from TNC. Fix INST-965
Currently the TNC etcd ign config is missing:
- etcd-memeber sevice (currently the ign config generated by the installer is different for each machine. We'll need to either unify them or have different TNC roles). 
- locksmith service

if TNC can serve the relevant services with the relevant domains for `initial_advertise_peer_urls`, `advertise_client_urls`, `initial cluster`, etc. via `ign/v1/role/etcd/N` (e.g `ign/v1/role/etcd/1`, `ign/v1/role/etcd/2`) that should work


enxebre:
cc @thorfour @alexsomesan @squat

thorfour:
TNC can serve an individual etcd at `ign/v1/role/etcdN` would that work? Or does it need to be `ign/v1/role/etcd/N` ?

installer/pkg/workflow/install.go
```diff
@@ -61,11 +61,13 @@ func installBootstrapStep(m *metadata) error {
 		return err
 	}
 
-	if err := waitForTNC(m); err != nil {
+	destroyCNAME(m.clusterDir)
+
```
squat: Ignition does not continuously retry to download forever, so I think this introduces a race. What if one of the etcd machines gets provisioned and tries to load it’s ignition before the TNC is up?

I think this should go after the waitForTNC step.

enxebre: yes there's a race which is currently only relying on retries. The problem here is that there's no obvious way for etcd to waitForTNC (when this is running as static pod) as the kubeclient won't be able to get answers until the server gets the state

squat:
Where are the retries? The retry logic would have to be *in* ignition for
this to work IMO

enxebre:
yeah I was referring to the ignition retries https://github.com/coreos/ignition/blob/master/internal/resource/http.go#L200

installer/pkg/workflow/install.go
```diff
@@ -61,11 +61,13 @@ func installBootstrapStep(m *metadata) error {
...
+	if err := runInstallStep(m.clusterDir, etcdStep); err != nil {
 		return err
 	}
 
-	return destroyCNAME(m.clusterDir)
+	return waitForTNC(m)
```
enxebre: By having it here we ensure that the step holds until etcd comes up (so the api server actually gives a response and waitForTNC finish when the TNC daemonset gets deployed)

thorfour:
@enxebre I was told to hold off on the etcd changes until the installer supports query strings. Is there a current issue that tracks that required work?

enxebre:
hey @thorfour thanks for the update! I can't see how cnames vs any other approach can block us from getting the TNC to serve the expected services regardless how they are requested ( e.g etcd, https://github.com/coreos-inc/tectonic-operators/pull/286#issuecomment-370883510, https://github.com/coreos/tectonic-installer/pull/3090#discussion_r173594526, is there a place where we track all these expected features for the tnc?). Also, although we need agree on an specific implementation (see INST-944) there's nothing stopping this PR from requesting via query strings so we can ensure etcd comes up correctly

thorfour:
@enxebre Sounds good. I've implemented query strings for ignition using a separate endpoint to unblock us. To use query strings we'll use the path `/ignition?role=etcd&id=N` sound good? I'll push that PR tomorrow and then add the etcd templates to see if we can get this working.

enxebre:
Cool thanks, sounds good. Just make sure you tag the docker image to a different id so we keep builds used by master hermetic

enxebre:
@thorfour any update on this, is there an image serving etcd services I can point to?

thorfour:
@enxebre PR with the etcd template is [here](https://github.com/coreos-inc/tectonic-operators/pull/307) 

I've pushed an image with that template here: `quay.io/coreos/tectonic-node-controller-dev:16fb00200ea86e6f31ce78e56225dec01f459921`

yifan-gu:
Any updates on this PR?

enxebre:
@yifan-gu I dropped few comments here https://github.com/coreos-inc/tectonic-operators/pull/307, should get clarified and merging soon

installer/pkg/workflow/install.go
alexsomesan: I second @squat's concern about the race condition. If we know there could be one, we should try to find a solution.
I don't have enough insight into the matter right now to suggest one here, but let me know if you need my help looking into it.

enxebre: @alexsomesan @squat this relies on ignition retries https://github.com/coreos/ignition/blob/master/internal/resource/http.go#L200 which will keep retrying until it times out, and if I'm not wrong by the look of `httpTotal` here https://coreos.com/ignition/docs/latest/configuration-v2_1.html it will keep retrying for ever by default

enxebre: @alexsomesan @squat it's verified that Ignition will never continue if it isn't completely successful.
`waitForTNC(m)` where it is now, will ensure that the step holds until etcd comes up (so the cluster get state, so the TNC daemonset gets deployed and so the api server actually gives a response, so waitForTNC can finish)
This is no strictly necessary as the every node will be able to get its config from the TNC pod, so the question is do we still want to `waitForTNC(m)` wdyt?

enxebre:
ok to test

config.tf
```diff
@@ -71,7 +71,7 @@ variable "tectonic_container_images" {
     awscli                               = "quay.io/coreos/awscli:025a357f05242fdad6a81e8a6b520098aa65a600"
     gcloudsdk                            = "google/cloud-sdk:178.0.0-alpine"
     bootkube                             = "quay.io/coreos/bootkube:v0.10.0"
-    tnc_bootstrap                        = "quay.io/coreos/tectonic-node-controller-dev:76a584680b7f39aa7b3c40cd742c736b30b5a89a"
+    tnc_bootstrap                        = "quay.io/alberto_lamela/tectonic-node-controller-dev:7092e1772378470e14d00b11198767edf28f9698-dirty"
```
yifan-gu: quay.io/coreos/tectonic-node-controller-bootstrap-dev:f6d5e710a97a8cd6f4cd2963f4426131f854a869

enxebre:
retest this please

modules/aws/etcd/ignition.tf
```diff
@@ -2,10 +2,17 @@ locals {
   ignition_etcd_keys = ["ignition_etcd_0.json", "ignition_etcd_1.json", "ignition_etcd_2.json"]
 }
 
-data "ignition_config" "s3" {
+data "ignition_config" "tnc" {
   count = "${length(var.external_endpoints) == 0 ? var.instance_count : 0}"
 
-  replace {
+  append {
+    source = "${format("http://${var.cluster_name}-tnc.${var.base_domain}/ignition?role=etcd&etcd_index=%d", count.index)}"
```
squat: :p it’s a little funny doing both interpolation and formatting but not a blocker

steps/etcd/inputs.tf
```diff
@@ -0,0 +1,28 @@
...
+
+  config {
+    path = "${path.cwd}/assets.tfstate"
+  }
+}
+
+locals {
+  container_linux_version = "${data.terraform_remote_state.bootstrap.container_linux_version}"
+  instance_count          = "${data.terraform_remote_state.bootstrap.etcd_instance_count}"
+  ignition_etcd           = "${data.terraform_remote_state.assets.ignition_etcd}"
```
squat: Just like instance_count, this could be just `ignition` since we are in the etcd step and know that everything is etcd-related

squat: Also for sanity we should standardize the output names to either be `prefix_value` or `value_suffix` but not both. That can be a separate cleanup pr

enxebre:
@squat thanks! I agree on your three points, I'll address that on a follow up as the append block is being moved into the cli any way after we merge this and https://github.com/coreos/tectonic-installer/pull/3120

PR #3085, opened 2018-03-08
Add validate Go package

kyoto:
Based on the frontend validators from [frontend/validate.js](https://github.com/coreos/tectonic-installer/blob/master/installer/frontend/validate.js) and [frontend/fields.js](https://github.com/coreos/tectonic-installer/blob/master/installer/frontend/fields.js).

### DIFF
--- a/installer/pkg/validate/validate.go
+++ b/installer/pkg/validate/validate.go
@@ -1,0 +1,364 @@
+package validate
+
+import (
+	"errors"
+	"fmt"
+	"net"
+	"regexp"
+	"strconv"
+	"strings"
+	"unicode/utf8"
+)
+
+func isMatch(re string, v string) bool {
+	return regexp.MustCompile(re).MatchString(v)
+}
+
+// NonEmpty checks if the given string contains at least one non-whitespace character and returns an error if not.
+func NonEmpty(v string) error {
+	if utf8.RuneCountInString(strings.TrimSpace(v)) == 0 {
+		return errors.New("cannot be empty")
+	}
+	return nil
+}
+
+// Int checks if the given string is a valid integer and returns an error if not.
+func Int(v string) error {
+	if err := NonEmpty(v); err != nil {
+		return err
+	}
+
+	if _, err := strconv.Atoi(v); err != nil {
+		return errors.New("invalid integer")
+	}
+	return nil
+}
+
+// IntRange checks if the given string is a valid integer between `min` and `max` and returns an error if not.
+func IntRange(v string, min int, max int) error {
+	i, err := strconv.Atoi(v)
+	if err != nil {
+		return Int(v)
+	}
+	if i < min {
+		return fmt.Errorf("cannot be less than %v", min)
+	}
+	if i > max {
+		return fmt.Errorf("cannot be greater than %v", max)
+	}
+	return nil
+}
+
+// IntOdd checks if the given string is a valid integer and that it is odd and returns an error if not.
+func IntOdd(v string) error {
+	i, err := strconv.Atoi(v)
+	if err != nil {
+		return Int(v)
+	}
+	if i%2 != 1 {
+		return errors.New("must be an odd integer")
+	}
+	return nil
+}
+
+// ClusterName checks if the given string is a valid name for a cluster and returns an error if not.
+func ClusterName(v string) error {
+	if err := NonEmpty(v); err != nil {
+		return err
+	}
+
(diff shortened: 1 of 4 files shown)


kyoto:
ok to test

installer/pkg/validate/validate.go
```diff
@@ -0,0 +1,360 @@
...
+	return nil
+}
+
+// IPv4 checks if the given string is a valid IP v4 address and returns an error if not.
+// Based on net.ParseIP.
+func IPv4(v string) error {
+	if err := NonEmpty(v); err != nil {
+		return err
+	}
+	if ip := net.ParseIP(v); ip == nil || !strings.Contains(v, ".") {
```
squat: this is to distinguish between ipv4 and ipv6, ya?

kyoto: yes

installer/pkg/validate/validate.go
```diff
@@ -0,0 +1,360 @@
...
+	return nil
+}
+
+// AWSSubnetCIDR checks if the given string is a valid CIDR for a master nodes or worker nodes subnet in an AWS VPC and returns an error if not.
+func AWSSubnetCIDR(v string) error {
+	if err := SubnetCIDR(v); err != nil {
+		return err
+	}
+
+	if mask, err := strconv.Atoi(strings.Split(v, "/")[1]); err != nil || mask < 16 || mask > 28 {
```
squat: this depends implicitly on the split slice having at least two elements. we know that this should be the case since we have validated the cidr already.

Instead of splitting and parsing the string, I think it would be better to:
```go
_, c, err := net.ParseCIDR(v)
if err != nil {
    return errors.New("not a valid CIDR")
}
if _, mask = c.Mask.Size(); mask < 16 || mask > 28 {
    ...
}
```

installer/pkg/validate/validate.go
```diff
@@ -0,0 +1,360 @@
...
+			continue
+		}
+		if !isMatch("^[a-zA-Z0-9-]{1,63}$", segment) {
+			return errors.New("invalid domain name")
+		}
+	}
+	return nil
+}
+
+// Host checks if the given string is either a valid v4 IP address or a valid domain name and returns an error if not.
```
squat: we normally just say IPv4

installer/pkg/validate/validate_test.go
```diff
@@ -0,0 +1,438 @@
...
+const caseMsg = "must be lower case"
+const emptyMsg = "cannot be empty"
+const invalidDomainMsg = "invalid domain name"
+const invalidHostMsg = "invalid host (must be a domain name or IP address)"
+const invalidIPMsg = "invalid IPv4 address"
+const invalidIntMsg = "invalid integer"
+const invalidPortMsg = "invalid port number"
+const noCIDRNetmaskMsg = "must provide a CIDR netmask (eg, /24)"
+
+type Test struct {
```
squat: should not be an exported type

installer/pkg/validate/validate_test.go
```diff
@@ -0,0 +1,438 @@
...
+		{"5", 2, 4, "cannot be greater than 4"},
+	}
+
+	for _, test := range tests {
+		err := IntRange(test.in, test.min, test.max)
+		if (err == nil && test.expected != "") || (err != nil && err.Error() != test.expected) {
+			t.Errorf("For IntRange(%q, %v, %v), expected %q, got %q", test.in, test.min, test.max, test.expected, err)
+		}
+	}
+
```
squat: extra newline here

installer/pkg/validate/validate.go
kyoto: 👍

installer/pkg/validate/validate.go
kyoto: 👍

installer/pkg/validate/validate_test.go
kyoto: 👍

installer/pkg/validate/validate.go
```diff
@@ -0,0 +1,360 @@
...
+	}
+	return nil
+}
+
+// Port checks if the given string is a valid port number and returns an error if not.
+func Port(v string) error {
+	if err := NonEmpty(v); err != nil {
+		return err
+	}
+	if i, err := strconv.Atoi(v); err != nil || i < 1 || i > 65535 {
```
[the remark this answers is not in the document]
kyoto: 👍

PR #3086, opened 2018-03-08
Introduce convert workflow

spangenberg:
Removing terraform-generator
Transformation is happening in config only
Convert command can parse tfvars files to new config


installer/cmd/tectonic/main.go
```diff
@@ -21,6 +21,9 @@ var (
 
 	clusterDestroyCommand = kingpin.Command("destroy", "Destroy an existing Tectonic cluster")
 	clusterDestroyDirFlag = clusterDestroyCommand.Flag("dir", "Cluster directory").Default(".").ExistingDir()
+
+	convertCommand    = kingpin.Command("convert", "Converts tfvars.json to config.yaml")
```
mxinden: Can you adjust the description to match the style of the destroy command? E.g.: "Convert a tfvars.json to a Tectonic config.yaml"

spangenberg:
ok to test

spangenberg:
ok to test

spangenberg:
ok to test

installer/pkg/config/aws/aws.go
```diff
@@ -0,0 +1,33 @@
+package aws
+
+import (
+	"github.com/coreos/tectonic-installer/installer/pkg/config/aws/etcd"
```
squat: Why are we modeling each of these fields as its own package? This seems like a lot of overhead to me. Some of these fields have identically named sub-fields that need to marshal to different names, but we can just name these types differently, e.g. RootVolumeMaster, RootVolumeEtcd, etcd

spangenberg: That was needed for the embedding, they name clash with other providers. This was the only solution.

spangenberg: The idea behind that was that the workflows can still access the structs the same way they're references in the config yml. We used to do that two weeks ago, but I can't see anything anymore that does that. I'll rename it.

installer/pkg/workflow/convert.go
```diff
@@ -0,0 +1,50 @@
...
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+
+	"github.com/coreos/tectonic-installer/installer/pkg/config"
+)
+
+// NewInitWorkflow creates new instances of the 'init' workflow,
+// responsible for initializing a new cluster.
+func NewConvertWorkflow(configFilePath string) Workflow {
```
enxebre: can we add a unit for NewConvertWorkflow?

spangenberg: Sure

tests/smoke/aws/vars/aws-basic.yaml
```diff
@@ -0,0 +1,45 @@
...
+    vpcCIDRBlock: 10.0.0.0/16
+    worker:
+      ec2Type: m4.large
+      rootVolume:
+        size: 32
+        type: gp2
+  baseDomain: tectonic-ci.de
+  containerLinux:
+    channel: beta
+  dnsName: aws-basic
```
mxinden: Please remove this line. Tectonic installer currently fails to bootstrap a cluster when cluster_name and dns_name differ. The test framework creates a random cluster name. I will report this issue as a separate one.

installer/pkg/config/aws/aws.go
```diff
@@ -0,0 +1,68 @@
+package aws
+
+type AWS struct {
```
squat: We should really use proper Godoc-style comments for all exported types. Otherwise, linting will be very sad :crying_cat_face:

installer/pkg/workflow/convert.go
```diff
@@ -0,0 +1,50 @@
...
+	config := config.Config{
+		Clusters: []config.Cluster{m.cluster},
+	}
+
+	yaml, err := config.YAML()
+	if err != nil {
+		return err
+	}
+
+	fmt.Println(yaml)
```
squat: Is this a leftover debug line?

spangenberg: No actually thats the thing that prints the revered config to your terminal.

tests/rspec/lib/config_file.rb
```diff
@@ -90,14 +90,15 @@ def pull_secret
...
+    new_data['clusters'][0]['admin'] ||= {}
     new_data['clusters'][0]['admin']['email'] = admin_email
     new_data['clusters'][0]['admin']['password'] = admin_passwd
     save(new_data)
   end
 
   def admin_credentials
-    admin_email = data['clusters'][0]['admin']['email']
-    admin_passwd = data['clusters'][0]['admin']['password']
+    admin_email = data.dig('clusters', 0, 'admin', 'email')
```
mxinden: Whats the benefit of traversing via `dig`?

installer/pkg/workflow/convert.go
mxinden: To address Lucas confusion, you could rename this function to `printYAMLConfigStep`.

tests/rspec/lib/config_file.rb
spangenberg: If admin is not set for example it doesn't explode.
nil doesn't have [].

mxinden: cool :+1:

PR #3088, opened 2018-03-08
installer,tests: stub out node pools

squat:
This commit adds very basic support for node pools. The only supported
node pools at the moment are:

- master
- worker
- etcd

Furthermore, the master and etcd node pools must be specified otherwise
the cluster will not build. This constraint is enforced with validation.

Fixes: INST-954

cc @spangenberg @alexsomesan


installer/pkg/config-generator/generator.go
```diff
@@ -19,6 +17,8 @@ import (
 	"github.com/ghodss/yaml"
 	"golang.org/x/crypto/bcrypt"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/coreos/tectonic-installer/installer/pkg/config"
```
squat: our imports were ordered incorrectly in several places

installer/pkg/config-generator/generator_test.go
```diff
@@ -65,16 +64,14 @@ func TestGetEtcdServersUrls(t *testing.T) {
 	for _, tc := range testCases {
 
 		config := initConfig(t, tc.configFile)
-		got := config.getEtcdServersUrls()
+		got := config.getEtcdServersURLs()
```
squat: some type names did not follow casing conventions

installer/pkg/config/parser.go
```diff
@@ -9,16 +9,15 @@ import (
 
 // Error codes returned by failures to parse a config.
 var (
-	ErrMultipleClusters = errors.New("Multiple cluster configurations are not supported at the moment")
-	ErrNoClusters       = errors.New("No clusters were defined")
+	ErrMultipleClusters = errors.New("multiple cluster configurations are not currently supported")
```
squat: Errors should not start with a capital letter to enable better legibility when wrapping

installer/pkg/config/validate.go
```diff
@@ -0,0 +1,106 @@
...
+	name   string
+	fields []string
+}
+
+// Error implements the error interface.
+func (e *ErrSharedNodePool) Error() string {
+	return fmt.Sprintf("node pools cannot be shared, but %q is used by %s", e.name, strings.Join(e.fields, ", "))
+}
+
+// Validate ensures that the Cluster is symantically correct and returns an error if not.
```
enxebre: typo semantically?

tests/smoke/aws/vars/aws.basic.yaml
```diff
@@ -29,18 +29,25 @@ clusters:
...
     master:
-      count: 2
+      nodePool: master
     name: aws-basic
     networking:
       mtu: 1480
       podCIDR: 10.2.0.0/16
       serviceCIDR: 10.3.0.0/16
       type: canal
+    nodePools:
```
enxebre: curious if you considered:
```yaml
nodePools:
  foo:
   - count: 3
  bar:
   - count: 1
```
and then unmarshal directly into a map

enxebre:
looks cool to me

tests/smoke/aws/vars/aws.basic.yaml
squat: Yes definitely considered. I decided against it because it does not follow the patterns in kubernetes manifests. This is more familiar. Plus, writing yaml is hard enough, and if arbitrary keys are allowed, it is harder to read IMO. last, modelling as map keys obfuscates the fact that any value is allowed for the name, not just certain “special” keys

installer/pkg/config/validate.go
squat: Thanks good catch

installer/pkg/config/validate.go
```diff
@@ -0,0 +1,106 @@
...
+	if err := c.validateSharedNodePool(); err != nil {
+		errs = append(errs, err)
+	}
+
+	return errs
+}
+
+func (c *Cluster) validateSharedNodePool() error {
+	var name string
+	fields := make(map[string]struct{})
```
mxinden: Why use a map with empty structs here, instead of a slice of strings?

installer/pkg/workflow/init_test.go
```diff
@@ -9,20 +11,26 @@ import (
 	"github.com/coreos/tectonic-installer/installer/pkg/config"
 )
 
-func initTestCluster(t *testing.T, file string) config.Cluster {
+func initTestCluster(file string) (*config.Cluster, error) {
 	testConfig, err := config.ParseFile(file)
 	if err != nil {
-		t.Errorf("Test case TestGenerateTerraformVariablesStep: failed to parse test config, %s", err)
+		return nil, fmt.Errorf("failed to parse test config, %s", err)
```
mxinden: Just a small nit: In the Prometheus project and as well with `https://github.com/pkg/errors`, the convention is to chain errors with colons and not commas. What do you think?

installer/cmd/tectonic/main.go
```diff
@@ -21,6 +22,8 @@ var (
 
 	clusterDestroyCommand = kingpin.Command("destroy", "Destroy an existing Tectonic cluster")
 	clusterDestroyDirFlag = clusterDestroyCommand.Flag("dir", "Cluster directory").Default(".").ExistingDir()
+
+	logLevel = kingpin.Flag("log-level", "log level (e.g. \"debug\")").Default("warn").Enum("debug", "info", "warn", "error", "fatal", "panic")
```
mxinden: Is there a way of only using a subset of the error levels, or do we have to buy in on all the levels of sirupsen/logrus?

installer/pkg/config-generator/generator_test.go
mxinden: I guess we should switch to `ETCD` as well, right?

installer/pkg/config/validate.go
```diff
@@ -0,0 +1,106 @@
...
+func (c *Cluster) Validate() []error {
+	var errs []error
+	errs = append(errs, c.validateNodePools()...)
+	return errs
+}
+
+func (c *Cluster) validateNodePools() []error {
+	var errs []error
+	n := c.NodePools.Map()
+	if c.Master.NodePool == "" {
```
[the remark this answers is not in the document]
squat: yeah we can do, but note that it needs to be an array of structs so that we can have a `NodePool` field, and also a string name for the error if the `NodePool` field is empty

installer/pkg/config/validate.go
squat: Imagine if all of your nodepools have the same name, then if you look for dupes you would get a list that says: `["master", "worker", "master", etcd", "worker", "etcd"]`. Using a map is an easy way to de-dupe this,

installer/pkg/workflow/init_test.go
squat: I agree, this is our convention everywher; this is carried over from before, as you can see. The change here was intended to fix the formatting of the error to not contain the extra prefix, but the mistake at the end slipped. thanks for catching

installer/cmd/tectonic/main.go
squat: Yes, we can force users to choose a subset by eliminating choices from the enum, but I don't see too much value in that at the moment. these verbosity choices are very common, IMO. In fact, Terraform also has many verbosity choices.

installer/pkg/config-generator/fixtures/test-aws.yaml
```diff
@@ -8,11 +8,11 @@ clusters:
       podCIDR: 10.2.0.0/16
       serviceCIDR: 10.3.0.0/16
     masters:
-      count: 2
+      nodePool: master
     workers:
-      count: 3
+      nodePool: worker
```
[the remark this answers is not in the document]
squat: yeah I agree. I'll implement this and for the moment force the list to be length 1

installer/pkg/config-generator/generator_test.go
squat: no, `etcd` is not an acronym

installer/pkg/config/types.go
```diff
@@ -72,8 +86,8 @@ type ddnsKey struct {
 }
 
 type etcd struct {
-	Count    int          `yaml:"count"`
 	External etcdExternal `yaml:"external"`
+	NodePool string       `yaml:"nodePool"`
```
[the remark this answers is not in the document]
squat: Hmm do you mean you expect users to provide a complete nodepool struct here as well as in masters as well as in workers? Or you want to automagically link the fields while unmarshaling?

installer/pkg/config/validate.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	if err := c.validateSharedNodePool(); err != nil {
+		errs = append(errs, err)
+	}
+
+	return errs
+}
+
+func (c *Cluster) validateSharedNodePool() error {
```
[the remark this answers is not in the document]
squat: sounds good :+1:

installer/pkg/config/types.go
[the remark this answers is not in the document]
mxinden: Automagic! 

But I guess implementing the custom unmarshaller and making sure, node pools are parsed before the master configuration ... is pretty over-engineered.

installer/pkg/config/validate.go
```diff
@@ -0,0 +1,106 @@
...
+	"fmt"
+	"strings"
+)
+
+// ErrMissingNodePool is returned when a nodePool was specified but not found in the nodePools list.
+type ErrMissingNodePool struct {
+	name string
+}
+
+// Error implements the error interface.
```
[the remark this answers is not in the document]
squat: thank you!

installer/pkg/config/types.go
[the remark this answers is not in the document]
squat: yeah I think it's a little too magic. If we look at kubernetes types that reference other structs by name, they are also implemented as `name string` rather than with magic

installer/pkg/config/validate.go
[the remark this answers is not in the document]
squat: good suggestion! it also cuts down on some length :)

installer/pkg/workflow/init_test.go
squat: ayy yeah thanks again for catching! I realize that I also let slip another format string mistake: we used `%s` instead of the conventional `%v`

mxinden:
@squat Oh, sorry about that. I only checked the Jenkins error messages. I am looking into it.

squat:
@mxinden your suggestion sent me doing some more hunting and I found that my methods were not working as expected in my branch. they were working fine in the master branch -_______- just pushed a fix :p

mxinden:
@squat Guess you were a bit faster. In case you want to include a unit test for this case: https://github.com/squat/tectonic-installer/pull/1

installer/pkg/config/validate_test.go
```diff
@@ -0,0 +1,338 @@
...
+		for _, err := range errs {
+			if _, ok := err.(*ErrSharedNodePool); ok {
+				n++
+			}
+		}
+
+		if n != c.errs {
+			t.Errorf("test case %d: expected %d shared node pool errors, got %d", i, c.errs, n)
+		}
+		//if c.fields != 0 {
```
enxebre: left over comment?

PR #3089, opened 2018-03-09
installer: For missing license / pull secret, log as INFO with filenames

kyoto:
(PR for `track-1` branch)

Change log messages output when the license and pull secret files are not found:
  - Instead of WARNING level, output as INFO level (not visible in terminal with default logging level)
  - Include the expected filenames, not just the expected directory

cc: @sym3tri, @aaronlevy

### DIFF
--- a/installer/api/handlers_tectonic.go
+++ b/installer/api/handlers_tectonic.go
@@ -337,16 +337,18 @@ func tectonicFactsHandler(w http.ResponseWriter, req *http.Request, ctx *Context
 		dir = filepath.Dir(ex)
 	}
 
-	license, err := ioutil.ReadFile(filepath.Join(dir, "license.txt"))
+	licensePath := filepath.Join(dir, "license.txt")
+	license, err := ioutil.ReadFile(licensePath)
 	if err != nil {
 		license = nil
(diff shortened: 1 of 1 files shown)


installer/api/handlers_tectonic.go
```diff
@@ -337,16 +337,18 @@ func tectonicFactsHandler(w http.ResponseWriter, req *http.Request, ctx *Context
 		dir = filepath.Dir(ex)
 	}
 
-	license, err := ioutil.ReadFile(filepath.Join(dir, "license.txt"))
+	licensePath := filepath.Join(dir, "license.txt")
```
squat: The issue is that this code expects a file that is named differently than the defaults that are downloaded from quay and account.tectonic.com. Why not look for tectonic-license.txt and config.json? Oytherwise, who would know to name the files that way?

kyoto: Actually, `license.txt` and `pull_secret.json` match the filenames we used in `assets.zip`.

squat: but assets.zip is only generated _after_ the installer has already run and the files were already not in the right place

PR #3070, opened 2018-03-05
platform/aws: add user custom ign config support

enxebre:
fix inst-906

PR #3090, opened 2018-03-09
platform/aws: add user manipulable raw ign config via user-data local files

enxebre:
add user manipulable raw ign config via user-data local files. Fix INST-906

### DIFF
--- a/steps/assets/resources/ignition-master.ign
+++ b/steps/assets/resources/ignition-master.ign
@@ -1,0 +1,3 @@
+{
+  "ignition": { "version": "2.1.0" }
+}
--- a/steps/bootstrap/main.tf
+++ b/steps/bootstrap/main.tf
@@ -121,6 +121,7 @@ module "masters" {
   subnet_ids                   = "${module.vpc.master_subnet_ids}"
   ec2_ami                      = "${var.tectonic_aws_ec2_ami_override}"
   kubeconfig_content           = "${local.kubeconfig_kubelet_content}"
+  user_ign                     = "${data.template_file.user_ign_master.rendered}"
 }
 
 module "workers" {
@@ -146,6 +147,15 @@ module "workers" {
   ec2_ami                      = "${var.tectonic_aws_ec2_ami_override}"
   base_domain                  = "${var.tectonic_base_domain}"
   kubeconfig_content           = "${local.kubeconfig_kubelet_content}"
+  user_ign                     = "${data.template_file.user_ign_worker.rendered}"
+}
+
+data "template_file" "user_ign_master" {
+  template = "${file("${path.module}/../assets/resources/ignition-master.ign")}"
+}
+
+data "template_file" "user_ign_worker" {
+  template = "${file("${path.module}/../assets/resources/ignition-worker.ign")}"
 }
 
 module "dns" {
--- a/modules/aws/master-asg/master.tf
+++ b/modules/aws/master-asg/master.tf
@@ -65,6 +65,10 @@ data "ignition_config" "tnc_master" {
   }
 
   files = ["${data.ignition_file.kubelet_master_kubeconfig.id}"]
+
+  append {
+    source = "data:text/plain;charset=utf-8;base64,${base64encode(var.user_ign)}"
+  }
 }
 
 data "ignition_file" "kubelet_master_kubeconfig" {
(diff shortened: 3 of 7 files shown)


enxebre:
cc @crawford

steps/assets/resources/ignition-master.ign
```diff
@@ -0,0 +1,32 @@
...
+    "version": "2.1.0"
+  },
+  "networkd": {},
+  "passwd": {},
+  "storage": {
+    "files": [
+      {
+        "filesystem": "root",
+        "group": {},
+        "path": "/etc/kubernetes/kubeconfig",
```
crawford: This doesn't seem right. The kubeconfig should be dynamically generated by TNC. This PR doesn't change the interaction though, but I think this should be removed eventually.

enxebre: thanks! I agree. It's unrelated to this PR. That would need to go hand in hand with the bootstrap token generation. Is there a place where we track what the TNC should do, and what it actually currently does?

squat:
I do not think this is the right approach. As t is currently implemented, users edit the very ignition config that we design. If they change the append, or the permissions on the kubeconfig then we are broken.

Instead, uses should be allowed to provide an entire ignition config by themselves. If they do, then we append it and our own ignition to the stub. If they do not, then we only append our ignition to the stub. Exposing our ignition is the wrong interface for this and is inviting more problems IMO.

Further, this append should really be performed in a step in the Tectonic CLI so that we can use the igntition golang packages to validate the config before sending to user data. Otherwise we will be booting nodes with ignition that was never going to work and this will be much harder to debug.

enxebre:
@squat if I'm understanding you correctly that's what the original PR https://github.com/coreos/tectonic-installer/pull/3070 was doing. For appending we'd need to upload the file and rely on either s3 or the TNC which was discarded

squat:
Why not append using inline url-encoded string?

enxebre:
using `data:, urlencode()` lead to parsing issues while bootstrapping aws
screen shot 2018-03-12 at 13 55 22
screen shot 2018-03-12 at 15 15 30

So using base64 here which also matches the default format that we are generating via tf for e.g filesystems/files/content 
@squat PTAL

enxebre:
This seems to be caused by discrepancies between tf `urlencode()` e.g replacing spaces with :"+" https://github.com/golang/go/blob/go1.9/src/net/url/url.go#L296 and how ignition decodes the url https://github.com/coreos/ignition/blob/master/vendor/github.com/vincent-petithory/dataurl/dataurl.go#L188

steps/assets/resources/ignition-master.ign
```diff
@@ -0,0 +1,3 @@
+{
+  "ignition": { "version": "2.1.0" }
```
squat: Rather than make this a file that is always in the repo, shouldn't this be a top level input that users can provide optionally? this would allow us to perform validation in the CLI as well rather than just appending whatever is given.

steps/bootstrap/main.tf
```diff
@@ -121,6 +121,7 @@ module "masters" {
   subnet_ids                   = "${module.vpc.master_subnet_ids}"
   ec2_ami                      = "${var.tectonic_aws_ec2_ami_override}"
   kubeconfig_content           = "${local.kubeconfig_kubelet_content}"
+  user_ign                     = "${data.template_file.user_ign_master.rendered}"
```
squat: And here we would do something like 
```hcl
"${var.tectonic_user_ignition_master}"
```

steps/bootstrap/main.tf
```diff
@@ -146,6 +147,15 @@ module "workers" {
   ec2_ami                      = "${var.tectonic_aws_ec2_ami_override}"
   base_domain                  = "${var.tectonic_base_domain}"
   kubeconfig_content           = "${local.kubeconfig_kubelet_content}"
+  user_ign                     = "${data.template_file.user_ign_worker.rendered}"
+}
+
+data "template_file" "user_ign_master" {
```
squat: and we would not need to template anything here.

squat: Even if we do not go with the top level config option, this should be 
```hcl
data "local_file" "foo" {
    filename = "${path.module}/foo.bar"
}
```
rather than `template_file`
https://www.terraform.io/docs/providers/local/d/file.html

enxebre:
closing in favour of https://github.com/coreos/tectonic-installer/pull/3120

PR #3093, opened 2018-03-12
tests: Print tf logs only on failure

mxinden:
Instead of logging the terraform output to Jenkins output by default,
only do so on failures. The successful logs will still be captured in
the log files, which are both archived on Jenkins as well as send to
Elastic search.

@cpanato what do you think?


cpanato:
thanks!!

mxinden:
@cpanato added a `puts` to inform the user not expect any output on success. Would you mind taking another look?

mxinden:
@cpanato Sorry for all the back and forth. Can you give this another review?

cpanato:
sorry for what? 😃

PR #3106, opened 2018-03-13
modules/bootkube-ut2: add cl update server url variable

sdemos:
tnc has to be able to configure machines to point to the in-cluster container linux updater for cl updates. this threads the configuration variable through into the installer to make sure we can get the right address, or not configure the update server at all.

I'm not super familiar with terraform so I'm not sure if I'm setting the variable in the most effective way. It should be referencing the [tectonic-clu service](https://github.com/coreos-inc/tectonic-operators/blob/master/operator/utility/manifests/clu/clu-service.yaml), deployed as part of the tectonic utility operator. 

The variable added to `tnc-config` is added to tnc in coreos-inc/tectonic-operators#306

/cc @thorfour


squat:
@sdemos Is the goal for the `cl_update_server_url` to be modifiable by the end user, right? In that case we need to add a variable declaration in `config.tf`, otherwise there is no way to modify the value from the `steps/assets` module.

modules/bootkube-ut2/variables.tf
```diff
@@ -255,3 +255,8 @@ EOF
 
   type = "string"
 }
+
+variable "update_server_url" {
```
squat: This variable declaration does not match its use in the module; it is missing the `cl_` prefix

squat:
@sdemos please rebase this PR to account for the recent cleanup in master and also address the review comments then we can get this merged.

sdemos:
oh, sorry, I should add an update here. This doesn't actually seem to work yet, the internal cluster dns name doesn't resolve correctly for components on the machine, and I have to figure out how to address them properly. I'll dig into it a little more and ping you when I have something that's actionable on your end.

alexsomesan:
@sdemos we already provision node-level DNS records for the TNC endpoint as we required it for bootstrapping.
You could make use of that to unblock your work until you figure out what is wrong with the cluster DNS.

Would that help?

sdemos:
We ended up just skipping dns entirely and just using a NodePort. That makes it unnecessary to merge any changes into the installer anymore. 

Sorry about the noise!

PR #3050, opened 2018-02-28
modules/tls: add uuid function to created certs

trawler:
Fixes #INST-896
rebase of pull #3031

PR #3031, opened 2018-02-26
modules/tls: add uuid function to created certs

kalmog:
Fixes #INST-896

PR #3108, opened 2018-03-13
Fixes that were missed in #2972

abhinavdahiya:
Fixes that were missed in #2972

* bootkube-ut2: update trust in kubeconfigs to root-ca
* tls: update user-provided variables and outputs
* also fix the duplicated organization for root-ca.

### DIFF
--- a/modules/bootkube-ut2/assets.tf
+++ b/modules/bootkube-ut2/assets.tf
@@ -56,7 +56,7 @@ data "template_file" "kubeconfig" {
   template = "${file("${path.module}/resources/kubeconfig")}"
 
   vars {
-    kube_ca_cert = "${base64encode(var.kube_ca_cert_pem)}"
+    root_ca_cert = "${base64encode(var.root_ca_cert_pem)}"
     admin_cert   = "${base64encode(var.admin_cert_pem)}"
     admin_key    = "${base64encode(var.admin_key_pem)}"
     server       = "${var.kube_apiserver_url}"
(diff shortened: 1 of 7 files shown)


abhinavdahiya:
retest this please, azure only

modules/bootkube-ut2/assets.tf
```diff
@@ -56,7 +56,7 @@ data "template_file" "kubeconfig" {
   template = "${file("${path.module}/resources/kubeconfig")}"
```
enxebre: Hey @abhinavdahiya You'll need to change the actual templates kubeconfig/kubeconfig-kubelet as well to use that value `root_ca_cert` too.
Certs questions, shouldn't the kube_ca_cert be enough here and used by both kubeconfig anyways?

modules/tls/ca/self-signed/main.tf
```diff
@@ -15,7 +15,6 @@ resource "tls_self_signed_cert" "root_ca" {
   subject {
     common_name  = "root-ca"
     organization = "${uuid()}"
```
trawler: The same change needed to be added to all the self generated certs (not CAs) - IE, identity, etcd, kube. you can see the list of of files and the changes in #3031

abhinavdahiya: :+1: Oh i didn't realize that. will do that.

abhinavdahiya: @trawler #3031 was closed in favor of #3050 and #3050 only updates the CAs with organisation and organizational_unit. ??

trawler: doh! sorry, you're right... got these two PRs confused. you're correct.

abhinavdahiya: so, since only root_ca is the only self-signed cert and kube_ca, aggregator_ca and etcd_ca are signed by root_ca, is adding uuid() to root_ca sufficient or should i add uuid() to all the all of them?

trawler: You can test this by verifying that the cluster always generates tls certs with a different hash: `openssl x509 -in $i -hash -noout`

abhinavdahiya:
retest this please. aws only

modules/tls/ca/self-signed/main.tf
abhinavdahiya: Added uuid to all CAs

abhinavdahiya:
retest this please. bare-metal only

abhinavdahiya:
retest this please. govcloud only

abhinavdahiya:
@enxebre @trawler all the tests are green.

PR #3111, opened 2018-03-14
tests: update certificates for metal tests (#2885) (#3102)

cpanato:
cherry-pick from: 45db8131234d11c073b70ad9767f7a32020579dd

certificates for matchbox was expired and we need to create new certificates for matchbox


mxinden:
@cpanato Both the PR description as well as the commit message explain why the certificates need to be updated. Would you mind adding that information?

PR #3112, opened 2018-03-14
tests: Raise error on failed tectonic / terraform command

mxinden:
Instead of returning a boolean as a success feedback, raise an error.
This seems to be the idiomatic way for both the framework as well as
Ruby.


mxinden:
@cpanato @squat Would you mind taking another look?

PR #3113, opened 2018-03-14
README: add instructions to build using Docker

squat:
Document how to build the repo using docker so that you do not need to install bazel to get started. This method has the benefit of caching build artifacts so subsequent invocations are faster.

cc @mxinden @spangenberg @sym3tri

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,10 +41,15 @@ These instructions can be used for the official stable platforms listed above, a
     bazel build tarball
     ```
 
-2. Unzip the tarball
+    *Note*: the project can optionally be built without installing Bazel, provided Docker is installed:
+    ```shell
+    docker run --rm -v $PWD:$PWD -w $PWD quay.io/coreos/tectonic-builder:bazel-v0.3 bazel --output_base=.cache build tarball
+    ```
+
+2. Extract the tarball
     ```shell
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -38,7 +38,7 @@ These instructions can be used for the official stable platforms listed above, a
 
 1. Build the project
     ```shell
-    bazel build tarball
+    docker run --rm -v $PWD:$PWD -w $PWD -u 0 quay.io/coreos/tectonic-builder:bazel-v0.3 bazel --output_base=.cache build tarball
```
mxinden: > `-u 0`

Does this has to be run as root?

squat: huh.. I thought it was failing without running as root, but now I'm testing again and I am finding it's fine. I'll make the change. Thanks for bringing it up

README.md
```diff
@@ -38,13 +38,13 @@ These instructions can be used for the official stable platforms listed above, a
 
 1. Build the project
     ```shell
-    bazel build tarball
```
sym3tri: No need to _replace_ the current instructions. I just meant to add running with Docker as _another_ option.

squat: PTAL again

Issue #3109, opened 2018-03-14
NET::ERR_CERT_AUTHORITY_INVALID


zhex900:
# BUG REPORT

### Tectonic version ([release](https://github.com/coreos/tectonic-installer/releases) or commit hash):
```
latest
```

### Terraform version (`terraform version`):

```
latest
```
### Platform:

```
AWS
```

### What happened?

I have followed the GUI cluster install. After the successful install is completed. The console page cannot be accessed. 
```
Your connection is not private
Attackers might be trying to steal your information from k8.ewhale.co (for example, passwords, messages or credit cards). Learn more
NET::ERR_CERT_AUTHORITY_INVALID
Subject: k8.ewhale.co

Issuer: kube-ca

Expires on: 13 Mar 2021

Current date: 14 Mar 2018

PEM encoded chain:
-----BEGIN CERTIFICATE-----
MIIDuDCCAqCgAwIBAgIQXzgfoWUsTyjNOs97s+LcVDANBgkqhkiG9w0BAQsFADBc
MQkwBwYDVQQGEwAxCTAHBgNVBAgTADEJMAcGA1UEBxMAMQkwBwYDVQQREwAxETAP
BgNVBAoTCGJvb3RrdWJlMQkwBwYDVQQLEwAxEDAOBgNVBAMTB2t1YmUtY2EwHhcN
MTgwMzEzMjMzMTEyWhcNMjEwMzEyMjMzMTEyWjBZMQkwBwYDVQQGEwAxCTAHBgNV
BAgTADEJMAcGA1UEBxMAMQkwBwYDVQQREwAxCTAHBgNVBAoTADEJMAcGA1UECxMA
MRUwEwYDVQQDEwxrOC5ld2hhbGUuY28wggEiMA0GCSqGSIb3DQEBAQUAA4IBDwAw
ggEKAoIBAQCsqyFzlB1RkCSxZNRmVXmj9ousuLrhN+YL337wqMI77VWRtImy4UkH
WgL38k9v8A3DCJQCKong42qUMLWp24r1noCvzl6FBN5KbGmoGfgGdF2Dm8qfoVoW
LIYk6BaB2s8XEiNHLfjLQTuOaS5x8t+Xh6RQvC+8gLEZhoxVmDrUwf9u7JcwNMRK
SweWwuK8G4Lk/v7VrjDDXmDteHZS9J/lyYN0WOpcOHlQPB6tG02iXrDds9Aafta+
[...]

PR #3114, opened 2018-03-14
config.tf: bump tls provider to fix CA

squat:
cc @alexsomesan @enxebre 

fixes: https://github.com/coreos/tectonic-installer/issues/3109


sym3tri:
@squat status?

PR #3115, opened 2018-03-15
cli: add gometalinter support

enxebre:
Bazel support for gometalinter

### DIFF
--- a/installer/BUILD.bazel
+++ b/installer/BUILD.bazel
@@ -89,6 +89,7 @@ genrule(
             "export HOME=test",
             "./installer/bin/gometalinter --deadline=240s --disable-all \
             --exclude=\"pkg\\/config\\/validate.*\\.go.*\\(dupl\\)$$\" \
+            --exclude=\"pkg\\/validate.*\\.go.*\\(dupl\\)$$\" \
             --cyclo-over=20 \
             --enable=golint \
             --enable=vet \
@@ -96,6 +97,7 @@ genrule(
             --enable=misspell \
             --enable=dupl \
             installer/cmd/tectonic \
+            installer/pkg/validate \
             installer/pkg/config \
             installer/pkg/config/aws \
             installer/pkg/config/azure \
@@ -105,18 +107,18 @@ genrule(
             installer/pkg/config/openstack \
             installer/pkg/config/vmware \
             installer/pkg/config-generator \
-            installer/pkg/terraform-generator \
             installer/pkg/workflow",
           ]) + "' > $@",
     testonly = 1,
 )
 
 sh_test(
-    name    = "cli_gometalinter",
-    size    = "large",
+    name    = "gometalinter",
+    size    = "small",
     srcs    = [":gometalinter_sh"],
     data    = [
         "//installer/cmd/tectonic:go_files",
+        "//installer/pkg/validate:go_files",
         "//installer/pkg/config:go_files",
         "//installer/pkg/config/aws:go_files",
         "//installer/pkg/config/azure:go_files",
@@ -126,7 +128,6 @@ sh_test(
         "//installer/pkg/config/openstack:go_files",
         "//installer/pkg/config/vmware:go_files",
         "//installer/pkg/config-generator:go_files",
-        "//installer/pkg/terraform-generator:go_files",
         "//installer/pkg/workflow:go_files",
         "//installer:gometalinter_runtime",
         "//installer:golint_runtime",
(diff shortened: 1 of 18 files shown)


enxebre:
cc @mxinden @cpanato @squat

installer/cmd/tectonic/BUILD.bazel
```diff
@@ -49,3 +49,11 @@ go_binary(
     goarch = "amd64",
     visibility = ["//visibility:public"],
 )
+
+filegroup(
```
mxinden: Does this need to be added to every sub-package, or could it be a top level definition?

installer/BUILD.bazel
```diff
@@ -37,3 +37,133 @@ test_suite(
...
+            installer/pkg/config/openstack \
+            installer/pkg/config/vmware \
+            installer/pkg/config-generator \
+            installer/pkg/terraform-generator \
+            installer/pkg/workflow",
+          ]) + "' > $@",
+    testonly = 1,
+)
+
+sh_test(
```
mxinden: You could use our custom `gen_test` here, instead of generating a shell script via a `genrule`. `gen_test` is in no way perfect, but might be a bit less verbose.

installer/BUILD.bazel
```diff
@@ -37,3 +37,133 @@ test_suite(
...
+    name    = "cli_gometalinter",
+    size    = "large",
```
mxinden: `large` is a bit exaggerated, resulting in: `There were tests whose specified size is too big. Use the --test_verbose_timeout_warnings command line option to see which ones these are.`. Would you mind adjusting it to a more reasonable level? (See Bazel level docs: https://docs.bazel.build/versions/master/test-encyclopedia.html#role-of-the-test-runner)

installer/BUILD.bazel
```diff
@@ -37,3 +37,133 @@ test_suite(
...
+            installer/pkg/config/aws \
+            installer/pkg/config/azure \
+            installer/pkg/config/gcp \
+            installer/pkg/config/govcloud \
+            installer/pkg/config/metal \
+            installer/pkg/config/openstack \
+            installer/pkg/config/vmware \
+            installer/pkg/config-generator \
+            installer/pkg/terraform-generator \
+            installer/pkg/workflow",
```
squat: Can we add the validate package? it is newly created to be used in the CLi for valdation: https://github.com/coreos/tectonic-installer/tree/master/installer/pkg/validate

installer/BUILD.bazel
```diff
@@ -37,3 +37,133 @@ test_suite(
...
+          ]) + "' > $@",
+    testonly = 1,
+)
+
+sh_test(
+    name    = "cli_gometalinter",
```
squat: can we remove the `cli_` prefix from these tests? I think it is clear enough with `bazel test installer:gometalinter` or even just `bazel test installer:lint`

installer/cmd/tectonic/BUILD.bazel
enxebre: Afaik it's needed. I can't find a way to filegroup crossing target boundaries, but my knowledge here is limited, would love to hear any ideas.

enxebre:
@squat @mxinden PTAL

mxinden:
@enxebre Is there a way to include the [go smoke test code](https://github.com/coreos/tectonic-installer/tree/master/tests/smoke) in the linting process as well?

enxebre:
thanks for the suggestions @mxinden, added a new commit with it and dupl improvements

enxebre:
ok to test

installer/BUILD.bazel
```diff
@@ -7,3 +7,137 @@ test_suite(
...
+            --exclude=\"pkg\\/config\\/validate.*\\.go.*\\(dupl\\)$$\" \
+            --exclude=\"pkg\\/validate.*\\.go.*\\(dupl\\)$$\" \
+            --cyclo-over=20 \
+            --dupl-threshold=75 \
+            --enable=golint \
+            --enable=vet \
+            --enable=gocyclo \
+            --enable=misspell \
+            --enable=dupl \
+            installer/cmd/tectonic \
```
squat: I worry that this is not flexible enough. Next time somebody adds a new package that has broken linting but forgets to add it here we will never know. We should use `go list` to grab all the packages.

enxebre: Yeah, I don't really like it much either...
A new package would need to add the filegroup
`
filegroup(
    name = "go_files",
    visibility = ["//visibility:public"],
    srcs = glob([
        "*.go",
    ]),
)
`
And then it has to be referenced here.
However I couldn't find nicer way to collect all the go files crossing target boundaries with Bazel.
For using something like go lint, wouldn't the packages need to be made available for it during bazel run time? so wouldn't it still require manual static reference of the packages?
I can't see a way in bazel to "collect all the files" without adding static targets

sym3tri:
needs rebase

sym3tri:
@enxebre still doing this? If not please close.

PR #3116, opened 2018-03-15
[WIP] *: Replace wrong usage of cluster_name with dns_name

mxinden:
By default the Tectonic installer uses the cluster name with the base
domain (`<cluster_name>.<base_domain>`) as the url for the console, api,
tnc, ... In addition the Tectonic installer has the feature to specify a
custom dns name, which will be prefixed to the base domain instead of
the cluster name when set.

This was not consistently done in the Terraform code. Instead, sometimes
the cluster name would be used even though the dns name was set. This
patch introduces consistent usage of dns name, which is either set to
the input dns name, or if empty to the cluster name.

**This is a work in progress implementation.** Waiting for proper scheduling into Sprint (INST-999).

//CC @squat @spangenberg

### DIFF
--- a/modules/aws/master-asg/master.tf
+++ b/modules/aws/master-asg/master.tf
@@ -61,7 +61,7 @@ resource "aws_autoscaling_group" "masters" {
 
 data "ignition_config" "tnc_master" {
   append {
-    source = "http://${var.cluster_name}-tnc.${var.base_domain}/ign/v1/role/master"
+    source = "http://${var.dns_name}-tnc.${var.base_domain}/ign/v1/role/master"
   }
 
   files = ["${data.ignition_file.kubelet_master_kubeconfig.id}"]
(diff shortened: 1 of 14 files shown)


modules/aws/master-asg/master.tf
```diff
@@ -61,7 +61,7 @@ resource "aws_autoscaling_group" "masters" {
 
 data "ignition_config" "tnc_master" {
   append {
```
enxebre: will need to change the pattern https://github.com/coreos/tectonic-installer/blob/master/steps/bootstrap/tnc-s3.tf#L16 and https://github.com/coreos/tectonic-installer/blob/master/steps/bootstrap/s3-assets.tf#L4

platforms/govcloud/main.tf
```diff
@@ -66,6 +66,7 @@ module "etcd" {
   ssh_key                    = "${var.tectonic_govcloud_ssh_key}"
   subnets                    = "${module.vpc.worker_subnet_ids}"
   dns_server_ip              = "${var.tectonic_govcloud_dns_server_ip}"
+  dns_name = "${var.tectonic_dns_name}"
```
enxebre: Can't find tectonic_dns_name in the config file

squat: https://github.com/coreos/tectonic-installer/blob/master/platforms/govcloud/variables.tf#L182

squat:
Closing since this was WIP and is out of date. We can still track this issue with INST-999.

PR #3120, opened 2018-03-16
cli: add ignition support

enxebre:
- Add a IgnitionFile field for node pools 
- the cli validate ign file and embed the tnc append block
- Terraform always receive the generated user-data ign file path as an input

Once we get https://github.com/coreos/tectonic-installer/pull/3079 will create a follow up PR for etcd to use the generated userdata


enxebre:
@thorfour seems that tnc is ignoring the KubeconfigFetchCmd property

config.tf
```diff
@@ -522,3 +522,19 @@ It is only supported for Azure cloud provider.
 Used to collect the boot logs for debug purposes.
 EOF
 }
+
+variable "tectonic_ignition_master" {
+  type = "string"
+
+  description = <<EOF
+(optional) Ignition config file path. This is automatically generated by the installer.
```
squat: Should we change these to `(internal)` since users should not touch it?

squat: What do you think about supplying the full json here rather than a file path? This way the tectonic CLI does not need to write anything to disk, just add the text to the terraform.tfvars.

installer/pkg/config-generator/ignition.go
```diff
@@ -0,0 +1,140 @@
...
+		poolToRole[n] = "worker"
+	}
+	for _, n := range c.Etcd.NodePools {
+		poolToRole[n] = "etcd"
+	}
+	return poolToRole
+}
+
+func (c ConfigGenerator) GenerateIgnConfig(clusterDir string) error {
+
```
squat: extra newline

installer/pkg/config-generator/ignition.go
```diff
@@ -0,0 +1,140 @@
...
+	for _, n := range c.Worker.NodePools {
+		poolToRole[n] = "worker"
+	}
+	for _, n := range c.Etcd.NodePools {
+		poolToRole[n] = "etcd"
+	}
+	return poolToRole
+}
+
+func (c ConfigGenerator) GenerateIgnConfig(clusterDir string) error {
```
squat: Can you add a comment to this exported func?

installer/pkg/config-generator/ignition.go
```diff
@@ -0,0 +1,140 @@
...
+	cfg, rpt, err := ignconfig.Parse(data)
+	if len(rpt.Entries) > 0 {
+		return nil, fmt.Errorf("failed to parse ignition file %s: %s", filePath, rpt.String())
+	}
+
+	return &cfg, nil
+}
+
+func (c ConfigGenerator) embedAppendBlock(ignCfg ignconfigtypes.Config, role string) *ignconfigtypes.Config {
+
```
squat: extra newline

installer/pkg/config-generator/ignition.go
```diff
@@ -0,0 +1,140 @@
...
+	appendBlock := &ignconfigtypes.ConfigReference{
```
squat: why do we generate a pointer to the struct and then de-reference it in the next statement?

installer/pkg/config-generator/ignition.go
```diff
@@ -0,0 +1,140 @@
...
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	ignconfig "github.com/coreos/ignition/config/v2_0"
+	ignconfigtypes "github.com/coreos/ignition/config/v2_0/types"
+)
+
+const (
+	IgnitionMaster = "ignition-master.ign"
```
squat: We redefine these exact same consts in https://github.com/coreos/tectonic-installer/pull/3120/files#diff-6db5d266dcce2af01d3bf830d357bc26R16

installer/pkg/config/types.go
```diff
@@ -91,6 +91,14 @@ type Proxy struct {
 	No    string `json:"tectonic_no_proxy,omitempty" yaml:"no,omitempty"`
 }
 
+type nodePools []nodePool
+
+type nodePool struct {
```
squat: It does not look like anyone is using this type? https://github.com/enxebre/tectonic-installer/blob/07fe3dab9cc74f8184bb23b821063c8f75eaacda/installer/pkg/config/cluster.go#L36

installer/pkg/config/validate.go
```diff
@@ -46,13 +50,69 @@ func (e *ErrSharedNodePool) Error() string {
 	return fmt.Sprintf("node pools cannot be shared, but %q is used by %s", e.name, strings.Join(e.fields, ", "))
 }
 
+// ErrWrongIgnConfig is returned when a wrong ign config is given.
```
squat: what do you mean by wrong? Invalid maybe?

installer/pkg/config/validate.go
```diff
@@ -46,13 +50,69 @@ func (e *ErrSharedNodePool) Error() string {
...
 // Validate ensures that the Cluster is semantically correct and returns an error if not.
 func (c *Cluster) Validate() []error {
 	var errs []error
 	errs = append(errs, c.validateNodePools()...)
+	errs = append(errs, c.validateIgnitionFiles()...)
+	return errs
+}
+
+func (c *Cluster) validateIgnitionFiles() []error {
+
```
squat: extra newline

installer/pkg/config/validate.go
```diff
@@ -46,13 +50,69 @@ func (e *ErrSharedNodePool) Error() string {
...
+		if err := validateIgnitionConfig(n.IgnitionFile); err != nil {
+			errs = append(errs, err)
+		}
+	}
 	return errs
 }
 
+func validateFileExist(ignitionFile string) error {
+	if _, err := os.Stat(ignitionFile); err != nil {
```
squat: there is no need for the `if` here. Just always `return err`. This way, if the error is nil, then you are returning nil and if the error is not nil then you are returning an error

installer/pkg/config/validate.go
```diff
@@ -46,13 +50,69 @@ func (e *ErrSharedNodePool) Error() string {
...
+		return err
+	}
+	return nil
+}
+
+func validateIgnitionConfig(filePath string) error {
+
```
squat: extra newline

config.tf
enxebre: Technically make sense, from a user perspective I like the file input

enxebre:
@squat addressed the comments PTAL. cc @crawford @alexsomesan

enxebre:
ok to test

enxebre:
test failing due to asg permissions:
```
* aws_autoscaling_group.workers: AccessDenied: User: arn:aws:sts::846518947292:assumed-role/tf-tectonic-installer/TECTONIC_INSTALLER_awsbas-pr-3120-8630a1 is not authorized to perform: autoscaling:DeleteTags on resource: arn:aws:autoscaling:us-east-2:846518947292:autoScalingGroup:64585609-4a53-4c26-9a37-238442081a0c:autoScalingGroupName/awsbas-pr-3120-8630a1-workers
```
cc @cpanato

squat:
@enxebre this PR adds too many files. Did you run glide-vc as directed in the README to reduce the number of vendors files?

installer/pkg/config-generator/ignition.go
```diff
@@ -0,0 +1,166 @@
...
+	"os"
+	"path/filepath"
+
+	ignconfig "github.com/coreos/ignition/config/v2_0"
+	ignconfigtypes "github.com/coreos/ignition/config/v2_0/types"
+	"github.com/coreos/tectonic-installer/installer/pkg/config"
+)
+
+var (
+	ignVersion = ignconfigtypes.IgnitionVersion{
```
squat: We should probably in-line the this struct for readability

installer/pkg/config-generator/ignition.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+// GenerateIgnConfig generates, if successful, files with the ign config for each role.
+func (c ConfigGenerator) GenerateIgnConfig(clusterDir string) error {
+	poolToRole := c.poolToRoleMap()
+	for _, p := range c.NodePools {
+		ignFile := p.IgnitionFile
+		ignCfg, err := parseIgnFile(ignFile)
+		if err != nil {
+			return err
```
squat: We should really wrap error so for improved context when debugging

installer/pkg/config-generator/ignition.go
```diff
@@ -0,0 +1,166 @@
...
+			Version: ignVersion,
+		}
+		return &ignconfigtypes.Config{Ignition: *ignition}, nil
+	}
+
+	data, err := ioutil.ReadFile(filePath)
+	if err != nil {
+		return nil, err
+	}
+
```
squat: This error seems to be completely ignored

installer/pkg/config/cluster.go
```diff
@@ -12,6 +12,12 @@ import (
 	"github.com/coreos/tectonic-installer/installer/pkg/config/vmware"
 )
 
+const (
+	IgnitionMaster = "ignition-master.ign"
```
squat: Could you document these exported values?

installer/pkg/config/cluster.go
```diff
@@ -41,6 +50,9 @@ type Cluster struct {
 	openstack.OpenStack `json:",inline" yaml:"openstack,omitempty"`
 	vmware.VMware       `json:",inline" yaml:"vmware,omitempty"`
 	Internal            `json:",inline" yaml:"-"`
+	IgnitionMaster      string `json:"tectonic_ignition_master,omitempty"`
```
spangenberg: Can you please ignore yaml in this and the below lines.

enxebre:
retest this please

enxebre:
retest this please

PR #3128, opened 2018-03-20
bazel: remove symlinks

squat:
cc @ggreer @crawford 

related: #3129 #3130

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,11 @@ bin_test/
 matchbox/
 modules/update-payload/generated/
 /contrib/govcloud/vpn.conf
+
+# Bazel
+bazel-bin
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -15,3 +15,11 @@ bin_test/
 matchbox/
 modules/update-payload/generated/
 /contrib/govcloud/vpn.conf
+
+# Bazel
```
crawford: I think you probably need `bazel-src` as well.

squat: I don't get that directory locally, but we can add it if other people do. JW, your working directory for the installer isn't named `src` is it? Because bazel will create a directory called `bazel-<working-dir>` by default.

PR #3131, opened 2018-03-21
Cli: get terraform binary from inside the package

trawler:
This PR resolves an issue whereby the tectonic binary didn't necessarily use the packages terraform binary

### DIFF
--- a/installer/pkg/workflow/executor.go
+++ b/installer/pkg/workflow/executor.go
@@ -1,0 +1,106 @@
+package workflow
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+)
+
+// executor enables calling TerraForm from Go, across platforms, with any
+// additional providers/provisioners that the currently executing binary
+// exposes.
+//
+// The TerraForm binary is expected to be in the executing binary's folder, in
+// the current working directory or in the PATH.
+type executor struct {
+	binaryPath string
+}
+
+// Set the binary names for different platforms
+const (
+	tfBinUnix    = "terraform"
+	tfBinWindows = "terraform.exe"
+)
+
+// errBinaryNotFound denotes the fact that the TerraForm binary could not be
+// found on disk.
+var errBinaryNotFound = errors.New(
+	"TerraForm not in executable's folder, cwd nor PATH",
+)
+
+// newExecutor initializes a new Executor.
+func newExecutor() (*executor, error) {
+	ex := new(executor)
+
+	// Find the TerraForm binary.
+	binPath, err := tfBinaryPath()
+	if err != nil {
+		return nil, err
+	}
+
+	ex.binaryPath = binPath
+	return ex, nil
+}
+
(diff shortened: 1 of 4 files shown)


trawler:
ok to test

installer/pkg/workflow/executor.go
```diff
@@ -0,0 +1,114 @@
...
+//
+// An error is returned if the TerraForm binary could not be found, or if the
+// TerraForm call itself failed, in which case, details can be found in the
+// output.
+func (ex *executor) execute(clusterDir string, args ...string) error {
+	// Prepare TerraForm command by setting up the command, configuration,
+	// and the working directory
+	cmd := exec.Command(ex.binaryPath, args...)
+
+	if clusterDir == "" {
```
enxebre: small hint: this should probably go at the top of the func before creating the cmd struct

installer/pkg/workflow/executor.go
```diff
@@ -0,0 +1,114 @@
...
+var ErrBinaryNotFound = errors.New(
+	"TerraForm not in executable's folder, cwd nor PATH",
+)
+
+// newExecutor initializes a new Executor.
+func newExecutor() (*executor, error) {
+	ex := new(executor)
+
+	// Find the TerraForm binary.
+	out, err := tfBinaryPath()
```
enxebre: `out` sounds a bit confusing to me, I'd rather use `bin` or `path`, wdyt?

installer/pkg/workflow/executor.go
```diff
@@ -0,0 +1,114 @@
...
+
+// Set the binary names for different platforms
+const (
+	tfBinUnix    = "terraform"
+	tfBinWindows = "terraform.exe"
+)
+
+// ErrBinaryNotFound denotes the fact that the TerraForm binary could not be
+// found on disk.
+var ErrBinaryNotFound = errors.New(
```
enxebre: no need to export this variable

installer/pkg/workflow/executor.go
```diff
@@ -0,0 +1,114 @@
...
+
+// executor enables calling TerraForm from Go, across platforms, with any
+// additional providers/provisioners that the currently executing binary
+// exposes.
+//
+// The TerraForm binary is expected to be in the executing binary's folder, in
+// the current working directory or in the PATH.
+type executor struct {
+	binaryPath   string
+	envVariables map[string]string
```
enxebre: this is no used anywhere

installer/pkg/workflow/executor.go
```diff
@@ -0,0 +1,113 @@
...
+
+	cmd := exec.Command(ex.binaryPath, args...)
+
+	cmd.Stdin = os.Stdin
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	cmd.Dir = clusterDir
+
+	// Start TerraForm.
+	err := cmd.Run()
```
enxebre: you can just:
`return cmd.Run()` here and avoid the `if err` verbosity

PR #3132, opened 2018-03-22
tests: fix role for aws govcloud

cpanato:

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -251,7 +251,7 @@ pipeline {
       }
       environment {
         TECTONIC_INSTALLER_ROLE = 'tf-tectonic-installer-track-2'
-        TECTONIC_GOVCLOUD_INSTALLER_ROLE = 'tf-tectonic-installer'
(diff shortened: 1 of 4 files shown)


Jenkinsfile
```diff
@@ -251,6 +251,7 @@ pipeline {
       }
       environment {
         TECTONIC_INSTALLER_ROLE = 'tf-tectonic-installer-track-2'
+        TECTONIC_GOVCLOUD_INSTALLER_ROLE = 'tf-tectonic-installer'
```
enxebre: should probably rather be named `TECTONIC_INSTALLER_GOVCLOUD_ROLE` wdyt?

enxebre: doesn't the value need to be 'tf-tectonic-installer-govcloud'?

cpanato: the role name in the govcloud accout is called `tf-tectonic-installer` :) 

I will update the first comment thanks!

cpanato:
@enxebre PTAL :)

cpanato:
this is not needed anymore

PR #3133, opened 2018-03-22
tests: save terraform logs when timeout

cpanato:


tests/rspec/lib/aws_cluster.rb
```diff
@@ -311,4 +296,12 @@ def describe_network_interfaces
   rescue => e
     puts e
   end
+
+  def save_terraform_logs(tectonic_logs, output)
```
enxebre: is there anything doing this terraform specific? should probably be just  `save_logs`?

tests/rspec/lib/aws_cluster.rb
```diff
@@ -207,43 +195,40 @@ def destroy
   end
 
   def run_tectonic_cli(env, cmd, flags = '')
```
enxebre: as a followup we could probably keep run_tectonic_cli purely as it is but returning the output, and add wrapper function that handles the timeout and call save_logs

cpanato: ok, since we are planning to refactor all the code, I dont want to spend to much time now

tests/rspec/lib/aws_cluster.rb
cpanato: this just save the terraform logs

enxebre: we use it in the reference above for the terraform logs. But there's nothing intrinsic in the function which makes it terraform specific. It just take a file path and a data stream, and fill the first one with the second one

PR #3134, opened 2018-03-23
Remove platforms - azure/gcp/govcloud/vmware/openstack

cpanato:
Cleaning master by removing `azure`, `govcloud`, `metal`, `gcp`, `openstack` and `vmware`.

Those platforms will be readded, when we advance in the new installer flow. For now we are focus on `AWS`.


cpanato:
@squat it uses the variables.bzl i changed that only. let me know if this is not good

enxebre:
as we are not supporting track-1 in this branch anymore we can remove bootkube module, rename bootkube-ut2 to bootkube and update the references in the steps folder now.

enxebre:
- can we also remove modules/openstack?
- Also I'd be in favour of removing non platform specific but unused modules e.g modules/bootstrap-ssh, dns/azure, dns/gcp wdyt?
- In the root BUILD.bazel there's glob pointing to "platforms" folder that should be removed now
- Also modules/ignition can be cleanup now, but let's do that in a different PR

sym3tri:
@squat I agree with @enxebre

bluecmd:
FYI, this broke a link on https://coreos.com/tectonic/docs/latest/install/vmware/vmware-terraform.html
> Edit the parameters with details of the VMware infrastructure. View all of the VMware specific options and the common Tectonic variables.

PR #3137, opened 2018-03-26
*: cleanup bazel rules

squat:
This commit cleans up several Bazel rules to remove mentions of code
that was previously in the Track-1 flow of the Tectonic Installer.

cc @enxebre @sym3tri @cpanato 
This PR builds off of #3134 and removes all remaining mentions of track1 resources.


squat:
most of the additions are actually the result of moving `modules/bootkube-ut2` to `modules/bootkube` and not actual additions of LoC

enxebre:
looking great to me, `make structure-check` needs to be dropped from Jenkinsfile

PR #3138, opened 2018-03-26
Documentation: add bazeless docs

squat:
cc @sym3tri @alexsomesan


Documentation/dev/bazel-in-depth.md
```diff
@@ -0,0 +1,51 @@
+# Bazel Under The Hood
+
+The goal of this document is to detail the steps taken by Bazel when building the Tectonic Installer project so that users better understand the process. Ultimately, a user building the project could elect to build the project without Bazel, either by hand or otherwise.
```
sym3tri: But this is not recommended because it would lead to a non-hermetic build which could lead to unpredictable results depending on the environment configured on the build host. Therefore we recommend using the build instructions outlined in the README for consistent reproducible builds.

squat:
Thanks @sym3tri. PTAL again when you can

PR #3140, opened 2018-03-27
Tf break down

enxebre:
Keeps breaking down tf steps to reduce impact surface and increase robustness:
- assets
- topology
- tnc_dns
- bootstrap
- joining_masters
- joining_workers

Ensures cli commands and terraform idempotency
Let's wait for https://github.com/coreos/tectonic-installer/pull/3079/files to go first


squat:
The description says this increases robustness, but I don’t think this is necessarily the case. We are introducing brittleness by depending on the interface between our terraform wrapper and terraform itself in more places now, depending on more external state imports etc. I am also skeptical of this approach because it makes it even more difficult for anyone to ever apply the installer manually, which was one of the goals of this redesign: a user should, if required, be able to use terraform manually to install a cluster, but with this many steps it will never happen.

enxebre:
Thanks for the feedback! I think the idea here is to make easier for the user, if required, to make targeted changes without impacting other parts of the cluster and potentially allow independent life cycles. The tf state boundaries here try to pursue real domain boundaries subject to be changed by the user in an isolated fashion, e.g infrastructure topology, master pools, worker pools, etc . Otherwise they can just run the cli to run the steps sequentially.
"depending on more external state imports" - That's by design. I don't think it's necessarily a bad thing. It defines contracts between domain boundaries which could even be leveraged for test driven individual parts of the cluster before an integration testing might take place.
"increase robustness" - Assuming the TNC domain flip is an invariant, this sanitise the tf state as it moves from imperatively destroying a targeted resource to a more declarative approach in which the state achieves an expected situation based on the input helping to be a bit more deterministic and achieve idempotency.

squat:
I like the idea of modularizing our code, but this change highlights the deficiencies of our design. Along with making the entire cluster creation more brittle because of the immaturity of our tool compared to Terraform, breaking the steps down more also makes the entire process take longer. Each time we break down cluster provisioning into further steps, we decrease Terraform's ability to exploit inherent parallelism in our resource graph. Moreover we can still enjoy all the benefits of improved modularity by creating logical modules in Terraform but executing them all as a single step. I do think that depending on more external state imports is inherently undesirable. It primarily exists as an escape hatch but we are using it as a principal part of our API. It makes it harder to write and maintain the Terraform configuration IMO. Furthermore, Terraform already has a best-practices way of defining API boundaries between modules, i.e. variables and outputs. The further we go in this direction the less value Terraform has as a tool to us. If that's our goal then that's fine, but in that case we should be approaching the problem differently.

enxebre:
"making the entire cluster creation more brittle because" I see this PR as actually the other way around, other than that I agree on mostly all your points:
"Each time we break down cluster provisioning into further steps, we decrease Terraform's ability to exploit inherent parallelism in our resource graph" - I agree, this is deliberately splitting and reducing the impact scope of Terraform. So we can still enjoy all of the terraform benefits within individual steps but avoiding unnecessary state coupling. This tries to focus on correctness and elucidating the right boundaries, once this is more mature we could parallelize terraform step runs to reduce deployment time.
Let's discuss internally if this is not the direction we wanna go.

enxebre:
ok to test

enxebre:
retest this please

steps/bootstrap/main.tf
```diff
@@ -9,49 +9,18 @@ provider "aws" {
...
-  private_master_endpoints = "${var.tectonic_aws_private_endpoints}"
-  public_master_endpoints  = "${var.tectonic_aws_public_endpoints}"
-}
-
 module "masters" {
   source = "../../modules/aws/master-asg"
 
   autoscaling_group_extra_tags = "${var.tectonic_autoscaling_group_extra_tags}"
-  aws_lbs                      = "${module.vpc.aws_lbs}"
+  aws_lbs                      = "${local.aws_lbs_masters}"
```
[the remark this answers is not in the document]
enxebre: `module.vpc.aws_lbs` along with all vpc infra related topology leaves in a different step - topology - then the input for the bootstrap step is consumed from the topology state and encapsulated as locals https://github.com/coreos/tectonic-installer/pull/3140/files#diff-236e14b6eaceb7ab872e1d017635b8f3R10

steps/bootstrap/main.tf
```diff
@@ -62,66 +31,14 @@ module "masters" {
...
-  master_sg_ids                = "${concat(var.tectonic_aws_master_extra_sg_ids, list(module.vpc.master_sg_id))}"
+  master_sg_ids                = "${concat(var.tectonic_aws_master_extra_sg_ids, list(local.master_sg_id))}"
   private_endpoints            = "${var.tectonic_aws_private_endpoints}"
   public_endpoints             = "${var.tectonic_aws_public_endpoints}"
   root_volume_iops             = "${var.tectonic_aws_master_root_volume_iops}"
   root_volume_size             = "${var.tectonic_aws_master_root_volume_size}"
   root_volume_type             = "${var.tectonic_aws_master_root_volume_type}"
   ssh_key                      = "${var.tectonic_aws_ssh_key}"
-  subnet_ids                   = "${module.vpc.master_subnet_ids}"
+  subnet_ids                   = "${local.subnet_ids_masters}"
```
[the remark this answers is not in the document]
enxebre: `module.vpc.master_subnet_ids` along with all vpc infra related topology leaves in a different step - topology - then the input for the bootstrap step is consumed from the topology state and encapsulated as locals https://github.com/coreos/tectonic-installer/pull/3140/files#diff-236e14b6eaceb7ab872e1d017635b8f3R10

steps/bootstrap/main.tf
```diff
@@ -62,66 +31,14 @@ module "masters" {
   extra_tags                   = "${var.tectonic_aws_extra_tags}"
   instance_count               = "1"
   master_iam_role              = "${var.tectonic_aws_master_iam_role_name}"
-  master_sg_ids                = "${concat(var.tectonic_aws_master_extra_sg_ids, list(module.vpc.master_sg_id))}"
+  master_sg_ids                = "${concat(var.tectonic_aws_master_extra_sg_ids, list(local.master_sg_id))}"
```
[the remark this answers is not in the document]
enxebre: `module.vpc.master_sg_id` along with all vpc infra related topology leaves in a different step - topology - then the input for the bootstrap step is consumed from the topology state and encapsulated as locals https://github.com/coreos/tectonic-installer/pull/3140/files#diff-236e14b6eaceb7ab872e1d017635b8f3R10

enxebre:
@trawler thanks for having a look! `module.vpc.*` is not available in the bootsrap step as explained in my comments.
Rebased and remove the redundant "master"/"worker" variable namespacing for the inputs as this is already implicit by the given step

installer/pkg/workflow/destroy.go
```diff
@@ -23,14 +23,26 @@ func destroyBootstrapStep(m *metadata) error {
 	if err := runDestroyStep(m.clusterDir, etcdStep); err != nil {
 		return err
 	}
-	return runDestroyStep(m.clusterDir, bootstrapStep)
+
+	if err := runDestroyStep(m.clusterDir, bootstrapStep); err != nil {
+		return err
+	}
+
+	if err := runDestroyStep(m.clusterDir, TNCDNSStep, "-var=tectonic_aws_bootstrap=false"); err != nil {
```
squat: the string `"-var=tectonic_aws_bootstrap=false"` has actual semantic meaning to this workflow engine. It's value is well-known and required for the step to work in an idempotent fashion. Could we consider making this a const? Or maybe abstract the `TNCDNSStep` into a func that accepts a bool and can generate two different steps based on the parameter value? The latter would be ideal IMO

enxebre: Good point. I already tried to abstract it away with `createTNCCNAME` and `createTNCARecord`. How about adding a `destroyTNCDNS` func. And in addition: 
`
const(
bootstrapOn = -var=tectonic_aws_bootstrap=true
bootstrapOff = -var=tectonic_aws_bootstrap=false
)
`
wdyt?

enxebre:
retest this please

enxebre:
retest this please

enxebre:
retest this please

enxebre:
retest this please

enxebre:
back to green, @trawler @squat PTAL

installer/pkg/workflow/utils.go
```diff
@@ -2,35 +2,33 @@ package workflow
...
-	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/tools/clientcmd"
 )
 
 const (
 	stepsBaseDir     = "steps"
 	assetsStep       = "assets"
+	topologyStep     = "topology"
+	TNCDNSStep       = "tnc_dns"
```
squat: Looks like this const is only used inside this package. Can we make it not-exported?

installer/pkg/workflow/destroy.go
```diff
@@ -23,14 +23,26 @@ func destroyBootstrapStep(m *metadata) error {
 	if err := runDestroyStep(m.clusterDir, etcdStep); err != nil {
 		return err
 	}
-	return runDestroyStep(m.clusterDir, bootstrapStep)
+
+	if err := runDestroyStep(m.clusterDir, bootstrapStep); err != nil {
```
squat: I'm confused why this single step `destroyBootstrapStep` is crossing the boundary of multiple Terraform `steps` modules. It is hard to reason about why they are distinct steps in Terraform configuration but single steps inside of this Go abstraction. Seems to me like each Terraform module should be its own step inside the workflows for both create and destroy.

PR #3141, opened 2018-03-28
tests: add parameter to -v and update jenkinsfile

cpanato:
also bump the conformance test image


cpanato:
it is failing due something in the cluster. Will merge this and we will investigate the reason

PR #3142, opened 2018-03-28
Simplified Bazel build

alexsomesan:
This change proposed a simplified build flow based on the current Bazel logic.
Notable changes introduced:

* Build only one platform a time. This implies releasing distinct tarballs for distinct architectures.
* Use Bazel datafiles syntax instead of genrule with shell-script to copy static resources.
* Remove the generated examples and replace with static cluster yaml examples.
* Remove some unnecessary rules and dependencies
* The release version is sourced from the `version.bzl` file which can be manipulated by RA.

Outstanding issues to be resolved:
* drop the terraform binary next to the installer in the local dev build (the tarball already has it).
* adapt documentation to explain the recommended developer workflow.


BUILD.bazel
```diff
@@ -16,31 +17,6 @@ config_setting(
...
 )
 
-genrule(
-    name = "terraform_runtime",
-    output_to_bindir = 1,
-    srcs = select({
-        "//:linux": ["@terraform_runtime_linux//:terraform"],
-        "//:darwin": ["@terraform_runtime_darwin//:terraform"],
-    }),
-    outs = ["bin/terraform"],
```
enxebre: terraform_fmt currently relies on this output so would need to be updated

WORKSPACE
```diff
@@ -1,28 +1,31 @@
...
+# https://github.com/bazelbuild/rules_go/pull/1393
+
 git_repository(
     name = "io_bazel_rules_go",
     remote = "https://github.com/bazelbuild/rules_go.git",
-    commit = "3f38260eda98d23e9142bb905caede5912508770"
+    commit = "361556b0d27318d1b8fa42c91a4baa4ab5ea1c58"
 )
 
+# http_archive(
```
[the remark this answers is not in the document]
alexsomesan: It's mentioned in the comment above. It's because we need a change from https://github.com/bazelbuild/rules_go/pull/1393 which is not yet released. Once a release is cut, we can revert back to http_archive retrieval. That's why I kept it commented.

BUILD.bazel
```diff
@@ -61,96 +37,55 @@ alias(
...
 alias(
     name = "cli",
-    actual = "//installer/cmd/tectonic:tectonic",
+    actual = "//installer:tectonic",
 )
 
-load("@bazel_tools//tools/build_defs/pkg:pkg.bzl", "pkg_tar")
+template_files = glob([
+    "modules/**/*",
+    "platforms/**/*",
```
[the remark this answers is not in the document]
alexsomesan: Cool. I'm going to remove it here too. Thanks!

BUILD.bazel
alexsomesan: ok, I'll put it back then 👍

BUILD.bazel
```diff
@@ -61,96 +49,56 @@ alias(
...
-
-pkg_tar(
-    name = "tarball_terraform_configuration",
-    mode = "0644",
-    srcs =  glob(["modules/**"]) + glob(["steps/**"]) + ["config.tf"],
-    strip_prefix = ".",
-)
+template_files = glob([
+    "modules/**/*",
+    "platforms/**/*",
```
enxebre: we can remove platforms

examples/BUILD.bazel
```diff
@@ -1,5 +1,9 @@
...
+
 filegroup(
-    name= "tectonic_cli_configs",
-    srcs = glob(["tectonic.*.yaml"]),
+    name= "tectonic_cli_examples",
+    srcs = example_cli_configs,
     visibility = ["//visibility:public"],
 )
+
+exports_files(example_cli_configs)
```
enxebre: missing line

enxebre (dismissed):

examples/BUILD.bazel
alexsomesan: Fixed it.

enxebre:
ok to test

enxebre (dismissed):

enxebre:
When running in mac `./tectonic-dev/installer/tectonic` is dropped with no execution perms 644, shouldn't it be 555?
Also could we the tarball name and the export binary path here https://github.com/coreos/tectonic-installer/blob/master/README.md#hacking?

squat (changes requested):
How does this PR change the tar ball layout? Can you update the Bazel README to reflect this? Also, one of @enxebre’s comments regarding globbing “platforms” should still be addressed I think

alexsomesan:
@squat Got around to amending the README. PTAL.
Also, smokes should be green now. I'll squash before merging.

alexsomesan:
Good to review. PTAL

enxebre (dismissed):

spangenberg (dismissed):

squat (approved):
Looks good! One more piece of clean up we need to do is to remove the hand-tailored gazelle generated Bazel files in vendor. These had to be modified by hand to support declarative cross-compilation in Bazel but since we are no longer doing this and instead the tarballs only have binaries for a single platform, we can remove those changes. As it stands they will *NOT* be regenerated by gazelle because of the special directive in the files. We simple need to `find ./installer/vendor/ -name *.BAZEL | xargs rm` or the like and then re-run gazelle to regenerate. We can do this in a follow up

squat:
We need to ensure we coordinate with RA team so they are aware of the new release process requiring one tarball per platform per release

alexsomesan:
@squat I agree with both points!
Would you say INST-922 would be good opportunity to handle the vendor build files? I'm still hoping we can schedule that one soon, now that the installer backend code is gone too.

On the topic of integration with RA, I'd say the release process as we know it no longer applies at this time. We definitely have to convey these changes to RA nonetheless. Sudha wants to meet this week to discuss the path forward, so we'll touch on these changes and others within the installer.

enxebre (approved):

WORKSPACE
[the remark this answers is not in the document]
wking: > Once a release is cut, we can revert back to http_archive retrieval.

I've filed openshift/installer#28 with a stab at this.

PR #3145, opened 2018-03-29
Restored update-payload module into tarball_terraform_configuration

coverprice:
The Release Automation Jenkins uses the Bazel tarball built by Installer Jenkins to create a release package, and requires the update-payload Terraform module to perform this. The update-payload module was previously excluded from the tarball, and this PR restores it.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -99,7 +99,7 @@ genrule(
 pkg_tar(
     name = "tarball_terraform_configuration",
     mode = "0644",
-    srcs =  glob(["modules/**"], exclude=["modules/update-payload/**"]) + glob(["steps/**"]) + ["config.tf"],
+    srcs =  glob(["modules/**"]) + glob(["steps/**"]) + ["config.tf"],
     strip_prefix = ".",
 )
 


coverprice:
ok to test

coverprice:
cc @colhom

squat:
This module is of no use to end-users of Tectonic. This should really be downloaded some other way IMO

alexsomesan:
I agree with @squat. If there is no foreseeable usage for this module by the end user, it does not belong in the release tarball. Let's maintain focus on keeping things at the simplest for the end user and figure out another way to use this module in RA.

One option I can suggest right away is to have a dedicated Bazel target that outputs this module (and any other required dependencies) to a separate artefact (another tarball?) that is then consumed by the RA pipeline only. This target would not be part of the build process for the release tarball, keeping it free of debris.

What do you think?

sym3tri:
I don't see the need to even have this 'update payload' module in the installer repo anymore. It's a separate concern that should live along-side the operators or release automation tooling IMO.

coverprice:
Ultimately RA needs to extract the Tectonic component manifest from `config.tf`, and IMHO the `update-payload` module is a overly-complex method of achieving that that I'd rather not use. Alternatives in my preferred order:
1. Move the manifest into a `manifest.json` file that's consumed by `config.tf` (AIUI Terraform supports this). `update-payload` module becomes obsolete, and RA can consume `manifest.json` directly from the tarball.
2. RA pipeline extracts the manifest using some Hashicorp Config Language-aware tool. This removes the need for `update-module`, but RA now depends on a new tool.
3. The module moves to the RA pipeline itself. I'd prefer not to since it retains the hacky method and seems brittle as it relies on a stable layout of Installer files.
4. A special build target is made just for the RA pipeline that would include the module.

I prefer the first option as it simplifies both Installer and RA. What do you think?

coverprice:
The payload generation is more complex than I originally thought, so the options I presented may not be viable. However I agree that the payload generation is a separate concern. I'll close this PR, migrate `update-payload` to the RA repo, and open a PR here to remove it from this repo.

sym3tri:
:+1: Thanks @coverprice

PR #3151, opened 2018-04-03
config.tf: Bump tnc version, also update the ignitoin endpoint URL.

yifan-gu:
Diff between the old and new tnc:

https://github.com/coreos-inc/tectonic-operators/compare/f6d5e710a97a8cd6f4cd2963f4426131f854a869...master


yifan-gu:
restest this please

enxebre:
Thanks for the PR @yifan-gu!
It’d be useful if we provide a link to a gh diff between both docker tags when bumping versions.
I added a commit to update the endpoint for masters/workers for the tnc (which is generated by the cli, etcd is being moving there as well), the bootstrap ign config path and livenessprobe too.
Can you please verify there's no stale properties for the new tag here https://github.com/coreos/tectonic-installer/blob/master/modules/bootkube/resources/tnc-config

yifan-gu:
@enxebre There are bunch of stale values there, but I'm removing them in my upcoming PR that adds the tnc operator.

PR #3157, opened 2018-04-04
bootkube: Add tnc operator to render and manage tnc.

yifan-gu:
Ref #https://github.com/coreos-inc/tectonic-operators/pull/316

Using the tnc operator image built from the above PR.
In `bootkube.sh`, tnc operator renders the bootstrap tnc, and its on-disk config during bootstrap.
There is also a selfhosted tnc operator manifest under bootkube manifest dir, which is for managing the long running tnc.


modules/bootkube/resources/tnco-config.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
```
enxebre: Is there anything stopping us from rendering this from golang with  c.tncoConfig() (or similar) so we completely remove boilerplate here?
Then we could also remove all related tf e.g ```data "template_file" "tnco_config"```

modules/bootkube/resources/bootkube.sh
```diff
@@ -9,8 +9,18 @@ set -e
...
+/usr/bin/docker run \
+    --user 0 \
+    --volume "$(pwd)":/assets \
+    "${tnc_operator_image}" \
+    --config=/assets/tnco-config.yaml \
+    --render-bootstrap=true \
+    --render-output=/assets/tnc-bootstrap
+
 mkdir -p /etc/kubernetes/manifests/
-mv /opt/tectonic/manifests/tectonic-node-controller-pod.yaml /etc/kubernetes/manifests/
```
enxebre: Just to clarify the casuistic here. This means tectonic-node-controller-pod.yaml will be owned by bootkube so it will be removed as part of the teardown as soon as it deploys the required pods (bootstrap cp). Then, there won't be TNC available until the TNCO gets deployed and does its job.

enxebre:
Hey @yifan-gu dropped a few comments. This is looking great!

modules/bootkube/resources/bootkube.sh
yifan-gu: Yes, but since the operator is also deployed by the bootkube, so the time window should be small. Do you think this is a potential problem?

modules/bootkube/resources/tnco-config.yaml
yifan-gu: Yes, I'd like to remove it, but not sure where to add the code for generating those on-disk configfile, same for the kco-config.

yifan-gu: OK, added the code to generate kco-config and tnco-config using go code.

modules/bootkube/resources/bootkube.sh
enxebre: I think is ok. Ignition consumers of the TNC will retry until it's reachable so it should be fine. We can discuss about having a golang entity to handle the switch over if we need it in upcoming sprints ( INST-1019 )

enxebre:
testKubeSystem is failing, at a quick glance seems aws is now uppercase

abhinavdahiya:
@yifan-gu 
add `tnco-config.yaml` to https://github.com/coreos/tectonic-installer/blob/c16b1ce6af65b2eb94b43479996f32be0e472a5b/tests/smoke/cluster_test.go#L44

installer/pkg/workflow/utils.go
```diff
@@ -75,8 +75,33 @@ func findTemplates(relativePath string) (string, error) {
 }
 
 func generateKubeConfigStep(m *metadata) error {
+	clusterGeneratedPath := filepath.Join(m.clusterDir, generatedPath)
```
enxebre: this is orthogonal for this PR, but just as a reminder for us, we should probably abstract this logic away into a config-generator packet func and just call it from here

installer/pkg/config-generator/generator.go
```diff
@@ -141,6 +162,39 @@ func (c ConfigGenerator) networkConfig() *tectonicnetwork.OperatorConfig {
...
+	}
+
+	tncoConfig.ControllerConfig = tnco.ControllerConfig{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: tnco.TNCConfigAPIVersion,
+			Kind:       tnco.TNCConfigKind,
+		},
+	}
+
+	cidrhost, err := cidrhost(c.Cluster.Networking.ServiceCIDR, 10)
```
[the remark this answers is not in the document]
yifan-gu: agree

PR #3159, opened 2018-04-05
modules: force etcd TLS client auth

coverprice:
Cherry-pick from track-1 PR (#3153)

Forces TLS client auth for etcd.
Fixes: https://jira.coreos.com/browse/INST-1026


coverprice:
ok to test

PR #3161, opened 2018-04-06
Cli ign bootstrap

enxebre:
- Move the bootstrap ign config generation from terraform into the golang ignition config-generator
- Update ign spec to 2.1.0
Follow ups:
- Once we merge this and https://github.com/coreos/tectonic-installer/pull/3157 we can clean up stale input variables from config.tf
- Bootkube and tectonic tf modules only have kube manifests now, keep moving those into the cli


enxebre:
ok to test

installer/pkg/config-generator/templates/files/bootkube.yaml
```diff
@@ -0,0 +1,28 @@
...
+        mkdir -p /etc/kubernetes/manifests/
+        mv /opt/tectonic/manifests/tectonic-node-controller-pod.yaml /etc/kubernetes/manifests/
+
+        # shellcheck disable=SC2154
+        /usr/bin/docker run \
+            --volume "$(pwd)":/assets \
+            --volume /etc/kubernetes:/etc/kubernetes \
+            --network=host \
+            --entrypoint=/bootkube \
+            "{{.BootkubeImage}}" \
```
sym3tri: Why do any of these values need to be templated at all?

installer/pkg/config-generator/ignition.go
```diff
@@ -71,6 +102,147 @@ func (c ConfigGenerator) GenerateIgnConfig(clusterDir string) error {
...
+	clusterDNSIP, err := c.getClusterDNSIP()
+	if err != nil {
+		return nil, err
+	}
+
+	bootstrapConfig := &bootstrapConfig{
+		AssetsS3Location:      c.getInitAssetsLocation(),
+		AwscliImage:           "quay.io/coreos/awscli:025a357f05242fdad6a81e8a6b520098aa65a600",
+		CloudProvider:         "aws",
+		ClusterDNSIP:          clusterDNSIP,
```
sym3tri: It seems to me this & the S3 location are the only things that need to be templated in from Go. No?

installer/pkg/config-generator/ignition.go
```diff
@@ -25,8 +30,27 @@ var (
 
 const (
 	kubeconfigKubeletPath = "generated/auth/kubeconfig-kubelet"
+	IgnTemplatesBaseDir   = "ign-templates"
```
squat: Add a comment to this exported type or keep it internal

installer/pkg/config-generator/ignition.go
enxebre: Yeah, you're totally right. I was going to open a discussion as part of the follow up for cleaning up input variables in case we still wanna allow an easy way for the user to override this images, e.g offline. I'll just remove the templating and hardcode the values for now

enxebre:
I'm moving this under the TNC repo

PR #3164, opened 2018-04-09
cli: plumb-through per-platform steps

squeed:
This lays the groundwork for per-platform steps.

If step FOO is desired, prefer `steps/FOO/aws`, but fall back to `steps/FOO` if missing.

### DIFF
--- a/installer/pkg/workflow/destroy.go
+++ b/installer/pkg/workflow/destroy.go
@@ -20,42 +20,42 @@ func NewDestroyWorkflow(clusterDir string) Workflow {
 }
 
 func destroyAssetsStep(m *metadata) error {
-	return runDestroyStep(m.clusterDir, assetsStep)
+	return runDestroyStep(m, assetsStep)
 }
 
 func destroyEtcdStep(m *metadata) error {
-	return runDestroyStep(m.clusterDir, etcdStep)
+	return runDestroyStep(m, etcdStep)
 }
 
 func destroyBootstrapStep(m *metadata) error {
-	return runDestroyStep(m.clusterDir, bootstrapStep)
+	return runDestroyStep(m, bootstrapStep)
 }
 
 func destroyTNCDNSStep(m *metadata) error {
-	return destroyTNCDNS(m.clusterDir)
+	return destroyTNCDNS(m)
 }
 
 func destroyTopologyStep(m *metadata) error {
-	return runDestroyStep(m.clusterDir, topologyStep)
+	return runDestroyStep(m, topologyStep)
 }
 
 func destroyJoinWorkersStep(m *metadata) error {
-	return runDestroyStep(m.clusterDir, joinWorkersStep)
+	return runDestroyStep(m, joinWorkersStep)
 }
 
 func destroyJoinMastersStep(m *metadata) error {
-	return runDestroyStep(m.clusterDir, joinMastersStep)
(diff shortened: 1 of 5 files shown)


enxebre:
hey @squeed thanks! I imagine there's a reason for adding this that I'm not aware of, would you mind to give some insight on that? otherwise I'm reluctant to add any abstraction until we actually need it, i.e per platform support should be added when we actually support more than one platform

squeed:
@enxebre We're currently adding libvirt as a second platform - the goal is to speed up the development cycle.

squeed:
@enxebre rebased on your PR, PTAL.

installer/pkg/workflow/utils.go
```diff
@@ -48,20 +48,31 @@ func copyFile(fromFilePath, toFilePath string) error {
 	return err
 }
 
-func findTemplates(relativePath string) (string, error) {
+// returns the directory containing templates for a given step. If platform is
```
enxebre: would be very nice to add a unit for this

enxebre:
ok to test

enxebre:
overall looks good to me. Would be easier to reason about with the addition of a new platform e.g tradeoffs of steps/aws/foo vs steps/foo/aws, etc

squeed:
Yeah, there's no obvious right answer w.r.t steps/foo/aws v.s steps/aws/foo. It depends on the abstraction boundary. Given that I don't expect steps to differ much between platforms (and after talking to alex), I picked steps/foo/aws.

installer/pkg/workflow/destroy.go
```diff
@@ -20,42 +20,42 @@ func NewDestroyWorkflow(clusterDir string) Workflow {
 }
 
 func destroyAssetsStep(m *metadata) error {
-	return runDestroyStep(m.clusterDir, assetsStep)
+	return runDestroyStep(m, assetsStep)
```
squat: ayayay this pattern is a lil bonkers. not this PR but something we need to clean up

squeed: Yes.

installer/pkg/workflow/utils.go
```diff
@@ -48,20 +48,31 @@ func copyFile(fromFilePath, toFilePath string) error {
...
+// specified, it looks for a subdirectory with platform first, falling back if
+// there are no platform-specific templates for that step
+func findStepTemplates(stepName, platform string) (string, error) {
 	base, err := baseLocation()
 	if err != nil {
-		return "", fmt.Errorf("unknown base path for '%s' templates: %s", relativePath, err)
+		return "", fmt.Errorf("Error looking up step %s templates : %v", stepName, err)
```
squat: would you mind making the `E` lowercase? by convention we typically do not capitalize error messages so they can be wrapped. Also, there is an extra space character between `templates` and `:`

installer/pkg/workflow/destroy.go
enxebre: we could also pass m.clusterDir, m.Platform here to runDestroyStep to make the dependencies explicit and less loose

installer/pkg/workflow/utils.go
squeed: ugh FINE but only if you eat more fruit.

squeed:
@squat updated, thanks.

PR #3171, opened 2018-04-10
examples: add missing required variables

squeed:
Some required variables are missing from the example yaml file. Add them.


alexsomesan:
It's by hand at this time.

PR #3148, opened 2018-04-03
module/bootkube: update k8s addons

squat:
Update kube-dns and kube-proxy to match upstream better match upstream.

cc @yifan-gu @aaronlevy 

Fixes: INST-1024

Related: #3178

PR #3178, opened 2018-04-13
module/bootkube: update k8s addons

squat:
Update kube-dns and kube-proxy to match upstream better match upstream.

cc @yifan-gu @aaronlevy 

Fixes: INST-1024

Related: #3148


squat:
ok to test

squat:
@cpanato can you take a look at this vpc test? It failed because we haven’t accepted some terms for the AMI.

PR #3183, opened 2018-04-16
Install assets only via Ignition

squeed:
This PR does the following:
- No longer zips up /generated and uploads it to S3
- Installs all needed files via Ignition
- Stops writing some files to /generated
- Cleans up the ignition "API"


squeed:
@abhinavdahiya 
In the old method, we uploaded both the master ignition file without secrets and a zip file containing secrets to S3. All this PR does is move everything that was in that zip file in to ignition. This was originally done because of size limitations on ignition files. It doesn't otherwise change the bootstrap process.

You are right that the .zip file has a different ACL. I'll change that.

The downloaded secret files are never deleted from the master - perhaps they should be, but that's not anything different in this PR. All "rm-assets.sh" does (and did) is overwrite the uploaded S3 files with empty files.

squeed:
@abhinavdahiya Changed the ignition flow to redirect to a s3:// url. Ignition can fetch private s3:// urls.

steps/assets/resources/rm-assets.sh
```diff
@@ -18,8 +18,7 @@ s3_clean() {
             set -e
             set -o pipefail
             REGION=$(wget -q -O - http://169.254.169.254/latest/meta-data/placement/availability-zone | sed '"'"'s/[a-zA-Z]$//'"'"')
-            /usr/bin/aws --region="$REGION" s3 cp /tmp/assets.zip s3://"$LOCATION/assets.zip"
-            /usr/bin/aws --region="$REGION" s3 cp /tmp/assets.zip s3://"$LOCATION/config/master"
+            /usr/bin/aws --region="$REGION" s3 cp /tmp/empty s3://"$LOCATION/config/bootstrap"
```
abhinavdahiya: Do we have 'config/bootstrap' ?

steps/topology/s3-assets.tf
```diff
@@ -39,32 +44,20 @@ resource "aws_s3_bucket_object" "tectonic_assets" {
...
-  # ID outputs, that are only computed once the assets have actually been
-  # written to disk. We re-hash the IDs (or dedicated module outputs, like module.bootkube.content_hash)
-  # to make the filename shorter, since there is no security nor collision risk anyways.
-  #
-  # Additionally, data sources do not support managing any lifecycle whatsoever,
-  # and therefore, the archive is never deleted. To avoid cluttering the module
-  # folder, we write it in the Terraform managed hidden folder `.terraform`.
-  output_path = "./.terraform/generated_${sha1("${var.tectonic_cluster_id}")}.zip"
+# The public ignition configuration
+data "ignition_config" "bootstrap_redirect" {
```
enxebre: I’ve been scratching my head to figure out how to make the http s3 endpoint more secure, this is just very nice and super simple, thanks!

squeed:
The tests have failed because it's looking for files in `/generated` that don't exist anymore.

squeed:
I'm stuck now. All of the tests expect to be able to compare `/generated/manifests` and `/generated/tectonic` with the state of the cluster. We don't write those any more (and there's no need to). Do we need to render those to disk just so tests can run?

enxebre:
The test failing is checking that some expected files are written to local disk, as this was part of the bootstrap workflow runtime https://github.com/coreos/tectonic-installer/blob/master/tests/rspec/lib/shared_examples/k8s.rb#L65
https://github.com/coreos/tectonic-installer/blob/master/tests/rspec/lib/operators.rb
Since this won't be part of the workflow anymore we don't need to test it so we can remove it.
Instead with could leverage `ssh_exec` to test those files are generated inside the machine (this would require the tests to know the bootstrap node ip)
@cpanato wdyt?
Also if we eventually move this out of tf, we can also unit test the generated ignition file to match an expected config

squeed:
The easiest answer is probably to enable a test mode that renders some files to disk.

squeed:
all green!

modules/tectonic/output.tf
```diff
@@ -35,6 +14,14 @@ output "systemd_path_unit_id" {
   value = "${data.ignition_systemd_unit.tectonic_path.id}"
 }
 
+output "ignition_file_id_list" {
+  value = [
+    "${data.ignition_file.tectonic_manifest_list.*.id}",
```
colhom: `flatten(list(...)))`

modules/tls/ca/self-signed/assets.tf
```diff
@@ -3,27 +3,92 @@ resource "local_file" "root_ca_cert" {
   filename = "./generated/tls/root-ca.crt"
 }
 
+data "ignition_file" "root_ca_cert" {
+  filesystem = "root"
+  mode       = "0644"
+  path       = "/opt/tectonic/tls/root-ca.crt"
+
+  content {
+    content = "${var.root_ca_cert_pem == "" ? join("", tls_self_signed_cert.root_ca.*.cert_pem) : var.root_ca_cert_pem}"
```
colhom: Ternary expressions conditionally consuming lists can make terraform behave "oddly". Would recommend doing something more generic like:

`${join("\n",compact(flatten(list(tls_self_signed_cert.root_ca.*.cert_pem, var.root_ca_cert_pem))))}`

Also- I'd think you'd want to join the certificate pem blocks using a newline, instead of empty string?

modules/tls/ca/self-signed/assets.tf
```diff
@@ -3,27 +3,92 @@ resource "local_file" "root_ca_cert" {
...
+}
+
 resource "local_file" "aggregator_ca_key" {
   content  = "${tls_private_key.aggregator_ca.private_key_pem}"
   filename = "./generated/tls/aggregator-ca.key"
 }
 
+data "ignition_file" "aggregator_ca_key" {
+  filesystem = "root"
+  mode       = "0644"
```
colhom: This (and all other private keys) need to be `0600` mode

modules/tls/kube/assets.tf
```diff
@@ -3,6 +3,17 @@ resource "local_file" "apiserver_key" {
   filename = "./generated/tls/apiserver.key"
 }
 
+data "ignition_file" "apiserver_key" {
+  filesystem = "root"
+  mode       = "0644"
```
colhom: 0600

modules/tls/ca/self-signed/assets.tf
```diff
@@ -3,27 +3,92 @@ resource "local_file" "root_ca_cert" {
...
+}
+
 resource "local_file" "kube_ca_key" {
   content  = "${tls_private_key.kube_ca.private_key_pem}"
   filename = "./generated/tls/kube-ca.key"
 }
 
+data "ignition_file" "kube_ca_key" {
+  filesystem = "root"
+  mode       = "0644"
```
colhom: 0600

modules/tls/etcd/assets.tf
```diff
@@ -3,11 +3,33 @@ resource "local_file" "etcd_client_cert" {
...
+}
+
 resource "local_file" "etcd_client_key" {
   content  = "${tls_private_key.etcd_client.private_key_pem}"
   filename = "./generated/tls/etcd-client.key"
 }
 
+data "ignition_file" "etcd_client_key" {
+  filesystem = "root"
+  mode       = "0644"
```
colhom: 0600

modules/tls/kube/assets.tf
```diff
@@ -12,22 +23,77 @@ resource "local_file" "apiserver_cert" {
...
+}
+
 resource "local_file" "apiserver_proxy_key" {
   content  = "${tls_private_key.apiserver_proxy.private_key_pem}"
   filename = "./generated/tls/apiserver-proxy.key"
 }
 
+data "ignition_file" "apiserver_proxy_key" {
+  filesystem = "root"
+  mode       = "0644"
```
colhom: 0600

modules/tls/kube/assets.tf
```diff
@@ -12,22 +23,77 @@ resource "local_file" "apiserver_cert" {
...
+}
+
 resource "local_file" "admin_key" {
   content  = "${tls_private_key.admin.private_key_pem}"
   filename = "./generated/tls/admin.key"
 }
 
+data "ignition_file" "admin_key" {
+  filesystem = "root"
+  mode       = "0644"
```
colhom: 0600

steps/assets/ignition-bootstrap.tf
```diff
@@ -92,12 +83,38 @@ data "template_file" "rm_assets" {
...
+  mode       = "0700"
 
   content {
-    content = "${data.template_file.rm_assets.rendered}"
+    content = "${data.template_file.rm_assets_sh.rendered}"
   }
 }
+
+data "ignition_config" "bootstrap" {
+  files = ["${compact(list(
```
colhom: This will also be a nested list situation, recommend flattening the whole list. (nice use of compact here).

steps/assets/ignition-bootstrap.tf
```diff
@@ -92,12 +83,38 @@ data "template_file" "rm_assets" {
...
+   ))}",
+    "${module.ignition_bootstrap.ignition_file_id_list}",
+    "${module.bootkube.ignition_file_id_list}",
+    "${module.tectonic.ignition_file_id_list}",
+    "${module.ca_certs.ignition_file_id_list}",
+    "${module.kube_certs.ignition_file_id_list}",
+    "${module.etcd_certs.ignition_file_id_list}",
+  ]
+
+  systemd = ["${compact(list(
```
colhom: flatten here as well

steps/assets/local-files.tf
```diff
@@ -0,0 +1,16 @@
+# some files we want rendered to disk for other use
+resource "local_file" "kubeconfig-kubelet" {
+  content  = "${module.bootkube.kubeconfig-kubelet_rendered}"
+  filename = "./generated/auth/kubeconfig-kubelet"
```
colhom: It would be nice if these generated file paths were canonically defined in a `locals` block.

steps/assets/resources/rm-assets.sh
colhom: Should also include a `set -u` here, so if `$LOCATION` is omitted bash will complain, and we don't just get a cryptic S3 API error back.

steps/topology/s3-assets.tf
```diff
@@ -39,32 +44,20 @@ resource "aws_s3_bucket_object" "tectonic_assets" {
...
+  replace {
```
colhom: What is the reason this a `replace` and not an `append` block?

squeed: Basically to keep people honest :-).

modules/bootkube/outputs.tf
```diff
@@ -1,53 +1,34 @@
...
+  value = "${data.template_file.kubeconfig.rendered}"
+}
+
+output "kubeconfig-kubelet_rendered" {
+  value = "${data.template_file.kubeconfig-kubelet.rendered}"
+}
+
+output "ignition_file_id_list" {
+  value = [
+    "${data.ignition_file.manifest_file_list.*.id}",
```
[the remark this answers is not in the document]
squeed: oh man, thanks!

modules/tls/ca/self-signed/assets.tf
squeed: I just copied this from the line above; while you're probably right, I'd rather not change what "works". I'm not sure that custom certificates work right now anyways.

steps/topology/s3-assets.tf
squeed: In other words, I don't want there any to be any chance that there's insecure ignition information out there.

squeed:
@colhom thanks a ton for the terraform tips! I reverted the directory structure change, there was no real reason for it.

modules/tls/ca/self-signed/assets.tf
colhom: a good rule of thumb w/ terraform is to only use ternary expressions when all inputs and outputs are primitives. we learned this the hard way in earlier versions of terraform, for all I know this could be perfectly fine with more current versions.

squat:
🎉

PR #3191, opened 2018-04-17
azure-terraform: custom vnet resource group in cloud config

torresdal:
If you specify custom vnet using `tectonic_azure_external_vnet_id`
and this belongs to a different resource group than where masters
and workers are located, Kubernetes Services of type LoadBalancer will
fail. Support for setting custom resource group for vnet in cloud config
was added in Kubernetes 1.6 using `vnetResourceGroup`. This will add
the `vnetResourceGroup` to the cloud config if
`tectonic_azure_external_vnet_id` is set.


torresdal:
ok to test

alexsomesan:
@torresdal Thanks for spotting this! Looks like a needed fix.
But looking over the PR I get a feeling it's missing some changes.
Looks like the part that templates the value into the cloud config is missing. Am I right?

torresdal:
@alexsomesan I don't think so. I've also tested this, and the cloud config has the new field. I added the field in `platforms/azure/main.tf` under the section `data "null_data_source" "cloud_provider"`. This has a inputs hash that later is converted to json (see extract below). Since it takes all the inputs fields and run `jsonencode` on them, that will in essence become the template.

Here's the section I'm referring to in `platforms/azure/main.tf` and the var `cloud_provider_config`:

```
module "ignition_masters" {
  source = "../../modules/ignition"

  base_domain               = "${var.tectonic_base_domain}"
  bootstrap_upgrade_cl      = "${var.tectonic_bootstrap_upgrade_cl}"
  cloud_provider            = "azure"
  cloud_provider_config     = "${jsonencode(data.null_data_source.cloud_provider.inputs)}"
  ...
}
```

Or am I missing something?

enxebre:
rebase seems incorrect, this PR should be only one commit on top of track-1

PR #3193, opened 2018-04-19
Scale test

enxebre:
- Enable scaling test - INST-1052
- Ensure `tfvars` file inputs stay consistent with `config.yaml` changes when re-running the cli, e.g changing number of masters - `refreshConfigStep` rather than `readConfigStep`
- It removes the tf import command necessity. Merge Bootstrap and join_master tf steps into one which receive a toggle parameter to dictate the number of machines. This avoids to scale in (bootstrap, 1 machine hardcoded value) and then scale out (join_masters) when rerunning `cli install`


installer/pkg/workflow/init.go
```diff
@@ -74,15 +75,33 @@ func prepareWorspaceStep(m *metadata) error {
...
 	}
 
-	m.clusterDir = filepath.Join(dir, m.cluster.Name)
-	if stat, err := os.Stat(m.clusterDir); err == nil && stat.IsDir() {
-		return fmt.Errorf("cluster directory already exists at %s", m.clusterDir)
+	if m.configFilePath == "" {
+		return errors.New("no configFilePath given for preparing the workspace")
+	}
+
+	// load initial cluster config to get cluster.Name
```
enxebre: In follow ups we'll simplify the init flow by moving the clusterDir Name to be a user input value, which makes UX sense as you could have same cluster name in different environment and we'll avoid to loadInitialClusterConfig

enxebre:
ok to test

enxebre:
retest this please

enxebre:
retest this please

enxebre:
retest this please

squeed:
This is definitely going to have some merge conflicts with the libvirt PR, but it shouldn't be too bad to fix.

steps/masters/masters.variables.tf
```diff
@@ -0,0 +1,3 @@
+variable "tectonic_aws_bootstrap" {
```
squeed: I think this should be renamed to drop the `aws`

squeed:
Very nice and much cleaner!

PR #3197, opened 2018-04-20
Refactor workflow

enxebre:
The workflow package with the metadata shared by reference across steps has given us high flexibility and ability for making changes and prototype very quickly. As product requirements get clearer and we introduce a second platform we should define clearer boundaries, stronger interfaces and refactor the package.
This proposes a refactor into a more object oriented and hermetic flavour, less prone to side effects. The main logic is in `cluster.go`


installer/pkg/config-generator/generator.go
```diff
@@ -290,6 +300,64 @@ func marshalYAML(obj interface{}) (string, error) {
 	return string(data), nil
 }
 
+// GenerateKubeConfig returns, if successful, the cluster kubernetes config
+func (c ConfigGenerator) GenerateKubeConfig(workspace string) error {
+	//configGenerator := configgenerator.New(c.config)
```
squeed: delete these comments?

installer/pkg/workflow/cluster.go
```diff
@@ -0,0 +1,312 @@
...
+	}
+	return tfApply(c.workspace, step, templateDir, extraArgs...)
+
+	return nil
+}
+
+func (c Cluster) runDestroyStep(step string, extraArgs ...string) error {
+	if !hasStateFile(c.workspace, step) {
+		// there is no statefile, therefore nothing to destroy for this step
+		return nil
```
squeed: does it make sense to log a warning here? or is this an expected state?

installer/pkg/workflow/cluster.go
```diff
@@ -0,0 +1,312 @@
...
+	templateDir, err := findStepTemplates(step, c.platform)
+	if err != nil {
+		return err
+	}
+	if err := tfInit(c.workspace, templateDir); err != nil {
+		return err
+	}
+	return tfApply(c.workspace, step, templateDir, extraArgs...)
+
+	return nil
```
squeed: extra return

installer/pkg/workflow/cluster.go
```diff
@@ -0,0 +1,312 @@
...
+	}
+	return &c, nil
+}
+
+// Assets generates, if successful, the cluster assets
+func (c Cluster) Assets() error {
+	if err := c.runInstallStep(assetsStep); err != nil {
+		return err
+	}
+	if err := c.generateKubeConfig(); err != nil {
```
squeed: just a FYI, I had to move this before the `runInstallStep(assetsStep)` as part of #3183.

alexsomesan:
Wearing the Ed hat, I'd have to ask which sprint / epic issue does this change tie into?

installer/pkg/workflow/terraform.go
```diff
@@ -51,3 +51,30 @@ func hasStateFile(stateDir string, stateName string) bool {
 	_, err := os.Stat(stepStateFile)
 	return !os.IsNotExist(err)
 }
+
+// returns the directory containing templates for a given step. If platform is
+// specified, it looks for a subdirectory with platform first, falling back if
+// there are no platform-specific templates for that step
+func findStepTemplates(stepName, platform string) (string, error) {
```
squeed: One little thing: could you convert platform to lower case? Otherwise I'll have to update that in a separate PR.

alexsomesan:
As a side note, I'd prefer that major changes like this be discussed briefly as proposals before actually putting any code down.
This is to minimize friction in incorporating the change and making sure it's in the right shape to get in.

installer/pkg/workflow/cluster.go
```diff
@@ -0,0 +1,312 @@
...
+		return err
+	}
+	if err := c.runInstallStep(joinMastersStep); err != nil {
+		return err
+	}
+	return c.runInstallStep(joinWorkersStep)
+}
+
+// Install runs, if successful, the steps to install a cluster
+func (c Cluster) Install() error {
```
squeed: This is a lot cleaner, nice.

enxebre:
hey @alexsomesan on the workspace questions, there's actually no new concept, it's just renaming clusterDir to workspace. In addition (but it's no necessary), I also added it as a flag so the user can use the same source config file to init different clusters

squeed:
As part of this, it might be nice to abstract Cluster behind an interface; that would help with multi-platform.

enxebre:
@squeed I thought about the interface but wanted to keep as simple as possible for now. Let's add the abstraction if needed with the introduction of the second platform. It also crossed my mind to have a `infraImpl` interface composing the `cluster` struct to implement the desired functionality. Once we have two will be easier to abstract more if required

installer/pkg/workflow/terraform.go
enxebre: https://github.com/coreos/tectonic-installer/pull/3197/files#diff-e22bb80d035913a0a75945c1e983a35bR88

installer/cmd/tectonic/main.go
```diff
@@ -30,33 +31,48 @@ var (
 )
 
 func main() {
-	var w workflow.Workflow
+	var c *workflow.Cluster
+	var err error
+
+	newCluster := func(clusterInstallDirFlag string) *workflow.Cluster {
```
squat: can we pull this function definition out of the `main` func? there is no need to do this inline.

squat: `clusterInstallDirFlag` is a funny name for this parameter. the function should not care if it is a flag or an argument or how the string was provided to the function. Could simply be named `dir`.

installer/cmd/tectonic/main.go
```diff
@@ -30,33 +31,48 @@ var (
...
+		l, err := log.ParseLevel(*logLevel)
```
squat: creating a cluster should be separate from parsing the log level. In fact, the way this is currently set up, log levels are not parsed for the `convert` command since that command does not execute this `newCluster` func. I think parsing the log level should be done outside of this func.

installer/pkg/config-generator/generator.go
```diff
@@ -290,6 +300,62 @@ func marshalYAML(obj interface{}) (string, error) {
 	return string(data), nil
 }
 
+// GenerateClusterConfigMaps returns, if successful, the cluster kubernetes configMaps
+func (c ConfigGenerator) GenerateClusterConfigMaps(workspace string) error {
+	clusterGeneratedPath := filepath.Join(workspace, generatedPath)
+	if err := os.MkdirAll(clusterGeneratedPath, os.ModeDir|0755); err != nil {
+		return fmt.Errorf("Failed to create cluster generated directory at %s", clusterGeneratedPath)
```
squat: We typically make errors lower-case so they can be wrapped.

installer/pkg/config-generator/generator.go
```diff
@@ -290,6 +300,62 @@ func marshalYAML(obj interface{}) (string, error) {
...
+	}
+
+	kubeSystem, err := c.KubeSystem()
+	if err != nil {
+		return err
+	}
+
+	kubePath := filepath.Join(workspace, kubeSystemPath)
+	if err := os.MkdirAll(kubePath, os.ModeDir|0755); err != nil {
+		return fmt.Errorf("Failed to create manifests directory at %s", kubePath)
```
squat: same as above

installer/pkg/config-generator/generator.go
```diff
@@ -290,6 +300,62 @@ func marshalYAML(obj interface{}) (string, error) {
...
+	}
+
+	tectonicSystem, err := c.TectonicSystem()
+	if err != nil {
+		return err
+	}
+
+	tectonicPath := filepath.Join(workspace, tectonicSystemPath)
+	if err := os.MkdirAll(tectonicPath, os.ModeDir|0755); err != nil {
+		return fmt.Errorf("Failed to create tectonic directory at %s", tectonicPath)
```
squat: same as above

installer/pkg/workflow/cluster.go
```diff
@@ -0,0 +1,311 @@
...
+type Cluster struct {
+	workspace string
+	config    config.Cluster
+	platform  string
+}
+
+// InitWorkspace generates, if successful, a workspace folder with the config.yaml
+func InitWorkspace(sourceConfigFilePath, workspaceName string) error {
+	if sourceConfigFilePath == "" {
+		errors.New("no cluster sourceConfigFilePath given for instantiating new cluster")
```
squat: should this be a return? currently this line evaluates the `errors.New` func but does nothing with the returned value

installer/pkg/workflow/cluster.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+	if workspaceName == "" {
+		errors.New("no cluster sourceConfigFilePath given for instantiating new cluster")
```
squat: same as above

installer/pkg/workflow/cluster.go
```diff
@@ -0,0 +1,311 @@
...
+
+	// generate the internal config file under the workspace folder
+	return buildInternalConfig(workspace)
+}
+
+// NewCluster creates a cluster struct from a workspace
+// It ensures cluster.config and the tfvars file are always up to date with config.yaml
+func NewCluster(workspace string) (*Cluster, error) {
+	if workspace == "" {
+		errors.New("no workspace dir given for new cluster")
```
squat: same as above

enxebre:
@squat thanks for having a look. Let's close this for now as more internal discussions is needed about the approach before going deeper into the code. cc @alexsomesan

installer/pkg/workflow/cluster.go
```diff
@@ -0,0 +1,311 @@
...
+		"aws_autoscaling_group.masters",
+		fmt.Sprintf("%s-masters", c.config.Name))
+}
+
+// readClusterConfig builds a config.Cluster from a workspace
+// it's not allowed to modify cluster.Config because we only
+// want that to happen atomically with generateTerraformVariables
+func readClusterConfig(workspace string) (*config.Cluster, error) {
+	if workspace == "" {
+		errors.New("no workspace dir given for reading config")
```
squat: This statement does not do anything with the returned value from `errors.New`. Did you mean to `return` here?

installer/pkg/workflow/cluster.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+	if err := tfInit(c.workspace, templateDir); err != nil {
+		return err
+	}
+	return tfApply(c.workspace, step, templateDir, extraArgs...)
+}
+
+func (c Cluster) runDestroyStep(step string, extraArgs ...string) error {
+	if !hasStateFile(c.workspace, step) {
+		log.Warningf("there is no statefile, therefore nothing to destroy for the step %s within %s", step, c.workspace)
```
squat: Log lines typically begin with a capital letter

installer/pkg/workflow/cluster.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+	internalCfg := config.Internal{
+		ClusterID: clusterID,
+	}
+
+	// store the content
+	yamlContent, err := yaml.Marshal(internalCfg)
+	internalFileContent := []byte("# Do not touch, auto-generated\n")
+	internalFileContent = append(internalFileContent, yamlContent...)
+	if err != nil {
```
squat: it's typical to put the error check immediately after the error declaration

PR #3204, opened 2018-04-25
Fix paths in .bazelrc for Bazel >0.12.x

alexsomesan:
This change is needed to build with Bazel versions 0.12.x and beyond.

The reason is that Bazel seems to have introduced a change in the way paths in .bazelrc are evaluated.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,2 +1,2 @@
-startup --output_base $PWD/build
-build --color=yes --workspace_status_command=$PWD/buildvars.sh
+startup --output_base build
+build --color=yes --workspace_status_command "${PWD}/buildvars.sh"
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,2 +1,2 @@
-startup --output_base ./build
-build --color=yes --workspace_status_command=./buildvars.sh
+startup --output_base $PWD/build
+build --color=yes --workspace_status_command=$PWD/buildvars.sh


.bazelrc
```diff
@@ -1,2 +1,2 @@
-startup --output_base ./build
-build --color=yes --workspace_status_command=./buildvars.sh
+startup --output_base $PWD/build
```
coverprice: Unfortunately this doesn't work, because `.bazelrc` does not expand environment vars. On my Bazel 0.12.0 it creates a directory tree `$PWD/build`. (And even if it did work, bazel would still have to be run from the workspace directory, it can't be run from any sub-directory).

`.bazelrc` has some limited support for interpolating the workspace directory using `%workspace%` macro. However it's not available for the build/startup directives.

.bazelrc
```diff
@@ -1,2 +1,2 @@
...
+build --color=yes --workspace_status_command=$PWD/buildvars.sh
```
coverprice: This change works by accident, I think. I don't believe it's expanding `$PWD` for the reasons stated above. There is some buggy path parsing in 0.12.0 so I think it's the presence of something other than '.' that's making this work.

I had success by moving `buildvars.sh` to `hack/buildvars.sh` and using that instead, albeit with the restriction that I only use the bazel client from the workspace directory.

alexsomesan:
@coverprice Thanks for spotting that! Does the new syntax I pushed just now work for you?
For me, this new one works well on macOS and Fedora.

.bazelrc
```diff
@@ -1,2 +1,2 @@
-startup --output_base $PWD/build
-build --color=yes --workspace_status_command=$PWD/buildvars.sh
+startup --output_base build
```
coverprice: This works, albeit the user must still run bazel from the `WORKSPACE` directory. I don't think there's anything we can do about that. :/

Be aware that `bazel run //:gazelle` breaks a lot of `BUILD.bazel` files within the `build` directory. This can be prevented by adding `# gazelle:exclude build` to the workspace's `BUILD.bazel`.

.bazelrc
```diff
@@ -1,2 +1,2 @@
...
+build --color=yes --workspace_status_command "${PWD}/buildvars.sh"
```
coverprice: Surprisingly, this works even when the user runs bazel outside the `WORKSPACE`. Bizarrely, `$PWD` is set to the `WORKSPACE` path when running the command, rather than the bazel client's actual PWD. I guess this is a lucky (and apparently undocumented) quirk. Here's hoping that future versions formalize this behavior.

PR #3211, opened 2018-04-27
steps: bootstrap cleanup

squat:
Cleanup bootstrap node
Fixes: INST-956

cc @enxebre @squeed

### DIFF
--- a/steps/assets/resources/rm-assets.sh
+++ b/steps/assets/resources/rm-assets.sh
@@ -23,7 +23,19 @@ s3_clean() {
         '
 }
 
+bootstrap_cleanup() {
+    rm /etc/systemd/system/multi-user.target.wants/tectonic.path
+    rm /etc/systemd/system/multi-user.target.wants/bootkube.path
+    rm /etc/systemd/system/multi-user.target.wants/rm-assets.service
+    rm /etc/systemd/system/tectonic.{path,service}
+    rm /etc/systemd/system/bootkube.{path,service}
+    rm /etc/systemd/system/rm-assets.service
+    rm -rf /opt/tectonic
+}
+
 until s3_clean; do
   echo "failed to clean up S3 assets. retrying in 5 seconds."
   sleep 5
 done
+
+bootstrap_cleanup
--- a/tests/rspec/lib/cluster.rb
+++ b/tests/rspec/lib/cluster.rb
@@ -313,7 +313,7 @@ class Cluster
   end
 
   def service_finished_bootstrapping?(ips, service)
-    command = "test -e /opt/tectonic/init_#{service}.done"
+    command = "systemctl is-active #{service} --quiet && [ $(systemctl show -p SubState --value #{service}) == \"exited\" ]"
     ips.each do |ip|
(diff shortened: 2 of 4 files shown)


steps/assets/resources/rm-assets.sh
```diff
@@ -23,7 +23,19 @@ s3_clean() {
...
 }
 
+bootstrap_cleanup() {
+    rm /etc/systemd/system/multi-user.target.wants/tectonic.path
+    rm /etc/systemd/system/multi-user.target.wants/bootkube.path
+    rm /etc/systemd/system/multi-user.target.wants/rm-assets.path
+    rm /etc/systemd/system/tectonic.{path,service}
+    rm /etc/systemd/system/bootkube.{path,service}
+    rm /etc/systemd/system/rm-assets.{path,service}
+    rm -rf /opt/tectonic
```
enxebre: The tests look for /opt/tectonic/init_tectonic.done and /opt/tectonic/init_bootkube.done to determine the services were run successfully, this might be introducing a race?

enxebre: Hum... yeh it's not clear to me why Jenkins always wins, unless the script breaks and exists before actually running `bootstrap_cleanup`. Also I can't see how we are ensuring the order since the "after" instruction was removed here https://github.com/coreos/tectonic-installer/commit/b2e0bcf2f26141706cbf96e508afd14ee351af7c#diff-1a97b193d989dc396fb00a024f0e42a8L4
We could add an spec (may be in a different PR) to verify files the node was actually cleaned up correctly.
Leaving the .done files sounds reasonable to me, It'd be nicer to leverage systemd instructions (status/is-active, etc) to check a expected status so we don't leave any clue but I'm not sure if there's simple way to do it

squat: It’s delcate since we are also removing the systemd service files. We do not run systemctl daemon-reload so it’s possible the exit state of the services is still available even without the unit files.

squeed: If all we need is a signal that the bootstrapping process was successful, then perhaps we can touch /opt/tectonic-done and keep it regardless?

tests/rspec/lib/cluster.rb
```diff
@@ -313,7 +313,7 @@ def wait_for_service(service, ips)
   end
 
   def service_finished_bootstrapping?(ips, service)
-    command = "test -e /opt/tectonic/init_#{service}.done"
+    command = "systemctl is-active #{service} --quiet && [ $(systemctl show -p SubState --value #{service}) == \"exited\" ]"
     ips.each do |ip|
```
enxebre: o you have a link to the systemd substate docs? I wonder if “exited” make a difference when the service exits because of a failure, otherwise looks good to me

squat: as written OOB: if it had failed then the service would not be active, so `systemctl is-active` would return non-zero

yifan-gu:
I think this PR causes the master to break now, can you take a look @squat ?

PR #3218, opened 2018-04-30
clean up obsolete config

enxebre:
clean up obsolete config


squeed:
seems reasonable, though the libvirt changes will definitely cause some conflicts. Watch out for the cloud provider when you merge.

squat:
@enxebre I think the real issue is not the steps/foo/pattern but rather the injection of config.tf as the default variable definition for every step. This is what is causing the duplication. It also makes it much harder to identify unused variables since we must check every step module before removing a variable declaration. We should consider breaking the config.tf into independent variables.tf for every module and construct a synthetic config.tf for config generation purposes by joining the files. This will also allow independent steps to be used in isolation, one of the benefits of the multi-step architecture, without defining useless variables.

enxebre:
@squat yep agreed, we are reaching the point where this is necessary

PR #3219, opened 2018-05-02
installer: Fix the plural definition of the aws struct.

yifan-gu:
Fixing some type definition.
Ref: https://jira.coreos.com/browse/INST-1029


enxebre:
@spangenberg PTAL

PR #3226, opened 2018-05-08
tectonic-alm-operator: update deployment for newer x-operator base image

ecordell:
This should fix tectonic-alm-operator crashlooping


squat:
ok to test

PR #3228, opened 2018-05-10
steps/joining_workers: add libvirt support

squeed:
Not immediately testable (since we don't have joining_masters), but I've tested it manually and it works.


installer/pkg/config-generator/generator.go
```diff
@@ -64,7 +64,7 @@ func New(cluster config.Cluster) ConfigGenerator {
 }
 
 // KubeSystem returns, if successful, a yaml string for the kube-system.
-func (c ConfigGenerator) KubeSystem() (string, error) {
+func (c *ConfigGenerator) KubeSystem() (string, error) {
```
enxebre: why does it need to be a pointer?

squeed: receiver methods should generally be pointers unless there's a good reason for them not to be. Otherwise, the underlying datastructure is copied for every call.

PR #3229, opened 2018-05-14
tectonic-alm-operator: Use command instead of args.

yifan-gu:
Since TCO is using 2-way merge for the containers in the manifest,
using `args` alone will not remove the old deployment's `command`,
thus new deployment spec end up with both `command` and `args`,
which will causes error during the upgrade.


yifan-gu:
test upgrade from 1.8.9-tectonic.2 1.8.9-tectonic.2

yifan-gu:
test upgrade from 1.8.9-tectonic.2 1.8.9-tectonic.2

squat:
ok to test

yifan-gu:
ok to test

PR #3231, opened 2018-05-16
modulus,steps: enable existing vpc

squat:
This commit enables the existing VPC functionality that was already
built into the installer by fixing several bugs related to the
processing of AWS private zone IDs.

Thanks to @yifan-gu's recent changes in https://github.com/coreos/tectonic-installer/pull/3219.

cc @yifan-gu @enxebre

### DIFF
--- a/modules/dns/route53/tectonic.tf
+++ b/modules/dns/route53/tectonic.tf
@@ -3,28 +3,11 @@ data "aws_route53_zone" "tectonic" {
 }
 
 locals {
-  private_zone_id = "${var.tectonic_external_private_zone == "" ?
-                        join("", aws_route53_zone.tectonic_int.*.zone_id) :
-                        var.tectonic_external_private_zone}"
-
   public_zone_id = "${join("", data.aws_route53_zone.tectonic.*.zone_id)}"
 
-  zone_id = "${var.tectonic_private_endpoints ?
-                      local.private_zone_id :
+  zone_id = "${var.private_endpoints ?
+                      var.private_zone_id :
                       local.public_zone_id}"
-}
-
-resource "aws_route53_zone" "tectonic_int" {
-  count         = "0"
-  vpc_id        = "${var.tectonic_external_vpc_id}"
-  name          = "${var.base_domain}"
-  force_destroy = true
-
-  tags = "${merge(map(
-      "Name", "${var.cluster_name}_tectonic_int_zone",
-      "KubernetesCluster", "${var.cluster_name}",
-      "tectonicClusterID", "${var.cluster_id}"
-    ), var.tectonic_extra_tags)}"
 }
 
 resource "aws_route53_record" "tectonic_api" {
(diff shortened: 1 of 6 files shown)


modules/dns/route53/tectonic.tf
```diff
@@ -3,30 +3,13 @@ data "aws_route53_zone" "tectonic" {
...
 
-  zone_id = "${var.tectonic_private_endpoints ?
-                      local.private_zone_id :
+  zone_id = "${var.private_endpoints ?
+                      var.private_zone_id :
                       local.public_zone_id}"
 }
 
-resource "aws_route53_zone" "tectonic_int" {
-  count         = "0"
```
squat: this resource was never used :/

steps/topology/aws/outputs.tf
```diff
@@ -31,7 +31,7 @@ output "worker_sg_id" {
 
 # TNC
 output "private_zone_id" {
-  value = "${aws_route53_zone.tectonic_int.id}"
+  value = "${join("", aws_route53_zone.tectonic_int.*.zone_id)}"
```
enxebre: I'm curious if you considered to use a ternary here to return either `tectonic_aws_external_private_zone` or `aws_route53_zone.tectonic_int.*.zone_id` then it would be transparent for consumers steps, so the `private_zone_id` input wouldn't need to do the check and could keep just `private_zone_id 
 = "${data.terraform_remote_state.topology.private_zone_id}"`

enxebre:
Nice! Thanks for putting this back. LGTM once is green.
Thought aside from this PR: In the future we might want to consider to avoid conditionals and to evolve the model to the point where deploying on a pre-existing vpc means just skip the "vpc" step and supply the expected user inputs for the remaining steps

steps/topology/aws/outputs.tf
squat: yes I did, but I decided against it for consistency and clarity. Otherwise, I foresee a near future where someone forgets that this variable is already the result of a ternary and adds another ternary on top. It wouldn't hurt the output but it is muddy.

dcondomitti:
I just tested this and deploying into an existing VPC worked except that my ELBs and masters have public IPs despite being in private subnets (autoassign public IP is off). I can open a separate issue for that if needed.

```
  publicEndpoints: false
  privateEndpoints: true
```

squat:
@dcondomitti yes please do; that is a separate issue

modules/dns/route53/tectonic.tf
alexsomesan: Does this mean we're not doing split horizon anymore?

squat: we still have split horizon, it was just moved somewhere else and this was never cleaned up: https://github.com/coreos/tectonic-installer/blob/master/steps/topology/aws/main.tf#L22

alexsomesan: Got it. Makes sense then. All cool!

PR #3232, opened 2018-05-16
move vendor to root folder

enxebre:
Fix INST-992
- Move installer/vendor and glide to the root folder
- remove  images/tectonic-error-server

This keeps the tests/smoke/vendor and glide files as the smoke tests currently rely on https://github.com/coreos/ktestutil which is read-only and requires old versions of kubernetes libraries: k8s.io/apimachinery, k8s.io/client-go, k8s.io/kubernetes (and their deps) - https://github.com/coreos/ktestutil/blob/master/glide.yaml
This collides with the libraries used by the cli to manage kubernetes objects https://github.com/coreos/tectonic-config/blob/master/glide.yaml and might include old undesired libraries in the final binary
Follow ups:
INST-1060
INST-1061
INST-1062


enxebre:
ok to test

squat:
alberto, how come there is a net addition of LOC if the only things that were done were

> Move installer/vendor and glide to the root folder
> remove images/tectonic-error-server

There should be a net reduction, no?

enxebre:
yep good point, indeed that captured my attention too.
The steps I followed are:
- Removed installer/vendor and glide files
- Created glide.yaml file in the top level folder adding exclude primitive for tests/ 
- Run glide install -v
- Run glide-vc --use-lock-file --no-tests --only-code
- Run bazel run //:gazelle
- Manually modify installer/BUILD.bazel to workaround INST-1060

My thinking is that the counterintuitive unbalance comes for the fact that by running `glide install -v` with no lock file, glide will update the scanned imports dependencies which are not pinned in the glide.yaml. E.g vendor/golang.org/x/sys:
https://github.com/coreos/tectonic-installer/pull/3232/files#diff-f16a80eae23d5b298c2652448ec420cfR106 vs https://github.com/coreos/tectonic-installer/pull/3232/files#diff-e63b53175840faf698629c97dbaf9dd9L105
Then if you check installer/vendor/golang.org/x/sys/unix/asm_linux_s390x.s (-28) vs vendor/golang.org/x/sys/unix/asm_linux_s390x.s (+56)
or:
installer/vendor/golang.org/x/sys/unix/zerrors_darwin_386.go (-1,673) vs vendor/golang.org/x/sys/unix/zerrors_darwin_386.go (+1,777)

enxebre:
Yeah we could, however I think this is a good chance to sanitise the vendoring process. The glide.yaml need to be modified and applied to remove k8s.io/client-go (unused) https://github.com/coreos/tectonic-installer/blob/master/installer/glide.yaml#L15 and also verify it works as expected in the new folder and with the `excludeDirs: - tests` directive. This would come up anyways the next time we wanna to update a libraries

BUILD.bazel
```diff
@@ -13,20 +13,20 @@ config_setting(
 
 config_setting(
     name = "linux",
-    values = {"cpu": "k8"}, # don't ask...
+    values = {"cpu": "k8"},  # don't ask...
```
squat: :p

PR #3233, opened 2018-05-16
installer: allow empty cidr to take default

squat:
This commit enables the Networking>xCIDR fields, which we declare to be
optional, to be left empty. It also enables validation of those fields.

cc @spangenberg @enxebre 
fixes: INST-969 and INST-1068


tests/smoke/aws/vars/aws-basic.yaml
```diff
@@ -29,9 +29,6 @@ master:
     - master
 name: aws-basic
 networking:
-  mtu: 1480
-  podCIDR: 10.2.0.0/16
-  serviceCIDR: 10.3.0.0/16
   type: canal
```
squat: The smoke tests currently rely on being able to read the networking type from the config file [0] so we cannot remove this value just yet.

[0] https://github.com/coreos/tectonic-installer/blob/master/tests/rspec/lib/config_file.rb#L21

squat:
cc @squeed @alexsomesan

installer/pkg/config/validate.go
```diff
@@ -66,6 +69,19 @@ func (c *Cluster) Validate() []error {
 	var errs []error
 	errs = append(errs, c.validateNodePools()...)
 	errs = append(errs, c.validateIgnitionFiles()...)
+	// https://en.wikipedia.org/wiki/Maximum_transmission_unit#MTUs_for_common_media
+	if err := validate.PrefixError("mtu", validate.IntRange(c.Networking.MTU, 68, 64*1024)); err != nil {
+		errs = append(errs, err)
+	}
+	if err := validate.PrefixError("podCIDR", validate.SubnetCIDR(c.Networking.PodCIDR)); err != nil {
```
squeed: People loooove to overlap podcidr and servicecidr. Can you check for that?

There is an implementation here you can crib: https://github.com/containernetworking/plugins/blob/master/plugins/ipam/host-local/backend/allocator/range.go#L125

squat: @squeed I just added overlap validation. PTAL when you can

installer/pkg/validate/validate.go
```diff
@@ -362,3 +371,48 @@ func OpenSSHPublicKey(v string) error {
...
+		*ip = ip.To16()
+		return nil
+	}
+	return fmt.Errorf("IP %q is of unknown type", ip)
+}
+
+func lastIP(cidr *net.IPNet) net.IP {
+	var last net.IP
+	for i := 0; i < len(cidr.IP); i++ {
+		last = append(last, cidr.IP[i]|^cidr.Mask[i])
```
squeed: very cute :-)

steps/assets/aws/main.tf
```diff
@@ -31,7 +31,9 @@ module assets_base {
   tectonic_ca_cert              = "${var.tectonic_ca_cert}"
   tectonic_ca_key_alg           = "${var.tectonic_ca_key_alg}"
   tectonic_ca_key               = "${var.tectonic_ca_key}"
+  tectonic_cluster_cidr         = "${var.tectonic_cluster_cidr}"
```
enxebre: why do we need to pass this value here?
I think this variable is actually not used by tf anymore so we should remove it in a different PR as part of further cleanup and per step vars review but no need to pass it here

steps/assets/aws/main.tf
```diff
@@ -31,7 +31,9 @@ module assets_base {
...
   tectonic_service_cidr         = "${var.tectonic_service_cidr}"
+  tectonic_networking           = "${var.tectonic_networking}"
```
enxebre: why do we need to pass this value here?
I think this variable is actually not used by tf anymore so we should remove it in a different PR as part of further cleanup and per step vars review but no need to pass it here

enxebre:
@squat this is nice! dropped some comments, let me know your thoughts

steps/assets/libvirt/main.tf
```diff
@@ -20,7 +20,9 @@ module assets_base {
   tectonic_ca_cert              = "${var.tectonic_ca_cert}"
   tectonic_ca_key_alg           = "${var.tectonic_ca_key_alg}"
   tectonic_ca_key               = "${var.tectonic_ca_key}"
+  tectonic_cluster_cidr         = "${var.tectonic_cluster_cidr}"
```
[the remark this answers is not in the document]
squat: We must pass this value here because the entire `/step/x/base` pattern is broken. Every variable that could potentially be used by the underlying step needs to have variables explicitly passed along. Otherwise: variables with defaults will use the defaults instead of the user-specified values and variables without defaults will cause the module to break.

steps/assets/libvirt/main.tf
```diff
@@ -20,7 +20,9 @@ module assets_base {
...
   tectonic_service_cidr         = "${var.tectonic_service_cidr}"
+  tectonic_networking           = "${var.tectonic_networking}"
```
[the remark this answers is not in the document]
squat: We must pass this value here because the entire `/step/x/base` pattern is broken. Every variable that could potentially be used by the underlying step needs to have variables explicitly passed along. Otherwise: variables with defaults will use the defaults instead of the user-specified values and variables without defaults will cause the module to break.

steps/assets/aws/main.tf
squat: We must pass this value here because the entire `/step/x/base` pattern is broken. Every variable that could potentially be used by the underlying step needs to have variables explicitly passed along. Otherwise: variables with defaults will use the defaults instead of the user-specified values and variables without defaults will cause the module to break.

steps/assets/aws/main.tf
squat: We must pass this value here because the entire `/step/x/base` pattern is broken. Every variable that could potentially be used by the underlying step needs to have variables explicitly passed along. Otherwise: variables with defaults will use the defaults instead of the user-specified values and variables without defaults will cause the module to break.

squat:
@enxebre thanks for taking a look. I prefer to do any extra cleanup of removing variables in a separate cleanup PR since this PR is already grown large enough

steps/assets/aws/main.tf
enxebre: yup, until we segregate variables per step we need to inject the ones needed into /base, however It seems pointless to inject the ones we know won't be used. Not a big deal anyways, let's approach it in different PR

PR #3234, opened 2018-05-16
rspec/aws_vpc: retry terraform apply, reduce flakiness

colhom:
\cc @coverprice


colhom:
I realize that the specs which use this block of code are now disabled. Closing this.

PR #3235, opened 2018-05-16
Cherry-pick of #3229 from 1.9.6 to track-1

coverprice:
Since TCO is using 2-way merge for the containers in the manifest,
using args alone will not remove the old deployment's command,
thus new deployment spec end up with both command and args,
which will causes error during the upgrade.


coverprice:
ok to test

squat:
ok to test

coverprice:
retest this please

PR #3236, opened 2018-05-17
 installer/pkg/config: validate cluster name

squat:
This commit adds cluster name validation and also ensures that running `tectonic init` performs config validation.

Fixes: INST-991

Should be merged AFTER #3233, as the first commit is just that PR
cc @enxebre @alexsomesan

### DIFF
--- a/installer/pkg/workflow/utils.go
+++ b/installer/pkg/workflow/utils.go
@@ -11,8 +11,6 @@ import (
 
 	"github.com/coreos/tectonic-installer/installer/pkg/config"
 	configgenerator "github.com/coreos/tectonic-installer/installer/pkg/config-generator"
-
-	log "github.com/Sirupsen/logrus"
 )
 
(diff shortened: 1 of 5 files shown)


installer/pkg/workflow/utils.go
```diff
@@ -149,17 +147,6 @@ func readClusterConfig(configFilePath string, internalFilePath string) (*config.
 	return cfg, nil
 }
 
-func getClusterNameFromConfig(configFilePath string) (*string, error) {
```
squat: this func was not necessary. we were already validating that `configFilePath` was not empty so the only thing this func did was access a struct member

enxebre: yeah I originally put the func there to make it clear for others that during the current `init` process it only need to read the config file to get the cluster name. This will change as we refine the `init` process

PR #3238, opened 2018-05-17
installer/pkg: validate s3 bucket

squat:
This commit adds S3 Bucket name validation and also adds some more structure to platform comparisons.

Fixes: INST-996

Should be merged AFTER #3233 and #3236, as the first two commits correspond to those PRs.
cc @enxebre @alexsomesan


installer/pkg/config/validate.go
```diff
@@ -151,6 +158,24 @@ func (c *Cluster) ValidateAndLog() error {
...
 
+// validateTNCS3Bucket does some basic validation to ensure that the TNC bucket
+// matches the S3 bucket naming rules. Not all rules are checked
+// because Tectonic controls the generation of S3 bucket names, creating
+// buckets of the form: <cluster-name>-<tnc>.<domain-name>
+func (c *Cluster) validateTNCS3Bucket() error {
+	if c.Platform != PlatformAWS {
+		return nil
+	}
+	bucket := strings.Join([]string{fmt.Sprintf("%s-tnc", c.Name), c.BaseDomain}, ".")
```
enxebre: nit: personally I'd find easy to read just `sprintf(%s-tnc.%s, c.Name, c.BaseDomain)`

installer/pkg/config/validate.go
```diff
@@ -151,6 +158,24 @@ func (c *Cluster) ValidateAndLog() error {
...
+	if len(bucket) > 63 {
```
enxebre: Might put 63 in a constant

installer/pkg/config/validate.go
```diff
@@ -151,6 +158,24 @@ func (c *Cluster) ValidateAndLog() error {
...
+		return errors.New("the S3 bucket name generated from the cluster name and base domain is too long; S3 bucket names must be less than 63 characters; please choose a shorter cluster name or base domain")
```
enxebre: printing `bucket` in the error would give immediate info to the user

installer/pkg/config/validate.go
```diff
@@ -151,6 +158,24 @@ func (c *Cluster) ValidateAndLog() error {
...
+	}
+	if !regexp.MustCompile("^[a-z0-9][a-z0-9-.]{1,61}[a-z0-9]$").MatchString(bucket) {
```
enxebre: Couldn't we check the length, then just `validate.DomainName(bucket)`?

squat: Well, there are valid domain names that are not valid bucket names. This was optimistic so we could later implement the stricter s3 bucket name rules though right now there is no difference.

squat:
thanks @enxebre. i made some updates based on your comments

installer/pkg/config/validate_test.go
```diff
@@ -331,6 +331,72 @@ func TestSharedNodePool(t *testing.T) {
...
+				Name:       ".foo",
+				BaseDomain: "example.com",
+				Platform:   PlatformAWS,
+			},
+			err: true,
+		},
+		{
+			cluster: Cluster{
+				Name:       "foo",
+				BaseDomain: "example.com.",
```
squat: valid domain name but invalid s3 bucket name

PR #3240, opened 2018-05-19
Remove public resources in private VPC deployment

dcondomitti:
False values for booleans are omitted when marshaling JSON in golang.
https://github.com/golang/go/issues/13284

fixes INST-1083


yifan-gu:
ok to test

yifan-gu:
ok to test

PR #3241, opened 2018-05-23
Gazelle cmd

enxebre:
Fix INST-1060. Add gazelle idempotency

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -49,7 +49,7 @@ alias(
 
 alias(
     name = "cli",
-    actual = "//installer:tectonic",
+    actual = "//installer/cmd/tectonic",
 )
 
 template_files = glob([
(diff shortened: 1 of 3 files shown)


enxebre:
ok to test

BUILD.bazel
```diff
@@ -77,12 +77,17 @@ pkg_tar(
...
-        "//installer:tectonic",
     ],
     extension = "tar.gz",
     package_dir = "tectonic-%s" % TECTONIC_VERSION,
     strip_prefix = ".",
-    deps = [":tf_bin"],
+    deps = [":tf_bin", ":cli_bin"],
+)
+
+pkg_tar(
```
enxebre: Because the binary is now generated by `installer/cmd/tectonic` (which is generated by gazelle) we ensure with pkg_tar that is dropped in the expected path at build time, i.e `installer/`

installer/BUILD.bazel
```diff
@@ -9,27 +7,3 @@ test_suite(
...
-    ],
-    visibility = ["//visibility:private"],
-    deps = [
-        "//installer/pkg/workflow:go_default_library",
-        "//vendor/github.com/Sirupsen/logrus:go_default_library",
-        "//vendor/gopkg.in/alecthomas/kingpin.v2:go_default_library",
-    ],
-)
-
-go_binary(
```
squat: Thanks for reverting this to how it used to be, this was very painful for Bazel/gazelle.

PR #3243, opened 2018-05-24
TNC: platform support changes

squeed:
Generate the new TNC configuration object with platforms.

Bumps the tnco to pick up the changes.


squeed:
do not merge until coreos-inc/tectonic-operators#372 is merged.

PR #3250, opened 2018-05-30
*: remove external etcd

squat:
Remove the external etcd option.

Part of: INST-974


modules/aws/etcd/nodes.tf
```diff
@@ -41,9 +40,8 @@ data "aws_iam_role" "etcd_role" {
 }
 
 resource "aws_iam_role" "etcd_role" {
-  count = "${length(var.external_endpoints) == 0 && var.etcd_iam_role == "" ? 1 : 0}"
```
enxebre: I think we need to keep the condition for `var.etcd_iam_role == "" ?` here

squat:
This PR has been around for almost a week so I had to rebase due to merge conflicts. Can you PTAL @enxebre @spangenberg

PR #3254, opened 2018-05-31
inst-1093: fix rm-assets race conditions

trawler:
this PR fixes inst-1093 rm-assets service runs even after tectonic.service failed.


trawler:
ok to test

PR #3259, opened 2018-06-01
config.tf: bump node controller operator

crawford:
This new version includes the removal of the following:

  - coreos-metadata.service - This metadata from this service (typically
        the public IP address) isn't actually used by etcd.
  - k8s-node-bootstrap.yaml - This service used to be responsible for
        installing the correct version of docker and kubelet. Now that
        TNC exists, it can inject the correct versions directly.
  - runtime-mappings.yaml - This was used by k8s-node-bootstrap as a
        fallback for when it was unable to contact the cluster.


squat:
ok to test

PR #3262, opened 2018-06-05
*: remove tls_validity_period

squat:
This commit removes the tls validity period option from the yaml config.
This options is not used anywhere in terraform.

cc @trawler @spangenberg  @enxebre

### DIFF
--- a/installer/pkg/config/cluster.go
+++ b/installer/pkg/config/cluster.go
@@ -45,26 +45,25 @@ var defaultCluster = Cluster{
 
 // Cluster defines the config for a cluster.
 type Cluster struct {
-	Admin             `json:",inline" yaml:"admin,omitempty"`
-	BaseDomain        string `json:"tectonic_base_domain,omitempty" yaml:"baseDomain,omitempty"`
-	CA                `json:",inline" yaml:"ca,omitempty"`
-	ContainerLinux    `json:",inline" yaml:"containerLinux,omitempty"`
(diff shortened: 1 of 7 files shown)


installer/pkg/config/cluster.go
```diff
@@ -45,26 +45,25 @@ var defaultCluster = Cluster{
...
-	Etcd              `json:",inline" yaml:"etcd,omitempty"`
-	Internal          `json:",inline" yaml:"-"`
-	LicensePath       string `json:"tectonic_license_path,omitempty" yaml:"licensePath,omitempty"`
-	Master            `json:",inline" yaml:"master,omitempty"`
-	Name              string `json:"tectonic_cluster_name,omitempty" yaml:"name,omitempty"`
-	Networking        `json:",inline" yaml:"networking,omitempty"`
-	NodePools         `json:"-" yaml:"nodePools"`
-	Platform          Platform `json:"tectonic_platform" yaml:"platform,omitempty"`
-	PullSecretPath    string   `json:"tectonic_pull_secret_path,omitempty" yaml:"pullSecretPath,omitempty"`
-	TLSValidityPeriod int      `json:"tectonic_tls_validity_period,omitempty" yaml:"tlsValidityPeriod,omitempty"`
```
squat: The only difference here is the removal of the `TLSValidityPeriod` field. The rest is white space.

PR #3265, opened 2018-06-06
installer: validate libvirt

squat:
This commit adds validation for all libvirt fields. Additionally, it
moves the default values for these validated fields from Terraform to
Golang so that the defaults are handled consistently and we do not
accidentally omit configuration values by forgetting to plumb values
through Terraform.

Addresses: INST-1089

CC @enxebre @squeed 

We need to set up our libvirt smoke tests so we can properly test future changes like this.

### DIFF
--- a/installer/pkg/config/libvirt/libvirt.go
+++ b/installer/pkg/config/libvirt/libvirt.go
@@ -7,32 +7,39 @@ import (
 	"github.com/apparentlymart/go-cidr/cidr"
 )
 
+const (
+	// DefaultDNSServer is the default DNS server for libvirt.
+	DefaultDNSServer = "8.8.8.8"
+	// DefaultIfName is the default interface name for libvirt.
+	DefaultIfName = "osbr0"
+)
+
 // Libvirt-specific configuration
 type Libvirt struct {
-	URI          string `json:"tectonic_libvirt_uri,omitempty" yaml:"uri"`
-	SshKey       string `json:"tectonic_libvirt_ssh_key,omitempty" yaml:"sshKey"`
-	QowImagePath string `json:"tectonic_coreos_qow_path,omitempty" yaml:"imagePath"`
-	Network      `json:",inline" yaml:"network"`
-	MasterIPs    []string `json:"tectonic_libvirt_master_ips,omitempty" yaml:"masterIps"`
+	URI           string `json:"tectonic_libvirt_uri,omitempty" yaml:"uri"`
+	SSHKey        string `json:"tectonic_libvirt_ssh_key,omitempty" yaml:"sshKey"`
+	QCOWImagePath string `json:"tectonic_coreos_qcow_path,omitempty" yaml:"imagePath"`
+	Network       `json:",inline" yaml:"network"`
+	MasterIPs     []string `json:"tectonic_libvirt_master_ips,omitempty" yaml:"masterIPs"`
 }
 
 type Network struct {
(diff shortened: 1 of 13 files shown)


installer/pkg/config/libvirt/libvirt.go
```diff
@@ -7,32 +7,39 @@ import (
 	"github.com/apparentlymart/go-cidr/cidr"
 )
 
+const (
+	// DefaultDNSServer is the default DNS server for libvirt.
+	DefaultDNSServer = "8.8.8.8"
+	// DefaultIfName is the default interface name for libvirt.
+	DefaultIfName = "tt0"
```
squeed: Let's get "tectonic" out of this - how about `osbr0`

installer/pkg/config/validate.go
```diff
@@ -104,6 +105,46 @@ func (c *Cluster) validateAWS() []error {
 	return errs
 }
 
+// validateLibvirt validates all fields specific to libvirt.
+func (c *Cluster) validateLibvirt() []error {
+	var errs []error
+	if c.Platform != PlatformLibvirt {
+		return errs
+	}
+	if err := validate.PrefixError("libvirt network ipRange", validate.SubnetCIDR(c.Libvirt.Network.IPRange)); err != nil {
```
squeed: It's not 100% necessary, but if you want to validate that this doesn't overlap with the cluster and service cidrs, that would be awesome!

installer/pkg/config/libvirt/libvirt.go
squat: sounds good. I should have confirmed with you earlier, I just grabbed what was in defaulted in the Terraform config.

installer/pkg/config/validate.go
squat: :+1:

installer/pkg/config/validate.go
```diff
@@ -104,6 +105,46 @@ func (c *Cluster) validateAWS() []error {
...
+		for i, ip := range c.Libvirt.MasterIPs {
+			if err := validate.PrefixError(fmt.Sprintf("libvirt masterIPs[%d] %q", i, ip), validate.IPv4(ip)); err != nil {
+				errs = append(errs, err)
+			}
+		}
+	}
+	if err := validate.PrefixError("libvirt uri", validate.NonEmpty(c.Libvirt.URI)); err != nil {
+		errs = append(errs, err)
+	}
+	if err := validate.PrefixError("libvirt imagePath", validate.NonEmpty(c.Libvirt.QOWImagePath)); err != nil {
```
squeed: Is it possible to validate that this is a valid path on disk? Even more amazing would be validating that the first four bytes match `QFI\xFB`

squeed:
A few suggestions, but looks great!

installer/pkg/config/validate.go
squat: cool idea!

installer/pkg/config/libvirt/libvirt.go
squeed: at this point, it's completely arbitrary. oh, did you know that Linux interface names have no restrictions on characters? Let's just call it :bridge:

squat:
Thanks @squeed! I made the requested additions. It meant some refactoring of the CIDR comparison code so the PR grew a little bit. Please take another look when you can.

squat:
I also replaced all occurrences of `qow` with `qcow`

squat:
rebased again hehe

PR #3267, opened 2018-06-07
Add a new CI Dockerfile for testing with CentOS and prow

smarterclayton:
The CI infra will use this as a shim for testing until we evolve the
installer image into a final form.

Hold briefly while I test the jobs on this.

### DIFF
--- a/images/tectonic-installer/Dockerfile.ci
+++ b/images/tectonic-installer/Dockerfile.ci
@@ -1,0 +1,28 @@
+# This Dockerfile is a used as a CI job to validate Tectonic installer on CentOS
+
+FROM openshift/origin-release:golang-1.9 as build
+WORKDIR /go/src/github.com/coreos/tectonic-installer
+COPY . .
+### Install Terraform
+ENV TERRAFORM_VERSION="0.11.1"
+# Install Terraform
+# TERRAFORM_URL enables us to build the upstream-terraform Tectonic builder
+# image (See README.md/##Upstream-and-CoreOS-Terraform)
+ARG TERRAFORM_URL=https://releases.hashicorp.com/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_linux_amd64.zip
+
+RUN go build -o ./installer/tectonic ./installer/cmd/tectonic && \
+    yum install -y unzip && \
+    yum clean all && \
+    curl -L ${TERRAFORM_URL} | funzip > ./installer/terraform && chmod +x ./installer/terraform && \
+    tar cvzf /opt/tectonic-dev.tar.gz ./installer/tectonic ./installer/terraform ./modules ./config.tf ./steps
+
+FROM centos:7
+COPY --from=build /opt/tectonic-dev.tar.gz /tmp/tectonic-dev.tar.gz
+
+ENV HOME /home/user
+ENV PATH $PATH:/home/user/installer
+RUN yum install -y gettext && \
+    yum clean all && \
+    mkdir -p ${HOME}/installer && chmod 777 -R ${HOME} && \
+    tar xzf /tmp/tectonic-dev.tar.gz -C ${HOME}
+WORKDIR /home/user


smarterclayton:
Here's an example of what the job output looks like when this runs

https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/coreos_tectonic-installer/3267/pull-ci-origin-installer-unit/1/

and here's what the definition looks like using this Dockerfile

https://github.com/openshift/release/blob/master/projects/openshift-installer/config.json

Multiple jobs are hooked up to the same definition - when a PR lands, prow kicks off each job individually, but the jobs all share the same common images / code.  My next step will be to try to use this to launch a cluster as an "add-on" job in the openshift accounts and see how everything runs.

images/tectonic-installer/Dockerfile.ci
```diff
@@ -0,0 +1,28 @@
...
+WORKDIR /go/src/github.com/coreos/tectonic-installer
+COPY . .
+### Install Terraform
+ENV TERRAFORM_VERSION="0.11.1"
+# Install Terraform
+# TERRAFORM_URL enables us to build the upstream-terraform Tectonic builder
+# image (See README.md/##Upstream-and-CoreOS-Terraform)
+ARG TERRAFORM_URL=https://releases.hashicorp.com/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_linux_amd64.zip
+
+RUN go build -o ./installer/tectonic ./installer/cmd/tectonic && \
```
enxebre: I'm confused about what is this actually doing?

smarterclayton: Performing a simple go build of the installer (just enough to run an install).  I'm trying to kill a few birds with one stone:

1. familiarize myself with the build toolchain so I know what is coming upstream
2. check for any places where bazel is needed (it doesn't look like it) since that's more effort to turn into official builds
3. get a simple single file dockerfile that we can potentially run identically with multiple bases in CI or out (fedora, centos, minimal image, rhel, etc)

smarterclayton:
Here's a link to the e2e run (something is failing I still have to debug, so the cluster never actually comes up but fails in bootkube sometime after the pivot) so you can see the log output and artifacts in a success case: https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/coreos_tectonic-installer/3267/pull-ci-origin-installer-e2e-aws/24/

images/tectonic-installer/Dockerfile.ci
alexsomesan: @smarterclayton Bazel was a decision we made to help us gain control of the user deliverables in the context of Tectonic as an end-user product. It gave us strong guarantees that the contents of what we ship come from where we expect them to (reproducible builds). Not once has it happened that we shipped wrong versions of our dependencies because of "handcrafted" release scripts.

That being said, that made sense back when Tectonic was shipping directly to customers and no other process was available. It's worth reconsidering if Bazel still makes sense in the RedHat context. I'm not familiar with the options we have now.

smarterclayton: Yeah, I had talked to Crawford a bit about it.  It's not that Bazel is bad on its own, it just doesn't integrate well into systems that already do reproducible builds (like the koji/brew build systems for Fedora/CentOS and debian).  Most of the assumptions bazel makes are either not helpful when layered on top of another reproducible build system (that for instance requires end to end source and rebuild capability, blocks external dependencies, already has a strong versioning and scope management tool, etc).  It's not that bazel can't eventually fit into that, but it will take some time.  I like a lot of bazel, am so-so on other parts at smaller scale.  

Generally the release scripts are "outside of" the repo context (for either images or the content within), but I agree that handrolled is bad.  This is more of a bridge step that allows us to simulate those environments (where we have hermetic inputs delivered via RPM) while still having a familiar Dockerfile.  So the "release" dockerfile might look like this but instead say "yum install terraform" where terraform was an external RPM managed dependency with its own hermetic seal, and the RPM is brought into scope by the build system at the point we tie the image (koji / brew have "tags" which bring together explicit versions of RPMs into a pool that can be used to build new content).

smarterclayton:
Any reason not to merge this for use with the new additional CI jobs?  I'll take point on keeping it up to date for the tests until we're further along on open tonic.

PR #3270, opened 2018-06-08
Spike opentonic green

enxebre:


modules/aws/vpc/sg-master.tf
```diff
@@ -65,7 +65,7 @@ resource "aws_security_group_rule" "master_ingress_https" {
   protocol    = "tcp"
   cidr_blocks = ["0.0.0.0/0"]
   from_port   = 6443
-  to_port     = 6443
+  to_port     = 6445
```
squat: what's running on 6444 and 6445?

modules/bootkube/resources/bootkube.sh
```diff
@@ -58,9 +58,10 @@ export ETCDCTL_API=
 
 # shellcheck disable=SC2154
 /usr/bin/docker kill $${signer_id}
-
 rm /etc/kubernetes/manifests/tectonic-node-controller-pod.yaml
 
+cp -r $(pwd)/bootstrap-configs /etc/kubernetes/bootstrap-configs
```
squat: why not just `cp -r bootstrap-configs /etc/kubernetes/bootstrap-conf` without the `$(pwd)`?

modules/bootkube/resources/manifests/03-openshift-web-console-namespace.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  # This is the namespace used to hold the openshift console.
+  # They require openshift console run in this namespace.
```
squat: Who is `they`?

modules/ignition/resources/kubernetes/runtime-mappings.yaml
```diff
@@ -9,3 +9,5 @@ versions:
         docker: [ "17.03", "1.12"]
     1.9:
         docker: [ "17.03", "1.12"]
+    3.10:
+        docker: [ "17.03", "1.12"]
```
squat: we should keep the newline

modules/ignition/resources/services/kubelet.service
```diff
@@ -2,44 +2,48 @@
...
   --volume=resolv,kind=host,source=/etc/resolv.conf \
   --mount volume=resolv,target=/etc/resolv.conf \
   --volume var-lib-cni,kind=host,source=/var/lib/cni \
   --mount volume=var-lib-cni,target=/var/lib/cni \
   --volume var-lib-kubelet,kind=host,source=/var/lib/kubelet \
   --mount volume=var-lib-kubelet,target=/var/lib/kubelet \
   --volume var-log,kind=host,source=/var/log \
-  --mount volume=var-log,target=/var/log"
+  --mount volume=var-log,target=/var/log \
+  --insecure-options=image"
```
squat: Why do we need to use `--insecure-options`? Can we instead add a task to make the image secure?

modules/tectonic/resources/manifests/secrets/ingress-tls.yaml
```diff
@@ -1,9 +1,10 @@
...
-  namespace: tectonic-system
+  name: tectonic-ingress-tls
+  namespace: tectonic-ingress
 type: Opaque
 data:
   tls.crt: ${ingress_tls_cert}
-  tls.key: ${ingress_tls_key}
\ No newline at end of file
+  tls.key: ${ingress_tls_key}
+  bundle.crt: ${ingress_tls_bundle}
```
squat: I see that this commit didn't introduce the mistake but we should add a newline here.

modules/ignition/resources/services/kubelet.service
squeed: because we're pulling a docker image instead of an ACI. No choice :-(

squeed: This goes away when we switch to quay.

squat: Right, this is my point. Can we either:
a: switch to quay right now
or
b: make a task to switch to quay and remove this flag

squeed:
I did a quick smoketest of this on libvirt, and it's good enough for now. There are some followup items:

1. tectonic-torcx (k8s-node-bootstrap.service) isn't working and should probably be deleted
2. I accidentally ran out of disk space on my testing VM and the openshift-apiserver was evicted. It needs the critical tag.

But this looks good! LGTM.

PR #3273, opened 2018-06-08
installer/pkg/tls: Create Package TLS

trawler:
This PR adds input vars + validation for the TLS vars that will later be used to generate the tls certificates. Fixes #INST-1106

### DIFF
--- a/installer/cmd/tectonic/main.go
+++ b/installer/cmd/tectonic/main.go
@@ -35,21 +35,21 @@ func main() {
 
 	switch kingpin.Parse() {
 	case clusterInitCommand.FullCommand():
-		w = workflow.NewInitWorkflow(*clusterInitConfigFlag)
+		w = workflow.InitWorkflow(*clusterInitConfigFlag)
 	case clusterInstallFullCommand.FullCommand():
-		w = workflow.NewInstallFullWorkflow(*clusterInstallDirFlag)
+		w = workflow.InstallFullWorkflow(*clusterInstallDirFlag)
 	case clusterInstallTLSCommand.FullCommand():
-		w = workflow.NewInstallTLSWorkflow(*clusterInstallDirFlag)
+		w = workflow.InstallTLSWorkflow(*clusterInstallDirFlag)
 	case clusterInstallAssetsCommand.FullCommand():
-		w = workflow.NewInstallAssetsWorkflow(*clusterInstallDirFlag)
+		w = workflow.InstallAssetsWorkflow(*clusterInstallDirFlag)
 	case clusterInstallBootstrapCommand.FullCommand():
-		w = workflow.NewInstallBootstrapWorkflow(*clusterInstallDirFlag)
+		w = workflow.InstallBootstrapWorkflow(*clusterInstallDirFlag)
 	case clusterInstallJoinCommand.FullCommand():
-		w = workflow.NewInstallJoinWorkflow(*clusterInstallDirFlag)
+		w = workflow.InstallJoinWorkflow(*clusterInstallDirFlag)
 	case clusterDestroyCommand.FullCommand():
-		w = workflow.NewDestroyWorkflow(*clusterDestroyDirFlag)
+		w = workflow.DestroyWorkflow(*clusterDestroyDirFlag)
 	case convertCommand.FullCommand():
-		w = workflow.NewConvertWorkflow(*convertConfigFlag)
+		w = workflow.ConvertWorkflow(*convertConfigFlag)
 	}
 
 	l, err := log.ParseLevel(*logLevel)
--- a/installer/pkg/workflow/convert.go
+++ b/installer/pkg/workflow/convert.go
@@ -8,9 +8,9 @@ import (
 	"github.com/coreos/tectonic-installer/installer/pkg/config"
 )
 
-// NewConvertWorkflow creates new instances of the 'convert' workflow,
+// ConvertWorkflow creates new instances of the 'convert' workflow,
 // responsible for converting an old cluster config.
-func NewConvertWorkflow(configFilePath string) Workflow {
+func ConvertWorkflow(configFilePath string) Workflow {
 	return Workflow{
 		metadata: metadata{configFilePath: configFilePath},
 		steps: []Step{
--- a/installer/pkg/workflow/destroy.go
+++ b/installer/pkg/workflow/destroy.go
@@ -1,9 +1,9 @@
 package workflow
 
-// NewDestroyWorkflow creates new instances of the 'destroy' workflow,
+// DestroyWorkflow creates new instances of the 'destroy' workflow,
 // responsible for running the actions required to remove resources
 // of an existing cluster and clean up any remaining artefacts.
-func NewDestroyWorkflow(clusterDir string) Workflow {
+func DestroyWorkflow(clusterDir string) Workflow {
 	return Workflow{
 		metadata: metadata{clusterDir: clusterDir},
 		steps: []Step{
--- a/installer/pkg/workflow/init.go
+++ b/installer/pkg/workflow/init.go
@@ -23,9 +23,9 @@ const (
 	terraformVariablesFileName = "terraform.tfvars"
 )
 
-// NewInitWorkflow creates new instances of the 'init' workflow,
+// InitWorkflow creates new instances of the 'init' workflow,
 // responsible for initializing a new cluster.
-func NewInitWorkflow(configFilePath string) Workflow {
+func InitWorkflow(configFilePath string) Workflow {
 	return Workflow{
 		metadata: metadata{configFilePath: configFilePath},
 		steps: []Step{
--- a/installer/pkg/workflow/install.go
+++ b/installer/pkg/workflow/install.go
@@ -2,9 +2,9 @@ package workflow
 
 import "github.com/coreos/tectonic-installer/installer/pkg/config-generator"
 
-// NewInstallFullWorkflow creates new instances of the 'install' workflow,
+// InstallFullWorkflow creates new instances of the 'install' workflow,
 // responsible for running the actions necessary to install a new cluster.
-func NewInstallFullWorkflow(clusterDir string) Workflow {
+func InstallFullWorkflow(clusterDir string) Workflow {
 	return Workflow{
 		metadata: metadata{clusterDir: clusterDir},
 		steps: []Step{
@@ -26,9 +26,9 @@ func NewInstallFullWorkflow(clusterDir string) Workflow {
 	}
 }
 
-// NewInstallTLSWorkflow creates the TLS assets, previously created by the
+// InstallTLSWorkflow creates the TLS assets, previously created by the
 // "assets" step
-func NewInstallTLSWorkflow(clusterDir string) Workflow {
+func InstallTLSWorkflow(clusterDir string) Workflow {
 	return Workflow{
 		metadata: metadata{clusterDir: clusterDir},
 		steps: []Step{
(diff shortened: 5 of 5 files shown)


installer/pkg/tls/tls.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+// ConfigGenerator defines the cluster config generation for a cluster.
+type ConfigGenerator struct {
+	config.Cluster
+}
+
+// Read and return the content of the Certificate Files
+func readCertFile(path string) (string, error) {
+	if err := validate.FileExists(path); err != nil {
```
enxebre: You probably don't want to validate here as it's redundant with .ioutil.ReadFile and does not belong to this responsibility boundary.

installer/pkg/tls/tls.go
```diff
@@ -0,0 +1,45 @@
...
+	f, err := ioutil.ReadFile(path)
+	if err != nil {
+		return "", fmt.Errorf("Error reading file (%s): %s", path, err)
+	}
+
+	pem := string(f)
+	return pem, nil
+}
+
+func (config *ConfigGenerator) getConfig() (string, string, string, error) {
```
enxebre: This is looking good!
having 
`
type ConfigGenerator struct {
 config.Cluster
}
`
feels a bit confusing as this is currently contained within its own package.
we should probably eventually align with it (so details from this library could be consumed from the configGenerator package) see the down path for https://github.com/coreos/tectonic-installer/blob/master/installer/pkg/workflow/install.go#L145 On the other hand might be good to keep this selfcontained for now wdyt?

trawler: Yeah, i'll probably move this section inside the `config/tls` package, as it belongs there anyways.

installer/pkg/tls/tls.go
trawler: Agreed. I'll validate the config parameters inside `config/tls` so as to remove redundancies.

installer/pkg/config-generator/tls.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"fmt"
+	"io/ioutil"
+
+	"github.com/coreos/tectonic-installer/installer/pkg/validate"
+)
+
+// If not provided, TLS pem contents are empty
+var (
+	rootCACert = ""
```
squat: These two variables are modified by the `GenerateTLSConfig` func but they are never used anywhere. Nothing actually uses the generated certs.
Also, why do we need global variables? `GenerateTLSConfig` should not need to store state anywhere other than generating certificates and writing them to disk, right?

installer/pkg/config-generator/tls.go
```diff
@@ -0,0 +1,71 @@
...
+	rootCAKey  = ""
+)
+
+// GenerateTLSConfig fetches and validates the TLS cert files
+// If no file paths were provided, the certs will be auto-generated
+func (c *ConfigGenerator) GenerateTLSConfig() error {
```
squat: This function suppresses all errors and always returns nil. How do we know if something is failing? We need to check all errors and surface them.

installer/pkg/config-generator/tls.go
```diff
@@ -0,0 +1,71 @@
...
+	rootCAKeyAlg := c.CA.RootCAKeyAlg
+	fmt.Printf("c.CA.RootCAKeyAlg equals: [%s]\n", rootCAKeyAlg)
```
squat: Were all of these `Printf`s for testing during development? We should either eliminate them all of replace them with logs of either trace or debug.

installer/pkg/config-generator/tls.go
```diff
@@ -0,0 +1,71 @@
...
+	rootCACertPath := c.CA.RootCACertPath
+	fmt.Printf("c.CA.RootCACertPath equals: [%s]\n", rootCACertPath)
+	if len(rootCACertPath) != 0 {
```
squat: The function description says that if no CA cert path is provided then a cert is generated but I don't see that here. Where does the generation occur?

installer/pkg/config-generator/tls.go
```diff
@@ -0,0 +1,71 @@
...
+		rootCACert, _ = parseCertFile(rootCACertPath)
+	}
+	rootCAKeyPath := c.CA.RootCAKeyPath
+	fmt.Printf("c.CA.RootCAKeyPath equals: [%s]\n", rootCAKeyPath)
+	if len(rootCAKeyPath) != 0 {
```
squat: Same with the case where no key is provided: where is the key generated?

installer/pkg/config-generator/tls.go
```diff
@@ -0,0 +1,71 @@
...
+		rootCAKey, _ = parseKeyFile(rootCAKeyPath)
+	}
+
+	return nil
+}
+
+// Validate and return the content of the Certificate File
+func parseCertFile(path string) (string, error) {
+	if err := validate.FileExists(path); err != nil {
+		return "", fmt.Errorf("Invalid file path provided (%s): %s", path, err)
```
squat: We like all errors to begin with a lowercase letter so that they can be wrapped by other functions (just like you are wrapping another error here).

installer/pkg/config-generator/tls.go
```diff
@@ -0,0 +1,71 @@
...
+	if err := validate.Certificate(pem); err != nil {
+		return "", fmt.Errorf("Provided file does not seem to contain a valid certificate (%s): %s", path, err)
+	}
+
+	return pem, nil
+}
+
+// Validate and return the content of the Private Key File
+func parseKeyFile(path string) (string, error) {
+	if err := validate.FileExists(path); err != nil {
```
squat: There is no need to validate that the file exists before trying to read the file; if you get an error when reading just wrap and return that, e.g.
```go
return fmt.Errorf("failed to ready certificate key file: %v", err)
```

In fact, we will already need to validate that these paths are either valid or null in `installer/pkg/config/validate.go` so this function should really just check the error returned by `ioutil.ReadFile`.

installer/pkg/config/cluster.go
```diff
@@ -53,24 +56,24 @@ var defaultCluster = Cluster{
...
 	Internal        `json:",inline" yaml:"-"`
+	libvirt.Libvirt `json:",inline" yaml:"libvirt,omitempty"`
 	LicensePath     string `json:"tectonic_license_path,omitempty" yaml:"licensePath,omitempty"`
 	Master          `json:",inline" yaml:"master,omitempty"`
 	Name            string `json:"tectonic_cluster_name,omitempty" yaml:"name,omitempty"`
 	Networking      `json:",inline" yaml:"networking,omitempty"`
 	NodePools       `json:"-" yaml:"nodePools"`
 	Platform        Platform `json:"tectonic_platform" yaml:"platform,omitempty"`
 	PullSecretPath  string   `json:"tectonic_pull_secret_path,omitempty" yaml:"pullSecretPath,omitempty"`
+	CA              `json:",inline" yaml:"CA,omitempty"`
```
squat: I see that you alphabetized all of the fields but put this one out of order. Was there a reason for that?

installer/cmd/tectonic/main.go
```diff
@@ -39,6 +40,8 @@ func main() {
 	case clusterInstallFullCommand.FullCommand():
 		w = workflow.NewInstallFullWorkflow(*clusterInstallDirFlag)
 	case clusterInstallTLSCommand.FullCommand():
+		w = workflow.InstallTLSWorkflow(*clusterInstallDirFlag)
+	case clusterInstallNewTLSCommand.FullCommand():
```
squat: Nit: these changes break the naming pattern that the other commands and workflows had. Previously it was:
command: X
workflow: workflow.NewX
We should try to be consistent so that future maintainers have an easier time.

installer/pkg/config-generator/tls.go
trawler: I was planning to generate the certs in a separate tls package `installer/pkg/tls` but If i called that package now, when it doesn't yet exist, It would fail the build. Hence, the printfs that will be replaced with the correct function call.

squat: I see. In that case we should work backwards. The first PR we should merge is a fully working `instaler/pkg/tls` package that can generate certificates. This is self-contained and feature complete. The second PR we merge is `installer/pkg/config-generator/tls.go`. which builds on the first one. In general, we should be merging the self-contained dependencies first and then merge the code that depends on it. Otherwise, we are merging known-broken code :/

enxebre: yeah let's try to get small batches of self-contained library details respecting the dependencies, then we add abstractions, integrate and expose to the user. i.e:
INST-1107..INST-1113 by order, then we get this in. Otherwise this one need to go in hand with an implementation for tls.generateRootCA

installer/pkg/config/validate.go
```diff
@@ -410,3 +411,25 @@ func (c *Cluster) validateNoSharedNodePools() []error {
...
+func (c *Cluster) validateTLSConfig() []error {
+	var errs []error
+
+	switch {
+	case c.CA.RootCAKeyPath == "" && c.CA.RootCACertPath != "":
+		errs = append(errs, fmt.Errorf("rootCAKeyPath cannot be empty if rootCACertPath is set"))
+		break
+	case c.CA.RootCACertPath == "" && c.CA.RootCAKeyPath != "":
+		errs = append(errs, fmt.Errorf("rootCACertPath cannot be empty if rootCAKeyPath is set"))
+		break
```
squat: we could summarize the two separate cases above into one single case like:
```go
case (c.CA.RootCACertPath == "") != (c.CA.RootCAKeyPath == ""):
    errs = append(errs, fmt.Errorf("rootCACertPath and rootCAKeyPath must both be set or empy"))
```

installer/pkg/config-generator/utils.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		return err
+	}
+	defer to.Close()
+
+	_, err = io.Copy(to, from)
+	return err
+}
+
+// Validate and return the content of the Certificate File
```
squat: nit: We should try to ensure that comments describing are full sentences with correct casing an identifier start with the name of the identifier, even if it is unexported. So this comment should look like:
```go
// parseCertFile validates and returns the content of a certificate file.
func parseCertFile(path string) (string, error) {
```

installer/pkg/config-generator/utils.go
```diff
@@ -0,0 +1,134 @@
...
+func parseCertFile(path string) (string, error) {
+	data, err := ioutil.ReadFile(path)
+	if err != nil {
+		return "", fmt.Errorf("failed to ready certificate file: %v", err)
+	}
+	pem := string(data)
+
+	// Validate that file content is a valid Certificate
+	if err := validate.Certificate(pem); err != nil {
```
squat: nit: lowercase `Certificate`

installer/pkg/config-generator/utils.go
```diff
@@ -0,0 +1,134 @@
...
+		return "", fmt.Errorf("provided file does not seem to contain a valid certificate (%s): %s", path, err)
+	}
+
+	return pem, nil
+}
+
+// Validate and return the content of the Private Key File
```
squat: nit: fix comment as described above.

installer/pkg/config-generator/utils.go
```diff
@@ -0,0 +1,134 @@
...
+
+	// Validate that file content is a valid Key
+	if err := validate.PrivateKey(pem); err != nil {
+		return "", fmt.Errorf("provided file does not seem to contain a valid private key (%s): %s", path, err)
+	}
+	return pem, nil
+}
+
+// Get the contents of the Private Key
+func privateKeytoPem(key *rsa.PrivateKey) string {
```
squat: nit: same as above

PR #3275, opened 2018-06-08
*: remove k8s-node-bootstrap.service

crawford:
This was broken by 4f8f4fc when the origin image was used instead. The
k8s-node-bootstrap service was failing to parse "latest" as valid
semver. Since we are eventually moving away from Container Linux, this
service is safe to remove in favor of some future mechanism.


squat:
:crying_cat_face:

squat:
ok to test

PR #3288, opened 2018-06-13
clean up ruby tests

enxebre:
remove tests unused code and cluster abstraction


enxebre:
retest this please

enxebre:
this is just a quick cleanup of old files on the solution which is currently in place, once we got a new solution working we can completely remove it

PR #3289, opened 2018-06-13
*: connect to API server on 6443 instead of 443

crawford:
The API server listens on 6443. Using 443 requires a load-balancer (AWS)
or IP Tables (libvirt) to direct the traffic properly. It's simpler to
just connect on the correct port.


crawford:
Continuation of https://github.com/coreos-inc/tectonic-operators/pull/389.

crawford:
ok to test

PR #3293, opened 2018-06-14
modules/tectonic/resources: add TNC key/cert pair

crawford:
TNC needs this key/cert pair in order to secure HTTPS traffic.


crawford:
ok to test

PR #3294, opened 2018-06-14
*: use HTTPS when serving Ignition configs

crawford:


crawford:
Continuation of https://github.com/coreos-inc/tectonic-operators/pull/393.

crawford:
retest this please

yifan-gu:
retest this please.

installer/pkg/config-generator/ignition.go
```diff
@@ -114,13 +135,15 @@ func (c *ConfigGenerator) getTNCURL(role string) ignconfigtypes.Url {
...
 	}
 	if role == "master" || role == "worker" {
-		url = ignconfigtypes.Url{
-			Scheme: "http",
-			Host:   fmt.Sprintf("%s-tnc.%s:%d", c.Name, c.BaseDomain, port),
-			Path:   fmt.Sprintf("/config/%s", role),
-		}
+		u = func() *url.URL {
+			return &url.URL{
+				Scheme: "https",
```
enxebre: This won't work for the bootstrap node request being resolved via cname to s3 as aws tls certificate does not know about this domain

steps/assets/base/tectonic.tf
```diff
@@ -40,6 +40,8 @@ module "bootkube" {
   kube_ca_key_pem              = "${local.kube_ca_key_pem}"
   kubelet_cert_pem             = "${local.kubelet_cert_pem}"
   kubelet_key_pem              = "${local.kubelet_key_pem}"
+  tnc_cert_pem                 = "${local.tnc_cert_pem}"
```
enxebre: where are these locals being defined?

installer/pkg/config-generator/ignition.go
enxebre: this won't work for the bootstrap node request resolved to s3 as the tls aws certs does not know about this domain

steps/assets/base/tectonic.tf
crawford: They were added in 37f29a5.

installer/pkg/config-generator/ignition.go
crawford: Christ, we need to kill off this S3-pivot insanity.

crawford:
The masters are still trying to use HTTPS to fetch the config. I have no idea why. This works on my local system.

installer/pkg/config-generator/ignition.go
```diff
@@ -91,36 +99,59 @@ func (c *ConfigGenerator) embedAppendBlock(ignCfg *ignconfigtypes.Config, role s
...
 	// libvirt doesn't do that - use the tnc port directly
 	port := 80
 	if c.Platform == "libvirt" {
 		port = 49500
 	}
+
+	// XXX: The bootstrap node on AWS uses a CNAME to redirect TNC-bound
+	// traffic to S3. Because of this, HTTPS cannot be used.
+	scheme := "https"
+	if c.Platform == "AWS" && role == "master" {
```
yifan-gu: aws?

yifan-gu: https://github.com/coreos/tectonic-installer/blob/master/installer/pkg/config/cluster.go#L25

yifan-gu: Also calling I think it's advised to `c.Platform.String()`

modules/bootkube/resources/bootkube.sh
```diff
@@ -44,24 +50,48 @@ signer_id=$(/usr/bin/docker run -d \
...
+    set -e
+
+    if [ "$status" -eq 0 ]; then
+      break
+    fi
+
+    i=$((i+1))
+    [ $i -eq 10 ] && echo "etcdctl failed too many times." && exit 1
+
+    echo "etcdctl failed. Retrying in 5 seconds..."
```
yifan-gu: Doesn't `etcdctl endpoint health` have its own retry logic?
cc @gyuho

crawford: I observed it failing to retry.

gyuho: The retry logic would depend on the error types. Since endpoint health is just a simple get request, it's safe to retry here.

crawford:
retest this please

yifan-gu:
retest this please

yifan-gu:
retest this please

enxebre:
retest this please

PR #3296, opened 2018-06-15
Platform cleanup

squat:
This PR cleans up the repository by removing references to old platforms, e.g. bare metal, azure, etc, by eliminating dead code, and updating documentation.

cc @enxebre

Addresses: INST-1124


contrib/internal-cluster/README.md
```diff
@@ -61,8 +61,8 @@ Terraform does not support changing SOA TTLs in Route 53. As a result, in order
 
 Once all the infrastructure is provisioned and the VPN connection is available, a Tectonic cluster can be installed in the VPC. When running the Tectonic installer, be sure to:
```
enxebre: Are we actually currently leveraging  internal-cluster somehow? should we remove the whole contrib?

squat: I looked through it and the code is still valid although unused. WDYT? I'm happy to eliminate it entirely

enxebre: Cool let’s remove it if you are good with it, we can always get back to the git history if we need it

enxebre:
retest this please

PR #3302, opened 2018-06-19
installer/pkg/tls: tls package for creating and signing certs

trawler:
Fixes INST-1107

### DIFF
--- a/installer/pkg/tls/tls_test.go
+++ b/installer/pkg/tls/tls_test.go
@@ -1,0 +1,58 @@
+package tls
+
+import (
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"testing"
+	"time"
+)
+
+func TestSelfSignedCACert(t *testing.T) {
+	key, err := GeneratePrivateKey()
+	if err != nil {
+		t.Fatalf("Failed to generate Private Key: %v", err)
+	}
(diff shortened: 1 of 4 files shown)


squat:
@trawler the tests are failing because the tls_test file was never finished:

```
installer/pkg/tls/tls_test.go:30:32: not enough arguments in call to SelfSignedCACert
installer/pkg/tls/tls_test.go:30:34: c.alg undefined (type struct { cfg CertCfg; err bool } has no field or method alg)
```

enxebre:
golang unit tests need to be added here to be run by bazel https://github.com/coreos/tectonic-installer/blob/master/installer/BUILD.bazel#L3

installer/pkg/tls/tls_test.go
```diff
@@ -0,0 +1,34 @@
...
+func TestSelfSignedCACert(t *testing.T) {
+	key, err := GeneratePrivateKey()
+	if err != nil {
+		t.Fatalf("Failed to generate Private Key: %v", err)
+	}
+	cases := []struct {
+		cfg *CertCfg
+		err bool
+	}{
+		{
```
squat: We should add some more test cases to really verify all the code paths. Can you add a test case that returns an error?

PR #3312, opened 2018-06-23
*: remove container linux specifics

crawford:


crawford:
Continuation of https://github.com/coreos-inc/tectonic-operators/pull/400.

PR #3319, opened 2018-07-02
WIP: Refactor init command to initialize and add import command which…

spangenberg:


installer/pkg/workflow/init.go
```diff
@@ -5,92 +5,43 @@ import (
...
 
-	if m.configFilePath == "" {
-		return errors.New("a path to a config file is required")
-	}
-
-	// load initial cluster config to get cluster.Name
-	cluster, err := readClusterConfig(m.configFilePath, "")
-	if err != nil {
-		return fmt.Errorf("failed to get configuration from file %q: %v", m.configFilePath, err)
+	if m.domain == "" {
```
enxebre: just to clarify by design this will never be the case as kingpin won't allow it right?

spangenberg: Yup, forgot to remove this. Didn't have Required() at first on the arg in kingpin.

PR #3320, opened 2018-07-02
modules/bootkube: explicitly depend on docker

crawford:
bootkube.sh makes heavy use of docker but doesn't wait until it's ready.
This worked on Container Linux because docker was socket activated.
That's not the case on RHCOS right now, so we need to be explicit.


squat:
ok to test

PR #3321, opened 2018-07-03
config.tf: bump operator images

crawford:
This removes a number of Container Linux specifics including the
following:

 - explicitly depend on docker.service when needed
 - remove unnecessary quotes from etcd TLS service
 - update the certificate location from /usr/share to /etc/ssl
 - change etcd-member to be a simple systemd service instead of notify


crawford:
Verified that this works with Container Linux and RHCOS (with some minor changes) on libvirt.

enxebre (approved):

PR #3322, opened 2018-07-04
installer/pkg/config-generator/tls: generate Kube  CAs

trawler:


installer/pkg/config-generator/tls/ca.go
```diff
@@ -0,0 +1,144 @@
...
+	certDst := filepath.Join(clusterDir, rootCACertPath)
+	if err := copyFile(certPath, certDst); err != nil {
+		return nil, nil, fmt.Errorf("failed to write file: %v", err)
+	}
+	// content validation occurs in pkg/config/validate.go
+	// so no err should be returned at this stage
+	certData, _ := ioutil.ReadFile(certPath)
+	certPem, _ := pem.Decode([]byte(string(certData)))
+
+	keyData, _ := ioutil.ReadFile(keyPath)
```
squat: same as above

installer/pkg/config-generator/tls/ca.go
```diff
@@ -0,0 +1,144 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error generating self signed certificate: %v", err)
+	}
+	if err := writeFile(fileTargetPath, certToPem(cert)); err != nil {
+		return nil, err
+	}
+	return cert, nil
+}
+
+func generateCSR(key *rsa.PrivateKey, subject *pkix.Name) (*x509.CertificateRequest, error) {
```
squat: this function does not seem very useful IMO

installer/pkg/tls/tls.go
```diff
@@ -88,3 +97,21 @@ func SignedCertificate(
...
+) (*x509.CertificateRequest, error) {
+	csrtTmpl := x509.CertificateRequest{
+		DNSNames:    cfg.DNSNames,
+		IPAddresses: cfg.IPAddresses,
+		Subject:     cfg.Subject,
+	}
+
+	csrBytes, err := x509.CreateCertificateRequest(rand.Reader, &csrtTmpl, key)
+	if err != nil {
+		return nil, fmt.Errorf("error creating signed certificate: %s", err)
```
squat: remember we like to format errors with `%v`

installer/pkg/tls/tls.go
```diff
@@ -88,3 +97,21 @@ func SignedCertificate(
 	}
 	return x509.ParseCertificate(certBytes)
 }
+
+// CertificateRequest generates a certificate request
+func CertificateRequest(
```
squat: This seems like an unnecessary level of abstraction. Users of this func still need to create a configuration (now a CSRCfg rather than a x509.CertificateRequest, and still need to call a func (now CertificateRequest rather than x509 CreateCertificateRequest). Given that more teams will be working on this and this function is quite trivial it seems better to maintain convention than to customize.

installer/pkg/config-generator/tls/utils.go
```diff
@@ -0,0 +1,89 @@
...
+	"encoding/pem"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+
+	"github.com/coreos/tectonic-installer/installer/pkg/tls"
+)
+
+func writeFile(path, content string) error {
```
squat: these functions  are now duplicated across multiple packages. We should avoid copy and pasting code across multiple packages and instead import those funcs from other packages.

installer/pkg/tls/tls.go
```diff
@@ -88,3 +97,21 @@ func SignedCertificate(
...
+	cfg *CSRCfg,
+	key *rsa.PrivateKey,
```
enxebre: just curious, is there a convention why we are injecting in multiline rather than a single line?

enxebre:
No more PRs will be accepted for master branch against https://github.com/coreos/tectonic-installer
Further development for the Tectonic and Openshift converged installer platform will now take place in https://github.com/openshift/installer
@trawler can you please reopen against that repo once the comments are addressed? thanks!

PR #3324, opened 2018-07-04
tls: support for certificate chains

spangenberg:
Fork of https://github.com/coreos/tectonic-installer/pull/3191 in order to rebase correctly.

If you specify custom vnet using tectonic_azure_external_vnet_id
and this belongs to a different resource group than where masters
and workers are located, Kubernetes Services of type LoadBalancer will
fail. Support for setting custom resource group for vnet in cloud config
was added in Kubernetes 1.6 using vnetResourceGroup. This will add
the vnetResourceGroup to the cloud config if
tectonic_azure_external_vnet_id is set.


modules/tls/identity/self-signed/outputs.tf
```diff
@@ -11,5 +15,5 @@ output "server_key_pem" {
 }
 
 output "server_cert_pem" {
-  value = "${tls_locally_signed_cert.identity_server.cert_pem}"
+  value = "${var.cert_chain == "" ? tls_locally_signed_cert.identity_server.cert_pem : local.server_crt_chain}"
```
squat, line 18: Why do we need a ternary at all? We have two strings a and b. c is defined as a+b. If b == “” then c == a, no? Can’t we simplify this to always output the local var?

modules/tls/ingress/self-signed/output.tf
```diff
@@ -7,5 +11,5 @@ output "key_pem" {
 }
 
 output "cert_pem" {
-  value = "${tls_locally_signed_cert.ingress.cert_pem}"
+  value = "${var.cert_chain == "" ? tls_locally_signed_cert.ingress.cert_pem : local.ingress_crt_chain}"
```
squat, line 14: Same as above

modules/tls/kube/self-signed/assets.tf
```diff
@@ -1,10 +1,15 @@
+locals {
+  apiserver_crt_chain = "${tls_locally_signed_cert.apiserver.cert_pem}${var.cert_chain}"
+  kubelet_crt_hain    = "${tls_locally_signed_cert.kubelet.cert_pem}${var.cert_chain}"
```
squat, line 3: Same for this file

modules/tls/kube/self-signed/outputs.tf
```diff
@@ -11,15 +11,15 @@ output "ca_key_pem" {
 }
 
 output "kubelet_cert_pem" {
-  value = "${tls_locally_signed_cert.kubelet.cert_pem}"
+  value = "${var.cert_chain == "" ? tls_locally_signed_cert.kubelet.cert_pem : local.kubelet_crt_hain}"
```
squat, line 14: Same for this file too

enxebre:
Thanks a lot for this, I'd refuse to add yet another top level config and a new feature at this point of the project life unless there's a critical strong justification so closing this

PR #3328, opened 2018-07-06
operator: Add toleration to the operators for everything

yifan-gu:
Customers will add their own taints to nodes, which can block
our operators, so we need to add the omnipotent toleration here
to make our operators schedulable.

Fix https://jira.coreos.com/browse/CORS-681


spangenberg (approved):
LGTM, please rebase

PR #3330, opened 2018-08-04
Fixes issue tectonic-forum issue 251

wongt:
Fix issue: https://github.com/coreos/tectonic-forum/issues/251

"data.aws_route53_zone.tectonic: no matching Route53Zone found" during install


yifan-gu:
ok-to-test

